PR #78, opened 2026-04-13
fix(spur-k8s): support SpurJobs across all namespaces

shiv-tyagi:
The operator was previously scoped to a single namespace (`--namespace`, default `spur`). As a result, any `SpurJob` created outside that namespace was never picked up by the controller.

---
#### What was broken

- The controller only watched one namespace, so jobs in other namespaces were ignored.
- Cleanup logic used the operator’s namespace instead of the job’s namespace, causing deletions to fail or target the wrong resources.

#### Why this matters

Running all workloads in the operator’s namespace is not a good security model. User jobs should run in their own namespaces so they can benefit from proper RBAC, quotas, and network policies. The previous setup forced everything into a shared namespace with no isolation. This forces job pods to run in the controller’s namespace. If those pods use a ServiceAccount with elevated permissions in that namespace, it can lead to unintended privilege escalation and potential security risks.

#### What’s changed

- The operator now watches `SpurJob`s cluster-wide.
- Jobs are created and managed in their own namespaces.
- Cleanup always happens in the correct namespace.

#### How it works

`spurctld` is namespace-agnostic and only deals with job IDs. To bridge that, the operator adds a `spur.ai/job-id` label to each `SpurJob`. When a job is dispatched, the system looks up the namespace using this label—making Kubernetes the source of truth and avoiding in-memory state.

#### Why cluster-wide scope

A single operator managing the whole cluster is the intended model (similar to Volcano or Argo). Running one operator per namespace would add unnecessary complexity, and the RBAC was already cluster-scoped anyway.

#### Before / After

**Before (job not admitted outside `spur`):**
Screenshot 2026-04-13 204651

**After (job admitted in its own namespace):**
Screenshot 2026-04-13 205140

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3122,7 +3122,6 @@ dependencies = [
  "chrono",
  "clap",
  "hostname",
- "libc",
  "nix",
  "prost-types",
  "serde",
--- a/crates/spur-k8s/src/agent.rs
+++ b/crates/spur-k8s/src/agent.rs
@@ -1,5 +1,8 @@
 use std::collections::BTreeMap;
+use std::time::Duration;
 
+use backoff::future::retry;
+use backoff::ExponentialBackoffBuilder;
 use k8s_openapi::api::core::v1::{
     Container, EnvVar, HostPathVolumeSource, Pod, PodSpec, ResourceRequirements, Service,
     ServicePort, ServiceSpec, Volume, VolumeMount,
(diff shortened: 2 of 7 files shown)


shiv-tyagi:
Will rebase and mark ready after #79 is merged.

shiv-tyagi:
This PR has a test file which would conflict with #81. I will rebase and merge post that. @powderluv Please approve if this looks okay to you. I will take care of merging on CI pass.

powderluv (approved):

shiv-tyagi:
Resolved the merge conflict with main. CI is looking good. Merging now. Thanks for the review @powderluv.

PR #81, opened 2026-04-14
 fix(spurctld): use correct namespace when creating leader election Lease

shiv-tyagi:
## Summary
- `try_acquire()` set the Lease `ObjectMeta.namespace` to `leases.resource_url()` which returns the full K8s API path (e.g. `/apis/coordination.k8s.io/v1/namespaces/spur/leases`) instead of just `"spur"`
- This caused a `400 BadRequest` ("the namespace of the provided object does not match the namespace sent on the request") on every Lease creation attempt, making leader election permanently broken on fresh deployments
- The bug was masked when a Lease already existed from a prior run (the PATCH/renew path doesn't include namespace in the body)

## Fix
Pass the namespace string through to `try_acquire()` and use it directly in the Lease `ObjectMeta` instead of calling `resource_url()`.

## Before (main branch)
spurctld stuck in infinite retry loop — never acquires leadership:
Screenshot 2026-04-14 122129

## After (this PR)
Lease created and acquired on first attempt:
Screenshot 2026-04-14 123150

## Testing
- [x] All 776 unit tests pass (`cargo test`)
- [x] Bug reproduced on live 3-node K8s cluster (v1.33)
- [x] Fix verified on same cluster — Lease created, leader acquired, nodes registered
- [x] Added K8s integration test (TEST 7 in `k8s_test.sh`) that verifies:
  - Controller starts with `--enable-leader-election`
  - `spurctld-leader` Lease is created in the correct namespace
  - No leader election errors in logs

### DIFF
--- a/crates/spurctld/src/leader_election.rs
+++ b/crates/spurctld/src/leader_election.rs
@@ -30,7 +30,7 @@ pub async fn acquire_lease(namespace: &str) -> anyhow::Result<()> {
 
     // Try to acquire the Lease in a loop
     loop {
-        match try_acquire(&leases, &identity).await {
+        match try_acquire(&leases, namespace, &identity).await {
             Ok(true) => {
                 info!("acquired leader Lease");
                 break;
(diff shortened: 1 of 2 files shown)


shiv-tyagi:
This fixes what we have today but I filed #82 to suggest some design improvements.

powderluv (approved):

PR #83, opened 2026-04-14
fix(ci): decouple k8s tests from cluster job via docker image artifact

shiv-tyagi:
The k8s integration step assumed binaries from the cluster job would be at `~/spur/bin/` on the same runner. Both jobs running on the same runner back to back is not guaranteed like another job can come in between the two jobs, causing intermittent failures. (I experienced this lately)

Fix: add a build-docker-image job that builds via `deploy/Dockerfile`, uploads the image as an artifact, and the k8s job downloads and uses it. `k8s_test.sh` now requires SPUR_CI_IMAGE instead of host binaries.

This also means the docker image build is validated on every CI run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,24 +149,47 @@ jobs:
           ssh mi300-2 '~/spur/etc/stop-agent.sh' || true
           ~/spur/etc/stop-all.sh || true
 
-  k8s:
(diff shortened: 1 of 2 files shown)


shiv-tyagi:
https://github.com/ROCm/spur/actions/runs/24402030397

This failure is an example of the problem we have. This job ran on stale binaries.

PR #89, opened 2026-04-14
fix(ci): run docker image build on github-hosted runners

shiv-tyagi:
Github runners are exempted from Dockerhub pull limits. This also helps us avoid too many open file issue which we hit on self hosted runners.

This aims to fix intermittent docker image build failure issues.
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -151,7 +151,7 @@ jobs:
 
   build-docker-image:
     name: Build Docker Image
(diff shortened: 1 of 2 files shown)


shiv-tyagi:
Cluster Tests failure is out of the scope of this PR. This can be merged.

Edit: This is fixed. Looks like a fix was done on the runner.

powderluv:
the runner was out of diskspace.

Issue #99, opened 2026-04-18
Bare-metal job isolation: user namespaces and cgroup enforcement


powderluv:
## Context
Several security issues have been filed against spur-cloud (powderluv/spur-cloud#3, #4, #5, #7, #8, #11) that trace back to insufficient isolation in bare-metal job execution.

## Current Behavior
When spurd runs jobs in bare-metal mode (no containers):
- Jobs run as root with full host access
- All sessions share the host filesystem
- Package installations in one session are visible to all others
- `who`, `ps`, etc. show all host processes
- Users can access Spur CLI tools (squeue, scancel) to view/cancel other users' jobs

## Required Changes
1. **User namespace isolation**: Run each job under a unique UID/GID, not root
2. **Filesystem isolation**: Per-session tmpfs overlay or mount namespace
3. **Process isolation**: PID namespace so `who`/`ps` only show session processes
4. **Spur CLI restriction**: Don't expose squeue/scancel in job environments, or scope them to the job's own user

## Related Issues
- powderluv/spur-cloud#3: Root access with all privileges
- powderluv/spur-cloud#4: Shared /nfsdata writable across users
- powderluv/spur-cloud#5: Session/process visibility
- powderluv/spur-cloud#7: Package installations shared
- powderluv/spur-cloud#8: Users can view/cancel other users' jobs
- powderluv/spur-cloud#11: 'who' shows host-level info

powderluv:
Implemented across 5 merged PRs:

- **PR #100**: UID/GID enforcement via setuid + pids.max fork bomb protection + memory.oom.group OOM isolation
- **PR #102**: PID + mount namespace isolation with selective GPU device bind-mounting
- **PR #103**: seccomp-BPF syscall whitelist (~150 syscalls, blocks ptrace/mount/bpf/unshare) — opt-in via SPUR_SECCOMP=1
- **PR #104**: Landlock LSM filesystem access control (kernel 5.13+) — opt-in via SPUR_LANDLOCK=1
- **PR #105**: `[isolation]` config section for operator control

All layers work with both bare-metal and container (Enroot-like) execution paths. Each degrades gracefully when the kernel doesn't support it or spurd isn't running as root.

PR #100, opened 2026-04-18
Add UID/GID enforcement, pids.max, and OOM isolation (#99 PR 1/6)

powderluv:
## Summary
First PR in the bare-metal job isolation series (#99). Adds three foundational cgroup + process controls.

## Changes
1. **setuid/setgid** — Jobs run as submitting user's UID/GID (not root). `Command::uid()/gid()` when spurd is root; graceful skip otherwise.
2. **pids.max** — Fork bomb protection. Limits processes per job to `max(cpus*256, 1024)`.
3. **memory.oom.group** — OOM kills entire job cgroup, not random process.

Works with both bare-metal and container (Enroot-like) paths — cgroup setup happens before process spawn in both cases.

## Isolation series roadmap
1. **This PR**: setuid + pids.max + OOM
2. Device cgroup (GPU enforcement at kernel level)
3. PID + mount namespace
4. seccomp-BPF syscall filter (from AXIS)
5. Landlock filesystem restrictions (from AXIS)
6. Config + CLI integration

## Test plan
- [x] Full test suite passes (0 failures)
- [ ] CI cluster test: submit job as non-root user, verify GPU access works
- [ ] Fork bomb test: `:(){ :|:& };:` → pids.max kills cgroup

### DIFF
--- a/crates/spurd/src/executor.rs
+++ b/crates/spurd/src/executor.rs
@@ -110,6 +110,8 @@ pub async fn launch_job(
     cpu_ids: &[u32],
     spank: Option<&SpankHost>,
     open_mode: Option<&str>,
+    uid: u32,
+    gid: u32,
 ) -> anyhow::Result<RunningJob> {
     info!(job_id, work_dir, "launching job");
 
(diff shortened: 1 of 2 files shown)


crates/spurd/src/executor.rs
```diff
@@ -279,6 +281,15 @@ pub async fn launch_job(
...
         .stdin(Stdio::null());
 
+    // Issue #99: Run job as the submitting user (not root).
+    // Requires spurd to run as root. Non-root spurd skips setuid.
+    if uid > 0 && nix::unistd::geteuid().is_root() {
+        use std::os::unix::process::CommandExt;
+        cmd.uid(uid);
+        cmd.gid(gid);
+        debug!(job_id, uid, gid, "job will run as non-root user");
+    }
```
shiv-tyagi, line 291: Should we also log a warning for spurd running as non-root user case that setuid and setgid cannot be performed?

shiv-tyagi:
Posted a doubt around gpu node permissions in #102. 

https://github.com/ROCm/spur/pull/102#pullrequestreview-4134946064

PR #102, opened 2026-04-18
Add PID + mount namespace isolation with GPU device restriction (#99 PR 2/6)

powderluv:
## Summary
Second PR in the isolation series (#99). Adds PID and mount namespace isolation for both bare-metal and container execution paths.

## Changes
- **PID namespace**: `unshare --pid --fork` — jobs see only their own processes
- **Mount namespace**: Private `/tmp`, `/dev/shm` per job; GPU `/dev/dri` restricted to allocated devices via selective bind-mount
- **Container path**: Added `--pid --fork` to existing `unshare --mount` in container.rs
- **Bare-metal path**: New namespace wrapper script in executor.rs
- **TrackedJob**: `has_pid_namespace` flag for correct nsenter behavior in exec_in_job

## GPU access verification
- Only allocated `renderD*` devices bind-mounted into private `/dev/dri`
- `/dev/kfd` preserved for AMD GPU access
- Unallocated GPU devices not visible in job's mount namespace

## Addresses
- spur-cloud #5 (process visibility)
- spur-cloud #11 (who shows host info)
- spur-cloud #4, #7 (partial — private /tmp prevents shared state)

## Test plan
- [ ] Job with PID namespace: `ps aux` shows only job processes (not host)
- [ ] Job with mount namespace: `ls /tmp` is empty (private tmpfs, not host)
- [ ] GPU device restriction: `ls /dev/dri/` inside 1-GPU job shows only 1 renderD device
- [ ] **GPU access after setuid** (per @shiv-tyagi): non-root user with `video`/`render` group can access `/dev/kfd` and `/dev/dri/renderD*`. PR #100 needs `initgroups()` to inherit supplementary groups — without it, setuid to a non-root UID loses `video`/`render` group membership and GPU device access fails with EACCES.
- [ ] Attach to namespaced job: `nsenter --mount --pid` enters correct namespaces
- [ ] Container path: `ps aux` inside container shows only container processes
- [ ] `who` command inside job: only shows current session (not host users)
- [ ] rocm-smi inside namespaced job: shows only allocated GPU
- [ ] HIP runtime test: `hipGetDeviceCount` returns correct count inside namespace
- [x] Full test suite passes (0 failures)

## Note on supplementary groups
As @shiv-tyagi pointed out, `Command::uid().gid()` sets primary UID/GID but does NOT set supplementary groups. GPU devices are typically `root:video` or `root:render`. PR #100 (setuid PR) needs to call `initgroups(username, gid)` in `pre_exec` to ensure the job process inherits `video`, `render`, and other groups. Without this, GPU access will fail for non-root users.

### DIFF
--- a/crates/spurd/src/agent_server.rs
+++ b/crates/spurd/src/agent_server.rs
@@ -34,6 +34,8 @@ struct TrackedJob {
     stdout_path: String,
     /// Stderr path for output streaming.
     stderr_path: String,
+    /// Whether this job has a PID namespace (for nsenter flags).
+    has_pid_namespace: bool,
 }
 
 pub struct AgentService {
@@ -615,6 +617,7 @@ impl SlurmAgent for AgentService {
                         allocation: alloc_result,
                         stdout_path,
                         stderr_path,
+                        has_pid_namespace: nix::unistd::geteuid().is_root(),
                     },
                 );
                 info!(job_id, gpus = ?gpu_devices, "job launched successfully");
@@ -721,9 +724,19 @@ impl SlurmAgent for AgentService {
             "exec into running job"
         );
 
-        // Use nsenter to enter the job's mount namespace and run the command
+        // Use nsenter to enter the job's namespace(s) and run the command
+        let has_pid_ns = {
+            let jobs = self.running.lock().await;
+            jobs.get(&req.job_id)
+                .map(|j| j.has_pid_namespace)
+                .unwrap_or(false)
+        };
         let mut cmd = tokio::process::Command::new("nsenter");
-        cmd.args(["--target", &pid.to_string(), "--mount", "--"]);
+        cmd.arg("--target").arg(pid.to_string()).arg("--mount");
+        if has_pid_ns {
+            cmd.arg("--pid");
+        }
+        cmd.arg("--");
         cmd.arg(&req.command[0]);
         for arg in &req.command[1..] {
             cmd.arg(arg);
--- a/crates/spurd/src/container.rs
+++ b/crates/spurd/src/container.rs
@@ -441,7 +441,7 @@ mkdir -p {rootfs}{home}
     script.push_str(&format!(
         r#"
 if [ "$(id -u)" = "0" ]; then
-  exec unshare --mount bash -c '
+  exec unshare --mount --pid --fork bash -c '
 set -e
 ROOTFS="{rootfs}"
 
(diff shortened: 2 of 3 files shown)


shiv-tyagi (commented):
In the test plan, we should also verify if a job process running under a non-root uid and gid is correctly able to access the gpu device nodes. 

The gpu nodes are restricted to `video` and `render` groups, we are setting uid and gid for the process, which doesn't seem to inherit supplementary groups of the user running the commands. I suspect that might lead to GPUs not being accessible to the process.

powderluv:
Good catch @shiv-tyagi — you're right that `Command::uid().gid()` doesn't set supplementary groups. GPU devices are typically owned by `root:video` or `root:render`, so the job process needs to be in those groups.

I'll update PR #100 (setuid PR) to call `initgroups()` in `pre_exec` to inherit all supplementary groups for the target user. This ensures the process gets `video`, `render`, and any other groups the user belongs to.

Updated test plan for this PR:
- [ ] Job with PID namespace: `ps aux` shows only job processes
- [ ] Job with mount namespace: private `/tmp`, only allocated GPU in `/dev/dri`
- [ ] **GPU access after setuid**: non-root user with `video`/`render` group membership can access `/dev/kfd` and `/dev/dri/renderD*`
- [ ] Attach to namespaced job: nsenter with `--pid` flag works correctly

Will add `initgroups()` call in a follow-up to PR #100.

shiv-tyagi:
> Good catch @shiv-tyagi — you're right that `Command::uid().gid()` doesn't set supplementary groups. GPU devices are typically owned by `root:video` or `root:render`, so the job process needs to be in those groups.
> 
> I'll update PR #100 (setuid PR) to call `initgroups()` in `pre_exec` to inherit all supplementary groups for the target user. This ensures the process gets `video`, `render`, and any other groups the user belongs to.
> 
> Updated test plan for this PR:
> 
> * [ ]  Job with PID namespace: `ps aux` shows only job processes
> * [ ]  Job with mount namespace: private `/tmp`, only allocated GPU in `/dev/dri`
> * [ ]  **GPU access after setuid**: non-root user with `video`/`render` group membership can access `/dev/kfd` and `/dev/dri/renderD*`
> * [ ]  Attach to namespaced job: nsenter with `--pid` flag works correctly
> 
> Will add `initgroups()` call in a follow-up to PR #100.

Submitted #107 to track this.

PR #108, opened 2026-04-19
Set supplementary groups for non-root job processes (#107)

### LINKED ISSUE #107: Ensure GPU nodes permissions to job process when running as non root
As mentioned in https://github.com/ROCm/spur/pull/102#issuecomment-4274454605 comment, when a job process is not running as root, it might not be able to access the GPU nodes due to missing permissions. We need to ensure the supplementary groups are passed to the process for it to be able to access them.

Submitting this to keep a track of this bug.

powderluv:
## Summary
Fixes #107 — calls `initgroups()` in `pre_exec` so non-root job processes inherit supplementary groups (video, render, etc.) needed for GPU device access.

## Problem
`Command::uid().gid()` only sets primary UID/GID. GPU devices are typically `root:video` or `root:render`. Without supplementary groups, non-root jobs get EACCES on `/dev/dri` and `/dev/kfd`.

## Fix
Added `libc::initgroups(username, gid)` call in `pre_exec` before the process execs. This reads `/etc/group` and sets all supplementary groups the user belongs to.

## Test plan
- [ ] Submit job as non-root user → verify `id` shows video/render groups
- [ ] Verify `rocm-smi` works inside non-root job
- [ ] Verify `stat /dev/kfd` accessible inside non-root job
- [x] Compiles clean

### DIFF
--- a/crates/spurd/src/executor.rs
+++ b/crates/spurd/src/executor.rs
@@ -338,12 +338,34 @@ pub async fn launch_job(
         .stderr(stderr_file.into_std().await)
         .stdin(Stdio::null());
 
-    // Issue #99: Run job as the submitting user (not root).
+    // Issue #99, #107: Run job as the submitting user (not root).
+    // Must set supplementary groups (video, render) via initgroups()
+    // so the process can access GPU device nodes.
     if uid > 0 && nix::unistd::geteuid().is_root() {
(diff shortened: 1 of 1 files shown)


shiv-tyagi (approved):
LGTM. Thanks @powderluv.

Issue #111, opened 2026-04-20
Test t50_106 fails when run from an interactive terminal


shiv-tyagi:
Test t50_106 fails when run from an interactive terminal because stdin is a real TTY, not a pipe. The test only passes on CI by accident.

Another important note,
The test also fails to test what it claims to. It reimplements the isatty() check that RawModeGuard::enter() does internally, rather than actually calling RawModeGuard::enter(). 

A test should run the real unit it's guarding, not simulate its behavior with standalone libc calls, otherwise we're just testing the OS and the test gives no confidence that Spur's actual code path works. 

The fix is to make RawModeGuard testable (e.g. pub(crate) or extract into a shared module) and assert directly that it returns Err for pipes and Ok for TTYs.

PR #112, opened 2026-04-20
CLAUDE.md: add don'ts for writing unit tests

shiv-tyagi:
To avoid a situation like #111, the agent should ensure the unit tests _actually test the unit_ instead of _simulating_ the behavior and then testing that simulation.


powderluv:
Exactly the kind of thing we need human oversight for

PR #114, opened 2026-04-22
Fix test t50_106: test real RawModeGuard, not simulation (#111)

### LINKED ISSUE #111: Test t50_106 fails when run from an interactive terminal

powderluv:
## Summary
Fixes #111 — replaces the broken test with one that exercises the real `RawModeGuard` code.

## Problem
Old test simulated `isatty()` with raw libc calls instead of calling `RawModeGuard::enter()`. It depended on stdin being a pipe, so it failed on interactive terminals.

## Fix
- `RawModeGuard::enter_on_fd(fd)` — new `pub(crate)` method that takes an explicit fd
- Test creates a pipe (always non-TTY), passes it to `enter_on_fd()`, asserts `Err("not a TTY")`
- Works identically in CI, interactive terminal, and IDE

Per CLAUDE.md:
> Do not write tests that simulate a unit's behavior instead of calling the actual unit.
> Do not write tests that depend on the test runner's environment.

### DIFF
--- a/crates/spur-cli/src/sattach.rs
+++ b/crates/spur-cli/src/sattach.rs
@@ -219,20 +219,27 @@ async fn interactive_attach(
 /// - Keystrokes are forwarded immediately (no buffering until Enter)
 /// - Special keys (Ctrl-C, Ctrl-Z) are sent as bytes instead of signals
 /// - No local echo (the remote shell handles echo)
-struct RawModeGuard {
+/// RAII guard that sets the terminal to raw mode and restores it on drop.
+pub(crate) struct RawModeGuard {
+    fd: i32,
     original: libc::termios,
 }
 
 impl RawModeGuard {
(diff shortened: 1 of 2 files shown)


shiv-tyagi (approved):
LGTM. Thanks @powderluv.

PR #151, opened 2026-04-30
fix(spurd): non-root container uses container's dynamic loader

### LINKED ISSUE #149: Container jobs submitted by non-root users fail with glibc version mismatch
## Summary:
When a non-root user submits a job with --container-image on a host where
the container's glibc is newer than the host's glibc (e.g., container on
Ubuntu 24 / glibc 2.38, host on Ubuntu 22 / glibc 2.35), the job fails
immediately with:

  bash: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found

The same job submitted as root succeeds. The same job run via docker by
a non-root user also succeeds.

powderluv:
## Summary

Closes #149.

When a non-root user submits a job with ``--container-image``, the original launch script did:

\`\`\`bash
/bin/bash \$ROOTFS/tmp/spur_job_X.sh
\`\`\`

This runs the **host's** ``/bin/bash``. Subsequent invocations of binaries from ``\$ROOTFS`` cause the kernel to honor each binary's ``PT_INTERP`` (e.g. ``/lib64/ld-linux-x86-64.so.2``), which resolves against the **host** filesystem since we never chrooted — so the host's loader loads the host's libc.

When the container's libc is newer than the host's (e.g., Ubuntu 24 container on Ubuntu 22 host), bash fails with:

\`\`\`
bash: /lib/x86_64-linux-gnu/libc.so.6: version 'GLIBC_2.38' not found
\`\`\`

The same job as root works because the root branch chroots into ``\$ROOTFS`` and the kernel transparently uses the container's loader.

## Fix

Detect the container's dynamic loader (glibc x86_64 paths first, then musl Alpine fallback) and invoke it explicitly with ``--library-path`` pointing at the container's libdirs. This bypasses the binary's ``PT_INTERP`` because we run the loader directly.

\`\`\`bash
exec \"\$SPUR_LOADER\" --library-path \"\$LD_LIBRARY_PATH\" \\
  \"\$ROOTFS/bin/bash\" \"\$ROOTFS/tmp/spur_job_\$JOB_ID.sh\"
\`\`\`

Also extends ``LD_LIBRARY_PATH`` to include ``/lib/x86_64-linux-gnu`` and ``/usr/lib/x86_64-linux-gnu`` (Debian/Ubuntu's libc location), which were missing from the original list.

Falls back to legacy ``/bin/bash`` invocation when no loader is found (static / scratch / non-glibc-non-musl images) — preserves prior behavior for the niche case.

## Why not just chroot or unshare?

Non-root chroot needs ``CAP_SYS_CHROOT``. ``unshare --user --mount --map-root-user`` would give us full container isolation but requires ``kernel.unprivileged_userns_clone=1`` which some hardened kernels disable. The explicit-loader approach works on every kernel and is sufficient for the reported bug.

A follow-up could add an opt-in user-namespace path when available; that's a separate, larger change.

## Tests

4 new tests in ``container::tests``:

- ``launch_script_non_root_invokes_container_loader_explicitly`` — verifies the loader-detection probe list and the explicit invocation pattern (\`\$SPUR_LOADER --library-path ...\`)
- ``launch_script_non_root_includes_debian_libdir_in_ld_path``
- ``launch_script_non_root_falls_back_to_host_bash_when_no_loader``
- ``launch_script_non_root_no_longer_uses_host_bash_unconditionally`` — regression guard

**Regression-catching verified**: with the fix bypassed (``if false && ...``), test 1 fails as expected.

## Test plan

- [x] ``cargo test -p spurd container::tests`` — all 49 pass
- [x] ``cargo fmt --check``

Closes #149

### DIFF
--- a/crates/spurd/src/container.rs
+++ b/crates/spurd/src/container.rs
@@ -665,15 +665,47 @@ done
 chroot $ROOTFS /bin/bash -c "cd {workdir} && {entrypoint}{script}"
 '
 else
-  # Non-root fallback: PATH-based execution, no namespace isolation
+  # Non-root path: no chroot/pivot_root capability. We can't run binaries
+  # from $ROOTFS naively because the kernel honors the binary's PT_INTERP
+  # (e.g. /lib64/ld-linux-x86-64.so.2) which resolves to the *host's*
+  # dynamic loader, which then loads the *host's* libc — producing a
+  # GLIBC version mismatch when the container's binaries are newer than
+  # the host's libc (#149).
+  #
+  # Fix: invoke the container's dynamic loader explicitly with
+  # --library-path so binaries see the container's libc + dependent libs.
   ROOTFS="{rootfs}"
   {env_exports}
   export PATH="$ROOTFS/usr/bin:$ROOTFS/bin:$ROOTFS/usr/sbin:$ROOTFS/sbin:$PATH"
-  export LD_LIBRARY_PATH="$ROOTFS/usr/lib:$ROOTFS/lib:$ROOTFS/usr/lib64:$ROOTFS/lib64:${{LD_LIBRARY_PATH:-}}"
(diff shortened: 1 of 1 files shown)


shiv-tyagi:
This is related to #152 and proper filesystem isolation in general.

PR #152, opened 2026-04-30
fix(spurd): GPU isolation in containers — only mount assigned renderD*

### LINKED ISSUE #148: GPU isolation not enforced — jobs can access all GPUs on the node
## Summary:
When a job requests a subset of GPUs (e.g., --gres=gpu:2 on an 8-GPU node),
all 8 GPUs remain accessible inside the job. rocm-smi shows every GPU, and
there is nothing preventing the job from using devices it was not allocated.

### Current behavior:
spurd sets no GPU visibility restrictions before launching the job. 
Setting ROCR_VISIBLE_DEVICES alone is insufficient — it can be unset or
overridden inside the job and provides no kernel-level enforcement

### Expected behavior:
Only allocated GPUs should be accessible.

### Proper fix:
Use cgroups v2 devices controller to whitelist only the assigned GPU device
files (/dev/dri/renderDXXX, /dev/kfd) for the job's process hierarchy.
This is how Slurm's task/cgroup plugin enforces GPU isolation. Env vars
(ROCR_VISIBLE_DEVICES, HIP_VISIBLE_DEVICES) should be set in addition as
a convenience hint for ROCm/CUDA runtimes, but cgroups is the enforcement
boundary.

powderluv:
## Summary

Closes #148 (partially — full kernel-level enforcement is a larger follow-up).

Two bugs:

1. ``agent_server.rs`` hard-coded ``gpu_devices: vec![]`` when building ``ContainerConfig`` (a ``// TODO: from GRES allocation`` left behind). Even when ``--gres=gpu:N`` was requested, the container received an empty gpu_devices list.

2. ``container.rs`` bulk bind-mounted **all** of ``/dev/dri`` into the container regardless of ``gpu_devices``, so even with the field populated, jobs could see every GPU on the host.

## Fix

**agent_server.rs**: move GPU/CPU allocation up so it runs *before* ``container_config`` is built. ``ContainerConfig.gpu_devices`` now reflects the actual GRES allocation.

**container.rs**: extract ``build_gpu_dri_mount_block()``. When ``gpu_devices`` is non-empty, emit a per-device bind mount for ``renderD(128 + id)``; otherwise emit nothing (CPU-only jobs don't touch ``/dev/dri``). The old bulk ``mount --bind /dev/dri $ROOTFS/dev/dri`` is gone.

Generated bash for ``--gres=gpu:2`` getting devices ``[0, 2]``:

\`\`\`bash
if [ -d /dev/dri ]; then
  mkdir -p \$ROOTFS/dev/dri
  if [ -e /dev/dri/renderD128 ]; then
    touch \$ROOTFS/dev/dri/renderD128 2>/dev/null
    mount --bind /dev/dri/renderD128 \$ROOTFS/dev/dri/renderD128 2>/dev/null || true
  fi
  if [ -e /dev/dri/renderD130 ]; then
    touch \$ROOTFS/dev/dri/renderD130 2>/dev/null
    mount --bind /dev/dri/renderD130 \$ROOTFS/dev/dri/renderD130 2>/dev/null || true
  fi
fi
\`\`\`

## Limitations / follow-up

This is **filesystem-level** isolation:

- ``/sys/class/drm`` is still visible — introspection tools like ``rocm-smi`` can still enumerate all GPUs on the node (they just can't open the unassigned device files).
- A privileged process inside the container with ``CAP_MKNOD`` could mknod its own renderD nodes given the major/minor numbers.

The original issue calls for a cgroup v2 BPF device controller for kernel-level enforcement. That's a meaningful, separate change (BPF program, attach via bpf() syscall) and is best as its own PR. I'll file a follow-up issue.

## Tests

5 new tests in ``container::tests``:

- ``gpu_dri_block_empty_when_no_gpus_allocated``
- ``gpu_dri_block_mounts_only_assigned_render_nodes`` — **#148 repro core**: gpu_devices=[0, 2] only mounts renderD128 + renderD130, not renderD129/131
- ``gpu_dri_block_uses_per_device_bind_mounts``
- ``launch_script_with_gpus_does_not_bulk_mount_dev_dri`` — end-to-end via the real launch script
- ``launch_script_no_gpus_skips_dri_mount_entirely`` — CPU-only safety

**Regression-catching verified**: with ``build_gpu_dri_mount_block`` reverted to the buggy bulk-mount, 4 of 5 tests fail with explicit \"regression of #148\" errors.

## Test plan

- [x] ``cargo test -p spurd container::tests`` — all 50 pass
- [x] ``cargo build -p spurd``
- [x] ``cargo fmt --check``

Closes #148

### DIFF
--- a/crates/spurd/src/agent_server.rs
+++ b/crates/spurd/src/agent_server.rs
@@ -365,6 +365,44 @@ impl SlurmAgent for AgentService {
             env.insert("SPUR_NODE_RANK".into(), node_rank.to_string());
         }
 
+        // Allocate GPU devices from the node's pool. Done up-front (before
+        // container setup) so the container path can selectively bind-mount
+        // only the assigned /dev/dri/renderD* devices (#148).
+        let mut gpu_count = 0u32;
+        let mut gpu_type: Option<String> = None;
+        for gres in &spec.gres {
+            if let Some((name, gtype, count)) = spur_core::resource::parse_gres(gres) {
+                if name == "gpu" {
+                    gpu_count += count;
+                    if let Some(t) = gtype {
+                        gpu_type = Some(t);
+                    }
+                }
+            }
+        }
+
+        let alloc_result = if gpu_count > 0 || spec.cpus_per_task > 0 {
+            let mut alloc = self.allocation.lock().await;
+            alloc.try_allocate(
+                spec.cpus_per_task.max(1),
+                spec.memory_per_node_mb,
+                gpu_count,
+                gpu_type.as_deref(),
+            )
+        } else {
+            None
+        };
+
+        let gpu_devices: Vec<u32> = alloc_result
+            .as_ref()
+            .map(|a| a.gpu_ids.clone())
+            .unwrap_or_default();
+
+        let cpu_ids: Vec<u32> = alloc_result
+            .as_ref()
+            .map(|a| a.cpu_ids.clone())
+            .unwrap_or_default();
+
         // If container image is specified, wrap the job in a container
         let (launch_script, rootfs_mode) = if !spec.container_image.is_empty() {
             info!(job_id, image = %spec.container_image, "launching containerized job");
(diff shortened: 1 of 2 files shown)


shiv-tyagi:
I am working on proper fs isolation part. Will raise a PR tomorrow.

shiv-tyagi:
I don’t have my laptop with me, so I can’t review this right now. Can we
hold off on merging until tomorrow if it’s not urgent?

powderluv:
will mark as draft

PR #154, opened 2026-05-01
Harden native container runtime with direct syscalls and integration tests

shiv-tyagi:
### Problem

The container runtime generated bash scripts at launch time — string-templating shell code for namespace setup, mounts, chroot, device passthrough, DNS, and privilege dropping. This was:

- **Hard to test**: every "test" was string inspection (`assert!(script.contains("mount --bind ..."))`) — verifying bash syntax, not behavior. Real bugs like #149 (glibc mismatch) and #148 (GPU isolation) slipped through because the generated script ran correctly in the author's environment but failed in others.
- **Fragile**: fixes were localized patches to string templates (#151 patched the dynamic loader path, #152 moved GPU allocation earlier to fix the mount block). Each fix addressed one symptom without solving the root cause — that bash string generation is inherently untestable and error-prone.
- **Incomplete isolation**: the non-root fallback ran container binaries via the host's PATH and dynamic loader, making glibc version mismatches inevitable (#149). The root path used chroot instead of pivot_root, leaving the old root accessible.

### Solution

Implement the container runtime as native Rust syscalls: `unshare`, `mount`, `pivot_root`, `execve`, `waitpid`, `pidfd_open`. Container setup now runs as compiled code in a forked child process — no shell involved, no string templating, no host binary dependencies inside the container.

This structurally eliminates the classes of bugs that #148, #149, #151, and #152 addressed:
- **#149 / #151** (glibc mismatch): gone — `pivot_root` makes the container's rootfs the real `/`, so its own loader and libc are always used. There is no PATH-based fallback.
- **#148 / #152** (GPU isolation): gone — `mount_dri_devices()` selectively bind-mounts only allocated renderD/card nodes via `nix::mount`, not a bash block that bulk-mounts `/dev/dri`.

### What changed

**container.rs** — native container init: `unshare` + `fork_into_pid_namespace` + `mount_filesystems` (proc/dev/shm/pts/sysfs) + `mount_hw_devices` (selective GPU/IB) + `mount_dns` (loopback stripping) + `run_hooks` ($SPUR_HOOK_ENVIRON) + `pivot_into_rootfs` + `drop_privileges`. Rootless path uses `CLONE_NEWUSER` with clear hard-error when unavailable. `close_inherited_fds` panic fix. Root and rootless paths share `set_mount_propagation_private()` and `fork_into_pid_namespace()` helpers.

**executor.rs** — `RunningJob` enum (`Managed` for tokio child, `Forked` for container pid+pidfd). `pidfd_open` prevents PID recycling. `kill_process_tree` recursively signals container process subtrees. `launch_container_job` does fork → container_init → execve directly.

**agent_server.rs** — `TrackedJob` simplified, monitoring via `try_wait`, cancel via `kill_signal` dispatching to correct PID variant. `gpu_devices` populated from GRES allocation.

**server.rs + scheduler_loop.rs** — `cancel_job` now sends `CancelJob` RPC to agents (was only changing WAL state, leaving container processes alive after scancel).

**cluster_test.sh** — 8 container integration tests (C1–C8): launch, exit codes, cancellation + orphan detection, DNS resolution via `getent`, /dev/shm write + size, PID namespace, env passthrough, bind mounts with readonly enforcement. All stateless (PID-suffixed paths, trap cleanup).

### Testing

- 74 unit tests pass (`cargo test -p spurd`)
- 17/17 container integration tests pass on bare-metal cluster
- Tested manually: rootless with AppArmor profile, fake GPU device injection, multi-node dispatch

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,7 +60,7 @@ k8s-openapi = { version = "0.23", features = ["latest"] }
 schemars = "0.8"
 
 # Unix
-nix = { version = "0.29", features = ["process", "signal", "fs", "user"] }
+nix = { version = "0.29", features = ["process", "signal", "fs", "user", "sched", "mount"] }
 
 # Auth
 jsonwebtoken = "9"
--- a/crates/spurctld/src/scheduler_loop.rs
+++ b/crates/spurctld/src/scheduler_loop.rs
@@ -835,7 +835,7 @@ async fn manage_power(cluster: Arc<ClusterManager>, raft: Arc<RaftHandle>) {
 }
 
 /// Send CancelJob RPC to all agents for a job with a specific signal.
-async fn send_cancel_to_agents(
+pub async fn send_cancel_to_agents(
     cluster: &Arc<ClusterManager>,
     job: &spur_core::job::Job,
     signal: i32,
--- a/crates/spurctld/src/server.rs
+++ b/crates/spurctld/src/server.rs
@@ -220,9 +220,23 @@ impl SlurmController for ControllerService {
         }
 
         let req = request.into_inner();
+        let job_id = req.job_id;
+
+        // Snapshot the job before cancelling so we have allocated_nodes
+        let job = self.cluster.get_job(job_id);
+
         self.cluster
-            .cancel_job(req.job_id, &req.user)
+            .cancel_job(job_id, &req.user)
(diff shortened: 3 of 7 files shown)


shiv-tyagi:
@powderluv this is good to be merged.

We can also close #152 as it is already covered in this work.

powderluv (approved):

PR #158, opened 2026-05-06
fix(spur-tests): remove unused imports and variables

shiv-tyagi:
Drop unused HashMap, Arc, JobId, and partition imports from the test harness and scheduler tests. Mark job binding as immutable in GPU test.


shiv-tyagi:
Reviewed myself and with copilot. Looking good. Merging.

PR #164, opened 2026-05-13
fix(spurd): improve container init error reporting

shiv-tyagi:
Use {:#} instead of {} when formatting the anyhow error from container_init so the full error chain (including the actual syscall failure) is propagated through the sync pipe. Previously only the outermost context was shown, hiding the root cause.

Also replace the misleading context message that pointed at a specific sysctl as the fix — the failure could be any of unshare, uid_map, setgroups, or gid_map.


shiv-tyagi:
This PR does its job. Will work on the error reported on CI failures on subsequent runs and find the root cause. Merging this.

PR #166, opened 2026-05-14
refactor(spur-net): use base64 crate for registry auth encoding

sgopinath1:
Replace hand-written Base64 in OCI registry credential paths with the standard base64 0.22 engine (RFC 4648 STANDARD). Docker config.json auth decoding and non-Docker-Hub Basic auth now use strict decode/encode.

Add decode_registry_auth_b64 helper and unit tests for valid input, whitespace trim, invalid or truncated blobs, non-standard alphabet, and encode/decode roundtrip.

## Motivation
spur-net implemented Docker config.json auth decoding and non–Docker Hub registry Basic auth using hand-written Base64 routines. That duplicated well-tested library behavior, was harder to audit, and accepted malformed input more leniently than a standards-based decoder. This change routes those paths through the maintained base64 crate so behavior matches common registry tooling and is easier to keep correct.

## Technical Details

- Added base64 0.22 as a workspace dependency and a direct dependency of spur-net ([Cargo.toml](vscode-file://vscode-app/c:/Users/sgopinath/AppData/Local/Programs/cursor/resources/app/out/vs/code/electron-sandbox/workbench/Cargo.toml), [crates/spur-net/Cargo.toml](vscode-file://vscode-app/c:/Users/sgopinath/AppData/Local/Programs/cursor/resources/app/out/vs/code/electron-sandbox/workbench/crates/spur-net/Cargo.toml)); [Cargo.lock](vscode-file://vscode-app/c:/Users/sgopinath/AppData/Local/Programs/cursor/resources/app/out/vs/code/electron-sandbox/workbench/Cargo.lock) updated accordingly.
- In [crates/spur-net/src/oci.rs](vscode-file://vscode-app/c:/Users/sgopinath/AppData/Local/Programs/cursor/resources/app/out/vs/code/electron-sandbox/workbench/crates/spur-net/src/oci.rs):
    - Introduced decode_registry_auth_b64, which decodes trimmed input with general_purpose::STANDARD and returns None on decode or non–UTF-8 failure. load_docker_config_auth uses it instead of the removed custom decoder.
    - Non–Docker Hub creds in get_auth_token now use STANDARD.encode on the user:password string for the Basic … payload (replacing the custom encoder).
- Behavior change: invalid characters, bad padding, or non-standard alphabet (for example URL-safe - / _ under the standard engine) now fail decode and surface as missing credentials, instead of being partially ignored by the old implementation.

## Test Plan
- cargo test -p spur-net (covers new oci unit tests: valid blob, trim, invalid/truncated input, non-standard alphabet, encode/decode roundtrip).
- cargo test (full workspace; requires protoc on the host, same as CI’s protobuf-compiler step).
- cargo fmt --all --check
- cargo clippy --workspace --exclude spur-ffi --all-targets -- -W clippy::all -D unused (CI-aligned; note spur-net may still emit pre-existing Clippy warnings unrelated to this change if the job is run with stricter flags locally).

## Scope / Out of Scope
- In scope: OCI/registry credential Base64 only in spur-net oci.rs.
- Out of scope: Docker Hub token request path (reqwest::basic_auth), URL-safe Base64 support, and any change to image pull logic beyond auth string encoding/decoding.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3302,6 +3302,7 @@ name = "spur-net"
 version = "0.3.0"
 dependencies = [
  "anyhow",
+ "base64",
  "bytes",
  "flate2",
  "futures-util",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,6 +86,7 @@ flate2 = "1"
 tar = "0.4"
 semver = "1"
 sha2 = "0.10"
+base64 = "0.22"
 
 # Internal
 spur-update = { path = "crates/spur-update" }
--- a/crates/spur-net/Cargo.toml
+++ b/crates/spur-net/Cargo.toml
@@ -15,3 +15,4 @@ futures-util = { workspace = true }
 flate2 = { workspace = true }
 tar = { workspace = true }
 bytes = { workspace = true }
+base64 = { workspace = true }
--- a/crates/spur-net/src/oci.rs
+++ b/crates/spur-net/src/oci.rs
@@ -15,6 +15,7 @@
 use std::path::{Path, PathBuf};
 
 use anyhow::{bail, Context};
+use base64::{engine::general_purpose::STANDARD, Engine as _};
 use flate2::read::GzDecoder;
 use reqwest::header::{ACCEPT, AUTHORIZATION};
 use serde::Deserialize;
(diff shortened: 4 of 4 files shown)


shiv-tyagi (approved):
LGTM. Thanks @sgopinath1.

powderluv (approved):

PR #169, opened 2026-05-14
Revert "ci: unpin container tests from main node"

shiv-tyagi:
Reverts ROCm/spur#168

The issue seems to be something else. Needs further investigation.


shiv-tyagi:
The runner process was using previous GID, had to be restarted. Looking good for now. Will reopen if the issue persists.

PR #189, opened 2026-05-19
perf(spur-tests): skip GPU venv provisioning when pre-existing venv available

shiv-tyagi:
Add SPUR_TEST_BM_GPU_VENV env var to point at a pre-provisioned venv on CI nodes, eliminating ~5 min/node of torch installation on every run. When unset, falls back to provisioning an ephemeral venv with torch in parallel across all nodes (previously sequential).


shiv-tyagi:
3x performance improvement

sgopinath1 (approved):

PR #202, opened 2026-05-21
chore(clippy): suppress result_large_err on 3 functions in spurctld

shiv-tyagi:
## Summary

Add per-function `#[allow(clippy::result_large_err)]` on 3 functions in `spurctld` where the error type exceeds clippy's default 128-byte threshold due to upstream dependencies (`tonic::Status` at 176 bytes, openraft's `RaftError`). These annotations can be removed when we upgrade to tonic 0.13+ which boxes `Status` internals (hyperium/tonic#2282).

## Affected functions

- `server.rs`: `check_leader`, `proto_to_job_spec` — `tonic::Status` (176 bytes)
- `cluster.rs`: `propose` — openraft `RaftError` intermediate


sgopinath1 (approved):
LGTM

PR #208, opened 2026-05-21
ci(clippy): enable all workspace lints and deny warnings in CI

shiv-tyagi:
Uncomment the 5 deferred clippy lints (should_implement_trait, too_many_arguments, large_enum_variant, result_large_err, type_complexity) now that all violations are fixed. Add global RUSTFLAGS="-D warnings" to the CI workflow so all warnings (rustc + clippy) are hard errors in CI while remaining non-blocking locally. Simplify the clippy command by dropping the now-redundant -- -W clippy::all -D unused flags.


shiv-tyagi:
We need to pin rust version. #209 should be merged first. We need to rebase after that.

sgopinath1 (approved):

PR #212, opened 2026-05-22
feat(spur-metrics): add job metrics collection in spurctld

sgopinath1:
### Summary
Add job-level metrics collection for spurctld. Introduces the spur-metrics crate and wires ClusterManager::job_metrics() to aggregate from the authoritative in-memory job map. Metrics are computed lazily on read (no separate cache, no WAL refresh hooks). OpenMetrics HTTP export (:6822) is follow-up work (14.1e).

### What changed
- crates/spur-metrics: JobMetricsSnapshot and JobMetricsSnapshot::collect() — counts by JobState, held pending jobs, and running/completing allocation sums (CPUs, memory, GPUs). Job-state buckets use proto wire indices via JobState::from_proto_i32.
- crates/spurctld: ClusterManager::job_metrics() scans jobs under a read lock on each call. Temporary log_job_metrics_debug() in the leader health loop (removed when export lands).
- Snapshot restore: log job/node counts after dropping write guards (avoids deadlock under info!).

### Test plan
- cargo test -p spur-metrics
- cargo test -p spurctld job_metrics
- cargo test -p spurctld snapshot_and_restore
- cargo fmt --check
- cargo clippy --workspace --exclude spur-ffi --all-targets --locked -- -D unused

### DIFF
--- a/crates/spur-metrics/src/job.rs
+++ b/crates/spur-metrics/src/job.rs
@@ -1,0 +1,168 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use spur_core::job::{Job, JobState, PendingReason};
+
+/// Number of [`JobState`] variants (index for `by_state`).
+pub const JOB_STATE_COUNT: usize = JobState::COUNT;
+
+/// Aggregated job metrics derived from the controller job map.
+///
+/// Built by scanning in-memory `Job` records (lazy, on read). Durable state lives
+/// in the Raft-backed job map on `ClusterManager`.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct JobMetricsSnapshot {
+    /// Total jobs in the controller map (includes terminal jobs).
+    pub total: u64,
+    /// Count per [`JobState`]; index via [`job_state_index`].
+    pub by_state: [u64; JOB_STATE_COUNT],
+    /// Pending jobs with `pending_reason == Held`.
+    pub held_pending: u64,
+    /// Sum of allocated CPUs for jobs in Running or Completing.
+    pub running_cpus: u64,
+    /// Sum of allocated memory (bytes) for Running or Completing.
+    pub running_memory_bytes: u64,
+    /// Sum of allocated GPU count for Running or Completing.
+    pub running_gpus: u64,
+}
+
+/// Map [`JobState`] to a stable index in [`JobMetricsSnapshot::by_state`]
+/// (proto wire discriminant via [`JobState::from_proto_i32`]).
+pub fn job_state_index(state: JobState) -> usize {
+    let wire = state.to_proto_i32();
+    debug_assert_eq!(JobState::from_proto_i32(wire), Some(state));
+    wire as usize
+}
+
+fn counts_toward_running_alloc(state: JobState) -> bool {
+    matches!(state, JobState::Running | JobState::Completing)
+}
+
+impl JobMetricsSnapshot {
+    /// Rebuild metrics by scanning all jobs.
+    pub fn collect<'a>(jobs: impl IntoIterator<Item = &'a Job>) -> Self {
+        let mut snap = Self::default();
+        for job in jobs {
+            snap.total += 1;
+            snap.by_state[job_state_index(job.state)] += 1;
+
+            if job.state == JobState::Pending && job.pending_reason == PendingReason::Held {
+                snap.held_pending += 1;
+            }
+
+            if counts_toward_running_alloc(job.state) {
+                if let Some(ref alloc) = job.allocated_resources {
+                    snap.running_cpus += u64::from(alloc.cpus);
+                    snap.running_memory_bytes += alloc.memory_mb.saturating_mul(1024 * 1024);
+                    snap.running_gpus += u64::from(alloc.total_gpus());
+                }
+            }
+        }
+        snap
+    }
+
+    /// Count for a single state.
+    pub fn count_state(&self, state: JobState) -> u64 {
+        self.by_state[job_state_index(state)]
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use chrono::Utc;
+    use spur_core::job::{Job, JobSpec, JobState, PendingReason};
+    use spur_core::resource::{GpuLinkType, GpuResource, ResourceSet};
+
+    use super::*;
+
+    fn job_with_state(id: u32, state: JobState, held: bool) -> Job {
+        let mut job = Job::new(id, JobSpec::default());
+        job.state = state;
+        if held {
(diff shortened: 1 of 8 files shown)


plans/phase-14.1-prometheus-metrics.md
```diff
@@ -0,0 +1,277 @@
+# Phase 14.1: Prometheus / OpenMetrics Metrics
```
shiv-tyagi: We don't need to commit this maybe.

crates/spur-metrics/src/job.rs
```diff
@@ -0,0 +1,167 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use spur_core::job::{Job, JobState, PendingReason};
+
+/// Number of [`JobState`] variants (index for `by_state`).
+pub const JOB_STATE_COUNT: usize = 10;
```
shiv-tyagi: this will be fragile and less maintainable.

crates/spur-metrics/src/job.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// Pending jobs with `pending_reason == Held`.
+    pub held_pending: u64,
+    /// Sum of allocated CPUs for jobs in Running or Completing.
+    pub running_cpus: u64,
+    /// Sum of allocated memory (bytes) for Running or Completing.
+    pub running_memory_bytes: u64,
+    /// Sum of allocated GPU count for Running or Completing.
+    pub running_gpus: u64,
+}
+
```
shiv-tyagi, line 28: Maybe we should move this out of here. I see at many places there we do similar mappings. I should find an alternative.

crates/spur-metrics/src/job.rs
```diff
@@ -0,0 +1,167 @@
...
+            }
+        }
+        snap
+    }
+
+    /// Count for a single state.
+    pub fn count_state(&self, state: JobState) -> u64 {
+        self.by_state[job_state_index(state)]
+    }
+}
```
shiv-tyagi, line 68: Should we just compute on scrape instead? The job map is already in memory, so a read-lock scan should be trivial. It also gives frequency control to the scraper rather than re-scanning on every WAL apply. We can introduce caching if we ever see performance issues.

shiv-tyagi (changes requested):
Some comments. PTAL.

Can we wire up a minimal /metrics endpoint with a couple of real Prometheus gauges (e.g. spur_jobs_total, spur_jobs_by_state) in this PR itself? A thin end-to-end slice would let us validate design choices (like compute-on-scrape vs caching) now rather than in a follow-up. We can add more metrics incrementally from there.

crates/spur-metrics/src/job.rs
```diff
@@ -0,0 +1,167 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use spur_core::job::{Job, JobState, PendingReason};
+
+/// Number of [`JobState`] variants (index for `by_state`).
+pub const JOB_STATE_COUNT: usize = 10;
+
```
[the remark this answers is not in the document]
shiv-tyagi: As said in one of my comments.

crates/spurctld/src/cluster.rs
```diff
@@ -1607,9 +1630,11 @@ impl StateMachineApply for ClusterManager {
...
 
             self.next_job_id.store(next_id, Ordering::Relaxed);
+            drop(jobs);
+            self.refresh_job_metrics();
             info!(
-                jobs = jobs.len(),
-                nodes = nodes.len(),
+                jobs = self.jobs.read().len(),
+                nodes = self.nodes.read().len(),
                 "restored cluster state from Raft snapshot"
```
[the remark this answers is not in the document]
shiv-tyagi: I think it will be better to not couple with WAL.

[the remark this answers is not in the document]
sgopinath1: WAL is required so that we get consistent metrics across process restarts and system reboots.

crates/spurctld/src/cluster.rs
```diff
@@ -1285,6 +1302,11 @@ impl ClusterManager {
     /// Apply a WalOperation to in-memory state.
     /// Called by Raft's `apply_to_state_machine` on commit.
     fn apply_operation(&self, op: &WalOperation) {
+        self.apply_operation_inner(op);
+        self.refresh_job_metrics();
+    }
+
```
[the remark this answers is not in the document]
sgopinath1: This will not be an issue when we support node metrics soon.

crates/spurctld/src/cluster.rs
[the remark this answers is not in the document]
sgopinath1: Removed cache and WAL.

crates/spur-metrics/src/job.rs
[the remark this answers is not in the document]
sgopinath1: Fixed.

plans/phase-14.1-prometheus-metrics.md
sgopinath1: Removed.

crates/spur-metrics/src/job.rs
sgopinath1: Fixed.

crates/spur-metrics/src/job.rs
sgopinath1: Fixed.

crates/spur-metrics/src/job.rs
sgopinath1: Yes, changed the design.

shiv-tyagi (commented):
Looking good mostly

crates/spurctld/src/cluster.rs
```diff
@@ -1606,10 +1636,15 @@ impl StateMachineApply for ClusterManager {
             *self.license_pool.write() = snap.license_pool;
             *self.hostname_aliases.write() = snap.hostname_aliases;
 
+            let jobs_len = jobs.len();
+            let nodes_len = nodes.len();
             self.next_job_id.store(next_id, Ordering::Relaxed);
+            drop(jobs);
+            drop(nodes);
+            drop(steps);
```
shiv-tyagi: Are these required now as a part of this PR?

crates/spur-metrics/src/job.rs
shiv-tyagi: Thanks

crates/spurctld/src/cluster.rs
sgopinath1: No, removed.

shiv-tyagi: thanls

shiv-tyagi (approved):
LGTM

PR #216, opened 2026-05-23
fix(spur-metrics): inherit publish=false for cargo-deny wildcard check

shiv-tyagi:
## Summary

`spur-metrics` was missing `publish.workspace = true`, so cargo-deny treated it as a publishable crate and failed on the unversioned path dependency on `spur-core`. Inheriting the workspace `publish = false` setting matches every other internal crate and fixes the CI `deny` job.

## Test plan

- [x] `cargo deny check`


sgopinath1 (approved):
LGTM

PR #249, opened 2026-06-03
fix(ci): use recursive glob for Dockerfile pathspecs in license checks

shiv-tyagi:
Use `**/Dockerfile*` instead of `*/Dockerfile` + `*/Dockerfile.*` so license header checks match Dockerfiles at any directory depth, not just one level deep.


sgopinath1 (approved):
LGTM

PR #256, opened 2026-06-05
feat(deploy): add Helm chart for Kubernetes deployment

### LINKED ISSUE #255: Package Spur as a Helm chart for K8s distribution
## Problem

Spur currently ships raw Kubernetes manifests under `examples/k8s/` (namespace, configmap, rbac, spurctld, spurd, spurrestd, spurdbd, operator, pdb, spurjob-crd). To deploy, users must:

1. Apply the manifests in the correct order
2. Hand-edit YAML to change replica counts, image refs, ports, the embedded postgres sidecar, ingress, etc.
3. Maintain a fork of the manifest set per cluster / environment

This is fine for the example walkthrough in `docs/deployment/kubernetes.rst`, but it doesn't scale to external distribution. There's no install-order automation, no parameterization, no upgrade/rollback story, no discoverability (Artifact Hub / OCI registries / `helm search`), and no convention for CRD lifecycle (the classic \"uninstall wipes the CRD and every CR with it\" footgun).

## Proposal

Ship a first-party Helm chart at `deploy/helm/spur/` that packages:

- **spurctld** — StatefulSet + headless Service, replica count drives the Raft peer list in `spur.conf`, opt-in PVC, PDB
- **spurd** — DaemonSet, ROCm `/dev/kfd` + `/dev/dri` toggle, configurable node selector / tolerations
- **spurrestd** — Deployment + Service + optional Ingress
- **spurdbd** — Deployment with two modes: external Postgres (recommended) or dev-mode sidecar
- **spur-k8s-operator** — Deployment, ClusterRole/Binding for SpurJob reconciliation
- **SpurJob CRD** — installed via `crds/` so Helm leaves it alone on upgrade/uninstall (avoids wiping user CRs)

### Key requirements

- Single `helm install` deploys everything in the right order
- Re-running `helm upgrade` is safe; config changes roll the controller via `checksum/config` annotation
- Every reasonable knob (replicas, image, resources, storage class, ingress, external DB) lives in `values.yaml`
- Free-form `config.extraToml` escape hatch so users don't need to fork the chart for a `spur.conf` field we don't model yet
- CI lints + templates + kubeconform-validates the chart on every PR touching it

### Why now

External users have started asking for K8s deployment guidance. The raw-manifest path forces them to learn the install order and edit YAML; a chart makes \"try Spur on K8s\" a one-liner and gives us a clean distribution channel via OCI / Artifact Hub later.

### Non-goals (for the first cut)

- Publishing the chart to a public registry (separate follow-up — needs `release.yml` changes and a CI image build)
- Rewriting `examples/k8s/` (keep as-is; chart is additive — the manifests are useful as a customization reference)
- Operator-only install pattern (the chart installs the operator alongside the daemons; a future iteration could split into umbrella + sub-charts)

## Implementation notes

The chart should mirror `examples/k8s/` so anyone reading the manifests can map them 1:1 to template files. Default `image.repository` will need to point at wherever the project publishes its image; today that's not set, so a default of `ghcr.io/rocm/spur` + `appVersion` tag is a reasonable placeholder until the publish workflow lands.

yansun1996:
Closes #255

## Summary

- New Helm chart at `deploy/helm/spur/` packaging spurctld (HA StatefulSet), spurd (DaemonSet), spurrestd, spurdbd, and the operator into a single `helm install`.
- `SpurJob` CRD lives under `crds/` so Helm installs it on first install and leaves it alone on upgrade/uninstall (avoids wiping user CRs).
- CI workflow (`.github/workflows/helm.yml`) runs `helm lint` + `helm template` across four scenarios (defaults / single-node dev / external DB / ingress) and validates the rendered output with `kubeconform --strict` against k8s 1.29.

## What the chart covers

| Component | Kind | Notes |
|---|---|---|
| `spurctld` | StatefulSet + headless Service + PDB | Raft peer list in `spur.conf` is generated from `controller.replicaCount` |
| `spurd` | DaemonSet | ROCm `/dev/kfd` + `/dev/dri` mounts toggleable via `agent.gpu.rocm` |
| `spurrestd` | Deployment + Service + optional Ingress | |
| `spurdbd` | Deployment | External Postgres via secret or dev-mode embedded sidecar |
| `spur-k8s-operator` | Deployment + ClusterRole/Binding | Reconciles `SpurJob` CRs |
| `SpurJob` CRD | `crds/spurjob-crd.yaml` | Helm never removes (by design) |

Templates mirror `examples/k8s/` 1:1 so the raw manifests remain a useful customization reference. `config.extraToml` escape hatch lets users append arbitrary `spur.conf` fields without forking the chart. Config changes roll the controller via `checksum/config` annotation. Post-install `NOTES.txt` prints the connect commands plus actionable `helm upgrade --reuse-values` remediation snippets when dev-mode defaults are in use (embedded postgres without PVC, controller `emptyDir`).

## Test plan

- [x] `helm lint deploy/helm/spur` passes
- [x] `helm template` renders cleanly for: defaults (3-replica Raft HA), single-node dev (1 replica, no PVC, no GPU, no operator), external Postgres via secret, REST API ingress
- [x] CI job runs all four scenarios + kubeconform-validates against k8s 1.29
- [x] Verified Raft peer list generation: `replicaCount=1` → 1 peer, `replicaCount=3` → 3 peers in `spur.conf`
- [ ] End-to-end install against a real cluster — blocked on `ghcr.io/rocm/spur` image not being published yet (today the default `image.repository` is a placeholder; users override with `--set image.repository=...`)

## Out of scope (follow-ups)

- Publishing the chart to a public registry (separate workflow change once the image is published)
- Updating `docs/deployment/kubernetes.rst` to add a Helm section alongside the existing raw-manifest walkthrough
- `helm package` + push to OCI as part of `release.yml`

### DIFF
--- a/deploy/helm/spur/Chart.yaml
+++ b/deploy/helm/spur/Chart.yaml
@@ -1,0 +1,27 @@
+apiVersion: v2
+name: spur
+description: |
+  Spur — AI-native job scheduler. Drop-in Slurm-compatible scheduler
+  with always-on Raft, GPU-first scheduling, and a Kubernetes operator
+  that reconciles SpurJob CRs into pods.
+type: application
+version: 0.1.0
+appVersion: "0.1.0"
+kubeVersion: ">=1.25.0-0"
+home: https://github.com/ROCm/spur
+sources:
+  - https://github.com/ROCm/spur
+maintainers:
+  - name: ROCm/spur maintainers
+    url: https://github.com/ROCm/spur
+keywords:
+  - hpc
+  - slurm
+  - scheduler
+  - gpu
+  - rocm
+  - ai
+icon: https://raw.githubusercontent.com/ROCm/spur/main/docs/_static/logo.png
+annotations:
+  artifacthub.io/category: skip-prediction
+  artifacthub.io/containsSecurityUpdates: "false"
--- a/deploy/helm/spur/values.yaml
+++ b/deploy/helm/spur/values.yaml
@@ -1,0 +1,223 @@
+# Default values for Spur.
+#
+# This file is the contract between the chart and its users. Override any value
+# with `--set` or `-f values-overrides.yaml`. Re-running `helm upgrade` after
+# changing values is safe; Helm will diff and reconcile.
+
+# -- Cluster name written into spur.conf
+clusterName: spur-k8s
+
+# Container image shared by all daemons (spurctld / spurd / spurrestd / spurdbd /
+# operator are produced by the same multi-call binary build).
+image:
+  repository: ghcr.io/rocm/spur
+  tag: ""                 # defaults to .Chart.AppVersion when empty
+  pullPolicy: IfNotPresent
+  pullSecrets: []
+
+# Common labels applied to every object the chart creates.
+commonLabels: {}
+commonAnnotations: {}
+
+# Namespace handling. By default the chart assumes the release namespace
+# already exists (Helm 3 won't create it without --create-namespace). Setting
+# this to true makes the chart render an explicit Namespace object.
+createNamespace: false
+
+# ServiceAccount used by all workloads (operator needs cluster RBAC).
+serviceAccount:
+  create: true
+  name: ""                # default: "<release>-spur"
+  annotations: {}
+
+# RBAC for the operator (ClusterRole / ClusterRoleBinding). Disable in
+# environments where cluster-scoped RBAC is provisioned out-of-band.
+rbac:
+  create: true
+
+# SpurJob CRD lifecycle. The CRD itself lives under crds/ so Helm installs it
+# on first install. Set install=false if you manage CRDs externally (e.g. ArgoCD
+# pre-sync hook). Helm never removes CRDs on uninstall — that's intentional to
+# avoid wiping user CRs.
+crds:
+  install: true
+
+# -----------------------------------------------------------------------------
+# Controller (spurctld) — Raft consensus, scheduler, job dispatch.
+# -----------------------------------------------------------------------------
+controller:
+  enabled: true
+  replicaCount: 3         # set to 1 for single-node dev clusters
+  image:
+    repository: ""        # falls back to .Values.image.repository
+    tag: ""
+    pullPolicy: ""
+  podManagementPolicy: Parallel
+  resources: {}
+  nodeSelector: {}
+  tolerations: []
+  affinity: {}
+  podLabels: {}
+  podAnnotations: {}
+  persistence:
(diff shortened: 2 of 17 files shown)


deploy/helm/spur/Chart.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: v2
+name: spur
+description: |
+  Spur — AI-native job scheduler. Drop-in Slurm-compatible scheduler
+  with always-on Raft, GPU-first scheduling, and a Kubernetes operator
+  that reconciles SpurJob CRs into pods.
+type: application
+version: 0.1.0
```
shiv-tyagi, line 8:
```suggestion
version: 0.0.1-dev
```

Let's keep the chart version as -dev till we are ready to release our first chart.

deploy/helm/spur/values.yaml
```diff
@@ -0,0 +1,223 @@
+# Default values for Spur.
+#
+# This file is the contract between the chart and its users. Override any value
+# with `--set` or `-f values-overrides.yaml`. Re-running `helm upgrade` after
+# changing values is safe; Helm will diff and reconcile.
+
+# -- Cluster name written into spur.conf
+clusterName: spur-k8s
```
shiv-tyagi, line 8: Do we need it here when there is a section for config itself?

deploy/helm/spur/values.yaml
```diff
@@ -0,0 +1,223 @@
...
+
+# Container image shared by all daemons (spurctld / spurd / spurrestd / spurdbd /
+# operator are produced by the same multi-call binary build).
+image:
+  repository: ghcr.io/rocm/spur
+  tag: ""                 # defaults to .Chart.AppVersion when empty
+  pullPolicy: IfNotPresent
+  pullSecrets: []
```
shiv-tyagi, line 16: We are not releasing images to public yet. We need to think about what to put here.

deploy/helm/spur/Chart.yaml
```diff
@@ -0,0 +1,27 @@
...
+appVersion: "0.1.0"
```
shiv-tyagi, line 9:
```suggestion
appVersion: "0.3.0"
```
appVersion does not need to match the chart version. 

There can be instances when we release the app but the chart doesn't need changing and vice-versa.

shiv-tyagi (changes requested):
Thanks @yansun1996.

We can do another round of review once you address the copilot comments.

yansun1996:
convert to draft, wait for the CI image to be hosted

PR #263, opened 2026-06-09
feat(spur): add DEADLINE job state and DeadLine reason for --deadline parity

### LINKED ISSUE #258: parity(slurm): close DEADLINE state gap — add JobState=DEADLINE and Reason=DeadLine enforcement
## Summary

Spur is missing the `DEADLINE` terminal job state and the corresponding `DeadLine` reason code. Submitting a job with `--deadline=<T>` that becomes unsatisfiable mis-reports as `JobState=CANCELLED Reason=NodeDown`, which is wrong on both fields.

This breaks Slurm parity for any workflow that relies on `--deadline` (CI gates, time-boxed batch pipelines, fair-share aging).

## Versions probed

| | Version | Source |
|---|---|---|
| Spur | `main @ ef2e265` (and `v0.3.0` release) | Built from source; behavior identical on both |
| Slurm | `25.11.6` (current stable; also re-checked on `21.08.5`) | Built from `slurm-25.11.6.tar.bz2` |

Confirmed on 2026-06-09: both `v0.3.0` and `upstream/main @ ef2e265` exhibit the broken behavior. No upstream fix has landed.

## Repro

```bash
# Slurm (control — 25.11.6 OR 21.08.5, identical result)
sbatch --parsable -N1 -c10 --wrap='sleep 300'  # x2 to saturate both nodes
NOW=$(date -u +%s)
DEAD=$(date -u -d "@$((NOW+10))" +%Y-%m-%dT%H:%M:%S)
jid=$(sbatch --parsable -N1 -c10 --deadline=$DEAD --wrap='sleep 300')
sleep 18
scontrol show job $jid | grep -E 'JobState|Reason|Deadline'
# → JobState=DEADLINE  Reason=DeadLine  Deadline=<DEAD>

# Spur (same intent — both v0.3.0 and main @ ef2e265)
cat > /tmp/sleep.sh <<'X'
#!/bin/bash
sleep 300
X
spur submit -c 10 /tmp/sleep.sh           # saturate
NOW=$(date -u +%s)
DEAD=$(date -u -d "@$((NOW+10))" +%Y-%m-%dT%H:%M:%SZ)
jid=$(spur submit -c 10 --deadline=$DEAD /tmp/sleep.sh | grep -oE '[0-9]+' | head -1)
sleep 18
spur show job $jid | grep -E 'JobState|Reason|Deadline'
# → JobState=CANCELLED  Reason=NodeDown   ← wrong state, wrong reason
```

1. Add `Deadline` to the job state enum and surface as `JobState=DEADLINE`.
2. Scheduler tick checks pending jobs; if `now > deadline` and the job hasn't started, transition to `Deadline` with `Reason=DeadLine`.
3. Reason emission on the "no resources at deadline" path must be `DeadLine`, not `NodeDown`.

## Scope / size

Single-PR change. Estimated **M (200–600 LOC)**:

- `spur-core/src/job.rs` — new `JobState::Deadline` variant + transition rules
- `spur-sched` — deadline check in scheduler tick
- `spurctld/src/server.rs` — reason emission fix + proto conversion
- Tests: state transition + scheduler tick at deadline
- CLI display (`spur show job`, `squeue` formatting)

## Priority

**P0** — Slurm-visible behavior, currently mis-reports rather than just being absent.

## Related

Part of broader Category-4 (Job Lifecycle & State Machine) parity work. Sibling: dependency-engine gap (#259).

---
Found during 2026-06-08 Slurm↔Spur parity testing; re-verified 2026-06-09 against Slurm 25.11.6 and Spur main @ ef2e265.

yansun1996:
## Summary

Closes #258.

Spur was missing the `DEADLINE` terminal job state and the `DeadLine` reason code. A job submitted with `--deadline=<T>` that became unsatisfiable while pending mis-reported as `JobState=CANCELLED Reason=NodeDown` — wrong on both fields, breaking Slurm parity for any `--deadline` workflow (CI gates, time-boxed batch pipelines, fair-share aging).

This PR adds the state and wires it through the stack:

- **Core** (`job.rs`): new `JobState::Deadline` (proto `JOB_DEADLINE=10`, code `DL`, display `DEADLINE`) — a terminal state reachable **only** from `Pending`. Renames `PendingReason::DeadlineReached → DeadLine` to match Slurm's exact reason string.
- **Scheduler** (`scheduler_loop.rs`): the deadline-enforcement path now calls `cluster.deadline_job()` instead of `cancel_job()`, so an expired pending job ends as `DEADLINE`/`Reason=DeadLine`.
- **Controller** (`cluster.rs`): new `deadline_job()` proposes a `JobComplete{ state: Deadline, exit_code: -1 }` WAL op and sets the reason before the transition; `update_pending_reasons()` skips jobs already marked `DeadLine` so the reason isn't clobbered mid-tick.
- **Plumbing**: dependency engine (`afterok`/`afternotok`/`afterany` treat `DEADLINE` as a failure terminal), `srun` terminal handling, metrics export, `sacct`/`sdiag` CLI, `spurdbd` accounting state mapping, proto enum, and metrics golden fixtures.

## State semantics

`DEADLINE` is a **pending-time** concept (deadline passed before the job started). A *running* job that overruns its `--time` is a `TIMEOUT`, not a `DEADLINE` — so the transition table permits only `Pending → Deadline` and rejects `Running → Deadline`. Unlike `Timeout`, `Deadline` is **not** requeue-able (requeuing an already-expired job would just re-expire).

## Tests

- **Unit**: transition table, terminal/reason semantics, `deadline_job()` controller path, `update_pending_reasons` race guard, proto round-trip.
- **Integration**: state machine (`t06_cancel`), dependency resolution against a DEADLINE'd job (`t56_deps`), `sacct` parse round-trip.
- **Goldens**: both `.prom` metrics fixtures updated.

`cargo build` and `cargo test` pass.

## Manual verification

Deployed the release build to a test VM and reproduced the issue's scenario (unschedulable pending job + near deadline):

```
JobId=26 JobState=DEADLINE Reason=DeadLine ExitCode=-1 StartTime=N/A
```

Controller log:
```
scheduler_loop: job deadline passed while still pending — transitioning to DEADLINE job_id=26
cluster: job deadline passed — transitioned to DEADLINE job_id=26
```

Previously this job would have shown `CANCELLED`.

### DIFF
--- a/crates/spur-core/src/job.rs
+++ b/crates/spur-core/src/job.rs
@@ -25,6 +25,10 @@ pub enum JobState {
     NodeFail,
     Preempted,
     Suspended,
+    /// Job's `--deadline` passed while still pending. Terminal state.
+    /// Distinct from `Cancelled` so users can tell timed-out submissions
+    /// apart from user/admin cancellations.
+    Deadline,
 }
 
 impl JobState {
@@ -41,6 +45,7 @@ impl JobState {
             Self::NodeFail => "NF",
             Self::Preempted => "PR",
             Self::Suspended => "S",
+            Self::Deadline => "DL",
         }
     }
 
@@ -57,13 +62,19 @@ impl JobState {
             Self::NodeFail => "NODE_FAIL",
             Self::Preempted => "PREEMPTED",
             Self::Suspended => "SUSPENDED",
+            Self::Deadline => "DEADLINE",
         }
     }
 
     pub fn is_terminal(&self) -> bool {
         matches!(
             self,
-            Self::Completed | Self::Failed | Self::Cancelled | Self::Timeout | Self::NodeFail
+            Self::Completed
+                | Self::Failed
+                | Self::Cancelled
+                | Self::Timeout
+                | Self::NodeFail
+                | Self::Deadline
         )
     }
 
@@ -105,7 +116,7 @@ impl JobState {
     }
 
     /// Every core variant, in proto discriminant order for iteration only.
-    pub const ALL: [JobState; 10] = [
+    pub const ALL: [JobState; 11] = [
         Self::Pending,
         Self::Running,
         Self::Completing,
@@ -116,6 +127,7 @@ impl JobState {
         Self::NodeFail,
         Self::Preempted,
         Self::Suspended,
+        Self::Deadline,
     ];
 
     pub const COUNT: usize = Self::ALL.len();
@@ -133,6 +145,7 @@ impl JobState {
             spur_proto::proto::JobState::JobNodeFail => Self::NodeFail,
             spur_proto::proto::JobState::JobPreempted => Self::Preempted,
             spur_proto::proto::JobState::JobSuspended => Self::Suspended,
+            spur_proto::proto::JobState::JobDeadline => Self::Deadline,
         }
     }
 
@@ -149,6 +162,7 @@ impl JobState {
             Self::NodeFail => spur_proto::proto::JobState::JobNodeFail,
             Self::Preempted => spur_proto::proto::JobState::JobPreempted,
             Self::Suspended => spur_proto::proto::JobState::JobSuspended,
+            Self::Deadline => spur_proto::proto::JobState::JobDeadline,
         }
     }
 
@@ -196,7 +210,9 @@ pub enum PendingReason {
     QoSMaxJobsPerUser,
     ReqNodeNotAvail,
     BeginTime,
-    DeadlineReached,
+    /// Job's `--deadline` passed. Matches Slurm's `DeadLine` reason string
+    /// (note the capitalisation).
+    DeadLine,
     Licenses,
 }
 
@@ -215,7 +231,7 @@ impl PendingReason {
             Self::QoSMaxJobsPerUser => "QOSMaxJobsPerUserLimit",
             Self::ReqNodeNotAvail => "ReqNodeNotAvail",
             Self::BeginTime => "BeginTime",
-            Self::DeadlineReached => "DeadlineReached",
+            Self::DeadLine => "DeadLine",
             Self::Licenses => "Licenses",
         }
     }
@@ -587,6 +603,7 @@ impl Job {
         let valid = match (self.state, to) {
             (JobState::Pending, JobState::Running) => true,
             (JobState::Pending, JobState::Cancelled) => true,
+            (JobState::Pending, JobState::Deadline) => true,
             (JobState::Running, JobState::Completing) => true,
             (JobState::Running, JobState::Completed) => true,
             (JobState::Running, JobState::Failed) => true,
@@ -770,6 +787,41 @@ mod tests {
     }
 
     #[test]
+    fn deadline_state_is_terminal_and_reachable_only_from_pending() {
+        // Terminal flag is what tells the dep engine and the requeue logic
+        // that this state can never come back to Pending.
+        assert!(JobState::Deadline.is_terminal());
+        assert!(!JobState::Deadline.is_active());
+        assert_eq!(JobState::Deadline.code(), "DL");
+        assert_eq!(JobState::Deadline.display(), "DEADLINE");
+
+        // Pending -> Deadline is the only legal entry. Running/Suspended/
+        // already-terminal jobs must NOT silently fall into DEADLINE — those
+        // would mask the real reason the job ended.
+        let mut p = make_job();
+        assert_eq!(p.state, JobState::Pending);
+        p.transition(JobState::Deadline).unwrap();
+        assert_eq!(p.state, JobState::Deadline);
+        assert!(p.end_time.is_some());
+
+        let mut r = make_job();
+        r.transition(JobState::Running).unwrap();
+        assert!(r.transition(JobState::Deadline).is_err());
+
+        let mut done = make_job();
+        done.transition(JobState::Running).unwrap();
+        done.transition(JobState::Completed).unwrap();
+        assert!(done.transition(JobState::Deadline).is_err());
+    }
+
+    #[test]
+    fn deadline_reason_displays_slurm_string() {
+        // Slurm reports this exact string ("DeadLine", note the cap D and L).
+        // squeue scrapers and Slurm-compat clients pattern-match on it.
+        assert_eq!(PendingReason::DeadLine.display(), "DeadLine");
+    }
+
+    #[test]
     fn test_path_resolution() {
         let mut job = make_job();
         job.job_id = 42;
@@ -805,6 +857,7 @@ mod tests {
             (P::JobNodeFail, JobState::NodeFail),
             (P::JobPreempted, JobState::Preempted),
             (P::JobSuspended, JobState::Suspended),
+            (P::JobDeadline, JobState::Deadline),
         ];
 
         assert_eq!(TABLE.len(), JobState::COUNT);
--- a/crates/spur-core/src/job.rs
+++ b/crates/spur-core/src/job.rs
@@ -25,9 +25,6 @@ pub enum JobState {
     NodeFail,
     Preempted,
     Suspended,
-    /// Job's `--deadline` passed while still pending. Terminal state.
-    /// Distinct from `Cancelled` so users can tell timed-out submissions
-    /// apart from user/admin cancellations.
     Deadline,
 }
 
@@ -210,10 +207,6 @@ pub enum PendingReason {
     QoSMaxJobsPerUser,
     ReqNodeNotAvail,
     BeginTime,
-    /// Job's `--deadline` passed. Matches Slurm's `DeadLine` reason string
-    /// (note the capitalisation). The serde alias keeps Raft snapshots/log
-    /// entries written before the rename (`DeadlineReached`) deserializable.
-    #[serde(alias = "DeadlineReached")]
     DeadLine,
     Licenses,
 }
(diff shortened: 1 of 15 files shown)


crates/spur-core/src/job.rs
```diff
@@ -25,6 +25,10 @@ pub enum JobState {
     NodeFail,
     Preempted,
     Suspended,
+    /// Job's `--deadline` passed while still pending. Terminal state.
+    /// Distinct from `Cancelled` so users can tell timed-out submissions
+    /// apart from user/admin cancellations.
```
shiv-tyagi: This comment adds less value. Maybe we should keep the enum clean.

crates/spur-core/src/job.rs
```diff
@@ -196,7 +210,9 @@ pub enum PendingReason {
     QoSMaxJobsPerUser,
     ReqNodeNotAvail,
     BeginTime,
-    DeadlineReached,
+    /// Job's `--deadline` passed. Matches Slurm's `DeadLine` reason string
+    /// (note the capitalisation).
+    DeadLine,
```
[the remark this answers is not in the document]
yansun1996: Good catch — fixed in 6027f45 by adding `#[serde(alias = "DeadlineReached")]` to `PendingReason::DeadLine`, so pre-rename Raft snapshots/log entries still deserialize on controller restart. Added a round-trip test (`deadline_reason_deserializes_pre_rename_snapshots`).

crates/spurctld/src/cluster.rs
```diff
@@ -237,6 +237,47 @@ impl ClusterManager {
             .collect()
     }
 
+    /// Mark a pending job as DEADLINE (Slurm parity for `--deadline`).
+    ///
+    /// Distinct from `cancel_job` so the terminal state and audit trail show
+    /// `JobState::Deadline / Reason=DeadLine` instead of `Cancelled / NodeDown`.
+    /// Only valid from `Pending`; if the job has already started or terminated
+    /// this is a no-op.
```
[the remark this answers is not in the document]
yansun1996: Fixed the doc in 6027f45. The error-returning behavior is intentional (running/terminal jobs must not be relabelled DEADLINE), so I corrected the comment: it returns `Err` for unknown/terminal/running jobs, and documented that the deadline enforcer treats that error as benign (the job left Pending between scan and call).

crates/spur-core/src/job.rs
```diff
@@ -196,7 +210,11 @@ pub enum PendingReason {
     QoSMaxJobsPerUser,
     ReqNodeNotAvail,
     BeginTime,
-    DeadlineReached,
+    /// Job's `--deadline` passed. Matches Slurm's `DeadLine` reason string
+    /// (note the capitalisation). The serde alias keeps Raft snapshots/log
+    /// entries written before the rename (`DeadlineReached`) deserializable.
+    #[serde(alias = "DeadlineReached")]
```
shiv-tyagi: We can cleanup the comment here as well.

Just wondering if we need to do the renaming? Is the enum entry user facing? We can also only do,
`Self::DeadlineReached => "Deadline",` where we set the display string for this. But if you want to make it consistent everywhere, no problems with that as well.

One more thought, the project is in pre 1.0.0 stage, if the renaming is really needed we can make breaking changes, without worrying about defaults and aliases (if that helps keeping things clean). Hence, we can directly change DeadlineExceeded to Deadline. There are other breaking changes as well apart from this PR.

I will let you decide what you want to do on this.

crates/spurctld/src/cluster.rs
```diff
@@ -237,6 +237,50 @@ impl ClusterManager {
...
 
+    /// Mark a pending job as DEADLINE (Slurm parity for `--deadline`).
+    ///
+    /// Distinct from `cancel_job` so the terminal state and audit trail show
+    /// `JobState::Deadline / Reason=DeadLine` instead of `Cancelled / NodeDown`.
+    /// Only valid from `Pending`: returns `Err` if the job is unknown, already
+    /// terminal, or has started running (those must not be relabelled DEADLINE).
+    /// Callers treat the error as non-fatal — the deadline enforcer logs and
+    /// moves on, since a job that left `Pending` between the scan and this call
+    /// no longer needs the deadline transition.
```
shiv-tyagi: We can reduce comment verbosity here as well. It is just repeating what it does.

crates/spurctld/src/cluster.rs
```diff
@@ -2833,6 +2884,79 @@ mod tests {
         assert_eq!(job.state, JobState::Cancelled);
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn deadline_job_transitions_pending_to_deadline_with_deadline_reason() {
+        // Mirrors the Slurm `JobState=DEADLINE Reason=DeadLine` parity check
+        // from issue #258 — a `--deadline` that passes while the job is still
+        // pending must produce a DEADLINE terminal state, not CANCELLED, and
+        // the surfaced reason must be `DeadLine`, not whatever
+        // `update_pending_reasons` last wrote (e.g. NodeDown, Resources).
```
shiv-tyagi: This comment provides no value.

crates/spurctld/src/cluster.rs
```diff
@@ -2833,6 +2884,79 @@ mod tests {
...
+        assert_eq!(job.state, JobState::Deadline);
+        assert_eq!(job.pending_reason, PendingReason::DeadLine);
+        assert_eq!(job.exit_code, Some(-1));
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn deadline_job_rejects_non_pending_states() {
+        // DEADLINE is a pending-time concept; once a job has actually started
+        // running, hitting the wall time is a Timeout, and hitting the
+        // deadline mid-run shouldn't silently rewrite Running → Deadline.
```
shiv-tyagi: Same. The test is self explanatory, the comment adds no value. You can consider trimming. Similar elsewhere.

crates/spur-core/src/job.rs
```diff
@@ -769,6 +788,57 @@ mod tests {
...
+
+    #[test]
+    fn deadline_reason_displays_slurm_string() {
+        // Slurm reports this exact string ("DeadLine", note the cap D and L).
+        // squeue scrapers and Slurm-compat clients pattern-match on it.
+        assert_eq!(PendingReason::DeadLine.display(), "DeadLine");
+    }
+
+    #[test]
+    fn deadline_reason_deserializes_pre_rename_snapshots() {
```
shiv-tyagi: This test can go away if you decide to do remove the serde alias.

crates/spurctld/src/scheduler_loop.rs
shiv-tyagi:
```suggestion
        // Deadline enforcement: mark pending jobs whose deadline has passed
```

crates/spurctld/src/scheduler_loop.rs
```diff
@@ -729,10 +729,10 @@ async fn enforce_time_limits(cluster: Arc<ClusterManager>, raft: Arc<RaftHandle>
                     if now > deadline {
                         info!(
                             job_id = job.job_id,
-                            "job deadline passed while still pending — cancelling"
+                            "job deadline passed while still pending — transitioning to DEADLINE"
                         );
```
shiv-tyagi: deadline_job() already logs on success, so this duplicates it. We can drop the caller log to stay consistent with how cancel_job does it.

shiv-tyagi (commented):
Thanks Yan.

Few nits. PTAL.

I have got the E2Es passing on your PR. It was a small infra issue.

A side note, you can rebase your PR on latest main branch's head for the codecov to compare the test coverage against the main branches latest commit.

crates/spur-core/src/job.rs
yansun1996: Done in ad97d70 — removed the comment, enum left clean.

crates/spur-core/src/job.rs
yansun1996: Decided to keep the variant names and drop the serde alias (done in ad97d70). I verified the strings against a live Slurm 25.11.6 instance: `JobState=DEADLINE` (short code `DL`) and `Reason=DeadLine` (note Slurm's own asymmetric capitalisation). Spur's `display()`/`code()` already emit those exact strings, so the wire/user-facing output is correct regardless of the Rust identifier. Since we're pre-1.0 and you OK'd breaking changes, I removed the `#[serde(alias = "DeadlineReached")]` rather than carrying back-compat. `DeadlineReached`/`DeadlineExceeded` no longer exist anywhere in the tree.

crates/spurctld/src/cluster.rs
yansun1996: Trimmed in ad97d70 — condensed the `deadline_job` doc to the eligibility/error contract.

crates/spurctld/src/cluster.rs
yansun1996: Removed in ad97d70.

crates/spurctld/src/cluster.rs
yansun1996: Removed in ad97d70 — trimmed this and the sibling test comment.

crates/spur-core/src/job.rs
yansun1996: Removed the alias and deleted this test (`deadline_reason_deserializes_pre_rename_snapshots`) in ad97d70.

crates/spurctld/src/scheduler_loop.rs
yansun1996: Applied in ad97d70.

crates/spurctld/src/scheduler_loop.rs
yansun1996: Done in ad97d70 — dropped the caller log; `deadline_job()` already logs on success, matching how `cancel_job` does it.

shiv-tyagi (approved):
LGTM

shiv-tyagi:
E2E runner is unreachable. Looks like a network issue, let's wait for it to come back.

PR #265, opened 2026-06-10
fix(spur-core): serialize hook tests to avoid ETXTBSY flakiness

shiv-tyagi:
## Problem

Hook tests are flaky (~9% failure rate) due to `ETXTBSY` — a multi-threaded fork+exec race where another test thread's `fork()` inherits a write fd to the temp script before it's closed. This is test-only; production never writes and execs a script in the same process.

## Before (100 runs of `cargo test -p spur-core`)

| Test | Failures |
|------|----------|
| `hook_receives_env_vars` | 3 |
| `hook_success` | 3 |
| `hook_stderr_does_not_prevent_success` | 3 |
| **Total suite failures** | **9 / 100** |

## After (100 runs of `cargo test -p spur-core`)

| Test | Failures |
|------|----------|
| `hook_receives_env_vars` | 0 |
| `hook_success` | 0 |
| `hook_stderr_does_not_prevent_success` | 0 |
| **Total suite failures** | **0 / 100** |

## Fix

Added `#[serial(run_hooks)]` to all hook tests. Only these 5 tests serialize with each other; the other ~119 tests remain parallel. Production `run_hook` is untouched.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -92,7 +92,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


sgopinath1 (approved):
LGTM

PR #271, opened 2026-06-10
fix(spur): resolve dependency-engine gaps against array parents

### LINKED ISSUE #259: parity(slurm): close dependency-engine gap — fix array-parent deadlock and implement aftercorr/after
## Summary

Spur's job-dependency engine has three related defects that all cause silent deadlock instead of either correct behavior or a clear error:

1. **`afterok:<array-parent>` deadlocks the child forever**, even when every array task of the parent succeeds.
2. **`aftercorr:<array-parent>` is silently accepted and never enforced** — child tasks stay PENDING indefinitely regardless of parent outcome.
3. **Time-based `after:<id>[+N]` accepted without validation** — never gated on parent start time.

Combined with the lack of dependency-type validation (separate issue), users get hung jobs instead of clear failures.

## Versions probed

| | Version | Source |
|---|---|---|
| Spur | `main @ ef2e265` (and `v0.3.0` release) | Built from source; behavior identical on both |
| Slurm | `25.11.6` (current stable; also re-checked on `21.08.5`) | Built from `slurm-25.11.6.tar.bz2` |

Confirmed on 2026-06-09 against the upstream `main` branch — no fix has landed since v0.3.0.

### Defect 1: afterok on array parent deadlocks the child

```bash
# Spur (main @ ef2e265)
cat > /tmp/ok.sh <<'X'
#!/bin/bash
echo ok job=$SPUR_JOB_ID
X
p=$(spur submit -a '0-2' /tmp/ok.sh | grep -oE '[0-9]+' | head -1)
sleep 8                                          # parent's 3 tasks all COMPLETE
c=$(spur submit -d afterok:$p /tmp/ok.sh | grep -oE '[0-9]+' | head -1)
sleep 10
spur show job $c                                 # → JobState=PENDING  (deadlocked)
```

### Defect 2: aftercorr never enforced

```bash
# Spur (main @ ef2e265)
p=$(spur submit -a '0-2' /tmp/ok.sh | grep -oE '[0-9]+' | head -1)
sleep 8                                          # parent tasks COMPLETED
c=$(spur submit -a '0-2' -d aftercorr:$p /tmp/ok.sh | grep -oE '[0-9]+' | head -1)
sleep 10
# Expect: each child task[N] in COMPLETED (parent task[N] succeeded)
# Actual: all 3 child tasks PENDING forever
for j in $(seq $c $((c+3))); do
  spur show job $j 2>/dev/null | grep -E 'JobState|Reason' | tr '\n' ' '; echo
done
```

### Defect 3: after:<id>+min accepted with no validation or enforcement

```bash
spur submit -d "after:$p+1" /tmp/ok.sh           # accepted, not gated on time
spur submit -d "expand:$p"  /tmp/ok.sh           # also accepted (bonus: no dep-type validation)
```

### Slurm 25.11.6 control (aftercorr with mixed-result parent)

```bash
cat > /tmp/mixed.sh <<'X'
#!/bin/bash
[ "$SLURM_ARRAY_TASK_ID" = "1" ] && exit 1
echo task $SLURM_ARRAY_TASK_ID ok
X
p=$(sbatch --parsable -a '0-2' /tmp/mixed.sh)
c=$(sbatch --parsable -a '0-2' -d aftercorr:$p /tmp/mixed.sh)
sleep 8
# Expected per-task correspondence — Slurm gets this right:
#   parent task 0 OK     → child task 0 COMPLETED
#   parent task 1 FAILED → child task 1 PENDING, Reason=DependencyNeverSatisfied
#   parent task 2 OK     → child task 2 COMPLETED
for k in 0 1 2; do
  scontrol show job ${c}_$k | grep -E 'JobState|Reason|Dependency'
done
```

## Root cause hypothesis

The array-job placeholder ID never reaches a terminal state visible to the dependency resolver (sibling tasks reach COMPLETED but the parent ID stays in a non-terminal aggregate state). This affects **every** dependency type when the parent is an array job, not just `aftercorr`.

Side note from re-testing on main: `spur show job <array_parent_id>` returns empty output (not even an error) — the parent placeholder appears to have no surfaced state at all, which is consistent with the dep-resolver having nothing to fire on.

1. **Parent aggregate state**: when all array tasks of `parent_array_job_id` reach a terminal state, the parent ID itself must be marked terminal so dep resolvers can fire. Aggregate state should follow Slurm: `COMPLETED` iff all tasks `COMPLETED`, else `FAILED` if any `FAILED`, etc.
2. **`aftercorr` semantics**: child task `[N]` releases when parent task `[N]` completes successfully (per-task correspondence, not all-or-nothing).
3. **`after:<id>[+N]`**: child eligible to start `N` minutes after parent's start time (regardless of parent outcome).
4. **Existing `afterok`/`afterany`/`afternotok`** should also work correctly against array parents once (1) is fixed.

## Scope / size

Estimated **L (600–1500 LOC)**, suggested 2 PRs:

- **PR (a):** Array-parent terminal-state propagation. Likely in `spur-core/src/array.rs` and the dep-resolver in `spur-sched`. Fixes defect 1 and unblocks all dep types against array parents. **Land first.**
- **PR (b):** Implement `aftercorr` per-task correspondence + `after:<id>[+N]` time gating. Adds new dep-type variants and per-task wiring.

Test coverage: matrix of {afterok, afterany, afternotok, aftercorr, after} × {scalar parent, array parent} × {parent success, parent failure, mixed-result array}.

## Priority

**P0** — silent deadlock is worse than rejection; affects every Slurm batch script that uses `-d` against an array job.

## Related

- Sibling P0: DEADLINE state gap (#258)
- Follow-up P1: dep-type validation at submit (will reject unknown keywords cleanly once these defects are fixed)
- Adjacent gap noticed during re-test: **`SLURM_ARRAY_TASK_ID` env var is not exported to array-job scripts on Spur**, breaking any Slurm script that branches on task index. May need its own ticket.

---
Found during 2026-06-08 Slurm↔Spur parity testing; re-verified 2026-06-09 against Slurm 25.11.6 and Spur main @ ef2e265.

yansun1996:
## Summary

Closes #259.

Fixes the dependency-engine gaps where job dependencies resolved against an array **parent** id silently deadlocked (the parent has no stored job record — Spur stores only per-task jobs), and where unknown dependency types were silently accepted then hung.

## What changed

- **Array-parent dependency resolution.** New `resolve_target_state()` aggregates per-task states when a dependency target is an array parent. `afterok`/`afterany`/`afternotok` against an array parent now resolve via `array::aggregate_array_state()` instead of waiting forever.
- **`aftercorr` per-task correspondence.** Child task[N] releases iff parent task[N] completed successfully (degrades to `afterok` for scalar parents/children).
- **Unsatisfiable deps are finalized, not stuck.** `cancel_unsatisfiable_dependency_jobs()` runs in the scheduler loop: jobs whose deps can never be met are Cancelled (`Reason=Dependency`) instead of sitting PENDING forever; still-waiting deps are tagged `Reason=Dependency`.
- **Submit-time validation.** `try_parse_dependencies()` rejects unknown types (e.g. `expand:N`) and malformed ids at submit, matching Slurm — instead of silent-accept-then-deadlock.
- **`after:N[+M]` time dependency.** New time-delay dependency (child eligible M minutes after parent start, regardless of outcome).
- **Array-parent display.** `scontrol show job <array_parent>` / `squeue` now return the synthesized aggregate (via both the single `get_job` and the `get_jobs` list RPC) instead of empty output.
- **Array task identity.** Proto `array_job_id`/`array_task_id` plumbed controller→agent; `SLURM_ARRAY_JOB_ID` / `SLURM_ARRAY_TASK_ID` (+ `SPUR_*`) exported to the job environment.

## Tests

- Unit: `array::aggregate_array_state` (6 cases), `dependency` resolver against array parents.
- Integration: `t28_arrays` (afterok all-pass / one-failed / aftercorr per-task), `t56_deps` (after-delay parse, unknown-type rejection, after-unknown-parent).
- e2e: `tests/e2e/native_host/test_single_node.py::TestArrayDependencies` (task env export, afterok-on-parent release, show-parent aggregate, unknown-dep rejection).
- All pass: `spur-core` 150, `spurctld` 345, `spur-tests` 97; `cargo fmt --check` clean.

## Live verification (single-node Spur)

Verified against a running single-node cluster — 24/24 checks pass:

- afterok on array parent → child released (the core deadlock, fixed)
- `scontrol show job <parent>` → aggregate PENDING while running → COMPLETED after
- unknown dep `expand:N` → rejected at submit
- afterok on failing/mixed array → child CANCELLED `Reason=Dependency`; parent aggregate FAILED
- aftercorr per-task: child t0 COMPLETED, t1 CANCELLED (parent t1 failed), t2 COMPLETED
- afterany/afternotok on array (success vs failure paths)
- `after:N` runs; `after:N+1` correctly gated until the delay elapses
- chained deps, multi-target `afterok:A:B` (both-ok vs one-fail), scalar-aftercorr fallback, scalar regression, nonexistent-target cancellation

### DIFF
--- a/crates/spur-core/src/dependency.rs
+++ b/crates/spur-core/src/dependency.rs
@@ -29,9 +29,8 @@ pub enum Dependency {
     Singleton,
 }
 
-/// Error returned when a dependency string cannot be parsed or names an
-/// unknown dependency type. Surfaced at submit time so users get a clear
-/// rejection instead of a silently-deadlocked job.
+/// Dependency string parse failure, surfaced at submit time so users get a
+/// clear rejection instead of a silently-deadlocked job.
 #[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
 pub enum DependencyParseError {
     #[error("unknown dependency type: '{0}'")]
@@ -42,12 +41,9 @@ pub enum DependencyParseError {
     InvalidSyntax(String),
 }
 
-/// Parse dependency strings, rejecting unknown types and malformed ids.
-///
-/// Accepts forms like `afterok:100`, `afterany:200:300`, `after:100+5`,
-/// `aftercorr:42`, `singleton`. Multiple entries may be comma-separated within
-/// one string. Returns an error on the first unrecognized type or unparseable
-/// id so the caller can reject the submission.
+/// Strict parse: rejects unknown types and malformed ids. Accepts `afterok:100`,
+/// `afterany:200:300`, `after:100+5`, `aftercorr:42`, `singleton`; entries
+/// comma-separated. Errors on the first bad type/id so the caller can reject.
 pub fn try_parse_dependencies(specs: &[String]) -> Result<Vec<Dependency>, DependencyParseError> {
     let mut deps = Vec::new();
     for spec in specs {
@@ -121,13 +117,11 @@ pub fn try_parse_dependencies(specs: &[String]) -> Result<Vec<Dependency>, Depen
     Ok(deps)
 }
 
-/// Lenient parse used on the resolution path, where the spec was already
-/// validated at submit time. Unknown/malformed entries are dropped rather than
-/// erroring so a single bad entry can't wedge the scheduler loop.
+/// Lenient parse for the resolution path (spec already validated at submit).
+/// Drops bad entries instead of erroring so one can't wedge the scheduler loop.
 pub fn parse_dependencies(specs: &[String]) -> Vec<Dependency> {
     try_parse_dependencies(specs).unwrap_or_else(|_| {
-        // Fall back to best-effort: parse each entry independently, skipping
-        // any that fail. This preserves behavior for callers that pre-validated.
+        // Best-effort: parse each entry alone, skipping failures.
         let mut deps = Vec::new();
         for spec in specs {
             for part in spec.split(',') {
@@ -140,19 +134,13 @@ pub fn parse_dependencies(specs: &[String]) -> Vec<Dependency> {
     })
 }
 
-/// Resolve a dependency target to its effective state, transparently handling
-/// array parents.
+/// Resolve a dependency target to its effective state, handling array parents
+/// (whose own job record never exists — Spur stores only per-task jobs) by
+/// aggregating task states.
 ///
-/// A dependency target id `P` may refer to:
-/// - a scalar job (returned directly by `get_job`), or
-/// - an array *parent* id whose own job record never exists (Spur stores only
-///   the per-task jobs). In that case we aggregate the task states.
-///
-/// Returns `None` when the target is unknown (neither a scalar job nor an array
-/// with any tasks). Otherwise returns `Some(state)`: the live scalar state, or
-/// the array's aggregate. An array that exists but has not finished aggregating
-/// is reported as the non-terminal sentinel `JobState::Running`, so callers
-/// treat it as still-waiting.
+/// `None` if the target is unknown. Otherwise `Some(state)`: the scalar state,
+/// or the array aggregate — an unfinished array reports the non-terminal
+/// sentinel `JobState::Running` so callers treat it as still-waiting.
 fn resolve_target_state(
     dep_id: JobId,
     get_job: &dyn Fn(JobId) -> Option<Job>,
@@ -165,8 +153,7 @@ fn resolve_target_state(
     if tasks.is_empty() {
         return None; // Genuinely unknown id.
     }
-    // Array parent: aggregate. `None` here means "not finished yet" — we encode
-    // that as a non-terminal sentinel (Running) so callers treat it as Waiting.
+    // Array parent: aggregate; unfinished -> Running sentinel (see doc).
     Some(
         aggregate_array_state(&tasks.iter().map(|t| t.state).collect::<Vec<_>>())
             .unwrap_or(JobState::Running),
@@ -195,9 +182,7 @@ pub fn check_dependencies(
                 job_id: dep_id,
                 delay_minutes,
             } => {
-                // Find the parent's start time. For an array parent, use the
-                // earliest task start. Eligible `delay_minutes` after start,
-                // regardless of outcome.
+                // Earliest start across the target (array parent: earliest task).
                 let start = get_job(*dep_id).and_then(|j| j.start_time).or_else(|| {
                     get_array_tasks(*dep_id)
                         .iter()
@@ -212,17 +197,12 @@ pub fn check_dependencies(
                         }
                     }
                     None => {
-                        // No recorded start time. Either the parent never
-                        // existed, or it reached a terminal state without ever
-                        // running (e.g. cancelled/failed while Pending —
-                        // start_time is only set on JobStart). In both cases the
-                        // parent will never start, so `after` can no longer
-                        // block: Slurm releases `after` dependents once the
-                        // parent leaves the pending/eligible set. Only a
-                        // known-but-still-non-terminal parent keeps us waiting.
+                        // No start time: a parent that never ran will never
+                        // start, so `after` releases (Slurm parity). Only a
+                        // still-non-terminal parent keeps us waiting.
                         match resolve_target_state(*dep_id, get_job, get_array_tasks) {
                             Some(s) if !s.is_terminal() => return DependencyResult::Waiting,
-                            _ => {} // Terminal, or unknown — satisfied.
+                            _ => {}
                         }
                     }
                 }
--- a/crates/spur-core/src/dependency.rs
+++ b/crates/spur-core/src/dependency.rs
@@ -64,14 +64,31 @@ pub fn try_parse_dependencies(specs: &[String]) -> Result<Vec<Dependency>, Depen
                 .split_once(':')
                 .ok_or_else(|| DependencyParseError::InvalidSyntax(part.to_string()))?;
             let dtype = dtype.to_lowercase();
+
+            // Reject unknown types up front, before the id loop — otherwise an
+            // id-less unknown type (e.g. `expand:`) would slip through because
+            // the loop body never runs.
+            let is_after = dtype == "after";
+            match dtype.as_str() {
+                "after" | "afterok" | "after_ok" | "afterany" | "after_any" | "afternotok"
+                | "after_not_ok" | "aftercorr" | "after_corr" => {}
+                other => return Err(DependencyParseError::UnknownType(other.to_string())),
+            }
+
+            // A known type with no parseable ids (e.g. `afterok:`) is malformed.
+            let mut saw_id = false;
             for id_str in ids.split(':') {
                 let id_str = id_str.trim();
                 if id_str.is_empty() {
                     continue;
                 }
-                // `after` allows an optional `+M` minute delay suffix.
+                // Only `after` accepts an optional `+M` minute delay suffix;
+                // any other type carrying `+M` is invalid syntax.
                 let (id_part, delay_minutes) = match id_str.split_once('+') {
                     Some((id, delay)) => {
+                        if !is_after {
+                            return Err(DependencyParseError::InvalidSyntax(id_str.to_string()));
+                        }
                         let delay = delay
                             .trim()
                             .parse::<u32>()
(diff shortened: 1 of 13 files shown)


crates/spur-core/src/dependency.rs
```diff
@@ -3,63 +3,165 @@
...
-                            }
-                            "afternotok" | "after_not_ok" => deps.push(Dependency::AfterNotOk(id)),
-                            "aftercorr" | "after_corr" => deps.push(Dependency::AfterCorr(id)),
-                            _ => {} // Unknown dependency type, skip
-                        }
+            let (dtype, ids) = part
+                .split_once(':')
+                .ok_or_else(|| DependencyParseError::InvalidSyntax(part.to_string()))?;
+            let dtype = dtype.to_lowercase();
+            for id_str in ids.split(':') {
```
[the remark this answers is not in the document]
yansun1996: Fixed in 1aaaa7c. The type is now validated *before* the id loop, and a known type with no parseable ids (`afterok:`) is rejected too — so both `afterok:` and `expand:` now error at submit instead of returning `Ok([])`. Added regression test `t56_18b_parse_rejects_idless_and_unknown_idless`.

crates/spur-core/src/dependency.rs
```diff
@@ -3,63 +3,165 @@
...
+                let (id_part, delay_minutes) = match id_str.split_once('+') {
+                    Some((id, delay)) => {
+                        let delay = delay
+                            .trim()
+                            .parse::<u32>()
+                            .map_err(|_| DependencyParseError::InvalidSyntax(id_str.to_string()))?;
+                        (id.trim(), delay)
+                    }
+                    None => (id_str, 0),
+                };
```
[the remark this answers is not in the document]
yansun1996: Fixed in 1aaaa7c. The `+M` suffix is now gated on `after`; any other type carrying `+M` (e.g. `afterok:123+5`) returns `InvalidSyntax`. Added test `t56_18c_parse_rejects_delay_on_non_after`.

crates/spur-core/src/dependency.rs
```diff
@@ -3,63 +3,165 @@
...
+///
+/// A dependency target id `P` may refer to:
+/// - a scalar job (returned directly by `get_job`), or
+/// - an array *parent* id whose own job record never exists (Spur stores only
+///   the per-task jobs). In that case we aggregate the task states.
+///
+/// Returns `None` when the target is unknown (neither a scalar job nor an array
+/// with any tasks). Returns `Some(None)` when the array exists but has not
+/// finished. Returns `Some(Some(state))` for a resolved terminal/aggregate
+/// state, or the live scalar state.
```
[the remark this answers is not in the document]
yansun1996: Fixed in 1aaaa7c — corrected the doc to match the actual `Option<JobState>` contract: `None` for unknown target, otherwise `Some(state)` with an unfinished array reported as the non-terminal `JobState::Running` sentinel.

crates/spurctld/src/cluster.rs
```diff
@@ -1144,6 +1240,90 @@ impl ClusterManager {
...
+
+        // Tag waiting jobs (write lock).
+        if !to_wait.is_empty() {
+            let mut jobs = self.jobs.write();
+            for id in &to_wait {
+                if let Some(j) = jobs.get_mut(id) {
+                    if j.state == JobState::Pending && j.pending_reason != PendingReason::Held {
+                        j.pending_reason = PendingReason::Dependency;
+                    }
+                }
```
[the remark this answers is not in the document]
yansun1996: Fixed in 1aaaa7c — the waiting-job tagging now also skips `PendingReason::DeadLine`, matching the guard in `update_pending_reasons()`, so deadline enforcement isn't briefly masked.

tests/e2e/native_host/test_single_node.py
```diff
@@ -208,3 +208,94 @@ def test_env_passthrough_export(self, cluster):
...
         assert "MULTIVAR=world456" in content, f"output:\n{content}"
+
+
+class TestArrayDependencies:
+    """Regression tests for issue #259: dependencies against array parents.
+
+    Array submission returns a parent placeholder id P; the individual task
+    jobs get ids P+1, P+2, ... each carrying array_job_id=P. These tests assert
+    that dependencies resolved against P (the array parent) no longer deadlock.
+    """
```
shiv-tyagi:
```suggestion
    """Array submission returns a parent placeholder id P; the individual task
    jobs get ids P+1, P+2, ... each carrying array_job_id=P. These tests assert
    that dependencies resolved against P (the array parent) no longer deadlock.
    """
```

AI tends to add issue and pr numbers to code, which I think is mostly noise.

tests/e2e/native_host/test_single_node.py
```diff
@@ -208,3 +208,94 @@ def test_env_passthrough_export(self, cluster):
...
+
+        # All parent tasks complete, then the child must run (not hang).
+        for tid in range(3):
+            wait_job(cluster, parent + 1 + tid, timeout=60)
+        state = wait_job(cluster, child, timeout=60)
+        assert state in ("CD", "GONE"), f"child should have run, got {state}"
+        content = cluster.read_output_on_any_node(out_c)
+        assert "CHILD_RAN" in content, f"child output:\n{content}"
+
+    def test_show_array_parent_reports_aggregate(self, cluster):
```
shiv-tyagi, line 272: Can we check if we can assert the things in first three tests in a single test?

Looks like the have the same pattern, submit an array job and assert stuff. Is there a scope of merging them without over complications?

proto/slurm.proto
```diff
@@ -105,6 +105,10 @@ message JobSpec {
 
   // Array
   string array_spec = 50;  // "0-99%10"
+  // Per-task array identity set by the controller on expansion. array_job_id
+  // is the parent id (0 = not an array task); exported as SLURM_ARRAY_*.
+  uint32 array_job_id = 81;
+  uint32 array_task_id = 82;
```
shiv-tyagi: Two things I would like to suggest here,

1. **Field numbers:** New fields should go at the end of the message, not jump to 81/82 mid-stream. Easier to track ordering that way.

2. **Wrong message:** These are controller-assigned at expansion time, not user-supplied. `JobSpec` is the submission contract; `proto_to_job_spec` already hardcodes these to `None` on receive. The only wire consumer is controller-to-agent dispatch, which is what `LaunchJobRequest` is for. It already carries `job_id`, `task_offset`, `peer_nodes`, `target_node` for the same reason. I'd suggest moving these to `LaunchJobRequest` and reading from `req.array_job_id` in the agent instead of `spec.array_job_id`.

shiv-tyagi (changes requested):
Thanks @yansun1996! This is looking very good.

Just a few comments and it is good to merge.

tests/e2e/native_host/test_single_node.py
yansun1996: Applied in e7b7759 — dropped the issue/PR reference from the docstring.

proto/slurm.proto
yansun1996: Both done in e7b7759. Moved `array_job_id`/`array_task_id` to the end of `LaunchJobRequest` (fields 7/8) and removed them from `JobSpec`. You're right that JobSpec is the submission contract — `proto_to_job_spec` already hardcoded them to `None` on receive, and the only consumer is controller→agent dispatch. The agent now reads `req.array_job_id` (alongside `task_offset`/`peer_nodes`/`target_node`) instead of `spec.array_job_id`. The core Rust `JobSpec` keeps the fields since the dependency/aftercorr resolver uses them; only the wire proto changed. Build + full tests + clippy green.

tests/e2e/native_host/test_single_node.py
yansun1996: I'd prefer to keep these three separate — they share the submit-an-array setup but assert genuinely different behaviors: (1) per-task SLURM_ARRAY_* env export, (2) afterok-on-array-parent releasing the child, (3) `scontrol show job <parent>` returning the aggregate. Merging would couple unrelated failure modes under one test, so a regression in one would be harder to localize. Happy to extract a small shared submit helper if you'd like to cut the duplication without merging the assertions — let me know.

proto/slurm.proto
yansun1996: Verified the proto move live on a single-node Spur cluster (the refactor's main risk is array identity now flowing via LaunchJobRequest instead of JobSpec):

- **SLURM_ARRAY_* env export** confirmed: a 3-task array (parent 218) produced `TID=0/1/2 JID=218 SPUR_JID=218 SPUR_TID=0/1/2` — distinct per task, correct parent id. The agent reading `req.array_job_id` works end-to-end.
- **Dependency scenarios**: full regression harness green — 10/10 base (afterok/aftercorr/show-parent-aggregate/unknown-dep-reject/failing-array-cancel) and 14/14 advanced (afterany/afternotok/mixed-array/after-delay/chains/multi-target/scalar-fallback).

shiv-tyagi (approved):
LGTM

Issue #273, opened 2026-06-11
bug(spurctld): openraft panic in handle_apply_result when Raft log grows without snapshot compaction


yansun1996:
## Summary

`spurctld` crashes with a panic inside openraft (`raft_core.rs:779: called Option::unwrap() on a None value`, in `handle_apply_result`) once the Raft log grows large without ever being compacted into a snapshot. After the panic the RaftCore task exits and every subsequent write fails with `raft propose failed: panicked` — the controller is permanently wedged until the state-dir is wiped.

## Environment

- Spur `main` (reproduced on the exit-code branch built from `main @ f65a029`).
- Single-node controller (1-member Raft), long-lived state-dir reused across many restarts.

## Symptoms

```
thread 'tokio-rt-worker' panicked at openraft-0.9.24/src/core/raft_core.rs:779:47:
called `Option::unwrap()` on a `None` value
openraft::raft::raft_inner: RaftCore exited res=Err(JoinError::Panic(..., "called `Option::unwrap()` on a `None` value"))
openraft::raft::raft_inner: failure sending RaftMsg to RaftCore; message: ClientWriteRequest core_result=Err(Panicked)
```
Job submission then returns: `Error: job submission failed — code: 'Internal error', message: "raft propose failed: panicked"`.

## Root cause (suspected)

The on-disk Raft store had **841 individual log-entry JSON files and zero snapshots** — `~/spur/state/raft/log/*.json` accumulated up to index 814+, with no `snapshot` directory present. Log compaction / snapshot-building never triggered, so the full log is replayed on every restart. openraft's `handle_apply_result` iterates `res.since..res.end` and calls `.next().unwrap()` on `applying_entries` / `apply_results`; with the oversized, never-compacted log (and apply/snapshot bookkeeping drifting across restarts) that iterator runs out and panics.

Spur's own `apply_to_state_machine` (`spurctld/src/raft.rs`) looks correct (one result pushed per entry), so this is openraft-internal bookkeeping triggered by the uncompacted-log state rather than a wrong result count from our applier.

## Impact

- Controllers that run a long time without compaction will eventually wedge and require a state-dir wipe (data loss).
- No automatic recovery; the panic is fatal to the Raft core.

## Likely fix direction

1. Ensure snapshot/log-compaction is actually triggered (config: `snapshot_policy` / `max_in_snapshot_log_to_keep`, or a periodic `trigger.snapshot()`), so the log is bounded.
[...]

PR #274, opened 2026-06-11
feat(spur): close exit-code reporting gap — exit:signal, DerivedExitCode, reasons

### LINKED ISSUE #269: parity(slurm): close exit-code reporting gap — add exit:signal encoding, DerivedExitCode, and NonZeroExitCode/RaisedSignal reasons
## Summary

Spur reports job outcomes with a lossy exit code and no failure reason, breaking parity with Slurm's `ExitCode=<exit>:<signal>` model:

- **No signal component** — a job killed by a signal reports `ExitCode=-1` instead of `ExitCode=0:9`.
- **No `DerivedExitCode`** — for multi-step batch jobs, Slurm surfaces the *max* step exit code separately from the last; Spur has no such field.
- **No failure reason** — failed jobs report `Reason=None` instead of `NonZeroExitCode` / `RaisedSignal:N(...)`.

This silently corrupts outcome data for anything scraping `scontrol show job` / `sacct` — CI gates, workflow engines (Nextflow/Snakemake-style), and monitoring all key off exit code and reason.

## Versions probed

| | Version | Source |
|---|---|---|
| Spur | `main @ 0c585bd` (and `v0.3.0`) | Built from source; behavior identical |
| Slurm | `25.11.6` | Built from source |

Confirmed on 2026-06-10.

## Repro

```bash
# multi-step: steps exit 0, 7, 3
cat > /tmp/multistep.sh <<'X'
#!/bin/bash
srun bash -c 'exit 0'
srun bash -c 'exit 7'
srun bash -c 'exit 3'
X

# signal: process kills itself
cat > /tmp/sig.sh <<'X'
#!/bin/bash
kill -9 $$
X
```

### Slurm 25.11.6 (control)
```
# multistep
JobState=FAILED  Reason=NonZeroExitCode
ExitCode=3:0     DerivedExitCode=7:0

# signal job
JobState=FAILED  Reason=RaisedSignal:9(Killed)
ExitCode=0:9
```

### Spur main @ 0c585bd
```
# multistep
JobState=FAILED  Reason=None
ExitCode=127            <- no :signal, no DerivedExitCode, wrong reason

# signal job
JobState=FAILED  Reason=None
ExitCode=-1             <- signal lost entirely
```

1. **Encode exit code as `exit:signal`.** A clean exit N -> `N:0`; death by signal S -> `0:S`. Stop collapsing to `-1`.
2. **Compute `DerivedExitCode`** = max exit code across all steps of a batch job; surface as its own field in `scontrol show job` and the proto job record.
3. **Set failure reason**: `NonZeroExitCode` when a step exits non-zero, `RaisedSignal:N(<name>)` when killed by signal N.
4. CLI (`spur show job`, `squeue`) renders the `exit:signal` form and `DerivedExitCode`.

## Scope / size

Single-PR change. Estimated **M (200-600 LOC)**:

- `proto/slurm.proto` — add `derived_exit_code` + signal component to the job record (additive field; no package/name change)
- `spurd/src/agent_server.rs` — capture `WaitStatus` (exit vs signal) from step completion; report both
- `spurctld/src/server.rs` — compute `DerivedExitCode` (max across steps), set `NonZeroExitCode`/`RaisedSignal` reason, proto conversion
- `spur-core/src/job.rs` — exit-code representation + exit-reason additions (`NonZeroExitCode`, `RaisedSignal`)
- `spur-cli` — `exit:signal` + `DerivedExitCode` formatting
- Tests: multi-step max-exit derivation, signal encoding, reason emission

## Related

Part of Job Lifecycle & State Machine parity. Siblings: DEADLINE (#258, merged), dependency-engine (#259).

---
Found during 2026-06-08 parity testing; re-verified 2026-06-10 against Slurm 25.11.6 and Spur main @ 0c585bd.

yansun1996:
## Summary

Closes #269. Brings Spur's job exit-code reporting to Slurm parity, end to end: `exit:signal` encoding, `DerivedExitCode`, and `NonZeroExitCode` / `RaisedSignal:N(name)` reasons, plus an expanded `PendingReason` vocabulary.

Previously Spur reported a bare lossy `ExitCode` (signal collapsed to `-1` / `128+sig`), had no `DerivedExitCode` field, and never set a completion reason — silently corrupting outcome data for anything scraping `scontrol show job` / `sacct`.

## What changed

Exit status is threaded from the agent's process wait status through the report RPC, WAL, core `Job`, display RPC, and CLI:

- Agent (`spurd`) decodes the raw `WaitStatus` into `(exit_code, signal)` instead of collapsing it; `WIFEXITED -> code:0`, `WIFSIGNALED -> 0:sig`.
- New `exit_signal` and `derived_exit_code` fields on the core `Job` and proto `JobInfo`; `JobComplete` / `JobNodeComplete` WAL ops carry the signal.
- Controller sets the failure reason from the batch/primary node's wait status: `NonZeroExitCode` on non-zero exit, `RaisedSignal` on signal death.
- `DerivedExitCode` is the running max over `srun` step exit codes (Slurm semantics), maintained live and durably via a new `JobStepComplete` WAL op; a job with no srun steps reports `0:0`.
- CLI renders `ExitCode=code:signal`, `DerivedExitCode=code:signal`, and composes the dynamic `RaisedSignal:N(name)` string at display time.
- Expanded `PendingReason` toward the Slurm vocabulary (`NonZeroExitCode`, `RaisedSignal`, `OutOfMemory`, `JobLaunchFailure`, ...).

Two adjacent fixes uncovered during live verification are included: jobs now start with default signal dispositions (so `kill -INT $$` is reported as `RaisedSignal:2` rather than being absorbed by spurd's inherited `SIG_IGN`), and the Spur CLI bin dir is added to the job `PATH` (derived from the agent's own location) so `srun` resolves inside batch scripts.

## Exact Slurm semantics implemented

`ExitCode` is the batch/primary process's raw wait status, not a max. A step killed by a signal surfaces to the batch shell as exit `128+sig`, so it appears as `NonZeroExitCode` with `ExitCode=128+sig:0`, not `RaisedSignal`. `RaisedSignal` only applies when the batch process itself is signaled. `DerivedExitCode` is the max exit code across steps.

## Testing

Unit and integration coverage across `spur-core`, `spurctld`, `spurd`, and `spur-cli` (derived-completion rules, signal decoding, reason emission, RPC-level signaled-completion recovery, WAL signal round-trip, step running-max). Full suite passes; `cargo fmt --check --all` and `clippy` clean.

Verified live against Slurm 25.11.6 on a single-node testbed: clean exit, single non-zero, multi-step `srun` ordering (`ExitCode=last`, `DerivedExitCode=max`), no-srun (`DerivedExitCode=0:0`), and self-signal cases (`SIGINT`/`SIGTERM`/`SIGABRT` -> `RaisedSignal:N(name)`), including the live running-max visible in `scontrol show job` mid-run.

## Out of scope (follow-ups)

`sacct` step history in `spurdbd` (live `DerivedExitCode` works via the controller; persisting per-step rows to the accounting DB is separate — tracked in #297). k8s backend signal reporting (currently `signal: 0`; the controller-side machinery already applies). A separately filed controller robustness issue (#273, openraft snapshot/log compaction) is unrelated to this change.

## Design docs

Design and implementation plan included under `docs/superpowers/`.

## Live verification vs Slurm 25.11.6

Ran an expanded matrix on a Slurm 25.11.6 cluster and a single-node Spur cluster, submitting sequentially (each job to a terminal state before the next) to avoid cluster-saturation races. Format per cell: `State / ExitCode / DerivedExitCode / Reason`.

| # | Case | Slurm | Spur | Match |
|---|------|-------|------|:---:|
| 1 | `exit 0` | COMPLETED / 0:0 / 0:0 / None | COMPLETED / 0:0 / 0:0 / None | ✅ |
| 2 | `exit 13` | FAILED / 13:0 / 0:0 / NonZeroExitCode | FAILED / 13:0 / 0:0 / NonZeroExitCode | ✅ |
| 3 | `exit 200` | FAILED / 200:0 / 0:0 / NonZeroExitCode | FAILED / 200:0 / 0:0 / NonZeroExitCode | ✅ |
| 4 | `exit 137` (plain) | FAILED / 137:0 / 0:0 / NonZeroExitCode | FAILED / 137:0 / 0:0 / NonZeroExitCode | ✅ |
| 5 | srun 1,4,9 | FAILED / 9:0 / 9:0 / NonZeroExitCode | FAILED / 9:0 / 9:0 / NonZeroExitCode | ✅ |
| 6 | srun 9,4,1 | FAILED / 1:0 / 9:0 / NonZeroExitCode | FAILED / 1:0 / 9:0 / NonZeroExitCode | ✅ |
| 7 | srun 0,8; batch exit 0 | COMPLETED / 0:0 / 8:0 / None | COMPLETED / 0:0 / 8:0 / None | ✅ |
| 8 | srun 0,0; exit 0 | COMPLETED / 0:0 / 0:0 / None | COMPLETED / 0:0 / 0:0 / None | ✅ |
| 9 | batch SIGINT | FAILED / 0:2 / 0:0 / RaisedSignal:2(Interrupt) | FAILED / 0:2 / 0:0 / RaisedSignal:2(Interrupt) | ✅ |
| 10 | batch SIGTERM | FAILED / 0:15 / 0:0 / RaisedSignal:15(Terminated) | FAILED / 0:15 / 0:0 / RaisedSignal:15(Terminated) | ✅ |
| 11 | batch SIGKILL | FAILED / 0:9 / 0:0 / RaisedSignal:9(Killed) | FAILED / 0:9 / 0:0 / RaisedSignal:9(Killed) | ✅ |
| 12 | batch SIGSEGV | FAILED / 0:11 / 0:0 / RaisedSignal:11(Segmentation_fault) | FAILED / 0:11 / 0:0 / RaisedSignal:11(Segmentation_fault) | ✅ |
| 13 | batch SIGABRT | FAILED / 0:6 / 0:0 / RaisedSignal:6(Aborted) | FAILED / 0:6 / 0:0 / RaisedSignal:6(Aborted) | ✅ |

Key behaviors confirmed identical: `ExitCode` follows the batch script's last command (case 5 = 9:0 vs case 6 = 1:0 — same steps, reordered), `DerivedExitCode` is the max over srun step exit codes (cases 5–7), a plain `exit 137` stays `137:0` (not mis-read as signal 9), and all batch signals report `0:sig` with `RaisedSignal:N(name)`.

### Note on DerivedExitCode signal half

Spur tracks `DerivedExitCode` as the running max over srun step *exit codes* (the exit half), which matches Slurm in every stable case above. Slurm additionally folds a signaled step's *signal* into the DerivedExitCode signal half in some cases, but that behavior is order-dependent and inconsistent in Slurm itself (e.g. a signaled step followed by an exiting step yields `0:0`; with shell control flow it instead surfaces as `ExitCode=0:sig`). Replicating it would mean matching undefined Slurm internals, so Spur deliberately does not — the exit-half parity is what consumers rely on.

### DIFF
--- a/crates/spur-cli/src/scontrol.rs
+++ b/crates/spur-cli/src/scontrol.rs
@@ -255,7 +255,7 @@ async fn show(controller: &str, entity: &str, name: Option<&str>) -> Result<()> 
                 println!(
                     "   JobState={} Reason={}",
                     state_name(job.state),
-                    job.state_reason
+                    render_reason(&job.state_reason, job.exit_signal),
                 );
                 println!(
                     "   NumNodes={} NumTasks={} CPUs/Task={}",
@@ -272,7 +272,12 @@ async fn show(controller: &str, entity: &str, name: Option<&str>) -> Result<()> 
                 );
                 println!("   WorkDir={}", job.work_dir);
                 println!("   StdOut={} StdErr={}", job.stdout_path, job.stderr_path);
-                println!("   ExitCode={} Priority={}", job.exit_code, job.priority);
+                println!(
+                    "   ExitCode={} DerivedExitCode={} Priority={}",
+                    format_exit(job.exit_code, job.exit_signal),
+                    format_exit(job.derived_exit_code, 0),
+                    job.priority
+                );
                 println!();
             }
         }
@@ -475,6 +480,34 @@ fn format_ts(ts: Option<&prost_types::Timestamp>) -> String {
     }
 }
 
+fn format_exit(code: i32, signal: i32) -> String {
+    format!("{}:{}", code, signal)
+}
+
+fn signal_name(sig: i32) -> String {
+    let name = match sig {
+        1 => "Hangup",
+        2 => "Interrupt",
+        6 => "Aborted",
+        9 => "Killed",
+        11 => "Segmentation fault",
+        13 => "Broken pipe",
+        15 => "Terminated",
+        _ => return format!("Signal {}", sig),
+    };
+    name.to_string()
+}
+
+/// Compose the displayed reason. For RaisedSignal, append `:N(name)` from the
+/// job's terminating signal, matching Slurm (`RaisedSignal:9(Killed)`).
+fn render_reason(reason: &str, signal: i32) -> String {
+    if reason == "RaisedSignal" && signal != 0 {
+        format!("RaisedSignal:{}({})", signal, signal_name(signal))
+    } else {
+        reason.to_string()
+    }
+}
+
 async fn send_job_update(controller: &str, req: spur_proto::proto::UpdateJobRequest) -> Result<()> {
     let hold = req.hold;
     let job_id = req.job_id;
@@ -656,4 +689,30 @@ async fn delete_reservation(controller: &str, name: &str) -> Result<()> {
 
     println!("Reservation {} deleted", name);
     Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn format_exit_code_pairs() {
+        assert_eq!(format_exit(2, 0), "2:0");
+        assert_eq!(format_exit(0, 9), "0:9");
+    }
+
+    #[test]
+    fn signal_name_lookup() {
+        assert_eq!(signal_name(9), "Killed");
+        assert_eq!(signal_name(15), "Terminated");
+        assert_eq!(signal_name(11), "Segmentation fault");
+        assert_eq!(signal_name(99), "Signal 99");
+    }
+
+    #[test]
+    fn reason_with_signal_composes_raisedsignal() {
+        assert_eq!(render_reason("RaisedSignal", 9), "RaisedSignal:9(Killed)");
+        assert_eq!(render_reason("NonZeroExitCode", 0), "NonZeroExitCode");
+        assert_eq!(render_reason("None", 0), "None");
+    }
 }
--- a/crates/spur-cli/src/scontrol.rs
+++ b/crates/spur-cli/src/scontrol.rs
@@ -5,6 +5,8 @@ use anyhow::{bail, Context, Result};
 use clap::{Parser, Subcommand};
 use spur_proto::proto::slurm_controller_client::SlurmControllerClient;
 
+use crate::exit_fmt::{format_exit, render_reason};
+
 /// Administrative control commands.
 #[derive(Parser, Debug)]
 #[command(name = "scontrol", about = "Administrative control for Spur")]
@@ -480,34 +482,6 @@ fn format_ts(ts: Option<&prost_types::Timestamp>) -> String {
     }
 }
 
-fn format_exit(code: i32, signal: i32) -> String {
-    format!("{}:{}", code, signal)
-}
-
-fn signal_name(sig: i32) -> String {
-    let name = match sig {
-        1 => "Hangup",
-        2 => "Interrupt",
-        6 => "Aborted",
-        9 => "Killed",
-        11 => "Segmentation fault",
-        13 => "Broken pipe",
-        15 => "Terminated",
-        _ => return format!("Signal {}", sig),
-    };
-    name.to_string()
-}
-
-/// Compose the displayed reason. For RaisedSignal, append `:N(name)` from the
-/// job's terminating signal, matching Slurm (`RaisedSignal:9(Killed)`).
-fn render_reason(reason: &str, signal: i32) -> String {
-    if reason == "RaisedSignal" && signal != 0 {
-        format!("RaisedSignal:{}({})", signal, signal_name(signal))
-    } else {
-        reason.to_string()
-    }
-}
-
 async fn send_job_update(controller: &str, req: spur_proto::proto::UpdateJobRequest) -> Result<()> {
     let hold = req.hold;
     let job_id = req.job_id;
@@ -689,30 +663,4 @@ async fn delete_reservation(controller: &str, name: &str) -> Result<()> {
 
     println!("Reservation {} deleted", name);
     Ok(())
-}
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-
-    #[test]
-    fn format_exit_code_pairs() {
-        assert_eq!(format_exit(2, 0), "2:0");
-        assert_eq!(format_exit(0, 9), "0:9");
-    }
-
-    #[test]
-    fn signal_name_lookup() {
-        assert_eq!(signal_name(9), "Killed");
-        assert_eq!(signal_name(15), "Terminated");
-        assert_eq!(signal_name(11), "Segmentation fault");
-        assert_eq!(signal_name(99), "Signal 99");
-    }
-
-    #[test]
-    fn reason_with_signal_composes_raisedsignal() {
-        assert_eq!(render_reason("RaisedSignal", 9), "RaisedSignal:9(Killed)");
-        assert_eq!(render_reason("NonZeroExitCode", 0), "NonZeroExitCode");
-        assert_eq!(render_reason("None", 0), "None");
-    }
 }
--- a/crates/spur-core/src/job.rs
+++ b/crates/spur-core/src/job.rs
@@ -209,6 +209,17 @@ pub enum PendingReason {
     BeginTime,
     DeadLine,
     Licenses,
+    NonZeroExitCode,
+    RaisedSignal,
+    JobLaunchFailure,
+    JobHeldAdmin,
+    BadConstraints,
+    PartitionInactive,
+    DependencyNeverSatisfied,
+    InvalidAccount,
+    InvalidQOS,
+    BootFail,
+    OutOfMemory,
 }
 
 impl PendingReason {
@@ -228,6 +239,17 @@ impl PendingReason {
             Self::BeginTime => "BeginTime",
             Self::DeadLine => "DeadLine",
             Self::Licenses => "Licenses",
+            Self::NonZeroExitCode => "NonZeroExitCode",
+            Self::RaisedSignal => "RaisedSignal",
+            Self::JobLaunchFailure => "JobLaunchFailure",
+            Self::JobHeldAdmin => "JobHeldAdmin",
+            Self::BadConstraints => "BadConstraints",
+            Self::PartitionInactive => "PartitionInactive",
+            Self::DependencyNeverSatisfied => "DependencyNeverSatisfied",
+            Self::InvalidAccount => "InvalidAccount",
+            Self::InvalidQOS => "InvalidQOS",
+            Self::BootFail => "BootFailure",
+            Self::OutOfMemory => "OutOfMemory",
         }
     }
 }
@@ -941,6 +963,18 @@ mod tests {
         // Slurm reports this exact string ("DeadLine", note the cap D and L).
         // squeue scrapers and Slurm-compat clients pattern-match on it.
         assert_eq!(PendingReason::DeadLine.display(), "DeadLine");
+    }
+
+    #[test]
+    fn pending_reason_exit_vocabulary_display() {
+        assert_eq!(PendingReason::NonZeroExitCode.display(), "NonZeroExitCode");
+        assert_eq!(PendingReason::RaisedSignal.display(), "RaisedSignal");
+        assert_eq!(
+            PendingReason::JobLaunchFailure.display(),
+            "JobLaunchFailure"
+        );
+        assert_eq!(PendingReason::OutOfMemory.display(), "OutOfMemory");
+        assert_eq!(PendingReason::BootFail.display(), "BootFailure");
     }
 
     #[test]
--- a/crates/spur-core/src/job.rs
+++ b/crates/spur-core/src/job.rs
@@ -475,9 +475,8 @@ pub struct Job {
     /// Terminating signal of the primary node's process (0 = none).
     #[serde(default)]
     pub exit_signal: Option<i32>,
-    /// Slurm `DerivedExitCode`: the running max over the job's srun step exit
-    /// codes, accumulated by `WalOperation::JobStepComplete`. `None`/0 for a job
-    /// with no srun steps.
+    /// Slurm `DerivedExitCode`: running max over srun step exit codes (via
+    /// `JobStepComplete`); `None`/0 when the job ran no srun steps.
     #[serde(default)]
     pub derived_exit_code: Option<i32>,
 
@@ -542,16 +541,12 @@ impl Job {
     /// signal); state is `Failed` if the primary exited non-zero or was
     /// signaled, else `Completed`.
     ///
-    /// If `primary_node` is absent from the map, falls back to the worst
-    /// completion (a signaled node, or the highest exit code) so a failure is
-    /// never reported as success.
+    /// If `primary_node` is absent, falls back to the worst completion (a
+    /// signaled node, or the highest exit code) so a failure isn't masked.
     ///
-    /// The 4th return value is a node-based max kept only for backward
-    /// compatibility; the job's real Slurm `DerivedExitCode` is the running max
-    /// over srun *steps*, maintained separately via `JobStepComplete`. Callers
-    /// finalizing a job should NOT use this value for `derived_exit_code`.
-    ///
-    /// Returns `(state, exit_code, exit_signal, node_max_exit)`.
+    /// Returns `(state, exit_code, exit_signal, node_max_exit)`. The 4th value
+    /// is a legacy node-based max — NOT the job's DerivedExitCode, which is the
+    /// running max over srun steps maintained via `JobStepComplete`.
     pub fn derived_completion(
         node_completions: &HashMap<String, NodeCompletion>,
         primary_node: &str,
--- a/crates/spur-core/src/job.rs
+++ b/crates/spur-core/src/job.rs
@@ -419,6 +419,16 @@ impl Default for JobSpec {
     }
 }
 
+/// One node's completion outcome for a job: the raw process wait status,
+/// split into exit code and terminating signal (0 = none).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Serialize, Deserialize)]
+pub struct NodeCompletion {
+    #[serde(default)]
+    pub code: i32,
+    #[serde(default)]
+    pub signal: i32,
+}
+
 /// Internal job record held by the controller.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Job {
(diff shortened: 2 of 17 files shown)


crates/spur-core/src/job.rs
```diff
@@ -487,23 +526,56 @@ impl Job {
...
             .max()
             .unwrap_or(0);
-        let state = JobState::completion_state_for_exit_code(exit_code);
-        (state, exit_code)
+
+        let primary = node_completions.get(primary_node).copied().or_else(|| {
+            node_completions
+                .values()
+                .filter(|c| c.code != 0 || c.signal != 0)
+                .max_by_key(|c| c.code)
```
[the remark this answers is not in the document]
yansun1996: Hardened in ee55207. The `None` arm only triggers when no node failed or was signaled (so reporting Completed there is correct), but you're right the fallback could drop a signal in favor of a higher exit code — the ranking now prefers a signaled node (`max_by_key((signal != 0, code, signal))`). Note this fallback only runs if the primary node's completion is missing, which can't happen once `all_nodes_completed()` holds; it's defensive. Added test `derived_completion_missing_primary_prefers_signaled`.

crates/spur-core/src/job.rs
```diff
@@ -440,6 +472,13 @@ pub struct Job {
 
     pub exit_code: Option<i32>,
 
+    /// Terminating signal of the primary node's process (0 = none).
+    #[serde(default)]
+    pub exit_signal: Option<i32>,
+    /// Max exit code across all node completions (Slurm DerivedExitCode).
+    #[serde(default)]
+    pub derived_exit_code: Option<i32>,
```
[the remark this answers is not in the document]
yansun1996: Fixed in ee55207 (tightened in 69b42f9). The field doc now says it's the running max over srun step exit codes, accumulated via `JobStepComplete`.

crates/spur-core/src/job.rs
```diff
@@ -487,23 +526,56 @@ impl Job {
...
     }
 
-    /// Derive final job state and exit code from per-node completion reports.
-    pub fn derived_completion(node_completions: &HashMap<String, i32>) -> (JobState, i32) {
-        let exit_code = node_completions
+    /// Derive the final job outcome from per-node completions, matching Slurm:
+    /// `ExitCode` is the primary node's raw wait status (exit_code, signal);
+    /// `DerivedExitCode` is the max exit code across all nodes. State is
+    /// `Failed` if the primary exited non-zero or was signaled, else `Completed`.
+    ///
```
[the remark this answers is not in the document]
yansun1996: Fixed in ee55207. The `derived_completion` docstring now states its 4th return value is a legacy node-based max, explicitly NOT the job's DerivedExitCode (which is the running max over srun steps via `JobStepComplete`), and that finalizing callers must not use it for `derived_exit_code`.

crates/spurd/src/executor.rs
```diff
@@ -409,6 +422,20 @@ async fn spawn_job_process(
...
+    // ignore mask — which would make a job's own `kill -INT $$` a no-op and break
+    // Slurm-parity signal reporting (e.g. SIGINT -> RaisedSignal:2). The job must
+    // start with default handlers.
+    unsafe {
+        cmd.pre_exec(|| {
+            for sig in [libc::SIGINT, libc::SIGQUIT, libc::SIGHUP, libc::SIGPIPE] {
+                libc::signal(sig, libc::SIG_DFL);
+            }
+            Ok(())
+        });
```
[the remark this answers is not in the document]
yansun1996: Fixed in ee55207 — switched to nix `sigaction` (async-signal-safe) instead of `signal()` for resetting dispositions in `pre_exec`, since it runs post-fork in a multi-threaded process.

proto/slurm.proto
```diff
@@ -183,6 +183,8 @@ message JobInfo {
   string command = 25;
 
   int32 exit_code = 30;
+  int32 exit_signal = 33;       // Terminating signal of primary node (0 = none)
+  int32 derived_exit_code = 34; // Max exit code across all job steps/nodes
```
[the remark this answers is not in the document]
yansun1996: Fixed in ee55207 — the proto comment now says "running max over srun step exit codes (0 if no steps)".

crates/spur-core/src/wal.rs
```diff
@@ -79,3 +91,34 @@ pub enum WalOperation {
...
+            WalOperation::JobNodeComplete {
+                signal, exit_code, ..
+            } => {
+                assert_eq!(signal, 0);
+                assert_eq!(exit_code, 3);
+            }
+            _ => panic!("wrong variant"),
+        }
+        // round-trip new form (WalOperation does not derive PartialEq)
+        let _ = (op, back);
```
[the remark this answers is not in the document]
yansun1996: Fixed in ee55207 — the test now matches on the deserialized `JobNodeComplete` and asserts job_id/node_name/exit_code/signal, so a serialization regression is caught (WalOperation has no PartialEq, hence field asserts rather than value equality).

crates/spur-core/src/job.rs
```diff
@@ -487,23 +527,60 @@ impl Job {
...
+    /// If `primary_node` is absent, falls back to the worst completion (a
+    /// signaled node, or the highest exit code) so a failure isn't masked.
+    ///
+    /// Returns `(state, exit_code, exit_signal, node_max_exit)`. The 4th value
+    /// is a legacy node-based max — NOT the job's DerivedExitCode, which is the
+    /// running max over srun steps maintained via `JobStepComplete`.
+    pub fn derived_completion(
+        node_completions: &HashMap<String, NodeCompletion>,
+        primary_node: &str,
+    ) -> (JobState, i32, i32, i32) {
```
shiv-tyagi: Returning a 4-tuple `(JobState, i32, i32, i32)` is asking for trouble. Every call site has to destructure positionally, and a reorder silently breaks things. The 4th value (`node_max_exit`) is unused in production code (both callers bind it to `_`), which makes the positional game even riskier.

No production caller reads the 4th field, so it's dead weight in the return type. Either drop it entirely, or return a named struct for the remaining fields so each one is self-documenting.

crates/spur-cli/src/scontrol.rs
```diff
@@ -475,6 +480,34 @@ fn format_ts(ts: Option<&prost_types::Timestamp>) -> String {
     }
 }
 
+fn format_exit(code: i32, signal: i32) -> String {
```
shiv-tyagi: `sacct.rs` still has its own `format_exit_code` that derives the signal half by negating the exit code (positive = `code:0`, negative = `0:-code`). That's the exact lossy encoding this PR eliminates everywhere else. `sacct` should use the new `exit_signal` / `derived_exit_code` proto fields and this `format_exit` helper.

Similarly, `squeue` renders `state_reason` raw without `render_reason`, so a `RaisedSignal` job will show bare `"RaisedSignal"` there instead of `RaisedSignal:9(Killed)`.

crates/spur-cli/src/scontrol.rs
```diff
@@ -475,6 +480,34 @@ fn format_ts(ts: Option<&prost_types::Timestamp>) -> String {
...
+        15 => "Terminated",
+        _ => return format!("Signal {}", sig),
+    };
+    name.to_string()
+}
+
+/// Compose the displayed reason. For RaisedSignal, append `:N(name)` from the
+/// job's terminating signal, matching Slurm (`RaisedSignal:9(Killed)`).
+fn render_reason(reason: &str, signal: i32) -> String {
+    if reason == "RaisedSignal" && signal != 0 {
```
shiv-tyagi: This string-compares the wire value `"RaisedSignal"` against a hardcoded literal. If `PendingReason::display()` changes its output or gets a typo, this silently stops composing the signal suffix and nobody notices.

Consider matching against the `PendingReason` enum directly (parse the wire string first), or at minimum pull the literal into a shared constant.

crates/spur-core/src/wal.rs
```diff
@@ -37,12 +37,24 @@ pub enum WalOperation {
     JobComplete {
         job_id: JobId,
         exit_code: i32,
+        // reserved; JobComplete paths carry no process signal (cancel/deadline/timeout)
+        #[serde(default)]
+        signal: i32,
```
shiv-tyagi: This `signal` field is hardcoded to `0` at every construction site and explicitly ignored (`signal: _`) in the apply handler. If it's reserved for future use, it shouldn't need `#[serde(default)]` (there are no old WAL entries with a `JobComplete` that ever carried a signal). If it's not planned for future use, just drop it.

As-is it adds noise to every `JobComplete` construction (7+ sites) for no functional purpose.

crates/spur-core/src/job.rs
```diff
@@ -440,6 +472,14 @@ pub struct Job {
 
     pub exit_code: Option<i32>,
 
+    /// Terminating signal of the primary node's process (0 = none).
+    #[serde(default)]
+    pub exit_signal: Option<i32>,
+    /// Slurm `DerivedExitCode`: running max over srun step exit codes (via
+    /// `JobStepComplete`); `None`/0 when the job ran no srun steps.
+    #[serde(default)]
+    pub derived_exit_code: Option<i32>,
```
shiv-tyagi: We're pre-1.0 with no deployed WAL/snapshot format to protect. These `#[serde(default)]` annotations and `Option` wrappers add complexity for a migration that doesn't exist:

1. `NodeCompletion` is a brand new struct introduced by this PR. There are no old serialized instances without `code`/`signal`. The `#[serde(default)]` on both fields is dead weight.
2. `exit_signal` and `derived_exit_code` are new fields. Every `Job` created after this lands will have them. Using `Option<i32>` forces `.unwrap_or(0)` at every read site and produces the awkward `Some(job.derived_exit_code.unwrap_or(0))` on finalization. Plain `i32` with `#[serde(default)]` (defaults to `0`) would eliminate all the `Option` gymnastics.
3. The `node_completions` type changed from `HashMap<String, i32>` to `HashMap<String, NodeCompletion>`, which is serde-incompatible anyway. An old snapshot with `{"n1": 42}` won't deserialize into the new type. So the `#[serde(default)]` on `NodeCompletion` fields isn't saving us from anything.

Same applies to the WAL: `JobNodeComplete.signal` gets `#[serde(default)]` and a dedicated compat test (`job_node_complete_carries_signal_and_old_logs_default_zero`) that constructs fake old-format JSON. There are no old WAL entries to replay.

Suggestion: drop the `Option` wrappers on the new fields (use plain `i32`), remove `#[serde(default)]` from `NodeCompletion`, and drop the old-format WAL deserialization test.

shiv-tyagi (commented):
Solid work. The signal threading end-to-end is clean, the Slurm parity matrix is thorough, and the adjacent fixes (signal disposition reset, PATH injection) are well-motivated.

crates/spur-cli/src/scontrol.rs
shiv-tyagi: To be fair, the string `"RaisedSignal"` is effectively part of the proto wire contract. The coupling is inherent to passing enums as strings over proto. Not blocking, but worth keeping in mind if the reason vocabulary grows.

proto/slurm.proto
```diff
@@ -183,6 +183,8 @@ message JobInfo {
   string command = 25;
 
   int32 exit_code = 30;
+  int32 exit_signal = 33;       // Terminating signal of primary node (0 = none)
+  int32 derived_exit_code = 34; // Running max over srun step exit codes (0 if no steps)
```
shiv-tyagi: Please move this down, after 31, 32

crates/spur-core/src/job.rs
yansun1996: Done in cd64a89. `derived_completion` now returns `(JobState, i32, i32)` — dropped the 4th node-based max, which no production caller read and which was easy to mistake for the job's DerivedExitCode (that's the running max over srun steps via `JobStepComplete`, not a node value). Both callers and the tests updated.

crates/spur-cli/src/scontrol.rs
yansun1996: Done in cd64a89. Factored `format_exit`/`signal_name`/`render_reason` into a new shared `spur-cli/src/exit_fmt.rs`. squeue `%r` now routes through `render_reason`, so a signaled job shows `RaisedSignal:9(Killed)` there too (verified live). sacct drops the negate-trick `format_exit_code` for the shared `code:signal` form. Caveat: sacct reads from spurdbd, which still hardcodes `exit_signal: 0` — so the signal half renders `:0` until per-step persistence lands (the out-of-scope spurdbd follow-up noted in the PR description).

crates/spur-cli/src/scontrol.rs
yansun1996: Addressed in cd64a89. `render_reason` now compares against `PendingReason::RaisedSignal.display()` instead of a bare `"RaisedSignal"` literal, so it tracks the enum. Agree with your follow-up that the string is inherently part of the proto wire contract once reasons cross as strings — this just removes the silent-drift risk if `display()` ever changes.

crates/spur-core/src/wal.rs
yansun1996: Done in cd64a89. Dropped the field entirely — it was hardcoded `0` at every site and ignored on apply, so it was pure noise. Removed from the variant, the apply destructure, and all 11 construction sites. (`JobNodeComplete.signal` stays — that one carries real data.)

crates/spur-core/src/job.rs
yansun1996: Done in cd64a89. `exit_signal` and `derived_exit_code` are now plain `i32` with `#[serde(default)]` — kills the `unwrap_or(0)`/`Some(...)` gymnastics. Dropped `#[serde(default)]` from the new `NodeCompletion` fields, and removed the fabricated old-format WAL test (replaced with a clean round-trip assertion, keeping the serialization-regression coverage Copilot asked for). Agreed there's no deployed format to protect pre-1.0.

proto/slurm.proto
yansun1996: Done in cd64a89. `exit_signal`/`derived_exit_code` (33/34) now follow `stdout_path`/`stderr_path` (31/32).

yansun1996:
Thanks for the review @shiv-tyagi — all six addressed in cd64a89 (replies inline). Net −33 LOC; full suite + clippy + fmt clean.

**Re-verified live** against Slurm 25.11.6 (single-node Spur vs Slurm cluster), re-running the full exit-code matrix — State / ExitCode / DerivedExitCode / Reason match across all 13 cases, and squeue `%r` now composes the signal suffix.

**One finding worth flagging:** live testing showed Slurm renders multi-word signal names with underscores — `RaisedSignal:11(Segmentation_fault)` and `RaisedSignal:13(Broken_pipe)` — whereas the original `signal_name` table (and matrix row 12 in the PR description, marked ✅) used spaces. That row was inaccurate; the space form was never a real match. Fixed `signal_name` to use underscores and re-verified live, so it's now truly identical. I'll correct row 12 in the PR description.

shiv-tyagi (approved):

PR #276, opened 2026-06-12
feat: node labels and dual-path partition matching

shiv-tyagi:
## Summary

- Nodes can carry arbitrary key-value labels, set at registration via `--label` / `SPUR_NODE_LABELS` or dynamically via `spur node label key=value`
- Partition membership is determined by a union of hostname pattern (`nodes=`) OR label selector (`selector = { key = "value" }`) — both can be specified
- `nodes = "ALL"` is treated as a true wildcard matching any node
- Labels are durable (WAL-backed), sync on re-registration, and trigger partition re-routing and NodeConfig feature/weight re-application when changed
- Labels are visible in `scontrol show node` output
- NodeInfo proto changed from `string partition` to `repeated string partitions` — one record per node with all partitions listed (matches Slurm REST API design, fixes double-counting in spur-cloud GPU pools)
- sinfo does client-side fan-out for partition display; scontrol shows `Partitions=x,y`

## Motivation

Static hostname-based partition membership requires config changes and controller restarts when nodes are added or repurposed. Label selectors allow dynamic partition routing — nodes self-describe their role at registration and the controller routes them to the correct partition without config changes.

## Test plan

- [x] 10 unit tests covering matching logic, label updates, re-registration, and NodeConfig application
- [x] Integration tested on 3-node bare-metal cluster: registration with labels, selector routing, admin label set/remove, re-registration sync, ALL wildcard
- [x] 4 e2e tests (pytest) covering label registration, selector partition routing, ALL wildcard inclusion, and admin label mutation
- [x] Verified sinfo/scontrol/REST output matches Slurm 25.11 behavior for multi-partition nodes

### DIFF
--- a/crates/spur-cli/src/node.rs
+++ b/crates/spur-cli/src/node.rs
@@ -60,16 +60,16 @@ fn parse_label_args(label_args: &[String]) -> Result<(HashMap<String, String>, V
     let mut remove_labels: Vec<String> = Vec::new();
 
     for arg in label_args {
-        if let Some(key) = arg.strip_suffix('-') {
-            if key.is_empty() {
-                bail!("invalid label removal: '{arg}'");
-            }
-            remove_labels.push(key.to_string());
-        } else if let Some((k, v)) = arg.split_once('=') {
+        if let Some((k, v)) = arg.split_once('=') {
             if k.is_empty() {
                 bail!("invalid label: '{arg}', key cannot be empty");
             }
             set_labels.insert(k.to_string(), v.to_string());
+        } else if let Some(key) = arg.strip_suffix('-') {
+            if key.is_empty() {
+                bail!("invalid label removal: '{arg}'");
+            }
+            remove_labels.push(key.to_string());
         } else {
             bail!("invalid label format: '{arg}', expected key=value or key-");
         }
@@ -147,5 +147,13 @@ mod tests {
     fn test_parse_label_invalid_format() {
         let args = vec!["noequalsnodash".to_string()];
         assert!(parse_label_args(&args).is_err());
+    }
+
+    #[test]
+    fn test_parse_label_value_ending_in_dash() {
+        let args = vec!["env=prod-".to_string()];
+        let (set, remove) = parse_label_args(&args).unwrap();
+        assert_eq!(set.get("env").unwrap(), "prod-");
+        assert!(remove.is_empty());
     }
 }
--- a/crates/spur-cli/src/node.rs
+++ b/crates/spur-cli/src/node.rs
@@ -1,0 +1,98 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! `spur node` subcommands for node lifecycle management.
+
+use std::collections::HashMap;
+
+use anyhow::{bail, Result};
+use clap::{Parser, Subcommand};
+
+use spur_proto::proto::slurm_controller_client::SlurmControllerClient;
+use spur_proto::proto::UpdateNodeRequest;
+
+/// Node management commands.
+#[derive(Parser, Debug)]
+#[command(name = "node", about = "Manage cluster nodes")]
+pub struct NodeArgs {
+    /// Controller address
+    #[arg(
+        long,
+        env = "SPUR_CONTROLLER_ADDR",
+        default_value = "http://localhost:6817",
+        global = true
+    )]
+    controller: String,
+
+    #[command(subcommand)]
+    pub command: NodeCommand,
+}
+
+#[derive(Subcommand, Debug)]
+pub enum NodeCommand {
+    /// Set or remove labels on a node.
+    ///
+    /// Labels are key=value pairs used for partition routing.
+    /// Append a trailing dash to remove a label (e.g., "pool-").
+    Label {
+        /// Node name
+        node: String,
+        /// Labels to set (key=value) or remove (key-)
+        #[arg(required = true)]
+        labels: Vec<String>,
+    },
+}
+
+pub async fn main() -> Result<()> {
+    main_with_args(std::env::args().collect()).await
+}
+
+pub async fn main_with_args(args: Vec<String>) -> Result<()> {
+    let parsed = NodeArgs::try_parse_from(args)?;
+    let controller = parsed.controller;
+    match parsed.command {
+        NodeCommand::Label { node, labels } => cmd_label(&controller, node, labels).await,
+    }
+}
+
+async fn cmd_label(controller: &str, node: String, label_args: Vec<String>) -> Result<()> {
+    let mut client = SlurmControllerClient::connect(controller.to_string()).await?;
+
+    let mut set_labels: HashMap<String, String> = HashMap::new();
+    let mut remove_labels: Vec<String> = Vec::new();
+
+    for arg in &label_args {
+        if let Some(key) = arg.strip_suffix('-') {
+            if key.is_empty() {
+                bail!("invalid label removal: '{arg}'");
+            }
+            remove_labels.push(key.to_string());
+        } else if let Some((k, v)) = arg.split_once('=') {
+            if k.is_empty() {
+                bail!("invalid label: '{arg}', key cannot be empty");
+            }
+            set_labels.insert(k.to_string(), v.to_string());
+        } else {
+            bail!("invalid label format: '{arg}', expected key=value or key-");
+        }
+    }
+
+    client
+        .update_node(UpdateNodeRequest {
+            name: node.clone(),
+            state: None,
+            reason: None,
+            labels: set_labels.clone(),
+            remove_labels: remove_labels.clone(),
+        })
+        .await?;
+
+    for (k, v) in &set_labels {
+        println!("  {node}: {k}={v}");
+    }
+    for k in &remove_labels {
+        println!("  {node}: {k} removed");
+    }
+
+    Ok(())
+}
--- a/crates/spur-cli/src/node.rs
+++ b/crates/spur-cli/src/node.rs
@@ -55,13 +55,11 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
     }
 }
 
-async fn cmd_label(controller: &str, node: String, label_args: Vec<String>) -> Result<()> {
-    let mut client = SlurmControllerClient::connect(controller.to_string()).await?;
-
+fn parse_label_args(label_args: &[String]) -> Result<(HashMap<String, String>, Vec<String>)> {
     let mut set_labels: HashMap<String, String> = HashMap::new();
     let mut remove_labels: Vec<String> = Vec::new();
 
-    for arg in &label_args {
+    for arg in label_args {
         if let Some(key) = arg.strip_suffix('-') {
             if key.is_empty() {
                 bail!("invalid label removal: '{arg}'");
@@ -76,6 +74,13 @@ async fn cmd_label(controller: &str, node: String, label_args: Vec<String>) -> R
             bail!("invalid label format: '{arg}', expected key=value or key-");
         }
     }
+
+    Ok((set_labels, remove_labels))
+}
+
+async fn cmd_label(controller: &str, node: String, label_args: Vec<String>) -> Result<()> {
+    let mut client = SlurmControllerClient::connect(controller.to_string()).await?;
+    let (set_labels, remove_labels) = parse_label_args(&label_args)?;
 
     client
         .update_node(UpdateNodeRequest {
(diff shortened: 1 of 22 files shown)


tests/e2e/native_host/cluster.py
```diff
@@ -661,12 +668,18 @@ def _spurd_start_cmd(self, node_index: int) -> str:
             if self.agent_as_root
             else f"'{self.bin_dir}/spurd'"
         )
+        label_args = ""
+        labels = self.agent_labels.get(node_index, {})
+        if labels:
+            label_args = " ".join(f"--label '{k}={v}'" for k, v in labels.items())
+            label_args = f" {label_args}"
```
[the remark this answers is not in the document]
shiv-tyagi: Acknowledged — however this is test-only code with fully controlled label values defined in our fixtures. No user input reaches this path. Skipping for now.

crates/spurctld/src/cluster.rs
```diff
@@ -2021,6 +2074,43 @@ impl ClusterManager {
...
+                    // Re-apply NodeConfig features/weight
+                    for nc in &self.config.nodes {
+                        if node_config_matches(nc, &node.name, &node.labels) {
+                            node.features = nc.features.clone();
+                            if nc.weight > 0 {
+                                node.weight = nc.weight;
+                            }
+                            break;
+                        }
+                    }
```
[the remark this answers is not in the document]
shiv-tyagi: Valid observation. Resetting features/weight to defaults when no NodeConfig matches requires defining what "defaults" are (zero? original hardware-reported values?). Tracking as a follow-up issue.

crates/spurctld/src/cluster.rs
```diff
@@ -2039,6 +2129,39 @@ struct ClusterSnapshot {
...
+            }
+            node.partitions = matched;
+
+            for nc in &self.config.nodes {
+                if node_config_matches(nc, &node.name, &node.labels) {
+                    node.features = nc.features.clone();
+                    node.weight = nc.weight;
+                    break;
+                }
+            }
```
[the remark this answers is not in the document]
shiv-tyagi: Same as above — tracked as follow-up.

crates/spur-cli/src/node.rs
```diff
@@ -0,0 +1,151 @@
...
+            remove_labels.push(key.to_string());
+        } else if let Some((k, v)) = arg.split_once('=') {
+            if k.is_empty() {
+                bail!("invalid label: '{arg}', key cannot be empty");
+            }
+            set_labels.insert(k.to_string(), v.to_string());
+        } else {
+            bail!("invalid label format: '{arg}', expected key=value or key-");
+        }
+    }
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed — reordered to check `split_once('=')` before `strip_suffix('-')`, matching kubectl semantics. Added a test for `env=prod-` confirming it's treated as a set operation.

crates/spurctld/src/cluster.rs
```diff
@@ -2124,6 +2252,55 @@ pub(crate) fn evaluate_registration(
...
+            .all(|(k, v)| labels.get(k) == Some(v));
+
+    let matches_hostlist = if partition.nodes.is_empty() {
+        false
+    } else if partition.nodes.eq_ignore_ascii_case("ALL") {
+        true
+    } else {
+        spur_core::hostlist::expand(&partition.nodes)
+            .map(|hosts| hosts.contains(&node_name.to_string()))
+            .unwrap_or(false)
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed — changed to `hosts.iter().any(|h| h == node_name)` which compares `&String` with `&str` via `PartialEq` without allocating.

crates/spurctld/src/cluster.rs
```diff
@@ -2124,6 +2252,55 @@ pub(crate) fn evaluate_registration(
...
+    labels: &HashMap<String, String>,
+) -> bool {
+    let matches_names = if nc.names.is_empty() {
+        false
+    } else if nc.names.eq_ignore_ascii_case("ALL") {
+        true
+    } else {
+        spur_core::hostlist::expand(&nc.names)
+            .map(|hosts| hosts.contains(&node_name.to_string()))
+            .unwrap_or(false)
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed — same change as above.

yansun1996 (commented):
_(Superseded — re-posted below with inline comments anchored to specific lines. See the newer review.)_

yansun1996 (commented):
## Code Review (re-posted with inline anchors)

Reviewed against the repo's `contributing.rst` process. Overall a **well-structured, high-quality PR** — CI is green across the board, commits are logically separated and follow conventional-commit style, and the feature is complete end-to-end (proto -> core -> controller -> agent -> CLI -> REST -> docs -> tests).

Inline comments anchor each finding to the relevant code. Summary of severities:
- **Blocking:** stale `features`/`weight` on label removal (cluster.rs); `weight` handling diverges between apply sites.
- **Should-fix:** `--label` comma delimiter footgun; duplicated label-sync block; non-atomic two-proposal node update.
- **Nits:** fixed sleeps in e2e; fragile partition-line filter.

**Proto compat — verified OK:** `string partition = 4` -> `repeated string partitions = 4` reuses field 4 with a wire-compatible cardinality change; package name `slurm` untouched. `spur-ffi` does not read the field; `spurrestd` now emits the real list (a fix); no stale singular reads remain.

**Verdict:** Approve with changes requested — the two blocking items are genuine correctness bugs; design (hostlist OR selector union, `ALL` wildcard, WAL-backed durability, re-reconcile on snapshot restore) is sound.

crates/spurctld/src/cluster.rs
yansun1996, line 2106: **Blocking (correctness): stale `features`/`weight` are never cleared when labels stop matching any NodeConfig.**

This loop only *overwrites* on a match. If a node previously matched a selector-based NodeConfig (gaining e.g. `features=["mi300x","rocm6"]`, `weight=10`) and the matching label is then removed via `spur node label <node> gpu-`, the loop finds no match, falls through, and the **old features/weight persist**. Partitions are correctly recomputed (empty -> default fallback) just above, but features/weight silently go stale — contradicting the PR's "NodeConfig feature/weight re-application when changed" claim.

Suggest resetting to a baseline before the loop:
```rust
node.features = Vec::new();
node.weight = <registration default>;
for nc in &self.config.nodes { ... }
```

crates/spurctld/src/cluster.rs
```diff
@@ -2021,6 +2074,43 @@ impl ClusterManager {
...
+                            matched.push(first.name.clone());
+                        }
+                    }
+                    node.partitions = matched;
+
+                    // Re-apply NodeConfig features/weight
+                    for nc in &self.config.nodes {
+                        if node_config_matches(nc, &node.name, &node.labels) {
+                            node.features = nc.features.clone();
+                            if nc.weight > 0 {
```
yansun1996: **Blocking (consistency): `weight` handling diverges between the two re-application sites.**

Here the label-update path guards with `if nc.weight > 0`, but `reconcile_partitions` (snapshot restore) and initial registration assign `node.weight = nc.weight` unconditionally. A label update and a controller restart can therefore leave the *same* node with *different* weights from the *same* config. Pick one rule and apply it in all three places.

crates/spurctld/src/cluster.rs
```diff
@@ -2039,6 +2129,39 @@ struct ClusterSnapshot {
...
+                } else if let Some(first) = partitions.first() {
+                    matched.push(first.name.clone());
+                }
+            }
+            node.partitions = matched;
+
+            for nc in &self.config.nodes {
+                if node_config_matches(nc, &node.name, &node.labels) {
+                    node.features = nc.features.clone();
+                    node.weight = nc.weight;
```
yansun1996, line 2152: For contrast with the label-update path above: here `weight` is assigned unconditionally (no `> 0` guard). This is the divergence flagged in the NodeLabelsUpdate handler — these two sites should match.

crates/spurctld/src/cluster.rs
```diff
@@ -769,6 +770,22 @@ impl ClusterManager {
...
+                            .cloned()
+                            .collect();
+                        self.propose(WalOperation::NodeLabelsUpdate {
+                            name: effective_name.clone(),
+                            set: labels,
+                            remove,
+                        })?;
+                        info!(node = %effective_name, "node labels synced on re-registration");
+                    }
+                }
```
yansun1996: **Should-fix: this label-sync block is duplicated verbatim in the `RegistrationAction::Update` arm just below.** Same `existing.labels != labels` -> compute `remove` -> propose `NodeLabelsUpdate`. Extract a helper (e.g. `sync_labels_if_changed(&effective_name, labels)`) so the two copies don't drift.

crates/spurctld/src/server.rs
```diff
@@ -382,6 +382,11 @@ impl SlurmController for ControllerService {
                 .update_node_state(&req.name, node_state, req.reason)
                 .map_err(|e| Status::internal(e.to_string()))?;
         }
+        if !req.labels.is_empty() || !req.remove_labels.is_empty() {
+            self.cluster
+                .update_node_labels(&req.name, req.labels, &req.remove_labels)
+                .map_err(|e| Status::internal(e.to_string()))?;
+        }
```
yansun1996, line 389: **Should-fix: non-atomic update.** A single `UpdateNodeRequest` can emit *two* separate Raft proposals — `update_node_state` (just above) and `update_node_labels` here. A crash between them applies a partial update. Low probability, but worth either a combined op or at least a comment acknowledging the non-atomicity.

crates/spurd/src/main.rs
```diff
@@ -49,6 +59,11 @@ struct Args {
     #[arg(long, env = "SPUR_NODE_ADDRESS")]
     address: Option<String>,
 
+    /// Node labels for partition routing (key=value pairs).
+    /// Can be specified multiple times: --label pool=gpu --label rack=a
+    #[arg(long = "label", value_parser = parse_label, env = "SPUR_NODE_LABELS", value_delimiter = ',')]
```
yansun1996: **Should-fix: `value_delimiter = ','` is a footgun for label *values*.** A value containing a comma (`--label zone=us,eu`) silently splits into two args, the second of which (`eu`) fails `parse_label` (no `=`). Consider dropping the delimiter (rely on repeated `--label`) or documenting that values may not contain commas.

tests/e2e/native_host/test_node_labels.py
```diff
@@ -0,0 +1,88 @@
...
+        out = label_cluster.sinfo()
+        gpu_lines = [l for l in out.splitlines() if "gpu" in l.split()[0:1]]
+        gpu_text = "\n".join(gpu_lines)
+        assert node1 not in gpu_text, (
+            f"precondition: {node1} should not be in gpu partition:\n{out}"
+        )
+
+        # Add label → node joins gpu partition
+        label_cluster.cli(["spur", "node", "label", node1, "gpu=mi300x"])
+        time.sleep(2)
```
yansun1996: **Nit: fixed `time.sleep(2)` after a label mutation is flaky.** Prefer polling `scontrol`/`sinfo` for the expected state, consistent with the recent move to polling in the suspend e2e suite. Same applies to the sleep at line 81.

crates/spurctld/src/cluster.rs
shiv-tyagi: Valid point. Reset needs a decision on what "no config matches" means (empty features? zero weight?). Registration doesn't report either today, so empty/zero is the only sane default. Deferring as follow-up since it touches all three re-application sites and needs consistent semantics.

crates/spurctld/src/cluster.rs
shiv-tyagi: Fixed. Removed the guard, both sites now assign unconditionally.

crates/spurctld/src/cluster.rs
shiv-tyagi: Addressed in the same fix. Label-update path now matches this.

crates/spurctld/src/cluster.rs
shiv-tyagi: Done. Extracted `sync_node_labels` helper, both arms call it now.

crates/spurctld/src/server.rs
shiv-tyagi: Agreed this is a gap. Making it atomic requires a composite WAL op or multi-op proposal, which is a non-trivial state machine change. The window between the two proposals is sub-millisecond in practice (single Raft round-trip). Deferring as follow-up given pre-1.0 status. Added a comment acknowledging the non-atomicity.

crates/spurd/src/main.rs
shiv-tyagi: Fixed. Removed the delimiter. Multiple labels use repeated `--label` flags, same as kubectl.

tests/e2e/native_host/test_node_labels.py
shiv-tyagi: Fixed. Replaced with a polling helper that checks sinfo every 0.5s, times out at 10s.

yansun1996 (approved):
lgtm pls file tickets to follow up on the deferred items

crates/spurctld/src/cluster.rs
shiv-tyagi: Marking resolved to enabling PR merging. Will file a ticket for this post that.

crates/spurctld/src/server.rs
shiv-tyagi: Resolving for now. Will file a ticket.

sgopinath1 (approved):
LGTM

PR #315, opened 2026-06-20
refactor(spurctld): remove hostname aliasing logic

shiv-tyagi:
## Summary

- Remove the hostname aliasing mechanism from `ClusterManager` that silently mapped agent hostnames to unmatched config entries during registration
- Agents now register under their actual hostname; mismatches between agent hostname and config are no longer hidden

## Motivation

The aliasing logic was a convenience for single-node dev setups where the config said `localhost` but the agent registered with its real hostname. It had several problems:

- **Hid misconfiguration** — agents with wrong hostnames were silently remapped instead of surfacing the error
- **Fragile first-match** — with multiple mismatched agents, it arbitrarily picked the first unmatched config entry
- **Unnecessary complexity** — persisted in Raft snapshots, adding serialization overhead for a dev convenience

With label-based partition selectors and `spurd --hostname` flag available, there are better ways to handle this. K8s deployments use deterministic pod names, and native-host deployments should have matching config entries.

## Changes

Single file: `crates/spurctld/src/cluster.rs` (11 insertions, 73 deletions)

- Removed `hostname_aliases` field from `ClusterManager` and `ClusterSnapshot`
- Removed 50-line aliasing block in `register_node` — registration now uses the agent-reported name directly
- Removed alias lookup in `update_heartbeat`
- Removed alias map from snapshot serialization/deserialization

## Test plan

- [x] `cargo build` passes
- [x] `cargo test` passes (all existing tests green)
- [x] No references to `hostname_alias` or `effective_name` remain in the codebase

### DIFF
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -52,7 +52,6 @@ pub struct ClusterManager {
     reservations: RwLock<Vec<Reservation>>,
     steps: RwLock<HashMap<(JobId, u32), JobStep>>,
     license_pool: RwLock<HashMap<String, u64>>,
-    hostname_aliases: RwLock<HashMap<String, String>>,
     raft: RwLock<Option<SpurRaft>>,
     accounting: RwLock<Option<AccountingNotifier>>,
     fairshare_cache: Arc<FairshareCache>,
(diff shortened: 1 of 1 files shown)


shiv-tyagi:
>  Patch coverage is 54.54545% with 5 lines in your changes missing coverage. Please review.

This  is okay and preexisting.

sgopinath1 (approved):
LGTM

PR #319, opened 2026-06-21
feat(spurctld): serve REST API directly, remove standalone spurrestd

shiv-tyagi:
## Why

Deploying the REST API as a separate daemon (`spurrestd`) means an extra binary to build, distribute, configure, and monitor. Every install — bare-metal or Kubernetes — has to coordinate two processes that share no state and talk over loopback gRPC. This adds friction to setup with no architectural benefit, since the REST layer is a stateless translation shim over the controller's gRPC API.

## What

- Add a `rest/` module to `spurctld` that serves the Slurm-compatible REST API on a separate port (default `6820`), using the same Axum/Tower stack.
- Gate it with `[rest_api] enabled` (default `true`) and `[controller] rest_addr` in `spur.conf` — setting `enabled = false` skips the listener entirely.
- Remove the `spurrestd` crate, its K8s manifest, Dockerfile build target, install script entry, and CI references.

## Impact

- **Distribution**: one fewer binary to build, package, and ship. The Docker image and install script are simpler.
- **Setup**: no separate process to configure or health-check. REST works out of the box when `spurctld` starts.
- **Kubernetes**: one fewer Deployment + Service. The `spurctld` Service now exposes port `6820` alongside `6817`.
- **Backwards compatibility**: REST endpoints, routes, and response schemas are unchanged. Existing clients work without modification.

## Test plan

- [x] `cargo test` — all tests pass
- [x] Bare-metal: REST enabled — smoke tested all endpoints + job submission
- [x] Bare-metal: REST disabled — verified no listener, gRPC unaffected
- [x] K8s: REST enabled — smoke tested all endpoints via port-forward
- [x] K8s: REST disabled — verified port 6820 connection refused


shiv-tyagi:
>  Patch coverage is 11.25828% with 268 lines

Low coverage is not actionable for this PR. This is pre-existing.

shiv-tyagi:
Points discussed in meeting:

**Q: Can the REST server block the scheduler or hurt performance?**
No. It runs as a separate tokio task on its own port (6820). Can't block the scheduler, Raft, or gRPC — they're all independent tasks on the same async runtime. REST handlers just read cluster state and return JSON, no heavy work.

**Q: Does embedding open us up to DDoS that a separate gateway would prevent?**
No. A decoupled REST daemon still proxies every request to the controller over gRPC, so flooding it floods the controller anyway. Embedded means one process to protect instead of two. Rate limiting is a tower middleware layer — same defense, less ops. Real DDoS mitigation happens at the network edge regardless.

**Q: In HA mode, do all 3 controllers run REST? Is that wasteful?**
Yes they all run it, no it's not wasteful. An idle listener costs a few KB of memory. Reads are served locally from Raft-synced state, so you get free load distribution. Writes go through the leader either way. You can set `enabled = false` on non-leaders if you want, but it's not worth the hassle.

**Q: Why not build a proper API gateway with auth, rate limiting, routing?**
Spur is an infra scheduler with a handful of REST consumers, not a SaaS API. Auth is a middleware layer, LB is handled by K8s/nginx, routing is trivial. The industry trend is consolidation (k3s, Consul, Nomad all embed their APIs). If we ever need a real gateway, we'd use Envoy or Kong — not maintain a custom one.

sgopinath1 (approved):
LGTM

Issue #307, opened 2026-06-15
parity(slurm): wire emission logic for pending-reason vocabulary landed in #301


yansun1996:
## Summary

PR #301 (closes #300) landed 33 Slurm-parity `PendingReason` variants with byte-exact display strings, but most are **vocabulary only** — the variant and its string exist so external tooling recognizes the reason, but Spur has no code path that computes/emits it. This issue tracks closing that emission gap.

Splitting vocabulary from emission was deliberate: landing the exact strings first means follow-up PRs add only logic, not strings (and can't drift the strings). This issue is the umbrella for that logic.

## Current emission status (after #301)

Of the 33 variants added:

- **1 production-active:** `Reservation` — emitted by `reservation_block()` in `pending_jobs()`/`tag_blocked_pending_reasons()` against real reservations (verified live vs Slurm 25.11.6).
- **5 wired but dormant:** `QOSMaxCpuPerJobLimit`, `QOSMaxWallDurationPerJobLimit`, `QOSMaxMemoryPerJob`, `QOSMaxCpuPerUserLimit`, `QOSMaxSubmitJobPerUserLimit`. `check_qos_limits()` returns them, but the only production caller `qos_block_for()` passes a default limitless `Qos`, so they never fire. Exercised today only by `qos.rs` unit tests.
- **27 display-string-only (no emission path):** `PartitionConfig`, `SystemFailure`, `AccountingPolicy`, all `Assoc*` (11), the remaining `QOS*` (`QOSJobLimit`, `QOSResourceLimit`, `QOSTimeLimit`, all `QOSGrp*`, `QOSMaxNodePerJobLimit`), and both `BurstBuffer*`. No code assigns these.

## What's missing / scope

The enforcement builds on the existing `fairshare_cache → spurdbd` (`SlurmAccountingClient`) pattern. The DB schema (`qos`, `associations`, `accounts` tables with limit columns) and the `QosLimits` / `AccountLimits` / `Association` structs already exist; what's missing is loading them into the controller and the per-QoS/per-association enforcement passes.

| Group | Reasons | Work | Est. (incl. tests) | Tracked by |
|---|---|---|---|---|
| Prereq: limits cache | — | Mirror `fairshare_cache`: refresh QoS + association + account limits from spurdbd into spurctld (may need one accounting RPC). | ~200 LOC | this issue |
[...]

PR #301, opened 2026-06-15
feat(spur): expand pending-reason vocabulary and surface reservation/license/QoS reasons

yansun1996:
Closes #300.

This PR closes the Slurm-parity gap where Spur shipped only ~14 `PendingReason` values against Slurm's ~50, and where the reasons Spur did compute often never reached the user. It has two parts: a small set of new reasons, and the wiring that makes them observable.

## Scope (revised after review)

An earlier revision of this PR added 33 `PendingReason` variants — 6 with a live emission path and 27 display-string-only placeholders landed ahead of their enforcement logic. Per review (@shiv-tyagi), the 27 dead variants were dropped: they carried match-arm bloat and, once serialized into Raft snapshots, would become a permanent deserialization obligation even if the corresponding feature shipped differently.

**This PR now lands only the 6 `PendingReason` variants that have a real emission path:**

- **`Reservation`** — production-active today. Emitted by `reservation_block()` in `pending_jobs()`/`tag_blocked_pending_reasons()` against real reservations (verified live vs Slurm 25.11.6).
- **5 QoS caps wired into `check_qos_limits()`**: `QOSMaxCpuPerJobLimit`, `QOSMaxWallDurationPerJobLimit`, `QOSMaxMemoryPerJob`, `QOSMaxCpuPerUserLimit`, `QOSMaxSubmitJobPerUserLimit`. `check_qos_limits()` returns these for the matching cap; today the only production caller (`qos_block_for`) passes a default limitless `Qos`, so they are exercised by `qos.rs`/`t21_acctmgr.rs` unit tests and become live end-to-end once QoS definitions are loaded from spurdbd (#282/#283).

The remaining Slurm reason vocabulary (associations, QoS group limits, partition config, burst buffer) will be added **alongside its enforcement logic** so a variant and its emitter land together — tracked in #307.

Each variant's display string matches Slurm 25.11.6's `job_reason_string()` (table `jsra[]` in `src/common/job_state_reason.c`) byte-for-byte, since these strings are scraped by CI gates and workflow engines.

## QoS emission correctness

[...]

PR #326, opened 2026-06-22
feat(spur): emit PartitionConfig/PartitionInactive pending reasons

yansun1996:
Group 3 of #307 (Partition/System pending reasons).

> ⚠️ **Stacked on #301** (`parity/reason-code-vocab`, unmerged). Until #301 lands, this diff includes #301's commits — review the `feat(spur): emit PartitionConfig/PartitionInactive` commit. Will rebase onto `main` once #301 merges.

## What
Hold jobs PENDING with a specific reason instead of rejecting at submit (Slurm parity — same pattern as reservation/license blocks):
- **`PartitionConfig`** — request structurally exceeds the partition config (`num_nodes > max_nodes`, `num_nodes < min_nodes`, `time_limit > max_time`).
- **`PartitionInactive`** — target partition is not `Up`. The variant existed (from #274) but was never emitted; now wired.

A shared `partition_block()` helper feeds BOTH `pending_jobs()` (drop) and `tag_blocked_pending_reasons()` (displayed reason). It's checked first in the chain — a structural partition block is permanent given current config. `validate_partition()` no longer rejects partition state / max_nodes / max_time at submit.

## Deliberately NOT added (would be dead variants)
- `SystemFailure` — no controller path holds a job *pending* on an internal error (launch failures route to terminal states).
- `AccountingPolicy` — QoS blocks already map to specific `QOS*` reasons; no real fallback site.

These land with their enforcement if/when a genuine trigger exists.

## Tests
`REASON_VOCAB` extended (Display + serde for both); integration test `tag_blocked_sets_partition_config_reason` (real ClusterManager, 2-node job vs `max_nodes=1` → `PartitionConfig`).

Static: build/clippy/fmt clean; spur-core 197, spurctld 140 pass.

## Live validation
Verified live on a Spur node: a 2-node job against a partition with `max_nodes=1` is held `PENDING Reason=PartitionConfig` (previously rejected at submit). `PartitionInactive` follows the same `partition_block` path; both Slurm strings are byte-exact from `job_state_reason.c`.

### DIFF
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -139,9 +139,6 @@ impl ClusterManager {
             None => anyhow::bail!("partition '{}' not found", partition_name),
         };
 
-        // Partition state and resource limits are not rejected here; the job
-        // pends with PartitionInactive/PartitionConfig instead (Slurm parity).
-
         // Check allow_accounts (if non-empty, user's account must be in the list)
         if !part.allow_accounts.is_empty() {
             let account = spec.account.as_deref().unwrap_or("");
@@ -2487,7 +2484,9 @@ fn partition_block(job: &Job, partitions: &[Partition]) -> Option<spur_core::job
     use spur_core::partition::PartitionState;
 
     let name = job.spec.partition.as_deref().filter(|p| !p.is_empty())?;
-    let part = partitions.iter().find(|p| p.name == name)?;
+    let Some(part) = partitions.iter().find(|p| p.name == name) else {
+        return Some(PendingReason::PartitionConfig);
+    };
 
     if part.state != PartitionState::Up {
         return Some(PendingReason::PartitionInactive);
@@ -4219,6 +4218,14 @@ mod tests {
             cm.get_job(n_id).unwrap().pending_reason,
             PendingReason::PartitionConfig,
             "num_nodes below partition min_nodes -> PartitionConfig"
+        );
+        assert!(
+            !cm.pending_jobs().iter().any(|j| j.job_id == t_id),
+            "time-blocked job must be dropped from scheduling"
+        );
+        assert!(
+            !cm.pending_jobs().iter().any(|j| j.job_id == n_id),
+            "min_nodes-blocked job must be dropped from scheduling"
         );
     }
 
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -139,11 +139,8 @@ impl ClusterManager {
             None => anyhow::bail!("partition '{}' not found", partition_name),
         };
 
-        // Partition state (Down/Drain/Inactive) and structural resource limits
-        // (max_nodes/max_time/min_nodes) are NOT rejected here: matching Slurm,
-        // the job is accepted and held PENDING with PartitionInactive /
-        // PartitionConfig (see partition_block + tag_blocked_pending_reasons), so
-        // it runs once the partition is brought back Up or its config changes.
+        // Partition state and resource limits are not rejected here; the job
+        // pends with PartitionInactive/PartitionConfig instead (Slurm parity).
 
         // Check allow_accounts (if non-empty, user's account must be in the list)
         if !part.allow_accounts.is_empty() {
@@ -1139,10 +1136,8 @@ impl ClusterManager {
             .cloned()
             .collect();
 
-        // Drop jobs structurally unschedulable in their target partition
-        // (inactive partition / request exceeds partition config). Shares
-        // partition_block() with tag_blocked_pending_reasons() so the drop
-        // decision and the shown PartitionInactive/PartitionConfig reason agree.
+        // Drop jobs blocked by partition state/config (shares partition_block()
+        // with tag_blocked_pending_reasons() so drop and shown reason agree).
         {
             let partitions = self.partitions.read();
             pending.retain(|job| partition_block(job, &partitions).is_none());
@@ -1464,10 +1459,8 @@ impl ClusterManager {
                         && job.pending_reason != PendingReason::DeadLine
                 })
                 .filter_map(|job| {
-                    // Same order pending_jobs() drops jobs, so the shown reason is
-                    // the one that actually removed it: Part -> Dep -> QoS -> Resv
-                    // -> Lic. partition_block is first: a structural partition
-                    // block is permanent, so it outranks the transient blocks.
+                    // Same drop order as pending_jobs(): Part -> Dep -> QoS ->
+                    // Resv -> Lic (partition block is permanent, so first).
                     partition_block(job, &partitions)
                         .or_else(|| dependency_block(job))
                         .or_else(|| qos_block_for(job, &jobs))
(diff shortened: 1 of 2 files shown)


yansun1996:
## Live-verified (Spur node)

Deployed this build's spurctld with partition `max_nodes=1`:
- 2-node job → `PENDING Reason=PartitionConfig` ✅ (previously rejected at submit; now held PENDING matching Slurm)

`PartitionInactive` follows the same `partition_block` path (partition state != Up); the `PartitionConfig` path being live confirms the wiring. Both Slurm strings are byte-exact from `job_state_reason.c`.

yansun1996:
## Rebased onto current main + review addressed

- **Rebased** onto latest `main` (now that #301 merged) — PR is a clean single feature commit plus a test commit.
- **Independent code review** (fresh context): **APPROVE-WITH-NITS**, no blocking issues. Reviewer verified the comparisons/units, precedence consistency between `pending_jobs()` and `tag_blocked_pending_reasons()`, lock ordering, the `validate_partition()` relaxation (nonexistent-partition still rejected), and that no prior tests were silently deleted.

### Test gaps from review — now closed (commit c0fba5b)
- `tag_blocked_sets_partition_inactive_when_not_up` — the previously-untested **PartitionInactive** path.
- `tag_blocked_sets_partition_config_for_time_and_min_nodes` — the **max_time** and **min_nodes** branches (only max_nodes was covered).
- `submit_still_rejects_nonexistent_partition` — confirms an unknown partition still errors at submit (only inactive/over-config pend).

Static: build/clippy/fmt clean; spur-core 197, spurctld 143 pass. Live-verified earlier (2-node job vs max_nodes=1 → PartitionConfig).

yansun1996:
## Live-verified on a Spur node — marking ready for review

All four scenarios confirmed against a running controller:
- Job exceeding partition `max_nodes` → `PENDING Reason=PartitionConfig` ✅
- Job exceeding partition `max_time` → `PENDING Reason=PartitionConfig` ✅
- Job in a non-`Up` (DOWN) partition → `PENDING Reason=PartitionInactive` ✅
- Job targeting a nonexistent partition → rejected at submit ✅ (not admitted/held)

Rebased onto current main (post-#301 merge), single clean feature commit + the review-feedback test commit. Static: build/clippy/fmt clean; spur-core 197, spurctld 143 pass.

yansun1996:
@sgopinath1 @shiv-tyagi could you please review this when you get a chance? It's rebased onto current main (post-#301 merge), live-verified on a Spur node (all four partition scenarios), and review-feedback test gaps are closed. Thanks!

crates/spurctld/src/cluster.rs
```diff
@@ -139,10 +139,8 @@ impl ClusterManager {
             None => anyhow::bail!("partition '{}' not found", partition_name),
         };
 
-        // Check partition state
-        if part.state != spur_core::partition::PartitionState::Up {
-            anyhow::bail!("partition '{}' is {}", partition_name, part.state.display());
-        }
+        // Partition state and resource limits are not rejected here; the job
+        // pends with PartitionInactive/PartitionConfig instead (Slurm parity).
```
shiv-tyagi:
```suggestion
```

crates/spurctld/src/cluster.rs
```diff
@@ -4150,6 +4165,99 @@ mod tests {
...
+        assert_eq!(
+            cm.get_job(t_id).unwrap().pending_reason,
+            PendingReason::PartitionConfig,
+            "time_limit over partition max_time -> PartitionConfig"
+        );
+        assert_eq!(
+            cm.get_job(n_id).unwrap().pending_reason,
+            PendingReason::PartitionConfig,
+            "num_nodes below partition min_nodes -> PartitionConfig"
+        );
```
shiv-tyagi, line 4221: The other two partition-block tests (`tag_blocked_sets_partition_config_reason`, `tag_blocked_sets_partition_inactive_when_not_up`) assert that `pending_jobs()` drops the blocked job. This test only checks `pending_reason`. If the `partition_block` retain call were accidentally removed from `pending_jobs()`, the job would display `PartitionConfig` but still enter scheduling every cycle.

```suggestion
        assert_eq!(
            cm.get_job(n_id).unwrap().pending_reason,
            PendingReason::PartitionConfig,
            "num_nodes below partition min_nodes -> PartitionConfig"
        );
        assert!(
            !cm.pending_jobs().iter().any(|j| j.job_id == t_id),
            "time-blocked job must be dropped from scheduling"
        );
        assert!(
            !cm.pending_jobs().iter().any(|j| j.job_id == n_id),
            "min_nodes-blocked job must be dropped from scheduling"
        );
```

crates/spurctld/src/cluster.rs
```diff
@@ -2497,6 +2479,35 @@ fn qos_block_for(job: &Job, jobs: &HashMap<JobId, Job>) -> Option<spur_core::job
...
 
+/// `PartitionInactive` if the partition is not Up, `PartitionConfig` if the
+/// request exceeds its node/time limits, else `None`. Shared by `pending_jobs()`
+/// and `tag_blocked_pending_reasons()` so drop and displayed reason agree.
+fn partition_block(job: &Job, partitions: &[Partition]) -> Option<spur_core::job::PendingReason> {
+    use spur_core::job::PendingReason;
+    use spur_core::partition::PartitionState;
+
+    let name = job.spec.partition.as_deref().filter(|p| !p.is_empty())?;
+    let part = partitions.iter().find(|p| p.name == name)?;
```
shiv-tyagi: `partition_block` returns `None` when the partition isn't in the list (the `?` on `find`), which lets the job pass through to scheduling where it sits with `Reason=Resources`. For correctness, returning `PartitionConfig` here is safer. Submit already rejects unknown partitions, so this path can only fire if a partition somehow disappears post-submission.

```suggestion
    let name = job.spec.partition.as_deref().filter(|p| !p.is_empty())?;
    let Some(part) = partitions.iter().find(|p| p.name == name) else {
        return Some(PendingReason::PartitionConfig);
    };
```

shiv-tyagi (commented):
Thanks @yansun1996. Few minor comments.

shiv-tyagi:
The two-pass design (drop in `pending_jobs()`, tag in `tag_blocked_pending_reasons()`) exists because `pending_jobs()` holds a read lock on jobs and can't write back reasons. This forces every block type (partition, dep, QoS, reservation, license) to be evaluated twice with a must-match ordering contract. Longer term, `pending_jobs()` should return the drop reasons alongside the filtered list so the caller can apply them in a single write pass, eliminating the duplicated scan and the ordering coupling.

Not for this PR but wanted to bring to your attention.

crates/spurctld/src/cluster.rs
yansun1996: Removed in 79497a9.

crates/spurctld/src/cluster.rs
yansun1996: Added both `pending_jobs()` drop assertions (for `t_id` and `n_id`) in 79497a9, so the test now fails if the `partition_block` retain is dropped from `pending_jobs()`.

crates/spurctld/src/cluster.rs
yansun1996: Agreed. Changed to return `PartitionConfig` on a missing partition in 79497a9, so a job whose partition vanished post-submit is dropped with a truthful reason instead of falling through to `Resources`.

yansun1996:
Addressed all three inline comments in 79497a9:
- `partition_block()` now returns `PartitionConfig` for a missing partition (was `None` → job fell through to `Resources`).
- Comment dropped.
- `tag_blocked_sets_partition_config_for_time_and_min_nodes` now also asserts `pending_jobs()` drops both jobs.

Re: the two-pass design — good point. The duplicate scan + must-match ordering between `pending_jobs()` and `tag_blocked_pending_reasons()` is real coupling, and having `pending_jobs()` return drop reasons for a single write pass is the right direction. Agree it's out of scope here; I'll file a follow-up so it isn't lost.

Static: build/clippy/fmt clean; spur-core 197, spurctld 143 pass. Re-verified live on a Spur node (all four partition scenarios).

yansun1996:
Filed the two-pass refactor as #349.

shiv-tyagi (approved):
LGTM. Thanks @yansun1996.

PR #342, opened 2026-06-24
fix(spurd): resolve stdout/stderr paths after work_dir fallback

shiv-tyagi:
## Summary

- Jobs failed with "failed to create stdout file" when the submitter's working directory didn't exist on the compute node. The agent server pre-resolved stdout/stderr to absolute paths using the submitter's `work_dir` before the executor's `/tmp` fallback ran, so the fallback never affected the output file paths.
- Moved stdout/stderr path resolution entirely into the executor (after the work_dir fallback) and introduced `LaunchResult` to return the resolved paths back for log streaming (`StreamOutput` RPC).

## Test plan

- [x] All unit tests pass (`cargo test --locked`)
- [x] `cargo clippy` and `cargo fmt --check` clean
- [x] Reproduced on bare-metal testbed: submitted jobs to nodes where submitter's work_dir doesn't exist — all stuck before fix, all RUNNING after
- [x] Verified default stdout (`/tmp/spur-<id>.out`) and explicit `--output=/tmp/custom_%j.log` both resolve correctly
- [x] Verified job reaches COMPLETED state
- [x] Audited for regressions: `TrackedJob` stores resolved absolute paths for `StreamOutput` RPC (log tailing)

### DIFF
--- a/crates/spurd/src/agent_server.rs
+++ b/crates/spurd/src/agent_server.rs
@@ -745,18 +745,6 @@ impl SlurmAgent for AgentService {
             .map(|a| a.cpu_ids.clone())
             .unwrap_or_default();
 
-        // Resolve stdout/stderr paths
-        let stdout_path = if spec.stdout_path.is_empty() {
-            format!("{}/spur-{}.out", work_dir, job_id)
-        } else {
-            spec.stdout_path.clone()
-        };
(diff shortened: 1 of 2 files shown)


sgopinath1 (approved):
LGTM

PR #348, opened 2026-06-24
perf(spur-sched): speed up backfill timeline queries

sgopinath1:
## Summary

- Bound `accumulated_at` and the `earliest_start` overlap scan with `partition_point` on sorted interval starts, cutting per-query work from O(n) to O(log n + k) for overlapping intervals.
- Add `AllocationSweep` so `earliest_start` advances a running allocation total forward with candidate time instead of recomputing `accumulated_at` on every loop iteration.
- Prune stale intervals via `gc(horizon)` at the start of each backfill cycle and before each `earliest_start` search, keeping per-node interval vectors smaller under sustained submit load.

Targets the backfill scheduler hot path identified in submit-throughput profiling (`NodeTimeline::accumulated_at` / `earliest_start`).

## Test plan

- [x] `cargo test -p spur-sched --locked`
- [x] `cargo test -p spur-tests t07_1 --locked` (timeline + scheduler integration)
- [x] Re-run submit throughput benchmark (gRPC direct submit, sustained 5K jobs) and compare against pre-change baseline

### DIFF
--- a/crates/spur-sched/src/backfill.rs
+++ b/crates/spur-sched/src/backfill.rs
@@ -184,6 +184,10 @@ impl Scheduler for BackfillScheduler {
             );
         }
 
+        for timeline in &mut self.timelines {
+            timeline.gc(now);
+        }
+
         // Identify het job groups: collect sets of jobs linked by het_job_id.
         // For each group, ALL components must be schedulable or NONE are scheduled.
         let mut het_groups: HashMap<JobId, Vec<usize>> = HashMap::new();
--- a/crates/spur-sched/src/backfill.rs
+++ b/crates/spur-sched/src/backfill.rs
@@ -184,10 +184,6 @@ impl Scheduler for BackfillScheduler {
             );
         }
 
-        for timeline in &mut self.timelines {
-            timeline.gc(now);
-        }
-
         // Identify het job groups: collect sets of jobs linked by het_job_id.
         // For each group, ALL components must be schedulable or NONE are scheduled.
         let mut het_groups: HashMap<JobId, Vec<usize>> = HashMap::new();
--- a/crates/spur-sched/src/timeline.rs
+++ b/crates/spur-sched/src/timeline.rs
@@ -1,6 +1,9 @@
 // Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::cmp::Reverse;
+use std::collections::BinaryHeap;
+
 use chrono::{DateTime, Duration, Utc};
 use spur_core::resource::{ResourceAllocations, ResourceSet};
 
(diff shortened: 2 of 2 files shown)


crates/spur-sched/src/backfill.rs
```diff
@@ -184,6 +184,10 @@ impl Scheduler for BackfillScheduler {
             );
         }
 
+        for timeline in &mut self.timelines {
+            timeline.gc(now);
+        }
```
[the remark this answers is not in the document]
sgopinath1: Fixed.

crates/spur-sched/src/timeline.rs
```diff
@@ -51,36 +135,31 @@ impl NodeTimeline {
     /// Find the earliest time at which `request` resources are available
     /// for `duration` contiguous time.
     pub fn earliest_start(
-        &self,
+        &mut self,
         request: &ResourceSet,
         duration: Duration,
         after: DateTime<Utc>,
     ) -> DateTime<Utc> {
+        self.gc(after);
```
shiv-tyagi: `gc(after)` is a no-op in the production code path. `init_timelines()` rebuilds all timelines from scratch each `schedule()` call, and every `reserve()` in `backfill.rs` uses `start >= now` with a positive duration, so `end > now` always holds. No interval can be stale.

This gc forces `&mut self` on what should be a pure query. For 1000 jobs x 3 nodes, that's 3000 `retain()` passes that iterate all intervals and remove nothing. It also prevents shared borrows and future parallelism over timelines.

Remove `gc(after)` and restore the signature to `&self`. Keep the `gc` method as a utility. If you want to guard the invariant, a debug assert after building timelines is cleaner:

```rust
#[cfg(debug_assertions)]
debug_assert!(
    self.intervals.iter().all(|i| i.end > after),
    "stale interval in timeline for {}",
    self.node_name,
);
```

`test_earliest_start_prunes_expired_intervals` would need to be rewritten as a standalone `gc()` unit test since it currently relies on gc being called as a side effect of `earliest_start`.

crates/spur-sched/src/timeline.rs
```diff
@@ -276,4 +358,362 @@ mod tests {
...
+
+        loop {
+            if candidate > max_check {
+                return max_check;
+            }
+
+            if tl.can_satisfy_at(candidate, request) {
+                let window_end = candidate + duration;
+                let mut ok = true;
+                for interval in &tl.intervals[..tl.overlap_prefix_end(window_end)] {
```
shiv-tyagi: `brute_earliest_start` is the reference oracle for differential tests, but this line uses `tl.overlap_prefix_end(window_end)`, the optimized binary search that's being tested. An off-by-one in the `<` vs `<=` boundary of `overlap_prefix_end` would be invisible because the oracle shares the same bug.

Iterate all intervals unconditionally so the oracle is truly independent:

```suggestion
                for interval in &tl.intervals {
```

crates/spur-sched/src/timeline.rs
```diff
@@ -51,36 +135,31 @@ impl NodeTimeline {
...
-                    return candidate;
+            sweep.advance_to(candidate);
+
+            if self.total.can_satisfy_with_allocated(sweep.used(), request) {
+                if let Some(jump) = self.duration_blocker(&mut sweep, request, candidate, duration)
+                {
+                    candidate = jump;
+                    continue;
                 }
+                return candidate;
```
shiv-tyagi, line 172: Nit: the `else` branch below (L164-169) is the last remaining linear scan in `earliest_start`. The sweep's `ending` min-heap already tracks active end times. Exposing a `next_end()` on `AllocationSweep` that peeks the heap would give the minimum end time in O(1) instead of O(n).

shiv-tyagi (commented):
Posted comments. PTAL.

shiv-tyagi:
image

I did a quick comparison against main, results look clearly better.

shiv-tyagi (approved):
LGTM

PR #351, opened 2026-06-25
refactor(spur-metrics): drop MetricsExpositionFormat, ship OpenMetrics 1.0 only

### LINKED ISSUE #233: refactor(spur-metrics): drop MetricsExpositionFormat enum, ship OpenMetrics 1.0 only
Ref: #217

`MetricsExpositionFormat` in `spur-core/src/config.rs` has two variants: `Slurm_0_0_4` (default) and `OpenMetrics_1_0`. We should simplify this to just OpenMetrics 1.0.

**The "Slurm_0_0_4" variant doesn't correspond to a real format spec.** Slurm 25.11 labels its own output "OpenMetrics" (`MetricsType=metrics/openmetrics`) but happens to emit Prometheus 0.0.4 on the wire without `# EOF`. Spur's compatibility contract covers the CLI, REST API, and FFI. The metrics endpoint lives on a separate port (6822) outside that surface, so there's no need to match Slurm's wire behavior here.

**For gauges the body is identical between the two formats.** The only difference today is the `Content-Type` header and the `# EOF` trailer. However, once we add counters and histograms (scheduler cycle counts, RPC latencies), the formats diverge in meaningful ways (counter `_total` handling, native histograms, exemplars). Maintaining and testing dual encoding for every new metric type is work we can avoid.

**OpenMetrics is what Prometheus prefers.** The default scrape negotiation order is `OpenMetricsText1.0.0 > 0.0.1 > PrometheusText1.0.0 > PrometheusText0.0.4`. Every major scraper and exporter (kube-state-metrics, node_exporter, cAdvisor, OTel Collector) supports it.

**No users of this config option exist yet.** The endpoint is new in #217, so simplifying now is straightforward. Doing it later would be a breaking config change.

The `prometheus-client` crate's `encode()` already produces spec-compliant OpenMetrics 1.0. We can use it directly instead of branching around it.

shiv-tyagi:
## Summary

- Remove the `MetricsExpositionFormat` enum and its two variants (`Slurm_0_0_4`, `OpenMetrics_1_0`), along with the `exposition_format` config field. All metrics endpoints now unconditionally emit OpenMetrics 1.0 text.
- Collapse `encode_*_with_format` functions into simpler `encode_*` equivalents across `spur-metrics`, and simplify the `spurctld` metrics HTTP server accordingly.
- Delete Slurm 0.0.4 golden test fixtures; rename OpenMetrics fixtures to format-agnostic names.

The `Slurm_0_0_4` variant didn't correspond to a real spec — Slurm itself labels its output "OpenMetrics" but emits Prometheus 0.0.4 on the wire. Since the metrics HTTP endpoint (port 6822) is outside Spur's Slurm-compatibility surface, there's no reason to replicate that behavior. Simplifying now avoids a future breaking config change and eliminates dual-encoding maintenance as counters and histograms are added.

Closes #233

## Test plan

- [x] `cargo fmt --check --all` — clean
- [x] `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` — zero warnings
- [x] `cargo test --locked` — all tests pass


sgopinath1 (approved):
LGTM

sajmera-pensando (approved):
LGTM

PR #352, opened 2026-06-25
test(spurctld): replace metrics_server unit tests with handler tests

### LINKED ISSUE #232: test(spurctld): replace metrics_server unit tests with meaningful handler tests
Ref: #217

The current tests in `crates/spurctld/src/metrics_server.rs` (`leader_returns_slurm_content_type`, `leader_returns_openmetrics_content_type`, `stub_nodes_endpoint_returns_200_on_leader`, `follower_returns_503`) test a `#[cfg(test)]`-only helper function `leader_metrics_response` that doesn't exist in production code. They verify that axum's tuple-to-response conversion returns the status code and content-type it was given, e.g.:

```rust
let response = leader_metrics_response(true, format, body);
assert_eq!(response.status(), StatusCode::OK);
```

This is testing axum, not Spur. The `leader_metrics_response` helper is just:

```rust
fn leader_metrics_response(is_leader: bool, format: MetricsExpositionFormat, body: String) -> Response {
    if !is_leader { return not_leader_response(); }
    metrics_exposition_response(format, body)
}
```

Both `not_leader_response` and `metrics_exposition_response` are trivial one-liner tuple constructions. There's no Spur logic under test.

## What to do

- Remove `leader_metrics_response` and the four tests that use it.
- Replace with tests that exercise the actual axum handlers (`metrics_jobs`, `metrics_nodes`, etc.) using `axum::test::TestClient` or by constructing a `Router` with test state. These would cover the real flow: Raft leader check, snapshot collection, encoding, and correct response format.
- If full handler tests are too heavy for this PR, even just removing the current tests and the dead helper is a net improvement since they add maintenance cost without catching real bugs.

shiv-tyagi:
## Summary

- Remove the `#[cfg(test)]`-only `leader_metrics_response` helper and its tests that only exercised axum's tuple-to-response conversion — zero Spur logic was under test.
- Replace with 5 handler-level tests using `tower::ServiceExt::oneshot` against the real axum router backed by a test `ClusterManager` + single-node `RaftHandle`. These cover the actual request flow: Raft leader check, snapshot collection, encoding, and correct response format.
- Add `tower` and `http-body-util` as dev-dependencies for handler testing.

Closes #232

## Test plan

- [x] `cargo test -p spurctld -- metrics_server` — all 5 tests pass
- [x] `cargo clippy -p spurctld --all-targets --locked` — zero warnings


sgopinath1 (approved):
LGTM

PR #353, opened 2026-06-25
docs(AGENTS-md): warn against #N references in PR/issue descriptions

shiv-tyagi:
GitHub interprets `#N` as a cross-reference to issue/PR N. When the intent is to reference a numbered point within the same PR or issue description (e.g. "see point `#1` above"), this silently creates a wrong link.

Adds a note to the Git Workflow section of AGENTS.md recommending `[N]` instead.


sgopinath1 (approved):
LGTM

PR #359, opened 2026-06-27
feat(spur-metrics): add scheduler cycle and lifecycle statistics

sgopinath1:
## Summary

- Add `SchedStatsCollector` as the single in-memory accumulator for scheduler metrics on the Raft leader: cycle wall time, `Scheduler::schedule()` time, and jobs submitted/started/completed since reset.
- Wire export through three read paths: `GET /metrics/scheduler`, `GetSchedStats` gRPC, and the `sdiag` scheduler block (matching the existing job/node/RPC metrics pattern).
- Add `ResetDiagStats` so `sdiag --reset` clears RPC and scheduler counters in one leader-side call.

## Approach

- **Cycle timing** is recorded in `scheduler_loop` when the leader runs a scheduling pass with pending work. Panic paths still record a cycle with zero starts.
- **Lifecycle counters** hook into `cluster`: submit after all array tasks propose, start after prolog succeeds, complete in `run_job_finalized_side_effects` so all terminal paths are counted.
- **HTTP encoding** lives in `spur-metrics`; spurctld holds the collector and converts snapshots for gRPC. Diagnostic RPCs are excluded from RPC timing middleware.

## Test plan

- [x] `cargo test --locked`
- [x] Golden fixture for `/metrics/scheduler` (`scheduler_export_golden`)
- [x] `metrics_proto` parity test (HTTP gauges vs gRPC `SchedStats`)
- [x] `sched_stats_track_submit_start_complete` cluster integration test
- [x] Follower 503 test for `/metrics/scheduler`

### DIFF
--- a/crates/spur-metrics/src/export/scheduler.rs
+++ b/crates/spur-metrics/src/export/scheduler.rs
@@ -38,9 +38,9 @@ pub fn register_scheduler(registry: &mut Registry, snap: &SchedStatsSnapshot) {
 
     register_gauge(
         registry,
-        "spur_scheduler_cycles",
-        "Scheduling cycles executed since reset",
-        snap.cycles,
+        "spur_scheduler_cycle_last_time_us",
+        "Most recent scheduling cycle wall time in microseconds",
+        snap.cycle_last_time_us,
     );
     register_gauge(
         registry,
@@ -50,21 +50,9 @@ pub fn register_scheduler(registry: &mut Registry, snap: &SchedStatsSnapshot) {
     );
     register_gauge(
         registry,
-        "spur_scheduler_cycle_last_time_us",
-        "Most recent scheduling cycle wall time in microseconds",
-        snap.cycle_last_time_us,
-    );
-    register_gauge(
-        registry,
-        "spur_scheduler_cycle_avg_time_us",
-        "Average scheduling cycle wall time in microseconds",
-        snap.cycle_avg_time_us(),
-    );
-    register_gauge(
-        registry,
-        "spur_scheduler_schedule_total_time_us",
-        "Cumulative Scheduler::schedule() time in microseconds",
-        snap.schedule_total_time_us,
+        "spur_scheduler_cycles_total",
+        "Scheduling cycles executed since reset",
+        snap.cycles,
     );
     register_gauge(
         registry,
@@ -74,33 +62,33 @@ pub fn register_scheduler(registry: &mut Registry, snap: &SchedStatsSnapshot) {
     );
     register_gauge(
         registry,
-        "spur_scheduler_schedule_avg_time_us",
-        "Average Scheduler::schedule() time in microseconds",
-        snap.schedule_avg_time_us(),
-    );
-    register_gauge(
-        registry,
-        "spur_scheduler_jobs_submitted",
-        "Jobs submitted since reset",
-        snap.jobs_submitted,
-    );
-    register_gauge(
-        registry,
-        "spur_scheduler_jobs_started",
-        "Jobs started since reset",
-        snap.jobs_started,
-    );
-    register_gauge(
-        registry,
-        "spur_scheduler_jobs_finalized",
-        "Jobs reaching a terminal state since reset",
-        snap.jobs_finalized,
+        "spur_scheduler_schedule_total_time_us",
+        "Cumulative Scheduler::schedule() time in microseconds",
+        snap.schedule_total_time_us,
     );
     register_gauge(
         registry,
         "spur_scheduler_jobs_started_last_cycle",
         "Jobs started in the most recent scheduling cycle",
         snap.jobs_started_last_cycle,
+    );
+    register_gauge(
+        registry,
+        "spur_scheduler_jobs_submitted_total",
+        "Jobs submitted since reset",
+        snap.jobs_submitted,
+    );
+    register_gauge(
+        registry,
+        "spur_scheduler_jobs_started_total",
+        "Jobs started since reset",
+        snap.jobs_started,
+    );
+    register_gauge(
+        registry,
+        "spur_scheduler_jobs_finalized_total",
+        "Jobs reaching a terminal state since reset",
+        snap.jobs_finalized,
     );
 }
 
@@ -132,10 +120,10 @@ mod tests {
     fn export_includes_scheduler_gauges() {
         let body = encode_scheduler_metrics(&sample_snapshot());
         assert!(body.contains("spur_scheduler_info{plugin=\"backfill\"} 1"));
-        assert!(body.contains("spur_scheduler_cycles 10"));
-        assert!(body.contains("spur_scheduler_cycle_avg_time_us 500"));
-        assert!(body.contains("spur_scheduler_schedule_avg_time_us 150"));
-        assert!(body.contains("spur_scheduler_jobs_submitted 42"));
+        assert!(body.contains("spur_scheduler_cycles_total 10"));
+        assert!(body.contains("spur_scheduler_cycle_total_time_us 5000"));
+        assert!(!body.contains("_avg_time_us"));
+        assert!(body.contains("spur_scheduler_jobs_submitted_total 42"));
         assert!(body.contains("spur_scheduler_jobs_started_last_cycle 3"));
         assert!(body.ends_with("# EOF\n"));
     }
@@ -147,7 +135,7 @@ mod tests {
             ..Default::default()
         };
         let body = encode_scheduler_metrics(&snap);
-        assert!(body.contains("spur_scheduler_cycles 0"));
+        assert!(body.contains("spur_scheduler_cycles_total 0"));
         assert!(body.ends_with("# EOF\n"));
     }
 }
--- a/crates/spur-metrics/src/export/scheduler.rs
+++ b/crates/spur-metrics/src/export/scheduler.rs
@@ -1,27 +1,153 @@
 // Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-//! Scheduler gauge registration for `/metrics/scheduler` (Layer 1c).
+//! Scheduler gauge registration for `/metrics/scheduler`.
 
+use prometheus_client::encoding::EncodeLabelSet;
+use prometheus_client::metrics::family::Family;
+use prometheus_client::metrics::gauge::Gauge;
 use prometheus_client::registry::Registry;
+use std::sync::atomic::AtomicU64;
 
 use crate::export::encode_registered;
+use crate::export::register_gauge;
+use crate::scheduler::SchedStatsSnapshot;
 
-/// Register scheduler catalog gauges (stub until scheduler snapshot exists).
-pub fn register_scheduler(_registry: &mut Registry) {}
+#[derive(Clone, Debug, Hash, PartialEq, Eq, EncodeLabelSet)]
+struct PluginLabel {
+    plugin: String,
+}
+
+fn set_plugin_info(registry: &mut Registry, snap: &SchedStatsSnapshot) {
+    let family = Family::<PluginLabel, Gauge<u64, AtomicU64>>::default();
+    family
+        .get_or_create(&PluginLabel {
+            plugin: snap.plugin.clone(),
+        })
+        .set(1);
+    registry.register(
+        "spur_scheduler_info",
+        "Active scheduler plugin (value is always 1)",
+        family,
+    );
+}
+
+/// Register scheduler gauges into `registry` from `snap`.
+pub fn register_scheduler(registry: &mut Registry, snap: &SchedStatsSnapshot) {
+    set_plugin_info(registry, snap);
+
+    register_gauge(
+        registry,
+        "spur_scheduler_cycles",
+        "Scheduling cycles executed since reset",
+        snap.cycles,
+    );
+    register_gauge(
+        registry,
+        "spur_scheduler_cycle_total_time_us",
+        "Cumulative scheduling cycle wall time in microseconds",
+        snap.cycle_total_time_us,
+    );
+    register_gauge(
+        registry,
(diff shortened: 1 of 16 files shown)


crates/spurctld/src/cluster.rs
```diff
@@ -594,6 +605,9 @@ impl ClusterManager {
     }
 
     fn run_job_finalized_side_effects(&self, finalized: JobFinalized) {
+        if let Some(stats) = self.sched_stats.read().as_ref() {
+            stats.record_completed();
+        }
         self.run_epilog_slurmctld(finalized.job_id);
```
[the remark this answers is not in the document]
sgopinath1: Fixed

crates/spurctld/src/cluster.rs
```diff
@@ -132,6 +136,13 @@ impl ClusterManager {
             })?;
         }
 
+        if submitted > 0 {
+            // Array expansion is all-or-nothing: count only after every JobSubmit propose succeeds.
```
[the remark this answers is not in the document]
sgopinath1: Fixed

crates/spurctld/src/sched_stats.rs
```diff
@@ -0,0 +1,132 @@
...
+    pub fn record_submitted(&self, count: u64) {
+        if count > 0 {
+            let mut accum = self.inner.lock();
+            accum.jobs_submitted = accum.jobs_submitted.saturating_add(count);
+        }
+    }
+
+    pub fn record_started(&self) {
+        self.inner.lock().jobs_started += 1;
+    }
```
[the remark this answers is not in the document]
sgopinath1: Fixed

crates/spurctld/src/sched_stats.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    pub fn record_completed(&self) {
+        self.inner.lock().jobs_completed += 1;
+    }
```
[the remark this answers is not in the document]
sgopinath1: Fixed

crates/spurctld/src/sched_stats.rs
```diff
@@ -0,0 +1,134 @@
...
+
+//! In-memory accumulator for scheduler cycle and lifecycle statistics.
+
+use parking_lot::Mutex;
+use spur_metrics::SchedStatsSnapshot;
+
+/// Leader-side scheduler statistics since process start or the last reset.
+#[derive(Debug)]
+pub struct SchedStatsCollector {
+    plugin: String,
```
yansun1996: **Suggestion**: The accumulator uses a `Mutex<SchedAccum>` for simple counters. Since these are independent u64 counters being incremented, consider using `AtomicU64` with `Relaxed` ordering (matching the pattern in spur-metrics RpcOperationSnapshot). This would eliminate lock contention in the hot path (`record_submitted`, `record_started`, `record_finalized` are called per-job).

However, `record_cycle` does update multiple related fields atomically (last + total + count), so a Mutex makes sense there. You could split: atomics for the simple lifecycle counters, Mutex just for cycle timing. Low priority since contention is unlikely at scheduler cycle granularity.

crates/spur-metrics/src/export/scheduler.rs
```diff
@@ -1,27 +1,153 @@
...
+        "Cumulative Scheduler::schedule() time in microseconds",
+        snap.schedule_total_time_us,
+    );
+    register_gauge(
+        registry,
+        "spur_scheduler_schedule_last_time_us",
+        "Most recent Scheduler::schedule() time in microseconds",
+        snap.schedule_last_time_us,
+    );
+    register_gauge(
```
yansun1996: **Nit (Prometheus conventions)**: Monotonically increasing counters should use `_total` suffix and `TYPE counter`, not `TYPE gauge`. `jobs_submitted`, `jobs_started`, `jobs_finalized`, and `cycles` are all monotonic (until reset), so they should be:

```
spur_scheduler_jobs_submitted_total  # TYPE counter
spur_scheduler_cycles_total          # TYPE counter
```

Gauges are for point-in-time values that can go up or down (like `jobs_started_last_cycle`, `cycle_last_time_us`). The `_avg_` metrics are also arguably gauges since they're derived.

This affects Prometheus rate() / increase() semantics. Not blocking but worth considering for semantic correctness.

yansun1996 (commented):
## Code Review: feat(spur-metrics): add scheduler cycle and lifecycle statistics

### Summary
This PR adds scheduler statistics collection and export through three paths: HTTP metrics endpoint (`/metrics/scheduler`), gRPC (`GetSchedStats`), and CLI (`sdiag`). The implementation follows existing patterns (RpcStatsCollector) and includes good test coverage.

### Findings

**Posted as inline comments:**
1. **Suggestion** (sched_stats.rs:12): Consider AtomicU64 for lifecycle counters instead of Mutex for the entire struct. Low priority since contention is unlikely.
2. **Nit** (export/scheduler.rs): Prometheus convention suggests `_total` suffix and `TYPE counter` for monotonic counters (cycles, jobs_submitted, jobs_started, jobs_finalized). Currently all are gauges.

**Additional findings:**

3. **Suggestion** (cluster.rs): `sched_stats: RwLock<Option<Arc<SchedStatsCollector>>>` adds read-lock overhead on hot paths. Since SchedStatsCollector has internal synchronization, consider `OnceLock<Arc<...>>` instead. Matches existing accounting pattern though, so consistency may be intentional.

4. **Good**: Saturating arithmetic used throughout (`saturating_add`) prevents overflow. Time conversion uses `.min(u64::MAX as u128) as u64` pattern correctly.

5. **Good**: Diagnostic RPCs excluded from RPC timing middleware (`GetSchedStats`, `ResetSchedStats`, `ResetDiagStats`).

6. **Good**: Panic handling in scheduler loop still records a cycle with zero starts, as documented in PR description.

7. **Good**: Test coverage includes golden fixture, proto parity test, and cluster integration test.

### Verdict
The implementation is correct and well-tested. The suggestions above are minor optimizations and convention nits - no blocking issues. **Approve** when ready.

crates/spur-metrics/src/export/scheduler.rs
```diff
@@ -1,27 +1,153 @@
...
+
+    register_gauge(
+        registry,
+        "spur_scheduler_cycles",
+        "Scheduling cycles executed since reset",
+        snap.cycles,
+    );
+    register_gauge(
+        registry,
+        "spur_scheduler_cycle_total_time_us",
```
shiv-tyagi, line 47: Total should be at the end

crates/spur-metrics/src/export/scheduler.rs
```diff
@@ -1,27 +1,153 @@
...
+    );
+}
+
+/// Register scheduler gauges into `registry` from `snap`.
+pub fn register_scheduler(registry: &mut Registry, snap: &SchedStatsSnapshot) {
+    set_plugin_info(registry, snap);
+
+    register_gauge(
+        registry,
+        "spur_scheduler_cycles",
```
shiv-tyagi: We should have _total suffix as pointed out by @yansun1996

crates/spur-metrics/src/export/scheduler.rs
```diff
@@ -1,27 +1,153 @@
...
+    );
+    register_gauge(
+        registry,
+        "spur_scheduler_cycle_last_time_us",
+        "Most recent scheduling cycle wall time in microseconds",
+        snap.cycle_last_time_us,
+    );
+    register_gauge(
+        registry,
+        "spur_scheduler_cycle_avg_time_us",
```
shiv-tyagi: Avg. is only good to have thing, we should check if prometheus standards allow it, if not we can drop it here and let only sdiag have it. 

Also if sdiag for slurm doesn't have it, we should surely get rid of it here.

crates/spur-metrics/tests/scheduler_export_golden.rs
```diff
@@ -0,0 +1,91 @@
...
+#[test]
+fn golden_scheduler_metrics() {
+    let body = normalize_exposition(&encode_scheduler_metrics(&sample_snapshot()));
+    let path = fixtures_dir().join("scheduler.prom");
+    let expected =
+        std::fs::read_to_string(&path).unwrap_or_else(|e| panic!("read {}: {e}", path.display()));
+    assert_eq!(body, expected);
+}
+
+/// Regenerate `tests/fixtures/scheduler.prom` after intentional encoder changes:
```
shiv-tyagi: Do we need a specific test for it or should we use a flag/env var the the very tests which do the matching so that when the flag is set, instead of comparing they update the fixture. 

Also if it is trivial to fix it AI on each drift, why not let AI do it while doing the PRs?

crates/spurctld/src/sched_stats.rs
```diff
@@ -0,0 +1,134 @@
...
+    pub fn new(plugin: impl Into<String>) -> Self {
+        Self {
+            plugin: plugin.into(),
+            inner: Mutex::new(SchedAccum::default()),
+        }
+    }
+
+    pub fn record_cycle(&self, cycle_time_us: u64, schedule_time_us: u64, jobs_started: u64) {
+        let mut accum = self.inner.lock();
+        accum.cycles += 1;
```
shiv-tyagi: `cycles += 1` is a plain add, but `jobs_started`, `jobs_finalized`, and both `*_total_time_us` fields use `saturating_add`. Same overflow-in-debug argument applies here.

```suggestion
        accum.cycles = accum.cycles.saturating_add(1);
```

crates/spurctld/src/scheduler_loop.rs
```diff
@@ -235,6 +245,9 @@ pub async fn run(cluster: Arc<ClusterManager>, raft: Arc<RaftHandle>) {
                 }
             }
 
+            cluster.record_sched_job_started();
```
shiv-tyagi: `record_sched_job_started()` is redundant. The loop already accumulates `jobs_started_cycle` and hands it to `record_cycle`, and cumulative `jobs_started` is just the running sum of that. Drop this per-job call (and `record_sched_job_started` / `record_started`) and update both fields in `record_cycle`:

```rust
accum.jobs_started = accum.jobs_started.saturating_add(jobs_started);
accum.jobs_started_last_cycle = jobs_started;
```

That removes a method and a per-job lock acquisition.

proto/slurm.proto
```diff
@@ -285,6 +285,11 @@ service SlurmController {
   rpc GetRpcStats(google.protobuf.Empty) returns (RpcStats);
   rpc ResetRpcStats(google.protobuf.Empty) returns (google.protobuf.Empty);
 
+  // Scheduler cycle and lifecycle statistics (since process start or last reset)
+  rpc GetSchedStats(google.protobuf.Empty) returns (SchedStats);
+  rpc ResetSchedStats(google.protobuf.Empty) returns (google.protobuf.Empty);
```
shiv-tyagi: Q: ResetSchedStats has no caller. `sdiag --reset` uses ResetDiagStats, which already resets both RPC and scheduler counters. Pre-1.0, is the standalone ResetSchedStats needed, or is it speculative API? (ResetRpcStats looks similarly orphaned now that ResetDiagStats exists.)

shiv-tyagi (changes requested):
Posted comments

crates/spurctld/src/sched_stats.rs
sgopinath1: Done.

crates/spur-metrics/src/export/scheduler.rs
sgopinath1: Adopted _total on monotonic fields (spur_scheduler_cycles_total, jobs_*_total, etc.). Keeping TYPE gauge though: these are since-reset accumulators cleared by ResetDiagStats, same as /metrics/rpc. Literal counters would decrease on reset and break rate()/increase() semantics unless we document counter-reset behavior.

sgopinath1:
@yansun1996 sched_stats is now OnceLock<Arc<SchedStatsCollector>>, set once at startup.

crates/spur-metrics/src/export/scheduler.rs
sgopinath1: Done.

crates/spur-metrics/src/export/scheduler.rs
sgopinath1: Done.

crates/spur-metrics/src/export/scheduler.rs
sgopinath1: Done.

crates/spur-metrics/tests/scheduler_export_golden.rs
sgopinath1: Done. The matcher test now honors UPDATE_GOLDEN=1 to rewrite the fixture instead of a separate #[ignore] refresh test. On “let AI fix drift in PRs”: we’d rather keep golden updates explicit and opt-in so fixture changes show up in the diff for human review.

crates/spurctld/src/sched_stats.rs
sgopinath1: Done.

crates/spurctld/src/scheduler_loop.rs
sgopinath1: Done.

proto/slurm.proto
sgopinath1: Agreed — removed both. ResetDiagStats is the only reset RPC (sdiag --reset); no callers needed the per-subsystem resets.

crates/spur-metrics/src/export/scheduler.rs
sgopinath1: Ignore my earlier comment. Adopted TYPE counter for monotonic since-reset fields (cycles, jobs_submitted, jobs_started, jobs_finalized). Cumulative timing and last-value metrics stay gauges. prometheus-client adds the _total sample suffix automatically.

shiv-tyagi (approved):
LGTM

PR #361, opened 2026-06-29
Ci/trivy json reports

karannathamd:
DO NOT MERGE. WIP

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -231,7 +231,7 @@ jobs:
       - name: Load image
         run: docker load -i /tmp/spur-image.tar
 
-      - name: Scan image for vulnerabilities and secrets
+      - name: Scan image for vulnerabilities and secrets (SARIF)
         id: scan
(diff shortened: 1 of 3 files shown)


shiv-tyagi:
- We have cargo deny for SBOM generation and dependency license and security advisory checks
- We do not need duplicate scan to upload scan results in a different format. 
- If we really need JIRA, we can use any Github-to-JIRA code scanning alert syncing action. There is also one native JIRA integration which pulls those from github as well.

[Integrate GitHub Advanced Security with Jira | Atlassian Support](https://support.atlassian.com/jira-cloud-administration/docs/integrate-github-for-security/)
[GitHub - github/ghas-jira-integration: Synchronize GitHub Code Scanning alerts to Jira issues](https://github.com/github/ghas-jira-integration)

Given that, I believe we are quite covered on Spur side with cargo-deny, dependabot and trivy. 

Also about JIRA, my personal opinion is that security issues are always P0 issues. We should aggressively fix them. We do not need time tracking for those, rather we need to ship fixes the fastest. Any open alert is equally urgent. So if time tracking on JIRA is the only reason for us to get into all this hassle, I would avoid it.

cc @sgopinath1 @shreyarajamd

PR #371, opened 2026-06-30
fix(spurctld): initialize next_job_id from first_job_id on cold start

### LINKED ISSUE #369: bug(spurctld): [controller] first_job_id ignored on cold start — only applied during snapshot restore
## Summary

The `[controller] first_job_id` config option is **not honored when a controller starts with fresh/empty Raft state**. The job-ID counter always begins at `1` on a cold start, regardless of `first_job_id`. The setting only takes effect later, if and when the cluster restores from a Raft snapshot.

This matters after any Raft-state wipe (e.g. a cluster re-form): operators expect `first_job_id = N` to make the new cluster start numbering at `N` (the classic Slurm `FirstJobId` use case — avoid colliding new job IDs with an existing accounting DB). Instead the counter restarts at 1.

## Expected

`first_job_id` should seed the job-ID counter on **every** startup path, including a cold start with no prior state — matching Slurm's `FirstJobId` semantics. Setting `first_job_id = 110` should make the first submitted job get id `110`.

## Actual

With `first_job_id = 110` set in `[controller]` on all controllers and a freshly-wiped state dir, the first submitted job gets id `1` (then 2, 3, …). The config value is silently ignored.

## Root cause (source)

In `crates/spurctld/src/cluster.rs`:

- The `ClusterManager` constructor hardcodes the counter:
  ```rust
  next_job_id: AtomicU32::new(1),   // line ~79
  ```
- `first_job_id` is read in exactly one place — `restore_from_snapshot()`:
  ```rust
  let mut next_id = self.config.controller.first_job_id;   // line ~2382
  for job in snap.jobs { next_id = next_id.max(job.job_id + 1); }
  ...
  self.next_job_id.store(next_id, Ordering::Relaxed);
  ```

So `first_job_id` is only consulted when a snapshot is restored. On a cold start (no snapshot yet — the controller recovers from an empty or log-only state), `restore_from_snapshot` never runs, and the counter stays at the constructor default of `1`. A grep of the crate confirms `first_job_id` has no other (non-test) reader.

## Reproduce

1. Set `first_job_id = 110` under `[controller]` in `spur.conf` on all controllers.
2. Wipe the Raft state dir on all controllers and (re)start them (fresh bootstrap, no snapshot).
3. `spur submit` a trivial job → it gets JobId `1`, not `110`.

## Impact

- After a cluster re-form / Raft wipe, new job IDs restart at 1 and **collide with historical rows in the accounting DB** (spurdbd keys `jobs` by `job_id INTEGER PRIMARY KEY` with `ON CONFLICT DO UPDATE`, so new jobs upsert over old accounting records as the counter climbs back through the previously-used range). `first_job_id` is exactly the knob meant to prevent this, but it doesn't work for the cold-start case where it's most needed.

## Suggested direction (not prescriptive)

Initialize the counter from `config.controller.first_job_id` at construction (or in the cold-start/init path), not only inside `restore_from_snapshot`. The snapshot-restore path can keep its `max(first_job_id, max_job_id + 1)` logic for the warm case.

## Environment

- spurctld 0.3.0
- 3-node HA, systemd-managed, persistent `--state-dir`

yansun1996:
## What

`[controller] first_job_id` in `spur.conf` was silently ignored on cold start. The job counter always started at 1 regardless of the configured value, causing job ID collisions with historical rows in the accounting DB after a cluster re-form.

## Root cause

`ClusterManager::new()` hardcoded `next_job_id: AtomicU32::new(1)`. `first_job_id` was only consulted in `restore_from_snapshot()`, which never runs on a cold start.

## Fix

Extract `config.controller.first_job_id` before moving `config` into the struct, and use it to initialize the counter. The snapshot restore path already does `max(first_job_id, max_job_id + 1)` and is unchanged.

## Testing

- New unit test `cold_start_respects_first_job_id`: sets `first_job_id = 500`, bootstraps a single-node Raft cluster from empty state, submits a job, asserts it receives ID 500.
- Live tested: `first_job_id = 1000` in config → first submitted job got ID 1000.

Fixes #369

### DIFF
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -3365,17 +3365,7 @@ mod tests {
         let dir = TempDir::new().unwrap();
         let mut config = test_config();
         config.controller.first_job_id = 500;
-        let cm = Arc::new(ClusterManager::new(config, dir.path()).unwrap());
-        let handle = crate::raft::start_raft(1, &["[::1]:0".into()], dir.path(), cm.clone())
-            .await
-            .unwrap();
-        handle
-            .raft
-            .wait(Some(std::time::Duration::from_secs(5)))
-            .metrics(|m| m.current_leader == Some(1), "leader elected")
-            .await
-            .expect("single-node raft did not self-elect within 5s");
-        cm.set_raft(handle.raft);
+        let cm = test_cluster_with_config(&dir, config).await;
 
         assert_eq!(cm.next_job_id.load(Ordering::Relaxed), 500);
 
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -79,6 +79,7 @@ impl ClusterManager {
         let license_pool = config.licenses.clone();
         let burst_buffer_total_gb = config.burst_buffer.total_gb;
         let fairshare_cache = Arc::new(FairshareCache::new());
+        let first_job_id = config.controller.first_job_id;
         let qos_cache = Arc::new(QosCache::new());
 
         let cm = Self {
@@ -88,7 +89,7 @@ impl ClusterManager {
             partitions: RwLock::new(partitions),
             reservations: RwLock::new(Vec::new()),
             steps: RwLock::new(HashMap::new()),
-            next_job_id: AtomicU32::new(1),
+            next_job_id: AtomicU32::new(first_job_id),
             license_pool: RwLock::new(license_pool),
             burst_buffer_total_gb: RwLock::new(burst_buffer_total_gb),
             tokens: RwLock::new(HashMap::new()),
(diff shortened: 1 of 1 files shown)


crates/spurctld/src/cluster.rs
```diff
@@ -3359,6 +3360,29 @@ mod tests {
...
+            .wait(Some(std::time::Duration::from_secs(5)))
+            .metrics(|m| m.current_leader == Some(1), "leader elected")
+            .await
+            .expect("single-node raft did not self-elect within 5s");
+        cm.set_raft(handle.raft);
+
+        assert_eq!(cm.next_job_id.load(Ordering::Relaxed), 500);
+
+        let job_id = submit_and_wait(&cm, basic_spec("first-job"));
+        assert_eq!(job_id, 500);
```
shiv-tyagi, line 3373: This copy-pastes the full Raft bring-up sequence (`start_raft` + `wait(...).metrics(leader elected)` + `set_raft`) that `test_cluster_with_config` right above already does verbatim. Use the helper:

```suggestion
    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
    async fn cold_start_respects_first_job_id() {
        let dir = TempDir::new().unwrap();
        let mut config = test_config();
        config.controller.first_job_id = 500;
        let cm = test_cluster_with_config(&dir, config).await;

        assert_eq!(cm.next_job_id.load(Ordering::Relaxed), 500);

        let job_id = submit_and_wait(&cm, basic_spec("first-job"));
        assert_eq!(job_id, 500);
    }
```

shiv-tyagi (commented):
PTAL

crates/spurctld/src/cluster.rs
yansun1996: Done in 4fb18c9 — the test now uses the `test_cluster_with_config` helper and the duplicated `start_raft` + leader-wait + `set_raft` block is removed. Assertions (counter seeded to `first_job_id`, first job id == 500) preserved; still a genuine cold-start path (fresh empty state, snapshot-restore never runs). `cargo test` + `clippy` clean, and live-verified on a single-node cluster (first submitted job got ID 500).

yansun1996:
@shiv-tyagi thanks for the review — addressed in 4fb18c9.

- Refactored `cold_start_respects_first_job_id` to use the existing `test_cluster_with_config` helper; removed the duplicated Raft bring-up. Assertions unchanged.
- `cargo test -p spurctld` + `cargo clippy --all-targets` clean.
- **Live-tested on a single-node cluster (node 147):** cold start with `first_job_id = 500` → first submitted job received JobId **500** (then 501, 502). Confirmed against a build without the fix, which returned JobId 1.

Review thread resolved. PTAL.

shiv-tyagi (approved):
Nice. Thanks @yansun1996.

PR #373, opened 2026-06-30
fix(spurctld): persist last_purged to survive restart after log compaction

### LINKED ISSUE #368: bug(spurctld): controller cannot restart after log compaction — last_purged not persisted → openraft "LogIndex(0) violates: try to get log at index 0 but got None"
## Summary

A `spurctld` controller whose Raft log has been snapshot-compacted (purged) **cannot restart**. On startup it panics/exits with:

```
Error: 'LogIndex(0)' violates: 'try to get log at index 0 but got None'
```

and systemd crash-loops it. Because compaction happens automatically as the cluster runs, **every controller becomes un-restartable over time**. In an HA deployment this means a routine `systemctl restart` (or a host reboot) permanently bricks that controller's Raft state. The surviving controllers only stay up because they have never re-read state from disk since the purge.

This is the post-compaction counterpart to #273 (which is the panic when the log grows *without* compaction).

## Correction — reopening was based on a disproven hypothesis; #373 is confirmed sufficient

This issue was briefly reopened with a claim that `install_snapshot()` (the path a follower takes when it receives a snapshot from the leader via replication) had the same unpersisted-`last_purged` defect as `purge_logs_upto()`, untouched by #373. That hypothesis looked right from reading the source — `install_snapshot()` sets `inner.last_applied` but never touches `inner.last_purged` or calls `persist_last_purged()` — but **live testing on a real 3-node cluster disproved it**:

- Built two binaries: `upstream/main` as of #373 (has the `purge_logs_upto()` fix, does NOT have `install_snapshot()` touching `last_purged`) and a pre-#373 commit, both with lowered snapshot/replication-lag thresholds to trigger the scenario quickly.
- Stood up a real 3-node Raft cluster (2-of-3 quorum, so one node can be down without blocking writes), stopped one follower, drove enough writes on the leader to force a snapshot + log purge past `max_in_snapshot_log_to_keep`, then restarted the follower so it had to receive an `InstallSnapshot` RPC.
- **Post-#373 binary**: follower received the snapshot, logged `installed snapshot from leader`, and on the *next* restart came back up cleanly. No crash.
- **Pre-#373 binary**, identical scenario: follower crashed on restart with the exact panic, `last_purged_log_id=None last_applied=T1-N3-39`, `load membership from log: [40..42)` — byte-for-byte the same shape as this issue's original report.

Root cause of the discrepancy: openraft's engine always issues an internal `purge_log` instruction immediately after *any* `install_snapshot` completes — regardless of whether the follower fell behind via replication lag or compacted locally — and that instruction is served by the *same* `purge_logs_upto()` storage method in both cases. #373 fixed `purge_logs_upto()` once, and that single fix covers both triggering paths. `install_snapshot()` itself never needed its own `last_purged` handling, because it was never the function actually responsible for persisting it.

**Conclusion**: the original production crash that prompted reopening this issue was almost certainly running a `spurctld` build predating #373 (merged 2026-07-06), not evidence of a second live gap. No code change is needed beyond #373. Re-closing.

---

## Impact (critical)

- HA controllers do not survive a restart → HA does not actually provide fault tolerance against the most basic operation (restarting/rebooting a controller).
- Observed on a live 3-controller cluster: restarting one controller to pick up a `spur.conf` change bricked it; the other two are now fragile (a restart/reboot of either would hit the same panic and break quorum).
- No data loss in the state machine (snapshot is intact), but the node is unrecoverable without a code fix.

## Root cause (confirmed in source)

`crates/spurctld/src/raft.rs`, `SpurStore`:

- `purge_logs_upto()` (~line 320) sets `inner.last_purged = Some(log_id)` **in memory only** and deletes the on-disk log files. `last_purged` is **never persisted** to disk.
- `SpurStore::new()` (~line 94) reloads `vote.json`, the log entry files, and `snapshot.json` on startup — but **does not restore `last_purged`** (there is nothing on disk to restore it from). It defaults to `None`.
- `get_log_state()` (~line 264) then returns `last_purged_log_id: None` while `last_log_id` is the first surviving entry (e.g. index 10000, snapshot at 9999).
- openraft, seeing `last_purged = None`, walks the log back toward index 0, finds nothing there, and aborts with `'LogIndex(0)' violates: 'try to get log at index 0 but got None'`.

Startup log from the affected node (snapshot at 9999, logs 10000–10246):

```
INFO spurctld::cluster: restored cluster state from Raft snapshot jobs=91 nodes=15
INFO spurctld::raft: raft store recovered from disk log_entries=1247 vote=Some(Vote { term: 1, node_id: 3, committed: true })
INFO openraft::storage::helper: get_initial_state vote=T1-N3:committed last_purged_log_id=None last_applied=T1-N3-9999 committed=None last_log_id=T1-N3-10246
INFO openraft::storage::helper: load membership from log: [10000..10247)
INFO openraft::storage::helper: load key log ids from (None,T1-N3-10246]
Error: 'LogIndex(0)' violates: 'try to get log at index 0 but got None'
spurctld.service: Main process exited, code=exited, status=1/FAILURE
```

## Reproduce

1. Run a controller long enough for openraft to take a snapshot and purge the log (or force a snapshot).
2. `systemctl restart spurctld` (or reboot the host).
3. spurctld exits 1 on every start with the `LogIndex(0)` error; systemd crash-loops it.

## Recovery paths that do NOT work (all reproduce the bug)

- **Plain restart** — hits the panic (log already purged).
- **Wipe state-dir and let it rejoin** — different but also fatal: with empty state the symmetric `raft.initialize(members)` in `start_raft()` (`raft.rs` ~line 663) actually bootstraps a *new* cluster and races the leader's `InstallSnapshot`, panicking in openraft `raft_core.rs:779` with `called Option::unwrap() on a None value`. So a wiped node cannot cleanly rejoin an existing cluster either.
- **Seed the node with the leader's state-dir** (snapshot + logs) — reproduces the identical `LogIndex(0)` panic, because the copied on-disk state also has no persisted `last_purged`. This confirms the defect is in the recovery code, not the data.

## Suggested fix direction (not prescriptive)

Persist `last_purged` durably (e.g. a `purged.json` / `log_state.json` written in `purge_logs_upto`, or derive it from snapshot meta) and restore it in `SpurStore::new()` so `get_log_state()` reports the correct `last_purged_log_id` after a compaction+restart. Separately, the empty-state rejoin path (unconditional `initialize()` in `start_raft`) should not bootstrap when the node is meant to join an existing cluster — that is a second, related crash.

## Environment

- spurctld version 0.3.0, openraft 0.9.24
- 3-node HA (node_ids 1/2/3), systemd-managed, persistent `--state-dir`

yansun1996:
## What

After openraft took a snapshot and compacted the log, `spurctld` could not restart:

```
Error: 'LogIndex(0)' violates: 'try to get log at index 0 but got None'
spurctld.service: Main process exited, code=exited, status=1/FAILURE
```

Since compaction happens automatically as the cluster runs, every controller eventually became permanently un-restartable — including on `systemctl restart` or host reboot. HA provided no real fault tolerance.

## Root cause

`purge_logs_upto()` set `inner.last_purged = Some(log_id)` in memory only and deleted the on-disk log files. `SpurStore::new()` had no `purged.json` to restore from, so `last_purged` defaulted to `None`. openraft, seeing `last_purged = None` with no log entries going back to index 0, aborted.

## Fix

- **`purge_logs_upto()`**: persist `last_purged` to `purged.json` before mutating in-memory state. Uses write-then-rename for atomicity — a crash mid-write can never leave a partially-written file.
- **`SpurStore::new()`**: load `purged.json` on startup if it exists. Missing file → `None` (correct for a fresh node). Parse error → hard failure (silently falling back to `None` would reproduce the panic).

## Testing

- Two new unit tests:
  - `store_persists_last_purged_across_restart`: calls public `purge_logs_upto()`, reconstructs store from same dir, asserts `get_log_state()` returns the correct `last_purged_log_id`.
  - `store_no_purged_file_recovers_as_none`: fresh store has `last_purged = None`.
- Live tested on a 1-node cluster: `systemctl restart spurctld` after compaction → recovered log entries, became leader, accepted new jobs.

Fixes #368

### DIFF
--- a/crates/spurctld/src/raft.rs
+++ b/crates/spurctld/src/raft.rs
@@ -127,15 +127,17 @@ impl SpurStore {
             }
         }
 
+        let mut snapshot_last_log_id: Option<LogId<NodeId>> = None;
         let snap_path = raft_dir.join("snapshot.json");
         if snap_path.exists() {
             // Soft-fail: a corrupt snapshot triggers re-snapshot on the next leader term.
-            // A missing/corrupt purged.json (below) cannot be recovered the same way.
+            // A present-but-corrupt purged.json (below) cannot be recovered the same way.
             match std::fs::read_to_string(&snap_path) {
                 Ok(data) => match serde_json::from_str::<PersistedSnapshot>(&data) {
                     Ok(ps) => {
                         inner.last_applied = ps.meta.last_log_id;
                         inner.last_membership = ps.meta.last_membership.clone();
+                        snapshot_last_log_id = ps.meta.last_log_id;
                         applier.restore_from_snapshot(&ps.data);
                     }
                     Err(e) => warn!("failed to parse snapshot.json: {e}"),
@@ -152,6 +154,18 @@ impl SpurStore {
                 serde_json::from_str::<LogId<NodeId>>(&data)
                     .map_err(|e| anyhow::anyhow!("failed to parse purged.json: {e}"))?,
             );
+        } else if let Some(snap_last) = snapshot_last_log_id {
+            // Backward compat: a node that compacted its log before purged.json
+            // existed has a snapshot but no purged.json. openraft purges up to the
+            // snapshot's last_log_id when it compacts, so that id is the correct
+            // last_purged lower bound. Without this, such a node recovers with
+            // last_purged=None and reproduces the startup panic the moment openraft
+            // reads a purged log range.
+            warn!(
+                last_purged = ?snap_last,
+                "purged.json missing; deriving last_purged from snapshot meta (pre-patch compat)"
+            );
+            inner.last_purged = Some(snap_last);
         }
 
         info!(
@@ -910,5 +924,38 @@ mod tests {
         let store = SpurStore::new(dir.path(), noop_applier()).unwrap();
         let inner = store.inner.read();
         assert!(inner.last_purged.is_none());
+    }
+
+    // A node that compacted its log before purged.json existed has a snapshot
+    // but no purged.json. Recovery must derive last_purged from the snapshot
+    // meta rather than defaulting to None (which reproduces the startup panic).
+    #[tokio::test]
+    async fn store_derives_last_purged_from_snapshot_when_purged_file_absent() {
+        use openraft::RaftStorage;
+        let dir = TempDir::new().unwrap();
+
+        let snap_last = LogId {
+            leader_id: openraft::LeaderId {
+                term: 3,
+                node_id: 1,
+            },
+            index: 100,
+        };
+
+        // Write a snapshot but no purged.json, mirroring a pre-patch compacted node.
+        {
+            let store = SpurStore::new(dir.path(), noop_applier()).unwrap();
+            let meta = SnapshotMeta {
+                last_log_id: Some(snap_last),
+                last_membership: StoredMembership::default(),
+                snapshot_id: "snap-100".into(),
+            };
+            store.persist_snapshot(&meta, b"snapshot data").unwrap();
+        }
+        assert!(!dir.path().join("raft").join("purged.json").exists());
+
+        let mut store = Arc::new(SpurStore::new(dir.path(), noop_applier()).unwrap());
+        let state = store.get_log_state().await.unwrap();
+        assert_eq!(state.last_purged_log_id, Some(snap_last));
     }
 }
--- a/crates/spurctld/src/raft.rs
+++ b/crates/spurctld/src/raft.rs
@@ -129,6 +129,8 @@ impl SpurStore {
 
         let snap_path = raft_dir.join("snapshot.json");
         if snap_path.exists() {
+            // Soft-fail: a corrupt snapshot triggers re-snapshot on the next leader term.
+            // A missing/corrupt purged.json (below) cannot be recovered the same way.
             match std::fs::read_to_string(&snap_path) {
                 Ok(data) => match serde_json::from_str::<PersistedSnapshot>(&data) {
                     Ok(ps) => {
@@ -142,6 +144,16 @@ impl SpurStore {
             }
         }
 
+        let purged_path = raft_dir.join("purged.json");
+        if purged_path.exists() {
+            // Hard-fail: silently falling back to None reproduces the startup panic.
+            let data = std::fs::read_to_string(&purged_path)?;
+            inner.last_purged = Some(
+                serde_json::from_str::<LogId<NodeId>>(&data)
+                    .map_err(|e| anyhow::anyhow!("failed to parse purged.json: {e}"))?,
+            );
+        }
+
         info!(
             log_entries = inner.log.len(),
             vote = ?inner.vote,
@@ -153,6 +165,18 @@ impl SpurStore {
             raft_dir,
             applier,
         })
+    }
+
+    fn persist_last_purged(&self, log_id: &LogId<NodeId>) -> Result<(), std::io::Error> {
+        let path = self.raft_dir.join("purged.json");
+        let tmp = self.raft_dir.join("purged.json.tmp");
+        let data = serde_json::to_vec(log_id)
+            .map_err(|e| std::io::Error::new(std::io::ErrorKind::InvalidData, e))?;
+        // Write-then-rename: a crash mid-write must never corrupt purged.json.
+        std::fs::write(&tmp, &data)
+            .map_err(|e| std::io::Error::new(e.kind(), format!("{tmp:?}: {e}")))?;
+        std::fs::rename(&tmp, &path)
+            .map_err(|e| std::io::Error::new(e.kind(), format!("{path:?}: {e}")))
     }
 
     fn persist_vote(&self, vote: &Vote<NodeId>) -> Result<(), std::io::Error> {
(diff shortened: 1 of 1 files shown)


shiv-tyagi (commented):
Looks good mostly. Please check if the copilot comment looks legit to you.

crates/spurctld/src/raft.rs
```diff
@@ -142,6 +144,16 @@ impl SpurStore {
...
 
+        let purged_path = raft_dir.join("purged.json");
+        if purged_path.exists() {
+            // Hard-fail: silently falling back to None reproduces the startup panic.
+            let data = std::fs::read_to_string(&purged_path)?;
+            inner.last_purged = Some(
+                serde_json::from_str::<LogId<NodeId>>(&data)
+                    .map_err(|e| anyhow::anyhow!("failed to parse purged.json: {e}"))?,
+            );
+        }
```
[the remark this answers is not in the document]
yansun1996: Good catch — fixed in 527ad0c. `SpurStore::new()` now derives `last_purged` from the snapshot meta's `last_log_id` when `purged.json` is absent (the pre-patch compacted-node case), while a present-but-corrupt `purged.json` still hard-fails (returns Err, never silent None). Added tests for both the snapshot-fallback and the corrupt-file hard-fail paths (a96e5ce). Note: the derived value is a safe upper bound for `last_purged` (openraft retains up to `max_in_snapshot_log_to_keep` entries below the snapshot, so the true value is slightly lower) — the extra retained logs are simply orphaned and cleaned on the next purge; no panic, no data loss. Live-verified on node 147: forced a real snapshot+compaction, deleted `purged.json` to simulate a pre-patch node, restarted → recovered via the snapshot fallback (`last_purged=4999` from snapshot), became leader, accepted new jobs — no `LogIndex(0)` panic.

yansun1996:
Thanks for the review — the pre-patch-compaction gap is fixed in 527ad0c, with follow-up test/comment cleanup in a96e5ce.

- `SpurStore::new()` now derives `last_purged` from the snapshot meta's `last_log_id` when `purged.json` is absent (pre-patch compacted node), while a present-but-corrupt `purged.json` hard-fails (returns Err, never silent None).
- Added tests for both paths: snapshot-fallback and corrupt-file hard-fail. `cargo test -p spurctld` (191 passed) + `clippy` clean.
- **Live-tested on node 147 (single-node), end-to-end:**
  1. Drove ~5k jobs → real openraft snapshot + log compaction (log 5144→1344, `purged.json` written at index 3999).
  2. Restart **with** `purged.json` → recovered, became leader, accepted new jobs — no `LogIndex(0)` panic (core #368 fix).
  3. Deleted `purged.json` to simulate a **pre-patch** node, restart → recovered via snapshot fallback (`last_purged` derived as 4999 from snapshot meta), became leader, accepted jobs — no panic (this review fix).

Note on semantics: the derived value is a safe upper bound for `last_purged` (openraft keeps up to `max_in_snapshot_log_to_keep` entries below the snapshot, so the true value is lower); the extra retained logs are orphaned and cleaned on the next purge — no data loss.

Review thread resolved. PTAL.

crates/spurctld/src/raft.rs
```diff
@@ -142,6 +146,31 @@ impl SpurStore {
...
+            // orphaned and cleaned on the next purge. This matches what openraft
+            // uses to reconcile a snapshot-vs-log hole. Without it, such a node
+            // recovers with last_purged=None and reproduces the startup panic the
+            // moment openraft reads a purged log range.
+            warn!(
+                last_purged = ?snap_last,
+                "purged.json missing; deriving last_purged from snapshot meta (pre-patch compat)"
+            );
+            inner.last_purged = Some(snap_last);
+        }
```
shiv-tyagi: Drop this backward-compat block.

- Pre-1.0, state resets are acceptable (AGENTS.md). Only nodes that compacted under the old code lack `purged.json`, and they recover cleanly by wiping state and re-syncing the snapshot from the leader.
- The derivation is wrong, not just approximate. openraft retains log entries below the snapshot, so `last_purged = snapshot.last_log_id` claims more was purged than actually was. On restart `inner.log` still holds entries at or below that index, violating the "nothing at or below `last_purged`" invariant. That swaps a loud panic for a silently wrong Raft state.

If in-place recovery is truly needed, compute the real `last_purged` (one below the lowest retained on-disk log index) and document it as a one-time migration, not a permanent startup path. Either way, the comment is too long.

shiv-tyagi:
Sorry for bugging you again @yansun1996. Just one last thing.

crates/spurctld/src/raft.rs
yansun1996: You're right on both counts — dropped in b22049f.

Agreed the derivation is wrong, not just approximate: openraft retains up to `max_in_snapshot_log_to_keep` entries below the snapshot, so `last_purged = snapshot.last_log_id` would claim more was purged than actually was, leaving `inner.log` with entries at/below `last_purged` — violating the invariant and swapping a loud panic for silently-wrong state. And per AGENTS.md pre-1.0 state resets are fine, so a node without `purged.json` can just wipe + re-sync the snapshot from the leader.

So I removed the whole backward-compat branch (and its `snapshot_last_log_id` tracking + the long comment). Missing `purged.json` → `last_purged = None` again. Kept the core fix (persist in `purge_logs_upto`, load in `new()`) and the corrupt-file hard-fail. Tests: 190 pass, clippy + fmt clean.

yansun1996:
@shiv-tyagi good catch — you're right, dropped the backward-compat fallback in b22049f.

The snapshot-meta derivation was genuinely incorrect (not just approximate): openraft retains entries below the snapshot, so `last_purged = snapshot.last_log_id` would violate the "nothing at or below last_purged" invariant and turn a loud panic into silently-wrong state. And pre-1.0 (AGENTS.md), a node lacking `purged.json` recovers fine by wiping + re-syncing the snapshot from the leader, so the fallback wasn't needed anyway.

**Now:** removed the fallback branch + its `snapshot_last_log_id` tracking + the long comment + the fallback test. Missing `purged.json` → `last_purged = None`. Kept the core fix (persist `last_purged` in `purge_logs_upto`, load in `SpurStore::new()`) and the corrupt-file hard-fail path + test. 190 tests pass, clippy + fmt clean.

Net for the PR: persist-and-reload `last_purged` so a compacted node restarts cleanly; pre-patch nodes without the file take the standard wipe+resync path. PTAL.

shiv-tyagi (approved):

Issue #278, opened 2026-06-15
Support SBATCH_* environment variable defaults for sbatch flags


shiv-tyagi:
## Problem

Slurm reads `SBATCH_*` environment variables as defaults for sbatch CLI flags (e.g., `SBATCH_PARTITION` sets the default partition). Spur's sbatch ignores these entirely, breaking workflows that rely on site-level defaults via environment modules or shell profiles.

Currently no `#[arg]` attribute in `SbatchArgs` has an `env = "SBATCH_*"` setting. The only env var read is `SPUR_CONTROLLER_ADDR` on the `--controller` flag.

## Expected Behavior

Each sbatch flag should optionally read from its corresponding `SBATCH_*` env var. CLI flags override env vars, which override `#SBATCH` directives. Variables include:

- `SBATCH_ACCOUNT`, `SBATCH_PARTITION`, `SBATCH_QOS`, `SBATCH_TIMELIMIT`
- `SBATCH_JOB_NAME`, `SBATCH_OUTPUT`, `SBATCH_ERROR`
- `SBATCH_GRES`, `SBATCH_CONSTRAINT`, `SBATCH_NODELIST`, `SBATCH_EXCLUDE`
- `SBATCH_RESERVATION`, `SBATCH_DISTRIBUTION`, `SBATCH_EXPORT`
- `SBATCH_MEM`, `SBATCH_MEM_PER_CPU`

Clap's built-in `env` support handles the precedence automatically.

## Acceptance Criteria

- [ ] Each relevant `#[arg]` in `SbatchArgs` has `env = "SBATCH_<FLAG>"` set
- [ ] Setting `SBATCH_PARTITION=gpu` causes sbatch to default to the `gpu` partition
- [ ] Explicit CLI flags override env var values
- [ ] Vec args like `--gres` work with env vars using comma-delimited values

biluriuday:
this gap is addressed via https://github.com/ROCm/spur/pull/374

PR #374, opened 2026-07-01
feat(spur-cli): read SBATCH_* env vars as sbatch flag defaults

biluriuday:
Add SBATCH_* environment-variable defaults for sbatch flags with Slurm precedence: CLI > env > #SBATCH directive. clap alone cannot rank a directive against an env var since directives are fed in as argv, so resolve_sbatch_args parses the CLI and directives separately and uses clap's ValueSource to pick the winning layer per field.

This PR addresses github issue https://github.com/ROCm/spur/issues/278

## Submission Checklist

### DIFF
--- a/crates/spur-cli/src/sbatch.rs
+++ b/crates/spur-cli/src/sbatch.rs
@@ -2,7 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use anyhow::{bail, Context, Result};
-use clap::Parser;
+use clap::parser::ValueSource;
+use clap::{CommandFactory, FromArgMatches, Parser};
 use spur_proto::proto::slurm_controller_client::SlurmControllerClient;
 use spur_proto::proto::{JobSpec, SubmitJobRequest};
 use std::collections::HashMap;
@@ -11,21 +12,31 @@ use std::collections::HashMap;
 #[derive(Parser, Debug)]
 #[command(name = "sbatch", about = "Submit a batch job script")]
 pub struct SbatchArgs {
-    // CLI args override #SBATCH directives. clap merges directives + CLI argv;
-    // `overrides_with = "self"` allows duplicates with last-wins, so a CLI flag
-    // appearing after a directive flag wins. Vec args (gres, licenses,
-    // container_mounts, container_env) accumulate by design and have no
-    // `overrides_with`. Closes #143.
+    // Precedence is CLI > env (SBATCH_*) > #SBATCH directive, matching Slurm;
+    // `resolve_sbatch_args` enforces it. `overrides_with = "self"` gives
+    // repeated scalar flags last-wins semantics. Vec args (gres, licenses,
+    // container_mounts, container_env) have no `overrides_with` because they
+    // accumulate by design.
     /// Job name
-    #[arg(short = 'J', long, overrides_with = "job_name")]
+    #[arg(
+        short = 'J',
+        long,
+        env = "SBATCH_JOB_NAME",
+        overrides_with = "job_name"
+    )]
     pub job_name: Option<String>,
 
     /// Partition
-    #[arg(short = 'p', long, overrides_with = "partition")]
+    #[arg(
+        short = 'p',
+        long,
+        env = "SBATCH_PARTITION",
+        overrides_with = "partition"
+    )]
     pub partition: Option<String>,
 
     /// Account
-    #[arg(short = 'A', long, overrides_with = "account")]
+    #[arg(short = 'A', long, env = "SBATCH_ACCOUNT", overrides_with = "account")]
     pub account: Option<String>,
 
     /// Number of nodes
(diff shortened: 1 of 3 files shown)


crates/spur-cli/src/sbatch.rs
```diff
@@ -264,16 +295,131 @@ pub fn parse_sbatch_directives(script: &str) -> Vec<String> {
...
+    directive: Vec<String>,
+    cli: Vec<String>,
+) -> Vec<String> {
+    match cli_matches.value_source("gres") {
+        Some(ValueSource::CommandLine) => concat_vec(directive, cli),
+        _ if !directive.is_empty() => directive,
+        Some(ValueSource::EnvVariable) => cli,
+        _ => Vec::new(),
+    }
 }
```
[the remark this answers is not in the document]
biluriuday: addressed the issue

crates/spur-cli/src/sbatch.rs
```diff
@@ -802,4 +946,102 @@ echo "hello world"
...
+        // When CLI/directive supply gres, SBATCH_GRES is not consulted.
+        clear_sbatch_env();
+        std::env::set_var("SBATCH_GRES", "env:1");
+        let args = parse_merged(
+            &["--gres=gpu:mi300x:8"],
+            &["sbatch", "--gres=license:fluent:1"],
+        );
+        clear_sbatch_env();
+        assert_eq!(args.gres, vec!["gpu:mi300x:8", "license:fluent:1"]);
+    }
```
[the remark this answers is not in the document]
biluriuday: added test

crates/spur-cli/src/sbatch.rs
```diff
@@ -264,16 +295,131 @@ pub fn parse_sbatch_directives(script: &str) -> Vec<String> {
...
-    let mut merged = vec!["sbatch".to_string()];
-    merged.extend(directives.iter().cloned());
-    merged.extend(cli_args.iter().skip(1).cloned()); // skip argv[0]
-    merged
+/// clap can resolve CLI > env > default within a single parse, but it cannot
+/// rank a directive against an env var, because directives are fed in as argv
+/// and would always outrank env. So we parse the CLI (with env enabled) and the
+/// directives independently, then per field keep the CLI value whenever clap
+/// reports its source as the command line or an env var, and otherwise fall
+/// back to the directive value.
```
shiv-tyagi: Can we tighten this comment?

```suggestion
/// Resolve `SbatchArgs` with Slurm precedence: CLI > env (SBATCH_*) > #SBATCH
/// directive > default.
///
/// clap resolves CLI > env in one parse, but directives fed as argv are tagged
/// `CommandLine` and would outrank env. So we parse CLI and directives
/// separately and merge per field: keep the CLI/env value when set, else the
/// directive.
```

crates/spur-cli/src/sbatch.rs
```diff
@@ -802,4 +946,129 @@ echo "hello world"
...
+        "SBATCH_GRES",
+        "SBATCH_CONSTRAINT",
+        "SBATCH_NODELIST",
+        "SBATCH_EXCLUDE",
+        "SBATCH_RESERVATION",
+        "SBATCH_DISTRIBUTION",
+        "SBATCH_EXPORT",
+        "SBATCH_MEM",
+        "SBATCH_MEM_PER_CPU",
+    ];
```
shiv-tyagi: `SbatchEnvGuard` handles the panic-leak cleanly. Two things still stand, both fixed by the same change: `SBATCH_ENV_VARS` is hand-maintained and drifts silently if a new `SBATCH_*` flag is added and not listed, and the tests still need `#[serial]`. Reading `SBATCH_*` into an explicit layer passed to `resolve_sbatch_args` removes the list and makes env injectable per call, so the tests can run in parallel.

crates/spur-cli/src/sbatch.rs
```diff
@@ -264,16 +295,131 @@ pub fn parse_sbatch_directives(script: &str) -> Vec<String> {
...
+/// `--gres` accumulates with directive values, since Vec args merge by design;
+/// but `SBATCH_GRES` replaces a directive so an env default can override
+/// `#SBATCH --gres`, consistent with the scalar flags.
+fn resolve_gres(
+    cli_matches: &clap::ArgMatches,
+    directive: Vec<String>,
+    cli: Vec<String>,
+) -> Vec<String> {
+    match cli_matches.value_source("gres") {
+        Some(ValueSource::CommandLine) => concat_vec(directive, cli),
```
shiv-tyagi: Q: an explicit CLI `--gres` accumulates with directive values here (documented as by design), but Slurm treats `--gres` as replace, not append. Intentional divergence?

shiv-tyagi (changes requested):
Nice work. Some comments. PTAL.

crates/spur-cli/src/sbatch.rs
biluriuday: done

crates/spur-cli/src/sbatch.rs
biluriuday: this was unintentional divergence. modified the code to match Slurm behaviour

crates/spur-cli/src/sbatch.rs
biluriuday: Removed the de-dup list. These env-var tests can't run in parallel: environment variables are process-global (not scoped per test), so running them concurrently would cause race conditions as one test's set_var/remove_var becomes visible to another mid-parse. That's why they're marked #[serial]

crates/spur-cli/src/sbatch.rs
```diff
@@ -802,4 +954,113 @@ echo "hello world"
...
+        let args = parse_merged(&["--gres=gpu:8"], &["sbatch"]);
+        assert_eq!(
+            args.gres,
+            vec!["gpu:2"],
+            "env gres must override #SBATCH directive"
+        );
+    }
+
+    #[test]
+    #[serial]
```
shiv-tyagi:
```suggestion
    #[serial(env_injection)]
```

Okay let us use explicit serial group then.

shiv-tyagi (changes requested):

crates/spur-cli/src/sbatch.rs
biluriuday: done

shiv-tyagi (approved):
LGTM. Thanks @biluriuday. Let's wait for E2Es before we can merge.

PR #376, opened 2026-07-01
refactor(spurctld): embed accounting service in-process

shiv-tyagi:
## Summary

- Removes the standalone `spurdbd` crate and folds its code into `spurctld` as a `crate::accounting` module (db, fairshare, grpc, notifier submodules).
- Accounting gRPC service is conditionally registered on port 6817 alongside the controller when `[accounting].database_url` is configured. No separate daemon, port, or binary.
- CLI tools (sacct, sacctmgr, sshare, sreport) unified on `SPUR_CONTROLLER_ADDR` / `--controller` instead of the removed `SPUR_ACCOUNTING_ADDR` / `--accounting`.

## Motivation

There was no reason to keep accounting as a separate daemon. Collapsing it into spurctld eliminates a network hop, simplifies deployment (one fewer binary, one fewer port), and removes the gRPC-client bridge that added latency to every job start/end recording.

## Key changes

- `crates/spurdbd/` deleted; code moved to `crates/spurctld/src/accounting/`
- `AccountingConfig` simplified to `database_url` + `fairshare_refresh_secs`
- `AccountingNotifier` replaces the old fire-and-forget gRPC calls with direct DB writes via shared `PgPool`
- CI, Dockerfile, install.sh, K8s manifests, and docs updated to reflect single-binary accounting
- E2E test infrastructure cleaned up (renamed helpers, removed dead constants)

## Test plan

- [x] `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` passes with no warnings
- [x] `cargo test --locked` passes
- [x] Bare-metal cluster smoke test (3 nodes): sinfo, sbatch, scancel, multi-node scheduling
- [x] Accounting E2E on bare-metal: sacctmgr add account/user/qos, sacct exit code tracking, sshare, sreport
- [x] Subagent audit confirmed zero stale references to spurdbd, port 6819, or removed env vars


shiv-tyagi:
>  Patch coverage is 7.86517% with 164 lines in your changes missing coverage. Please review.

Low patch coverage is expected and pre existing. As we are only moving the code and the existing one did not have coverage.

crates/spur-core/src/config.rs
```diff
@@ -323,39 +323,27 @@ impl Default for ControllerConfig {
...
     #[serde(default = "default_fairshare_refresh_secs")]
     pub fairshare_refresh_secs: u32,
 }
 
-fn default_accounting_host() -> String {
-    "localhost:6819".into()
-}
 fn default_database_url() -> String {
     "postgresql://spur:spur@localhost/spur".into()
 }
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed. database_url now defaults to empty; accounting is opt-in.

crates/spurctld/src/main.rs
```diff
@@ -145,36 +145,47 @@ async fn main() -> anyhow::Result<()> {
...
+                    config.accounting.fairshare_refresh_secs as u64,
+                );
+
+                cluster.qos_cache().spawn_refresh_loop(
+                    pool.clone(),
+                    config.accounting.fairshare_refresh_secs as u64,
+                );
+
+                Some(pool)
+            }
```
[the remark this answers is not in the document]
shiv-tyagi: Addressed. Migration failure now disables accounting entirely.

crates/spurctld/src/accounting/notifier.rs
```diff
@@ -0,0 +1,79 @@
...
+                cpus,
+                1,
+                memory_mb,
+                start_time,
+            )
+            .await
+            {
+                warn!(job_id, error = %e, "failed to record job start in accounting");
+            }
+        });
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed. Notifier now takes real num_nodes/num_tasks/cpus_per_task from JobSpec.

tests/native_host/e2e/cluster.py
```diff
@@ -127,13 +124,12 @@ def ensure_bins(nodes: list[SshNode], binaries_dir: str, bin_dir: str,
     Upload binaries to all nodes if not already present (or size differs).
     bin_dir is the remote directory where binaries are installed.
 
-    When *with_accounting* is True, also upload spurdbd and create the
-    sacct/sacctmgr symlinks.
+    When *with_accounting* is True, also create sacct/sacctmgr symlinks.
     """
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed.

crates/spurctld/src/limits_cache.rs
```diff
@@ -139,85 +117,35 @@ mod tests {
...
+            limits: QosLimits::default(),
             usage_factor: 1.0,
-            max_jobs_per_user: 0,
-            max_wall_minutes: 0,
-            max_tres_per_job: String::new(),
-            max_submit_jobs_per_user: 0,
-            max_tres_per_user: String::new(),
-            grp_tres: String::new(),
         }
     }
```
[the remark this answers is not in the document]
shiv-tyagi: Added qos_from_record unit tests covering limit parsing and zero/None edge cases.

crates/spurctld/src/accounting/notifier.rs
```diff
@@ -0,0 +1,84 @@
...
+        job_id: JobId,
+        state: JobState,
+        exit_code: i32,
+        end_time: DateTime<Utc>,
+        exit_signal: i32,
+        derived_exit_code: i32,
+    ) {
+        let pool = self.pool.clone();
+        let state_str = state.display().to_owned();
+        tokio::spawn(async move {
```
yansun1996, line 68: **Start/end write race → lost usage/fairshare and a row stuck in `RUNNING`.**

`notify_job_start` (line 37) and `notify_job_end` (here, line 68) each `tokio::spawn` an independent Postgres write with no ordering between them. The *call sites* are ordered (a job is recorded started before it can finalize through Raft), but the two spawned tasks are not — for a sub-second job the end write can land first.

When it does: `record_job_end` is UPDATE-only, matches 0 rows, and `db.rs:185-191` swallows it with `return Ok(())`, so `update_usage` (and therefore the fairshare/usage contribution) never runs. The start task then inserts the row afterward with `state = 'RUNNING'`, leaving the job permanently un-finalized in accounting.

Two notes to keep the framing accurate:
- This is **not a regression introduced by this PR** — the identical DB logic existed in `spurdbd` on `main`, but nothing ever called it (the controller used a stubbed default). This PR wires it up, making the latent flaw reachable for the first time.
- The start upsert only sets `start_time` + `state`; it does **not** clear `end_time`. So the row is stuck `RUNNING` with a stale/absent end_time (not a fully reset row).

Suggested fix: make `record_job_end` an upsert (mirroring the start path), or serialize start-then-end per job so the end write can't precede the insert.

crates/spurctld/src/accounting/notifier.rs
```diff
@@ -0,0 +1,84 @@
...
+    ) {
+        let pool = self.pool.clone();
+        let num_nodes = num_nodes as i32;
+        let num_tasks = num_tasks as i32;
+        let cpus_per_task = cpus_per_task as i32;
+        let memory_mb = memory_mb as i64;
+        tokio::spawn(async move {
+            if let Err(e) = super::db::record_job_start(
+                &pool,
+                job_id as i32,
```
yansun1996, line 40: **`job_id as i32` truncation (informational, pre-existing).**

`JobId` is `u32` but this casts to `i32` (also at line 71), and the schema column is `job_id INTEGER`. IDs > 2^31 would wrap negative. In practice this is not reachable — IDs come from an `AtomicU32` incremented by 1 per job, so ~2.1B jobs would be needed — and it's pre-existing (the old `spurdbd` schema and the `uint32` proto field are unchanged). Noting only; if the schema is ever revisited, `BIGINT` + `i64` would remove the sharp edge.

crates/spurctld/src/server.rs
```diff
@@ -1472,11 +1473,14 @@ pub async fn serve(
...
 
-    tonic::transport::Server::builder()
-        .layer(stats_layer)
-        .add_service(SlurmControllerServer::new(service))
-        .serve(addr)
-        .await?;
+    let mut builder = tonic::transport::Server::builder().layer(stats_layer);
+
+    let mut router = builder.add_service(SlurmControllerServer::new(service));
+    if let Some(pool) = accounting_pool {
```
yansun1996, line 1479: **Confusing CLI error when accounting is disabled (low UX).**

The `SlurmAccounting` service is only registered when `accounting_pool` is `Some` (i.e. `[accounting].database_url` set). That's the right gate, but when it's unset the accounting CLI tools (`sacct`/`sshare`/`sreport`/`sacctmgr`) hit a raw tonic `Unimplemented` status, which surfaces via `.context("failed to get job history")` as a cryptic error rather than something like "accounting is not enabled on this controller (set `accounting.database_url`)." Consider mapping `Code::Unimplemented` to a friendly message in the CLI tools.

crates/spur-core/src/config.rs
```diff
@@ -323,39 +323,24 @@ impl Default for ControllerConfig {
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct AccountingConfig {
```
yansun1996, line 325: **Silent config migration (low).**

`AccountingConfig` drops the old `host` and `purge_after_days` fields, and no config struct sets `deny_unknown_fields`, so pre-existing configs that still specify them are silently ignored with no warning. This is functionally harmless — `purge_after_days` had no runtime effect on `main` (only its own field/default; the `purge` references in `raft.rs` are unrelated Raft-log purging) and `host` had no server-side consumer — so it's genuinely just cleanup. Worth a one-line changelog/migration note so operators aren't surprised those keys became no-ops.

yansun1996 (changes requested):
## Review — accounting consolidation

Overall the consolidation is clean: the stale-reference audit passes (no live `spurdbd` / port `6819` / `SPUR_ACCOUNTING_ADDR` references remain), `package slurm;` is unchanged, and both gRPC services are correctly muxed on 6817. Findings below are mostly low-severity; the one substantive correctness issue is the start/end write race (inline on `notifier.rs`).

Each finding was cross-validated by a second pass, so severities are deliberately calibrated (several are pre-existing/low, called out as such).

### Items that GitHub won't let me anchor inline

(`crates/spurctld/src/accounting/db.rs` is a large new file whose diff the API won't accept line comments on, and `install.sh` is a pure mode change — so these two go here.)

**1. `unwrap()` in library code — `db.rs:225` (low, convention only).** `.and_hms_opt(start_time.hour(), 0, 0).unwrap()` violates the AGENTS.md "no `unwrap()` in library code" rule. It's infallible in practice (`hour` from `.hour()` is 0–23, so `and_hms_opt` can't return `None`) and pre-existing (identical on `main`), so this is a cleanup nit, not a panic risk. Prefer `ok_or_else(...)?` or `expect("hour in 0..=23")`.

> Note: a related first-pass concern — that `db.rs:214` reading `start_time` into a non-`Option<DateTime>` could panic on a NULL column — was checked and **refuted**: every write path binds `start_time` non-null and `record_job_end`/`update_usage` only run after `record_job_start`, so a NULL row is unreachable. No action needed there.

**2. `install.sh` lost its executable bit.** The diff shows `old mode 100755 → new mode 100644`, so `./install.sh` now fails with permission denied and users must invoke `bash install.sh`. Trivially restored with `git update-index --chmod=+x install.sh`. Minor but worth fixing before merge.

### Out-of-scope heads-up

Unrelated to accounting, but noticed while reviewing: `NodeInfo`'s `resources` / `alloc_resources` appear retyped on the same proto field numbers, which is wire-incompatible for those fields. Flagging for whoever owns FFI/REST compat — not something this PR needs to address.

crates/spurctld/src/server.rs
shiv-tyagi: This requires some larger refactoring. Maybe a proper accounting client. We would need to distinguish between actual errors and disabled accounting. So I will leave it as is for now.

crates/spurctld/src/accounting/notifier.rs
shiv-tyagi: Addressed

crates/spur-core/src/config.rs
shiv-tyagi: We are pre-1.0.0. It is okay.

crates/spurctld/src/accounting/notifier.rs
shiv-tyagi: This is pre-existing. wont fix in this PR.

sgopinath1 (approved):
LGTM

yansun1996 (approved):
LGTM

PR #381, opened 2026-07-01
feat(spur-cli): add sbatch --wrap for inline command submission

shiv-tyagi:
## Summary

- Adds the `--wrap` flag to `sbatch`, matching Slurm's behavior for inline command submission without a script file. The command is wrapped in a minimal `#!/bin/sh` script with the same byte layout as Slurm's `_script_wrap`.
- Extracts pure, testable helpers: `choose_body_source` (body resolution with wrap > file > stdin precedence), `candidate_script_path` (pre-parse guard that skips directive scanning in wrap mode), `default_job_name` (name fallback: `-J` > script name > `"sbatch"`).
- `--wrap` conflicts with a positional script argument at the clap level, matching Slurm's mutual exclusivity.

## Test plan

- [x] 17 unit tests covering `wrap_command_body`, `choose_body_source`, `candidate_script_path`, `default_job_name`, flag conflicts, and directive non-interference
- [x] `cargo test -p spur-cli --locked` — 76 tests pass
- [x] `cargo clippy -p spur-cli --all-targets --locked` — no warnings
- [x] E2E on 4-node bare-metal testbed:
  - `--wrap="echo hello"` completes with correct output and default name `sbatch`
  - `--wrap` with `-J` override sets custom job name
  - `--wrap` + script argument rejected with conflict error
  - `--wrap` value matching an existing file with `#SBATCH` directives does not inject those directives (pre-parse fix verified)
  - Normal script submission with `#SBATCH` directives still works (no regression)

### DIFF
--- a/crates/spur-cli/src/sbatch.rs
+++ b/crates/spur-cli/src/sbatch.rs
@@ -269,8 +269,39 @@ pub struct SbatchArgs {
     )]
     pub controller: String,
 
+    /// Wrap the given command in a minimal shell script (mutually exclusive with a script file)
+    #[arg(long, conflicts_with = "script")]
+    pub wrap: Option<String>,
+
     /// The batch script file
     pub script: Option<String>,
+}
+
+#[derive(Debug, PartialEq)]
+enum BodySource {
+    Wrap(String),
+    File(String),
+    Stdin,
+}
+
+/// Decide where the batch script body comes from.
+///
+/// `--wrap` wins; else a script file; else stdin (when piped); else error.
+fn choose_body_source(
+    wrap: Option<String>,
+    script: Option<String>,
+    stdin_is_terminal: bool,
+) -> Result<BodySource> {
+    match (wrap, script) {
+        (Some(cmd), _) => Ok(BodySource::Wrap(cmd)),
+        (None, Some(path)) => Ok(BodySource::File(path)),
+        (None, None) if stdin_is_terminal => bail!("sbatch: no script file specified"),
+        (None, None) => Ok(BodySource::Stdin),
+    }
+}
+
+fn wrap_command_body(cmd: &str) -> String {
+    format!("#!/bin/sh\n# This script was created by sbatch --wrap.\n\n{cmd}\n")
 }
 
 /// Parse #SBATCH directives from a script, returning them as argv-style strings.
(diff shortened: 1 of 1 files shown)


crates/spur-cli/src/sbatch.rs
```diff
@@ -465,35 +496,47 @@ pub async fn main() -> Result<()> {
...
+        .iter()
+        .any(|a| a == "--wrap" || a.starts_with("--wrap="))
+    {
+        return None;
+    }
+    match cli_args.last() {
+        Some(last) if !last.starts_with('-') && last != "sbatch" => Some(last.as_str()),
+        _ => None,
+    }
+}
```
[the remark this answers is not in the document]
shiv-tyagi: Pre-existing behavior, not introduced by this PR. The last-token heuristic existed before `--wrap`. A preliminary clap parse to find the positional arg creates a chicken-and-egg problem: the pre-parse exists to extract `#SBATCH` directives that must be merged into argv *before* clap runs. Worth improving separately, but out of scope here.

crates/spur-cli/src/sbatch.rs
```diff
@@ -465,35 +496,47 @@ pub async fn main() -> Result<()> {
...
+
+/// Default job name: explicit `-J`, else the script filename, else `"sbatch"`.
+fn default_job_name(job_name: Option<&str>, script: Option<&str>) -> String {
+    job_name.or(script).unwrap_or("sbatch").to_string()
```
yansun1996: **Minor Slurm-parity nit: a `--wrap` job with no `-J` is named `"sbatch"`; Slurm names it `"wrap"`.**

`default_job_name(job_name, script)` returns `job_name.or(script).unwrap_or("sbatch")`. In wrap mode both args are `None` (wrap has no script path), so the fallback is `"sbatch"`. Slurm's `sbatch` sets the job name to `"wrap"` in this case (`opt.c`: `if (!opt.job_name && sbopt.wrap) opt.job_name = xstrdup("wrap")`).

This is cosmetic — only affects the displayed job name in `squeue`, no functional impact. Note the current signature can't distinguish wrap-mode from stdin-mode (both pass `(None, None)`), so emitting `"wrap"` would require threading a wrap indicator into this helper. Only worth doing if strict Slurm parity on the default name is a goal; if so, `test_default_job_name_wrap_or_stdin` would need updating too.

yansun1996 (commented):
## Review — sbatch --wrap

Solid PR. Verified in detail:

- **`wrap_command_body` byte layout is byte-exact** vs Slurm's `_script_wrap` (`#!/bin/sh\n# This script was created by sbatch --wrap.\n\n<cmd>\n`), and the command is embedded into the script *body* (not a shell-eval string), so quotes/`$`/backticks/newlines are literal — no injection/escaping concern.
- **`choose_body_source` precedence** (wrap > file > stdin) is total across all combinations; stdin is correctly ignored when wrap is present; tty-with-no-input bails cleanly.
- **`candidate_script_path` pre-parse guard** correctly short-circuits so a `--wrap` value that happens to match a filename never gets its `#SBATCH` directives scanned.
- **clap conflict** (`--wrap` vs positional script) is declared correctly and errors cleanly.
- The helper extraction is genuinely pure/testable; 17 new tests are well-targeted.

One low-severity parity note inline, plus a minor test-gap suggestion. Nothing blocking.

### Suggestion (non-blocking)

- No test for an **empty** wrap string (`--wrap=""`), which produces a body of `#!/bin/sh\n# ...\n\n\n`. Worth a case to pin the behavior.

crates/spur-cli/src/sbatch.rs
shiv-tyagi: fixed

sgopinath1 (approved):
LGTM

PR #382, opened 2026-07-02
feat(e2e): add perf test

sgopinath1:
## Summary

Adds a Spur scheduling/ingestion perf test with held-job submit for metrics 1–2, release + drain for metric 3 (queue wait from controller timestamps, not client RTT).

- **Shell scripts** (`perf_tests/`): `run_perf.sh` for one tier, `run_all.sh` for multi-tier sweeps on an existing cluster. Metrics are printed as a stdout KEY=VALUE block; the script fails on partial submits and drain timeout.
- **Pytest harness** (`tests/native_host/e2e/perf_harness/`): deploys an ephemeral E2E cluster, runs `run_perf.sh` over SSH, parses results into JSON, and prints a markdown summary. Multi-tier runs redeploy the cluster between tiers so later tiers are not skewed by accumulated job state.
- **Compare tool** (`perf_harness.compare`): diff two suite JSON files (e.g. PR vs nightly) with configurable regression threshold and tier-mismatch warnings.
- **CI**: `@pytest.mark.perf` test excluded from default E2E (`-m "not perf"`). New manual `perf.yml` workflow builds PR binaries vs latest `nightly` release on the bare-metal cluster and prints a comparison report to the workflow log.

## Design notes

| Metric | What it measures |
|--------|------------------|
| 1 — submit throughput | Client wall clock during parallel `spur submit -H` |
| 2 — SubmitJob RPC avg | `sdiag` PRE/POST delta after `sdiag --reset` |
| 3 — queue wait p50/p95 | `StartTime − SubmitTime` on the same held-then-released jobs |

Shell-only sweeps (`run_all.sh`) intentionally reuse one controller (warm state after tier 1). The pytest harness redeploys between tiers for comparable numbers.


perf_tests/run_perf.sh
```diff
@@ -0,0 +1,298 @@
...
+echo "    release_wall=${RELEASE_WALL}s"
+
+echo "==> Waiting for queue to drain (timeout ${DRAIN_TIMEOUT}s)..."
+t_drain_deadline=$(( $(date +%s) + DRAIN_TIMEOUT ))
+peak_running=0
+drain_timed_out=0
+q=0
+while :; do
+  q="$("$SPUR_CLI" jobs 2>/dev/null | awk 'NR > 1 && $4 == "'"$USER"'"' | wc -l | tr -d ' ')"
+  [ "$q" -gt "$peak_running" ] && peak_running="$q"
```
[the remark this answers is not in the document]
sgopinath1: Fixed.

tests/native_host/e2e/perf_harness/harness.py
```diff
@@ -0,0 +1,435 @@
...
+
+
+def parse_tiers_from_env() -> list[int]:
+    raw = os.environ.get("SPUR_PERF_TIERS", _DEFAULT_TIERS).strip()
+    tiers: list[int] = []
+    for part in raw.replace(",", " ").split():
+        part = part.strip()
+        if not part:
+            continue
+        tiers.append(_parse_int_env("SPUR_PERF_TIERS", part))
```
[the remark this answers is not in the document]
sgopinath1: Fixed.

.github/workflows/perf.yml
```diff
@@ -0,0 +1,192 @@
...
+        run: |
+          set -euo pipefail
+          docker build -f Dockerfile --target dist -t spur:perf-pr .
+          CID=$(docker create spur:perf-pr)
+          mkdir -p /tmp/perf-pr-bins
+          for bin in spur spurctld spurd spurdbd; do
+            docker cp "$CID:/usr/local/bin/$bin" "/tmp/perf-pr-bins/$bin"
+          done
+          docker rm "$CID"
+          chmod +x /tmp/perf-pr-bins/*
```
shiv-tyagi: Don't check out and build the PR on the self-hosted bare-metal runner. Two problems, one fix.

Security: this checks out an arbitrary PR head and `docker build`s it (running the PR's `build.rs`/proc-macros), then runs the PR's own perf scripts, all on the self-hosted runner while `secrets.BM_SSH_KEY` is live in the agent. `e2e.yml` deliberately avoids exactly this ("No untrusted code is checked out or compiled here"). `workflow_dispatch` is not covered by `checkout@v7`'s fork-checkout guard, so nothing blocks it.

Broken build: even ignoring security, this step can't run. `docker create` fails on the `scratch`-based `dist` target (`Error response from daemon: no command specified`), and `docker cp` reads `/usr/local/bin/$bin` while the `dist` stage writes binaries to `/bin/`.

Fix: reuse CI's already-built artifacts (CI's Build Docker Image job uploads `release-binaries` from `/usr/local/bin` and `e2e-assets`), so perf does zero builds. Replace the checkout + build + stage-assets steps with:

```yaml
      - name: Resolve PR head SHA and CI run
        id: pr
        env:
          GH_TOKEN: ${{ github.token }}
        run: |
          set -euo pipefail
          PR="${{ inputs.pr_number }}"
          HEAD_SHA=$(gh pr view "$PR" --repo "${{ github.repository }}" --json headRefOid -q .headRefOid)
          echo "head_sha=${HEAD_SHA}" >> "$GITHUB_OUTPUT"
          RUN_ID=$(gh run list --repo "${{ github.repository }}" --workflow CI \
            --commit "$HEAD_SHA" --status success --json databaseId -q '.[0].databaseId')
          test -n "$RUN_ID" || { echo "no successful CI run for $HEAD_SHA"; exit 1; }
          echo "ci_run_id=${RUN_ID}" >> "$GITHUB_OUTPUT"

      - name: Download PR binaries from CI
        uses: actions/download-artifact@v8
        with:
          name: release-binaries
          path: /tmp/perf-pr-bins
          run-id: ${{ steps.pr.outputs.ci_run_id }}
          github-token: ${{ secrets.GITHUB_TOKEN }}

      - name: Download PR perf assets from CI
        uses: actions/download-artifact@v8
        with:
          name: e2e-assets
          path: ${{ env.PERF_ASSETS_DIR }}
          run-id: ${{ steps.pr.outputs.ci_run_id }}
          github-token: ${{ secrets.GITHUB_TOKEN }}
```

This removes the need for `docker`, the `pr-src` checkout, and the leftover `spur:perf-pr` image. Keep the nightly download as-is. Only caveat: perf must be dispatched after CI has succeeded for that head SHA (the resolve step fails fast otherwise), and artifact retention must cover the gap.

tests/pytest.ini
```diff
@@ -1,10 +1,11 @@
 [pytest]
-pythonpath = native_host/e2e k8s/e2e
-testpaths = native_host/e2e k8s/e2e
+pythonpath = native_host k8s/e2e
+testpaths = native_host/e2e native_host/perf k8s/e2e
```
shiv-tyagi, line 3: These new perf `testpaths` add `native_host/perf`, but the pure-Python tests there (`test_perf_report.py`, `test_perf_parse.py`) run in no CI job: `ci.yml` runs no pytest, `e2e.yml` runs only `native_host/e2e/`, and `perf.yml` runs only `test_scheduler_perf.py`. The comparison/verdict logic will bitrot. Wire these into a lightweight unit job (or the existing test job) with `-m "not perf"`. Note they also pull in the full e2e deps just to collect, since the parent `conftest.py` imports `cluster.py` at module load.

docs/developer/building.rst
```diff
@@ -125,14 +125,61 @@ Running the Tests
...
+     - Seconds each benchmark job sleeps.
+     - ``0``
+   * - ``SPUR_PERF_RUN_LABEL``
+     - Label stored in suite JSON.
+     - ``pr-382``
+   * - ``SPUR_PERF_COMPARE_THRESHOLD_PCT``
+     - Regression threshold for ``python -m perf.report``.
+     - ``10``
+
+See ``tests/native_host/perf/README.md`` and ``docs/developer/ingestion-benchmark.md`` for metric definitions.
```
shiv-tyagi: `docs/developer/ingestion-benchmark.md` doesn't exist in this PR. Either add the doc or drop the reference (also referenced in `tests/native_host/perf/README.md`).

tests/native_host/perf/README.md
```diff
@@ -0,0 +1,82 @@
...
+## CI
+
+Manual workflow [`.github/workflows/perf.yml`](../../../.github/workflows/perf.yml)
+builds PR and nightly binaries, then calls `run_perf_compare.sh`. First merge of
+this layout must land before the workflow paths resolve on the default branch.
+
+## Further reading
+
+- Metric definitions: header comment in [`scripts/run_perf.sh`](scripts/run_perf.sh)
+- Ingestion benchmark doc: [`docs/developer/ingestion-benchmark.md`](../../../docs/developer/ingestion-benchmark.md)
```
shiv-tyagi: `docs/developer/ingestion-benchmark.md` doesn't exist in this PR. Add it or drop the reference (same link is in `docs/developer/building.rst`).

.github/workflows/perf.yml
```diff
@@ -0,0 +1,192 @@
...
+      - name: Download latest nightly binaries
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          mkdir -p /tmp/nightly-dl
+          gh release download nightly --repo "${{ github.repository }}" \
+            --pattern 'spur-*.tar.gz' --dir /tmp/nightly-dl
+          tar xzf /tmp/nightly-dl/spur-*.tar.gz -C /tmp/nightly-dl
+          BINDIR=$(find /tmp/nightly-dl -path '*/bin' -type d | head -1)
```
shiv-tyagi: Nit: under `set -euo pipefail`, `BINDIR=$(find ... | head -1)` can abort this step. When `find` emits more than one line, `head` closes early, `find` gets SIGPIPE, and `pipefail` makes the pipeline exit 141. Use `-quit` or tolerate the pipe:

```suggestion
          BINDIR=$(find /tmp/nightly-dl -path '*/bin' -type d | head -1 || true)
```

tests/native_host/perf/run_perf_compare.sh
```diff
@@ -0,0 +1,142 @@
...
+  local remote_dir="${1:-}"
+  [ -n "${SPUR_TEST_NODES:-}" ] || return 0
+  [ -n "${SPUR_TEST_SSH_USER:-}" ] || return 0
+
+  IFS=',' read -ra NODES <<< "$SPUR_TEST_NODES"
+  for node in "${NODES[@]}"; do
+    node="${node#"${node%%[![:space:]]*}"}"
+    node="${node%"${node##*[![:space:]]}"}"
+    [ -n "$node" ] || continue
+    ssh "${SPUR_TEST_SSH_USER}@${node}" "
```
shiv-tyagi: Nit: `cleanup_cluster()` uses bare `ssh` here, so it ignores `SPUR_TEST_SSH_PASSWORD` (the conftest honors it via paramiko) and omits `-o BatchMode=yes`, so on a runner without an agent it hangs on a password prompt instead of failing fast. In a password-auth environment this spams `Permission denied` per node. The pytest teardown already tears down the ephemeral cluster, so this is belt-and-suspenders, but worth aligning: add `-o BatchMode=yes` and honor the password path (e.g. via `sshpass` when `SPUR_TEST_SSH_PASSWORD` is set).

.github/workflows/perf.yml
```diff
@@ -0,0 +1,192 @@
...
+              "${BM_SSH_USER}@${node}" "
+              sudo pkill -f '${BM_REMOTE_BIN_DIR_PR}/spurctld' 2>/dev/null || true
+              sudo pkill -f '${BM_REMOTE_BIN_DIR_PR}/spurd' 2>/dev/null || true
+              sudo pkill -f '${BM_REMOTE_BIN_DIR_NIGHTLY}/spurctld' 2>/dev/null || true
+              sudo pkill -f '${BM_REMOTE_BIN_DIR_NIGHTLY}/spurd' 2>/dev/null || true
+              sudo pkill -f '${BM_REMOTE_BIN_DIR}/spurctld' 2>/dev/null || true
+              sudo pkill -f '${BM_REMOTE_BIN_DIR}/spurd' 2>/dev/null || true
+              sudo rm -rf /tmp/spur-e2e-* 2>/dev/null || true
+            " || true
+          done
```
shiv-tyagi, line 186: This cleanup `pkill`s the binaries and removes `/tmp/spur-e2e-*`, but never removes the remote bin dirs `BM_REMOTE_BIN_DIR_PR` / `BM_REMOTE_BIN_DIR_NIGHTLY`, so they're left on every node after each run (I confirmed this leak running the harness). Add `rm -rf` for both to the per-node cleanup:

```bash
              sudo rm -rf '${BM_REMOTE_BIN_DIR_PR}' '${BM_REMOTE_BIN_DIR_NIGHTLY}' /tmp/spur-e2e-* 2>/dev/null || true
```

shiv-tyagi (changes requested):
posted comments. PTAL.

shiv-tyagi (dismissed):

PR #384, opened 2026-07-05
feat(spur-client): add client-side controller failover with endpoint rotation

### LINKED ISSUE #338: feat(spurd,spur-cli): support multiple controller endpoints with client-side failover
## Problem

`spurd --controller` accepts a single address (`String`), and every CLI subcommand does the same via `SPUR_CONTROLLER_ADDR`. In an HA deployment with a 3-node Raft quorum, if the specific controller node in that string goes down (crash, network partition), the agent and all CLI tools get a connection error with no fallback — even though other controllers are healthy.

**Note:** Spur already has server-side leader forwarding — any non-leader controller transparently proxies requests to the current leader via `get_leader_client()` (see `crates/spurctld/src/server.rs`). This means pointing at any _running_ controller works correctly regardless of which node is the leader. The gap is purely client-side: if the one configured controller is unreachable at the TCP level, forwarding never gets a chance to run, and the client has no other address to try.

Currently the only workaround is an external L4 VIP or DNS round-robin, which adds deployment complexity that shouldn't be required.

## Scope

- `crates/spurd/src/main.rs` — `controller: String` (line 47)
- `crates/spur-cli/src/` — every subcommand has `controller: String` with `env = "SPUR_CONTROLLER_ADDR"` (sbatch, squeue, sinfo, scancel, scontrol, srun, salloc, sdiag, smd, sprio, sstat, sattach, strigger, exec, node, token)
- `crates/spur-cli/src/main.rs` — `load_controller_addr_from_config()` reads a single host/port from config

## Expected Behavior

- `--controller` accepts a comma-separated list of endpoints (e.g. `http://ctrl1:6817,http://ctrl2:6817,http://ctrl3:6817`)
- `SPUR_CONTROLLER_ADDR` env var supports the same comma-separated format
- Config file supports a list of controller addresses
- Client-side retry logic: on connection failure or gRPC unavailable, rotate to the next endpoint
- Agents (spurd) should reconnect to a surviving controller automatically on leader failover

## Acceptance Criteria

- [ ] `--controller` and `SPUR_CONTROLLER_ADDR` accept multiple comma-separated endpoints
- [ ] Config file supports a list of controller addresses
- [ ] gRPC client retries with rotation on connection failure / unavailable
- [ ] spurd agent reconnects to a surviving controller without manual restart
- [ ] CLI commands fail over transparently to the user
- [ ] Backward compatible: a single address still works as before

shiv-tyagi:
## What

Closes #338.

`spurd` and the CLI could only be pointed at a single controller address, so an HA quorum offered no client-side resilience: if the one configured controller was down, connections failed even when other peers were healthy. This adds client-side failover across a list of controller endpoints.

## Approach

New `spur-client` crate provides a service-agnostic dialer:

- `connect_channel(endpoints)` takes a comma-separated list, tries each in order, and rotates to the next on connection failure, returning a `tonic::transport::Channel`. If every endpoint fails, the last error is returned.
- Callers wrap the channel in whatever generated client they need (`SlurmControllerClient::new(channel)`), so the crate depends on neither `spur-proto` nor `spur-core`.

A single reachable controller is sufficient: server-side Raft leader forwarding already routes writes to the current leader, so the client only needs to reach any live peer.

Config plumbing:

- `ControllerConfig::endpoints()` expands `hosts` + the `listen_addr` port into an endpoint list.
- The CLI exports it as a comma-joined `SPUR_CONTROLLER_ADDR`; the same env var and `--controller` flag accept a comma-separated list.

Call sites in `spur-cli`, `spurd`, and `spur-ffi` switch from `SlurmControllerClient::connect(addr)` to building a client from `connect_channel(&addr)`.

## Design notes

- Endpoints keep the `http://` scheme because tonic requires a full URI for the h2c transport; entries without a scheme are normalized to `http://`.
- The crate is intentionally minimal (only `tonic` + `tracing`) to keep layering clean and avoid pulling proto/config into the transport helper.

## Testing

- `spur-client` unit tests: endpoint parsing (scheme normalization, whitespace, empty entries) and rotation (dead-first reaches a live server, all-down returns an error).
- `spur-core` unit tests for `ControllerConfig::endpoints()`.
- Native-host e2e (`test_controller_failover.py`) run against the 4-node bare-metal cluster, all passing: rotation past a dead first endpoint, live-first list unaffected by a trailing dead endpoint, all-down list fails cleanly with a connection error (no hang), and a job submitted through a failover list completes.
- `cargo clippy --workspace --exclude spur-ffi --all-targets` clean; `cargo fmt` applied.
- Manual 3-peer Raft failover on the bare-metal testbed: killing the first controller in the list, the CLI rotates to a survivor and commands succeed after re-election.

Full multi-controller Raft failover in the automated native-host harness is deferred; the harness needs Raft support first.

### DIFF
--- a/crates/spur-client/src/lib.rs
+++ b/crates/spur-client/src/lib.rs
@@ -8,8 +8,10 @@
 //! is service-agnostic: callers wrap the returned channel in whatever generated
 //! client they need (e.g. `SlurmControllerClient::new(channel)`).
 
+use std::time::Duration;
+
 use tonic::transport::{Channel, Endpoint};
-use tracing::warn;
+use tracing::debug;
 
 /// Parse a comma-separated endpoint string into normalized `http(s)://host:port`
 /// entries. Whitespace is trimmed and empty entries dropped. Entries without a
@@ -55,7 +57,7 @@ pub async fn connect_channel(endpoints: &str) -> Result<Channel, tonic::transpor
     for endpoint in &list[..last] {
         match try_connect(endpoint).await {
             Ok(channel) => return Ok(channel),
-            Err(e) => warn!(
+            Err(e) => debug!(
                 %endpoint,
                 error = %e,
                 "controller endpoint unreachable, trying next"
@@ -67,7 +69,10 @@ pub async fn connect_channel(endpoints: &str) -> Result<Channel, tonic::transpor
 }
 
 async fn try_connect(endpoint: &str) -> Result<Channel, tonic::transport::Error> {
-    Endpoint::from_shared(endpoint.to_string())?.connect().await
+    Endpoint::from_shared(endpoint.to_string())?
+        .connect_timeout(Duration::from_secs(2))
+        .connect()
+        .await
 }
 
 #[cfg(test)]
@@ -135,6 +140,15 @@ mod tests {
         let down = free_addr().await;
         let up = spawn_server().await;
         let endpoints = format!("http://{down},http://{up}");
+        assert!(connect_channel(&endpoints).await.is_ok());
+    }
+
+    #[tokio::test]
+    async fn connect_rotates_past_multiple_dead() {
+        let down1 = free_addr().await;
+        let down2 = free_addr().await;
+        let up = spawn_server().await;
+        let endpoints = format!("http://{down1},http://{down2},http://{up}");
         assert!(connect_channel(&endpoints).await.is_ok());
     }
 
--- a/crates/spur-client/src/lib.rs
+++ b/crates/spur-client/src/lib.rs
@@ -1,0 +1,168 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Transport dialing with client-side failover.
+//!
+//! Establishes a [`tonic::transport::Channel`] from a comma-separated list of
+//! endpoints, rotating to the next endpoint when one is unreachable. The helper
+//! is service-agnostic: callers wrap the returned channel in whatever generated
+//! client they need (e.g. `SlurmControllerClient::new(channel)`).
+
+use tonic::transport::{Channel, Endpoint};
+use tracing::warn;
+
+/// Parse a comma-separated endpoint string into normalized `http(s)://host:port`
+/// entries. Whitespace is trimmed and empty entries dropped. Entries without a
+/// scheme are prefixed with `http://`.
+pub fn parse_endpoints(raw: &str) -> Vec<String> {
+    raw.split(',')
+        .map(str::trim)
+        .filter(|s| !s.is_empty())
+        .map(normalize_endpoint)
+        .collect()
+}
+
+fn normalize_endpoint(addr: &str) -> String {
+    if addr.starts_with("http://") || addr.starts_with("https://") {
+        addr.to_string()
+    } else {
+        format!("http://{addr}")
+    }
+}
+
+/// Always yields at least one entry so callers get a real transport error
+/// instead of a silent no-op when handed an empty string.
+fn endpoint_list(raw: &str) -> Vec<String> {
+    let parsed = parse_endpoints(raw);
+    if parsed.is_empty() {
+        vec![normalize_endpoint(raw.trim())]
+    } else {
+        parsed
+    }
+}
+
+/// Connect to the first reachable controller endpoint.
+///
+/// Endpoints are tried in the order given. On connection failure the next
+/// endpoint is attempted; if every endpoint fails, the last error is returned.
+/// Once connected to any running controller, server-side leader forwarding
+/// routes writes to the current Raft leader, so a single reachable node is
+/// sufficient regardless of which one is the leader.
+pub async fn connect_channel(endpoints: &str) -> Result<Channel, tonic::transport::Error> {
+    let list = endpoint_list(endpoints);
+    let last = list.len() - 1;
+
+    for endpoint in &list[..last] {
+        match try_connect(endpoint).await {
+            Ok(channel) => return Ok(channel),
+            Err(e) => warn!(
+                %endpoint,
+                error = %e,
+                "controller endpoint unreachable, trying next"
+            ),
+        }
+    }
+
+    try_connect(&list[last]).await
+}
+
+async fn try_connect(endpoint: &str) -> Result<Channel, tonic::transport::Error> {
+    Endpoint::from_shared(endpoint.to_string())?.connect().await
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn parse_single_endpoint() {
+        assert_eq!(
+            parse_endpoints("http://ctrl:6817"),
+            vec!["http://ctrl:6817"]
+        );
+    }
+
+    #[test]
+    fn parse_adds_scheme() {
+        assert_eq!(parse_endpoints("ctrl:6817"), vec!["http://ctrl:6817"]);
+    }
+
+    #[test]
+    fn parse_preserves_https() {
+        assert_eq!(
+            parse_endpoints("https://ctrl:6817"),
+            vec!["https://ctrl:6817"]
+        );
+    }
+
+    #[test]
+    fn parse_comma_list_with_whitespace() {
+        assert_eq!(
+            parse_endpoints(" ctrl1:6817 , http://ctrl2:6817 ,ctrl3:6817"),
+            vec![
+                "http://ctrl1:6817",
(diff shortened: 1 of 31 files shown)


crates/spur-client/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+                "controller endpoint unreachable, trying next"
+            ),
+        }
+    }
+
+    try_connect(&list[last]).await
+}
+
+async fn try_connect(endpoint: &str) -> Result<Channel, tonic::transport::Error> {
+    Endpoint::from_shared(endpoint.to_string())?.connect().await
```
yansun1996: `Endpoint::from_shared(...).connect()` has no connect timeout. tonic/hyper set none by default, so an endpoint unreachable via a network partition (dropped packets, not an active refusal) can hang for the OS TCP timeout (~2 min on Linux) before rotating to the next endpoint — defeating the point of fast failover. `crates/spurctld/src/raft.rs` already calls `.connect_timeout(Duration::from_secs(2))` for exactly this reason; worth mirroring that here.

crates/spur-client/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+/// Connect to the first reachable controller endpoint.
+///
+/// Endpoints are tried in the order given. On connection failure the next
+/// endpoint is attempted; if every endpoint fails, the last error is returned.
+/// Once connected to any running controller, server-side leader forwarding
+/// routes writes to the current Raft leader, so a single reachable node is
+/// sufficient regardless of which one is the leader.
+pub async fn connect_channel(endpoints: &str) -> Result<Channel, tonic::transport::Error> {
```
yansun1996, line 53: Nit/follow-up: no memoization of the last-known-good endpoint — every call starts at index 0, so long-lived periodic callers (e.g. the heartbeat loop) will keep re-dialing a dead first endpoint every cycle instead of remembering the one that worked last time. Not a correctness issue given the intentionally stateless design, but worth tracking as a follow-up for steady-state HA efficiency.

crates/spur-core/src/config.rs
```diff
@@ -321,6 +321,19 @@ impl Default for ControllerConfig {
...
 
+impl ControllerConfig {
+    /// Client-facing controller endpoints, one per `hosts` entry, each combined
+    /// with the port from `listen_addr`. Used for client-side failover across an
+    /// HA quorum.
+    pub fn endpoints(&self) -> Vec<String> {
+        let port = self.listen_addr.rsplit(':').next().unwrap_or("6817");
+        self.hosts
+            .iter()
+            .map(|host| format!("http://{host}:{port}"))
```
yansun1996: Nit: `format!("http://{host}:{port}")` produces a malformed URI if `host` is a bare IPv6 literal (e.g. `::1`), since it isn't bracketed (needs `[::1]:6817`). Low priority if IPv6 controller hosts aren't a supported deployment target, but worth a doc note or bracket-check otherwise.

crates/spur-client/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+    async fn connect_all_down_returns_error() {
+        // Two ports that are bound then released, so connections are refused.
+        let down1 = free_addr().await;
+        let down2 = free_addr().await;
+        let endpoints = format!("http://{down1},http://{down2}");
+        assert!(connect_channel(&endpoints).await.is_err());
+    }
+
+    #[tokio::test]
+    async fn connect_rotates_to_first_reachable() {
```
yansun1996, line 139: Test coverage: this only exercises rotation past a single dead endpoint. Worth adding a case with two consecutive dead endpoints before the live one (`dead1, dead2, up`), closer to the 3-node HA topology this feature targets. Can reuse `free_addr()`/`spawn_server()` as-is — no new scaffolding needed.

crates/spur-core/src/config.rs
```diff
@@ -947,6 +960,39 @@ pub fn format_time(total_minutes: Option<u32>) -> String {
 mod tests {
     use super::*;
 
+    #[test]
+    fn test_controller_endpoints_single_host() {
```
yansun1996, line 970: Test coverage: no case for `hosts: vec![]`, which is the runtime state `main.rs`'s `if !endpoints.is_empty()` branch guards against. A `test_controller_endpoints_empty_hosts` asserting `endpoints()` returns `[]` for empty `hosts` would close that gap.

yansun1996 (commented):
Please see the inline comments below.

crates/spur-client/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+            ),
+        }
+    }
+
+    try_connect(&list[last]).await
+}
+
+async fn try_connect(endpoint: &str) -> Result<Channel, tonic::transport::Error> {
+    Endpoint::from_shared(endpoint.to_string())?.connect().await
+}
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed in 617fe35. `try_connect` now sets `.connect_timeout(Duration::from_secs(2))`, matching the precedent in `spurctld/src/raft.rs`. No request-level `.timeout()` added: this channel carries long-lived streams (`srun`/`sattach` attach), so a blanket request timeout would break those.

crates/spur-core/src/config.rs
```diff
@@ -321,6 +321,19 @@ impl Default for ControllerConfig {
...
+    /// Client-facing controller endpoints, one per `hosts` entry, each combined
+    /// with the port from `listen_addr`. Used for client-side failover across an
+    /// HA quorum.
+    pub fn endpoints(&self) -> Vec<String> {
+        let port = self.listen_addr.rsplit(':').next().unwrap_or("6817");
+        self.hosts
+            .iter()
+            .map(|host| format!("http://{host}:{port}"))
+            .collect()
+    }
```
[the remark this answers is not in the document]
shiv-tyagi: The `rsplit(':')` example is slightly off: it already handles `[::]:6817` correctly (yields `6817`). The real gap was unbracketed IPv6 in `hosts`, not in `listen_addr`. That is addressed in 617fe35: bare IPv6 hosts (containing `:`, not already bracketed) get wrapped in `[]`. Parsing as `SocketAddr` would not fit here since `hosts` entries are hostnames, not socket addresses.

crates/spur-client/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+    let last = list.len() - 1;
+
+    for endpoint in &list[..last] {
+        match try_connect(endpoint).await {
+            Ok(channel) => return Ok(channel),
+            Err(e) => warn!(
+                %endpoint,
+                error = %e,
+                "controller endpoint unreachable, trying next"
+            ),
```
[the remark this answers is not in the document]
shiv-tyagi: Agreed. Downgraded to `debug!` in 617fe35. The all-endpoints-failed case already surfaces as the returned error, so no information is lost at WARN level.

crates/spur-client/src/lib.rs
shiv-tyagi: Fixed in 617fe35. Added `.connect_timeout(Duration::from_secs(2))` matching `spurctld/src/raft.rs`. Omitted request-level `.timeout()` because this channel also carries long-lived streams (`srun`/`sattach`).

crates/spur-client/src/lib.rs
shiv-tyagi: Agree it would help steady-state HA, but it changes the stateless contract and needs shared state across calls. With `connect_timeout(2s)` added, re-dialing a dead first endpoint costs at most 2s per cycle. Tracking as a follow-up.

crates/spur-core/src/config.rs
shiv-tyagi: Fixed in 617fe35. `endpoints()` now brackets bare IPv6 hosts (contains `:` and not already `[`-prefixed), plus tests for `::1` and pre-bracketed `[::1]` passthrough.

crates/spur-client/src/lib.rs
shiv-tyagi: Added `connect_rotates_past_multiple_dead` (dead1, dead2, up) in 617fe35.

crates/spur-core/src/config.rs
shiv-tyagi: Added `test_controller_endpoints_empty_hosts` in 617fe35.

sgopinath1 (approved):
LGTM

yansun1996 (approved):
All three findings addressed and verified: connect_timeout added, IPv6 hosts correctly bracketed, both proposed tests present. CI green across the board. LGTM.

PR #385, opened 2026-07-05
feat(reservations): persist CRUD via raft WAL and enforce scheduling parity

sgopinath1:
## Summary

Adds durable node reservations with scheduling enforcement, and preemption interaction.

- **Durability** — Reservation create/update/delete go through `WalOperation` (Raft log replay), not snapshot-only state.
- **Flags** — `MAINT`, `IGNORE_JOBS`, `NO_HOLD_JOBS`, and `OVERLAP` with validation (unknown flags rejected at API boundary).
- **Scheduling** — Prospective backfill blocking for reserved nodes, reservation-targeting jobs prioritized in `pending_jobs()`, future-start window fit via `now.max(res.start_time)`, and lifecycle hooks (`purge_expired`, end-time cancel with `resv_overrun_minutes` grace, hold pending jobs on delete/purge unless `NO_HOLD_JOBS`).
- **Preemption** — Honors partition `PreemptMode`; skips jobs in active reservations unless the preemptor has a higher `priority_tier`; only considers running jobs on nodes the pending job could use.
- **Display / API** — `scontrol`/`sinfo`/`smd` reservation state; `PendingReason::ReservedMaintenance` / `ReservationDeleted`; proto `JobInfo.reservation`, `reservation_maint`, `ReservationInfo.state`; `squeue %v` column.

## Design notes

- Spur always runs Raft; reservations now follow the same propose → apply path as jobs and nodes.
- `hold_jobs_for_deleted_reservation` sets `PendingReason::ReservationDeleted` after `hold_job` (not WAL-persisted; same pattern as other hold reasons).
- `IGNORE_JOBS` bypasses running-job overlap checks at create time; `OVERLAP` / `MAINT` allow storage overlap between reservations.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` (`RUSTFLAGS=-D warnings`)
- [x] `cargo test --locked` (unit tests across `spur-core`, `spur-sched`, `spurctld`, `spur-cli`, `spurdbd`)
- [x] Unit: overlap validation, priority ordering, `resv_overrun` grace, preemption immunity, node-overlap gating, `purge_expired` hold behavior, WAL apply for reservation CRUD
- [ ] E2E: `tests/native_host/e2e/test_reservations.py` on bare-metal lab (create/list/delete, unauthorized block, authorized schedule, busy-node rejection)


crates/spur-core/src/reservation.rs
```diff
@@ -44,10 +125,167 @@ impl Reservation {
...
+        return false;
+    }
+    let job_end = now + job_duration;
+    if reservation.is_active(now) {
+        return job_end > reservation.end_time || now < reservation.start_time;
+    }
+    if reservation.is_future(now) {
+        return job_end > reservation.start_time;
+    }
+    false
```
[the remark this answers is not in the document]
sgopinath1: Fixed. Unauthorized jobs on an active reservation are always blocked (not only when duration exceeds the window). Added unit test.

crates/spur-core/src/reservation.rs
```diff
@@ -44,10 +125,167 @@ impl Reservation {
...
+        let end = running_job_end(job, Utc::now()).unwrap_or(start_time);
+        if end > start_time {
+            let node = job
+                .allocated_nodes
+                .iter()
+                .find(|n| node_set.contains(n.as_str()))
+                .cloned()
+                .unwrap_or_default();
+            return Some((job.job_id, node));
+        }
```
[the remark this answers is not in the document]
sgopinath1: Fixed. Running jobs with unknown end time are treated as still occupying the node. Added unit test.

crates/spurctld/src/server.rs
```diff
@@ -1182,18 +1182,22 @@ impl SlurmController for ControllerService {
...
             nodes: req.nodes,
             accounts: req.accounts,
             users: req.users,
+            flags,
         };
 
         self.cluster
             .create_reservation(reservation)
-            .map_err(|e| Status::internal(e.to_string()))?;
+            .map_err(|e| Status::invalid_argument(e.to_string()))?;
```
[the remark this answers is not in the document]
sgopinath1: Added reservation_rpc_status():

raft propose failed → INTERNAL
not found → NOT_FOUND
validation errors → INVALID_ARGUMENT
Applied to create/update/delete reservation handlers.

crates/spurctld/src/server.rs
```diff
@@ -1913,9 +1922,11 @@ fn annotate_nodes_with_reservations(
...
 ) {
     for node_info in nodes.iter_mut() {
+        node_info.reservation_maint = false;
         for res in reservations {
             if res.is_active(now) && res.covers_node(&node_info.name) {
                 node_info.active_reservation = res.name.clone();
+                node_info.reservation_maint = res.flags.maint;
                 break;
             }
         }
```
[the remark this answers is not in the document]
sgopinath1: Fixed. Scans all active reservations; ORs reservation_maint when any overlapping res has MAINT. Added unit test for OVERLAP + MAINT case.

tests/native_host/e2e/test_reservations.py
```diff
@@ -0,0 +1,127 @@
...
+        sb = cluster.sbatch(["-N", "1", "-w", node, "-t", "10", long_script])
+        job_id = parse_job_id(sb)
+        assert job_id is not None
+
+        deadline = time.time() + 30
+        while time.time() < deadline:
+            sq = cluster.squeue_all()
+            if str(job_id) in sq and " R " in sq:
+                break
+            time.sleep(2)
```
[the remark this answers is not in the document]
sgopinath1: Added wait_job_state() using job_state(); unauthorized-block and busy-node tests now wait for the specific job’s PD/R state.

crates/spurctld/src/server.rs
```diff
@@ -1229,7 +1233,7 @@ impl SlurmController for ControllerService {
                 &req.add_accounts,
                 &req.remove_accounts,
             )
-            .map_err(|e| Status::internal(e.to_string()))?;
+            .map_err(|e| Status::invalid_argument(e.to_string()))?;
         Ok(Response::new(()))
```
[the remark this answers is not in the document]
sgopinath1: Fixed

crates/spurctld/src/cluster.rs
```diff
@@ -1786,18 +1822,20 @@ impl ClusterManager {
         }
     }
 
-    /// Create a new reservation.
-    pub fn create_reservation(&self, res: Reservation) -> anyhow::Result<()> {
-        let mut reservations = self.reservations.write();
-        if reservations.iter().any(|r| r.name == res.name) {
+    /// Create a new reservation (validated, persisted via Raft).
+    pub fn create_reservation(&self, mut res: Reservation) -> anyhow::Result<()> {
+        if self.reservations.read().iter().any(|r| r.name == res.name) {
```
yansun1996, line 2022: **Blocking:** TOCTOU race on reservation name uniqueness. This check reads `self.reservations` before `propose()`; two concurrent `create_reservation` calls for the same name can both pass it before either commits. `apply_operation`'s `ReservationCreate` arm never re-checks for an existing name and unconditionally pushes, so duplicates can land. Worse, `ReservationDelete`'s apply arm uses `retain(|r| r.name != *name)`, which would then remove *both* duplicates on a single delete. The uniqueness invariant needs to be enforced inside `apply_operation` — the single serialized point all replicas agree on — not just at propose time. Once fixed, a regression test spawning two same-name creates and asserting only one survives would pin down the invariant.

crates/spurctld/src/cluster.rs
```diff
@@ -1810,12 +1848,251 @@ impl ClusterManager {
...
+                    job.job_id
+                );
+            }
+        }
+
+        self.propose(WalOperation::ReservationDelete {
+            name: name.to_string(),
+        })?;
+
+        if !res.flags.no_hold_jobs {
```
yansun1996: **Blocking:** `no_hold_jobs` doesn't actually let jobs be rescheduled. Skipping `hold_jobs_for_deleted_reservation` here (and in `purge_expired_reservations`) never clears `job.spec.reservation` — no code path does. `reservation_block()` unconditionally returns `Some(PendingReason::Reservation)` when the named reservation no longer exists, and `pending_jobs()` filters those out entirely. So jobs from a `no_hold_jobs`-deleted reservation are still permanently excluded from scheduling — functionally identical to being held, just without the `Held` label. A minimal regression test (extend `purge_expired_holds_pending_reservation_jobs` below at line ~5363): create a reservation with `no_hold_jobs`, submit a job against it, delete/expire it, assert the job is still eligible via `pending_jobs()`.

crates/spurctld/src/cluster.rs
```diff
@@ -2852,6 +3155,47 @@ impl StateMachineApply for ClusterManager {
     }
 }
 
+fn reservation_fence_reason(
```
yansun1996, line 3567: Suggestion: this fences on *any* node being reservation-blocked, even if 99 other free nodes would satisfy the job, because it `continue`s before the more accurate `Resources`/`NodeDown`/`Priority` classification runs. It's diagnostics-only (actual placement uses `prospective_overlap` independently in `spur-sched`), but in any cluster with an active/upcoming reservation covering even one node, unrelated pending jobs can be mislabeled `ReqNodeNotAvail`/`ReservedMaintenance` in `squeue`/`scontrol` when the real reason is unrelated.

crates/spurctld/src/cluster.rs
```diff
@@ -714,6 +714,36 @@ impl ClusterManager {
...
 
+    /// Return a preempted job to the pending queue.
+    pub fn requeue_preempted_job(&self, job_id: JobId, force: bool) -> anyhow::Result<()> {
+        const MAX_REQUEUE: u32 = 3;
+        let old_state = {
+            let jobs = self.jobs.read();
+            let Some(job) = jobs.get(&job_id) else {
+                return Ok(());
+            };
+            if job.requeue_count >= MAX_REQUEUE {
```
yansun1996: Suggestion (pre-existing gap, not introduced by this PR): once `requeue_count >= MAX_REQUEUE`, the job returns `Ok(())` and stays in `JobState::Preempted` forever — neither `is_terminal()` nor `is_active()` — so it never completes, gets cancelled, or gets picked up by any further scheduler hook. Consider transitioning to `Cancelled` with a distinguishing exit reason once retries are exhausted.

crates/spurctld/src/server.rs
```diff
@@ -1913,15 +1922,33 @@ fn annotate_nodes_with_reservations(
...
+                if res.flags.maint {
+                    node_info.reservation_maint = true;
+                }
             }
         }
     }
 }
 
+#[allow(clippy::result_large_err)]
+fn reservation_rpc_status(err: impl std::fmt::Display) -> Status {
```
yansun1996: Nit: classifying gRPC status by substring-matching `anyhow::Error::to_string()` is fragile — any future rewording of an error message (including ones from `hostlist::expand()`/`normalize_node_list()`, which this function doesn't control) silently changes the client-facing status code. Consider a typed error enum or `downcast_ref` at the `ClusterManager` boundary instead.

crates/spurctld/src/server.rs
```diff
@@ -1913,15 +1922,33 @@ fn annotate_nodes_with_reservations(
...
+                }
+                if res.flags.maint {
+                    node_info.reservation_maint = true;
+                }
             }
         }
     }
 }
 
+#[allow(clippy::result_large_err)]
```
yansun1996: Nit: `reservation_rpc_status` returns a bare `Status`, not a `Result<_, E>`, so `#[allow(clippy::result_large_err)]` can never fire here — looks like a copy-paste leftover, safe to remove.

crates/spur-sched/src/backfill.rs
```diff
@@ -45,6 +45,33 @@ impl BackfillScheduler {
             .collect();
     }
 
+    /// Returns true when `[start, start+duration)` intersects a reservation on `node`
+    /// and the job is not authorized for that reservation.
+    fn start_overlaps_reservation(
```
yansun1996, line 51: Suggestion: this duplicates `prospective_overlap` in `spur-core/src/reservation.rs` — both implement "does `[start, start+duration)` intersect a reservation window this job isn't authorized for", just parameterized differently (`now` vs. an arbitrary future `start`). Consider generalizing `prospective_overlap` to take an explicit `at: DateTime<Utc>` and having this call it, so future changes to the authorization rule can't drift between the two call sites.

proto/slurm.proto
```diff
@@ -196,6 +196,7 @@ message JobInfo {
   // Array jobs
   uint32 array_job_id = 50;
   uint32 array_task_id = 51;
+  string reservation = 52;
```
yansun1996, line 201: Suggestion: `reservation = 52` is grouped under the `// Array jobs` comment above `array_job_id`/`array_task_id`, but it isn't array-related. Consider splitting into its own `// Reservation` comment line for readability — the field number itself (52) is fine as-is.

crates/spurctld/src/cluster.rs
```diff
@@ -4824,6 +5168,273 @@ mod tests {
...
+        cm.purge_expired_reservations();
+        let job = cm.get_job(job_id).unwrap();
+        assert_eq!(job.state, JobState::Pending);
+        assert_eq!(job.pending_reason, PendingReason::ReservationDeleted);
+        assert_eq!(job.priority, 0);
+        assert!(cm.get_reservations().is_empty());
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn apply_reservation_create_update_delete() {
```
yansun1996, line 6376: Test coverage: nearby tests (`apply_reservation_create_update_delete`, `snapshot_and_restore`) prove CRUD round-trips and snapshot/restore work, but none proves durability across an actual restart via WAL replay, which is this PR's headline claim. No reservation test drops a `ClusterManager` and starts a new one against the *same* on-disk raft log directory to assert the reservation reappears from replay (as opposed to a hand-invoked snapshot restore into a fresh `dir2`). Worth adding a `reservation_survives_wal_replay` test reusing this same `TempDir`/`test_cluster` scaffolding but restarting against `dir.path()`.

crates/spurctld/src/cluster.rs
```diff
@@ -4824,6 +5168,273 @@ mod tests {
...
+        high.nodelist = Some("n1".into());
+        let high_id = submit_and_wait(&cm, high);
+        let high_job = cm.get_job(high_id).unwrap();
+
+        crate::scheduler_loop::try_preempt(&cm, &[&high_job]);
+        assert_eq!(cm.get_job(low_id).unwrap().state, JobState::Running);
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn purge_expired_holds_pending_reservation_jobs() {
```
yansun1996, line 6346: Test coverage: only the default (holding) path is tested here. A minimal `no_hold_jobs_reservation_delete_does_not_permanently_block_job` test — create a reservation with `no_hold_jobs` set, submit a job against it, delete/expire the reservation, assert the job is still eligible via `pending_jobs()` — would catch the gap flagged above on `delete_reservation`/`purge_expired_reservations`.

yansun1996 (commented):
Please see the inline comments below.

crates/spurctld/src/cluster.rs
sgopinath1: Fixed. apply_operation now dedupes ReservationCreate at the serialized apply point; ClientResponse.reservation_created reports whether it applied, and the RPC returns AlreadyExists otherwise. Added concurrent_reservation_create_keeps_single_entry test.

crates/spurctld/src/cluster.rs
sgopinath1: Fixed. Delete/purge now clears job.spec.reservation via detach_jobs_from_deleted_reservation_jobs in the WAL apply path, so reservation_block() no longer filters those jobs. Added no_hold_jobs_reservation_delete_does_not_permanently_block_job and purge variant tests.

crates/spurctld/src/cluster.rs
sgopinath1: Fixed. Fence reasons are only set when all candidate nodes are reservation-blocked (job_candidate_node_names). Diagnostics-only path; scheduling still uses prospective_overlap independently.

crates/spurctld/src/cluster.rs
sgopinath1: Fixed. Exhausted requeues now transition to held Pending with JobHoldMaxRequeue (configurable via max_batch_requeue in spur.conf) instead of staying in Preempted forever. Admin can release with scontrol release.

crates/spurctld/src/server.rs
sgopinath1: Fixed. Introduced typed ReservationError; reservation_rpc_status maps variants to gRPC codes directly — no to_string() substring matching.

crates/spurctld/src/server.rs
sgopinath1: Fixed — removed the unused allow.

crates/spur-sched/src/backfill.rs
sgopinath1: Fixed. Added prospective_overlap_at(job, res, node, at, duration) in spur-core; backfill calls it so the authorization rule lives in one place.

proto/slurm.proto
sgopinath1: Fixed — moved under its own // Reservation comment, field number unchanged.

crates/spurctld/src/cluster.rs
sgopinath1: Added reservation_survives_wal_replay: drops ClusterManager, restarts against the same Raft dir, asserts reservation survives replay.

crates/spurctld/src/cluster.rs
sgopinath1: Added no_hold_jobs_reservation_delete_does_not_permanently_block_job and purge variant — create with no_hold_jobs, submit job, delete/expire, assert job is still in pending_jobs().

sgopinath1:
> Please see the inline comments below.

Thanks for reviewing @yansun1996. Please check the fixes and responses.

yansun1996 (dismissed):
All 10 of my original findings are genuinely fixed in the latest push (d3cd08c) — verified by reading the actual code, not just the reply text:

1. TOCTOU race on reservation names — fixed via a proper re-check in `apply_operation`'s serialized `ReservationCreate` arm (`reservation_created` flag)
2. `no_hold_jobs` permanent block — fixed via `detach_jobs_from_deleted_reservation_jobs` clearing `job.spec.reservation`
3. `reservation_fence_reason` over-fencing — fixed, now requires all candidate nodes blocked
4. Preempted jobs stuck past max requeue — fixed via `hold_job_at_max_requeue`
5. Brittle string-matched RPC status — fixed with a typed `ReservationError` enum
6. Dead clippy attribute — removed
7. Duplicated overlap logic — unified via `prospective_overlap_at`
8. Proto field grouping — `reservation = 52` now has its own comment header
9. WAL-replay durability — new test properly restarts against the same on-disk dir
10. `no_hold_jobs` regression coverage — dedicated tests added

Also confirmed Copilot's earlier 5 findings were genuinely addressed in the prior commit.

CI hasn't finished running against this commit yet — merge once it's green.

PR #401, opened 2026-07-07
fix(spurd): chown job stdout/stderr and output dirs to submitting user

biluriuday:
This PR makes spurd run jobs correctly as the submitting user and brings `srun` I/O behavior in line with Slurm.

- **Own job output as the submitting user.** spurd runs as root but created stdout/stderr files and output dirs as root, so users couldn't read or manage their own output. Files are now opened/created in a forked child that drops to the user's credentials, with the open fds passed back over a socketpair (`SCM_RIGHTS`). Resolving paths as the user also closes a symlink TOCTOU where a job could coerce root into creating/truncating/owning files outside its reach.
- **Keep job scratch off NFS.** The job script and namespace wrapper now go in a spurd-owned per-job dir under `/var/spool/spur` (mirroring Slurm's `SlurmdSpoolDir`) instead of the user's `work_dir`. Root-side writes into a `root_squash` NFS `work_dir` were squashed to `nobody` and failed, hanging jobs in PENDING. Cleaned up on completion.
- **Fix shared-output interleaving.** When stderr follows stdout (`srun -o` with no `-e`), stderr now shares stdout's open file description via `dup` so both streams advance one offset; previously the path was opened twice with independent offsets, so stdout writes clobbered stderr.
- **Preserve `srun` argv quoting.** `srun`'s command is now shell-escaped when wrapped into the batch script, so spaces and shell metacharacters stay literal arguments (`srun touch "my file"` → one file; `srun bash -c 'a; b'` → single `-c` arg), matching Slurm's direct-exec semantics.

## Test plan

- `cargo test` and `cargo clippy` green across `spurd` and `spur-cli`.
- Verified end-to-end on the bare-metal cluster: I/O redirection (`-o`/`-e`/`-i`), `%j` expansion, shared-file stdout/stderr interleaving, stdin overlap/permission guards, and `srun` argv quoting.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3475,6 +3475,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serial_test",
+ "shlex",
  "spur-client",
  "spur-core",
  "spur-net",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2054,6 +2054,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "88904434abc2901f197fe8cc55f0445e7ded921dba5911dad2e2b39b48e663c4"
 
 [[package]]
+name = "memoffset"
+version = "0.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


shiv-tyagi (approved):
LGTM

PR #407, opened 2026-07-08
feat(spur-cli): add -o/--output, -e/--error, -i/--input to srun

shiv-tyagi:
## Summary

- Add `-o/--output`, `-e/--error`, and `-i/--input` flags to `srun`, matching Slurm's I/O redirection interface. Previously, `srun -o /shared_nfs/uptime.out uptime` was silently ignored.
- `-o` and `-e` wire into the existing `JobSpec.stdout_path` / `stderr_path` proto fields that `sbatch` already uses. In step mode (inside `salloc`/`sbatch`), the RPC response is written to the specified files client-side.
- `-i` adds a new `stdin_path` field end-to-end (proto, core model, controller, scheduler, agent). The agent opens the file and passes it as the batch process's stdin instead of `/dev/null`.
- When `-o` is set, terminal streaming is suppressed since the agent writes to the file on the compute node. Without `-o`, behavior is unchanged.
- `scontrol show job` displays `StdIn=` when set.
- Step mode warns when `-i` is used since `RunStepRequest` has no stdin mechanism.

## Test plan

- [x] Unit tests: flag parsing (short/long), defaults, `resolve_output_path_client` (empty, pattern, absolute)
- [x] `cargo clippy` clean, `cargo test` green (full suite, 0 failures)
- [x] Bare-metal cluster (4-node):
  - `srun hostname` (no flags): output to default `spur-{id}.out` on compute node, no regression
  - `srun -o /tmp/test-output.out hostname`: output in file, nothing streamed to terminal
  - `srun -e /tmp/test-error.err bash -c "echo stderr >&2"`: stderr separated into file
  - `srun -o ... -e ...` together: both files written correctly
  - `srun -i /tmp/stdin.txt cat`: `cat` reads input file, writes to output file
  - `srun -o /tmp/job-%j.out hostname`: `%j` expanded to job ID in filename
  - `scontrol show job`: displays `StdOut=`, `StdErr=`, `StdIn=` correctly
  - `srun -o` / `-i` inside `sbatch` script (step mode): `-o` writes to file, `-i` prints warning

### DIFF
--- a/crates/spur-cli/src/sbatch.rs
+++ b/crates/spur-cli/src/sbatch.rs
@@ -758,6 +758,7 @@ pub async fn main_with_args(cli_args: Vec<String>) -> Result<()> {
         work_dir,
         stdout_path: args.output.unwrap_or_default(),
         stderr_path: args.error.unwrap_or_default(),
+        stdin_path: String::new(),
         environment,
         time_limit,
         time_min: None,
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -196,9 +196,7 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
         }
     }
 
-    let stdout_path = args.output.clone().unwrap_or_default();
-    let stderr_path = args.error.clone().unwrap_or_default();
-    let stdin_path = args.input.clone().unwrap_or_default();
+    let io = resolve_io_paths(&args);
 
     let name = args.job_name.unwrap_or_else(|| args.command[0].clone());
 
@@ -261,9 +259,9 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
         gres,
         script,
         work_dir: work_dir.clone(),
-        stdout_path: stdout_path.clone(),
-        stderr_path: stderr_path.clone(),
-        stdin_path: stdin_path.clone(),
+        stdout_path: io.stdout.clone(),
+        stderr_path: io.stderr.clone(),
+        stdin_path: io.stdin.clone(),
         environment,
         time_limit,
         constraint: args.constraint.unwrap_or_default(),
@@ -349,7 +347,7 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
                             job_id,
                             job.exit_code,
                             &work_dir,
-                            &stdout_path,
+                            &io.stdout,
                             &hooks,
                             false,
                         )
@@ -374,7 +372,7 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
     }
 
     // Stream output to terminal only when the user didn't ask for file output.
-    let output_streamed = if stdout_path.is_empty() {
+    let output_streamed = if io.stdout.is_empty() {
         try_stream_output(&mut client, &nodelist, job_id).await
     } else {
         false
@@ -383,7 +381,7 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
         &mut client,
         job_id,
         &work_dir,
-        &stdout_path,
+        &io.stdout,
         &hooks,
         output_streamed,
     )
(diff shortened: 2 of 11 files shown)


crates/spurd/src/executor.rs
```diff
@@ -418,12 +420,20 @@ async fn spawn_job_process(
 
     // Launch the process
     let mut cmd = Command::new(&launch_cmd);
+    let stdin_stdio: Stdio = if stdin_path.is_empty() {
```
yansun1996, line 446: This stdin wiring only covers the non-container path. `launch_container_job` (used whenever `--container-image` is set) never receives `stdin_path`, and its child only dup2s stdout/stderr — `-i` silently becomes a no-op for containerized jobs, with no warning (unlike the explicit warning printed for the step-mode case).

crates/spurd/src/executor.rs
```diff
@@ -418,12 +420,20 @@ async fn spawn_job_process(
...
+        Stdio::null()
+    } else {
+        let resolved = resolve_output_path(stdin_path, job_id, work_dir);
+        let f = std::fs::File::open(&resolved)
```
yansun1996, line 465: This opens the user-supplied `-i` path with spurd's own privileges (root in the standard deployment) — `cmd.uid()/cmd.gid()` only drops privilege in the forked child at exec time, not for this parent-side open. There's no check that the resolved path is readable by the job's actual uid, so `srun -i /etc/shadow cat` can read arbitrary root-readable files back through the job's stdout. Needs a permission check against the job's uid before opening, or defer the open until after the privilege drop.

crates/spurd/src/executor.rs
```diff
@@ -418,12 +420,20 @@ async fn spawn_job_process(
...
+            .with_context(|| format!("failed to open stdin file: {}", resolved))?;
+        Stdio::from(f)
+    };
     cmd.args(&launch_args)
         .current_dir(work_dir)
         .envs(&env)
         .stdout(stdout_file.into_std().await)
```
yansun1996, line 472: `stdout_file`/`stderr_file` are opened with `File::create` (truncating) earlier in this function, before the stdin file above is opened for reading. If `-o` and `-i` target the same path, the input file gets truncated to empty before it's ever read — silent data loss with no diagnostic.

yansun1996 (commented):
Went through the srun I/O redirection changes in detail — flagged a few things inline, mainly around the container path and the new file handling order.

crates/spur-cli/src/srun.rs
```diff
@@ -184,6 +196,10 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
         }
     }
 
+    let stdout_path = args.output.clone().unwrap_or_default();
+    let stderr_path = args.error.clone().unwrap_or_default();
+    let stdin_path = args.input.clone().unwrap_or_default();
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed in 6725f17. Added `resolve_io_paths()` helper that defaults stderr to stdout when `-o` is set but `-e` is not. Used in both `main_with_args` and `run_as_step`.

crates/spur-cli/src/srun.rs
```diff
@@ -586,11 +647,38 @@ async fn run_as_step(
...
+        } else {
+            print!("{}", resp.stdout);
+        }
     }
     if !resp.stderr.is_empty() {
-        eprint!("{}", resp.stderr);
+        if let Some(ref path) = args.error {
+            let resolved = resolve_output_path_client(path, job_id, work_dir);
+            if let Some(parent) = std::path::Path::new(&resolved).parent() {
+                tokio::fs::create_dir_all(parent).await.ok();
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed in 6725f17. Step mode now uses `resolve_io_paths()` for the same stderr-follows-stdout default. When stderr and stdout resolve to the same file, stderr is appended via `write_step_file(..., append: true)` instead of truncating.

crates/spur-cli/src/sbatch.rs
```diff
@@ -758,6 +758,7 @@ pub async fn main_with_args(cli_args: Vec<String>) -> Result<()> {
         work_dir,
         stdout_path: args.output.unwrap_or_default(),
         stderr_path: args.error.unwrap_or_default(),
+        stdin_path: String::new(),
```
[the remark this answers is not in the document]
shiv-tyagi: Pre-existing behavior in `sbatch`, not introduced by this PR. The only `sbatch.rs` change here adds `stdin_path: String::new()` for the new proto field. Will track sbatch stderr-follows-stdout separately.

crates/spurd/src/executor.rs
```diff
@@ -418,12 +420,20 @@ async fn spawn_job_process(
...
     // Launch the process
     let mut cmd = Command::new(&launch_cmd);
+    let stdin_stdio: Stdio = if stdin_path.is_empty() {
+        Stdio::null()
+    } else {
+        let resolved = resolve_output_path(stdin_path, job_id, work_dir);
+        let f = std::fs::File::open(&resolved)
+            .with_context(|| format!("failed to open stdin file: {}", resolved))?;
+        Stdio::from(f)
+    };
```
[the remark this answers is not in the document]
shiv-tyagi: Addressed in 6725f17. Container path now logs a warning when `stdin_path` is set. Wiring stdin into the container runtime is a larger change and deferred.

crates/spurctld/src/server.rs
```diff
@@ -1777,6 +1782,7 @@ fn job_to_proto(job: &spur_core::job::Job) -> JobInfo {
         derived_exit_code: job.derived_exit_code,
         stdout_path: job.resolved_stdout(),
         stderr_path: job.resolved_stderr(),
+        stdin_path: job.spec.stdin_path.clone().unwrap_or_default(),
         resources: job.allocated_resources.as_ref().map(allocations_to_proto),
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed in 6725f17. Added `Job::resolved_stdin()` that expands `%j`/`%N`/etc, and `job_to_proto` now uses it instead of the raw spec value.

crates/spurd/src/executor.rs
shiv-tyagi: Fixed in 6725f17. Container path now warns when `stdin_path` is non-empty. Wiring stdin into `launch_container_job` requires plumbing through the fork/exec path and is deferred to a follow-up.

crates/spurd/src/executor.rs
shiv-tyagi: Fixed in 6725f17. Before `File::open`, we now check owner/group/other read bits against the job's uid/gid. Jobs submitted by non-root users cannot open files they don't have read access to.

Known limitation: supplementary groups and ACLs are not checked. A more robust approach would open the fd in a `pre_exec` closure after `setuid`/`initgroups`, but that requires consolidating the existing `pre_exec` closures and is deferred.

crates/spurd/src/executor.rs
shiv-tyagi: Fixed in 6725f17. Two changes:

1. Stdin overlap guard: if the resolved stdin path matches stdout or stderr, we bail before any `File::create` runs.
2. Executor stderr fd reuse: when `stdout_resolved == stderr_resolved` (stderr-follows-stdout case), stderr is opened with `O_APPEND` instead of truncating, so both streams merge correctly.

sgopinath1 (approved):
LGTM

PR #413, opened 2026-07-09
feat(config): enforce partition allow_accounts and deny_accounts at s…

sgopinath1:
## Summary

Partition account access control (`allow_accounts` / `deny_accounts`) now works end-to-end from `spur.conf` through submit-time enforcement.

- Fix `build_partitions()` so `allow_accounts` and `deny_accounts` are copied from config into runtime `Partition` structs (they were previously dropped via `..Default::default()`).
- Enforce both lists in `validate_partition()` at job submit.
- Resolve the submitter's default association account before validation when `--account` is omitted.
- Add `SubmitError` so client mistakes (bad partition, disallowed/denied account, invalid dependency/QOS) map to gRPC `INVALID_ARGUMENT` and REST HTTP 400 instead of `INTERNAL`.
- Append `deny_accounts` to `PartitionInfo` proto (field 42) and expose account access fields in `scontrol`, `sinfo -o %u`, REST `/partitions`, and `examples/spur.conf`.

`allow_groups` is intentionally out of scope for this PR (schema exists but is not enforced).

## Test plan

- [x] `cargo test --locked`
- [x] Unit tests: config propagation, submit accept/reject for allow/deny accounts, default account resolution, `SubmitError` gRPC/REST mapping

### DIFF
--- a/crates/spur-cli/src/format_engine.rs
+++ b/crates/spur-cli/src/format_engine.rs
@@ -217,6 +217,7 @@ pub fn sinfo_header(spec: char) -> &'static str {
         'n' => "HOSTNAMES",
         'O' => "CPU_LOAD",
         'e' => "FREE_MEM",
+        'u' => "ALLOW_ACCOUNTS",
         _ => "?",
     }
 }
@@ -224,6 +225,11 @@ pub fn sinfo_header(spec: char) -> &'static str {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn sinfo_header_includes_allow_accounts() {
+        assert_eq!(super::sinfo_header('u'), "ALLOW_ACCOUNTS");
+    }
 
     #[test]
     fn test_parse_default_squeue_format() {
--- a/crates/spur-cli/src/format_engine.rs
+++ b/crates/spur-cli/src/format_engine.rs
@@ -217,7 +217,6 @@ pub fn sinfo_header(spec: char) -> &'static str {
         'n' => "HOSTNAMES",
         'O' => "CPU_LOAD",
         'e' => "FREE_MEM",
-        'u' => "ALLOW_ACCOUNTS",
         _ => "?",
     }
 }
@@ -225,11 +224,6 @@ pub fn sinfo_header(spec: char) -> &'static str {
 #[cfg(test)]
 mod tests {
     use super::*;
-
-    #[test]
-    fn sinfo_header_includes_allow_accounts() {
-        assert_eq!(super::sinfo_header('u'), "ALLOW_ACCOUNTS");
-    }
 
     #[test]
     fn test_parse_default_squeue_format() {
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -12,6 +12,7 @@ use tokio::sync::Notify;
 use tracing::{debug, info, warn};
 
 use spur_core::accounting::{Qos, TresRecord, TresType};
+use spur_core::auth::AuthError;
 use spur_core::burst_buffer::BbStageState;
 use spur_core::config::SlurmConfig;
 use spur_core::job::{
@@ -463,8 +464,23 @@ impl ClusterManager {
         Ok(())
     }
 
-    /// Cancel a job.
-    pub fn cancel_job(&self, job_id: JobId, _user: &str) -> anyhow::Result<()> {
+    /// Check that `user` is allowed to perform `action` on a job owned by `owner`.
+    /// Empty user (internal/daemon calls) and root are always allowed.
+    fn check_job_owner(user: &str, owner: &str, action: &str) -> anyhow::Result<()> {
+        if user.is_empty() || user == "root" || user == owner {
+            return Ok(());
+        }
+        Err(AuthError::NotJobOwner {
+            user: user.into(),
+            owner: owner.into(),
+            action: action.into(),
+        }
+        .into())
+    }
+
+    /// Cancel a job. The requesting `user` must be the job owner, root, or
+    /// empty (trusted internal/daemon calls).
+    pub fn cancel_job(&self, job_id: JobId, user: &str) -> anyhow::Result<()> {
         {
             let jobs = self.jobs.read();
             let job = jobs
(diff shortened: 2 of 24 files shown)


crates/spurctld/src/cluster.rs
```diff
@@ -164,8 +191,9 @@ impl ClusterManager {
     }
 
     /// Submit a new job. If it has an array spec, expand into individual tasks.
-    pub fn submit_job(&self, mut spec: JobSpec) -> anyhow::Result<JobId> {
+    pub fn submit_job(&self, mut spec: JobSpec) -> Result<JobId, SubmitError> {
         apply_default_partition(&mut spec, &self.partitions.read());
+        apply_default_account(&mut spec, &self.association_cache);
         self.validate_partition(&spec)?;
         apply_default_qos(&mut spec, &self.association_cache, &self.qos_cache)?;
```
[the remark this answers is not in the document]
sgopinath1: Fixed.

crates/spurctld/src/cluster.rs
```diff
@@ -205,39 +235,36 @@ impl ClusterManager {
...
             Some(p) => p,
-            None => anyhow::bail!("partition '{}' not found", partition_name),
+            None => {
+                return Err(SubmitError::invalid(format!(
+                    "partition '{partition_name}' not found"
+                )));
+            }
         };
 
+        let account = spec.account.as_deref().unwrap_or("");
```
yansun1996: `account` here is whatever the client sent, with no check that the submitting user is actually a member of it — `AssociationCache` only tracks one *default* account per user, not the full membership set. Anyone can bypass `allow_accounts`/`deny_accounts` by just passing `--account=<allowed-account>`, which defeats the point of this PR. Needs a real membership check before this is used for access control.

yansun1996 (commented):
Reviewed the account-ACL enforcement path. Left inline comments on the blocker and major items — the core one being that the claimed `account` is never checked against the submitter's real associations, so the new allow/deny lists are bypassable by just passing a different `--account`.

crates/spurctld/src/cluster.rs
```diff
@@ -205,39 +235,36 @@ impl ClusterManager {
...
         // Check deny_accounts
-        if let Some(ref account) = spec.account {
-            if part.deny_accounts.iter().any(|a| a == account) {
-                anyhow::bail!(
-                    "account '{}' denied on partition '{}'",
-                    account,
-                    partition_name
-                );
-            }
+        if !account.is_empty() && part.deny_accounts.iter().any(|a| a == account) {
```
yansun1996: If an account is in both `allow_accounts` and `deny_accounts`, deny wins here. Real Slurm does the opposite — when `AllowAccounts` is set, `DenyAccounts` is ignored entirely. Worth matching that precedence so a migrated `slurm.conf` doesn't silently behave differently.

crates/spur-cli/src/format_engine.rs
```diff
@@ -217,6 +217,7 @@ pub fn sinfo_header(spec: char) -> &'static str {
         'n' => "HOSTNAMES",
         'O' => "CPU_LOAD",
         'e' => "FREE_MEM",
+        'u' => "ALLOW_ACCOUNTS",
```
yansun1996: `%u` already means something in real Slurm `sinfo` (user who set a node's unavailable reason) — repurposing it for `ALLOW_ACCOUNTS` breaks drop-in compat for any script using stock `sinfo -o %u`. Slurm has no native sinfo specifier for this at all; consider dropping the sinfo exposure (scontrol/REST already cover it) or picking an unused letter and documenting it as a Spur extension.

crates/spurctld/src/rest/handlers.rs
```diff
@@ -95,14 +95,18 @@ pub async fn submit_job(
         ..Default::default()
```
yansun1996, line 96: `SubmitJobFields` has no `user` field, so REST-submitted jobs always get `spec.user == ""`. Since default-account resolution keys off `spec.user`, a REST client that omits `account` will always be rejected on any `allow_accounts` partition — even though the same user submitting via CLI/gRPC would succeed via their default account. REST and CLI/gRPC now diverge in behavior for the same user.

crates/spurctld/src/cluster.rs
```diff
@@ -3937,17 +3967,29 @@ fn apply_default_partition(spec: &mut JobSpec, partitions: &[Partition]) {
     }
 }
 
+/// Resolve the submitting user's default account from the association cache
+/// when `--account` was not provided (mirrors `apply_default_partition`).
+fn apply_default_account(spec: &mut JobSpec, assoc_cache: &AssociationCache) {
+    if spec.account.as_deref().is_some_and(|a| !a.is_empty()) {
+        return;
+    }
+    let (account, _) = assoc_cache.resolve(&spec.user, None);
```
yansun1996, line 4000: If the accounting DB is unreachable at startup (or `accounting.database_url` is empty), the refresh loop that populates this cache never runs, so this resolves to `None` for every user permanently, not just during a startup race. Combined with `allow_accounts`, that means every submission omitting `--account` fails outright on those partitions. `apply_default_qos` right below explicitly degrades gracefully in the equivalent case — this doesn't.

yansun1996 (approved):
All 4 comments addressed with real code changes, verified against the diff:
- Account membership now checked against a real association-cache membership set — spoofing an unauthorized account is rejected.
- Allow/deny precedence fixed to match Slurm (allow set → deny ignored).
- `sinfo -o %u` reverted rather than repurposed, preserving compat.
- Default-account resolution now fails open (matching `apply_default_qos`) with a warning instead of hard-failing every submission when the accounting DB is unreachable.

Ran clippy + the new tests against the current head — clean. Good to merge once E2E finishes green.

crates/spurctld/src/cluster.rs
sgopinath1: Fixed.

crates/spurctld/src/cluster.rs
sgopinath1: Fixed.

crates/spur-cli/src/format_engine.rs
sgopinath1: Fixed.

crates/spurctld/src/rest/handlers.rs
sgopinath1: Fixed

crates/spurctld/src/cluster.rs
sgopinath1: Fixed.

PR #415, opened 2026-07-10
fix(spur-cli): use connect_channel for accounting CLI commands

shiv-tyagi:
## Summary

- `sacct`, `sacctmgr`, `sreport`, and `sshare` passed the raw `SPUR_CONTROLLER_ADDR` value directly to `tonic::Endpoint::connect()`, which treats the entire string as a single URI. When multiple controllers are configured (comma-separated for HA), this produced "invalid URI / invalid authority" errors.
- All four commands now route through `spur_client::connect_channel()`, which splits on commas and does ordered failover with a 2-second connect timeout per endpoint. This matches the behavior of every other CLI command.

## Test plan

- [x] Reproduced "invalid authority" error on bare-metal cluster with multi-endpoint `SPUR_CONTROLLER_ADDR`
- [x] Verified fix: `sacct` and `sacctmgr` connect successfully with comma-separated endpoints
- [x] Verified failover: killed first controller, commands skip dead endpoint and connect to next live one
- [x] `cargo clippy` passes with 0 warnings
- [x] `cargo test` passes

### DIFF
--- a/crates/spur-cli/src/sacct.rs
+++ b/crates/spur-cli/src/sacct.rs
@@ -153,9 +153,10 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
         })
         .unwrap_or_default();
 
-    let mut client = SlurmAccountingClient::connect(args.controller)
+    let channel = spur_client::connect_channel(&args.controller)
         .await
         .context("failed to connect to controller")?;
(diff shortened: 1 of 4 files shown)


sgopinath1 (approved):
LGTM

PR #421, opened 2026-07-10
feat(spur-cli): honor SRUN_/SALLOC_/SLURM_ env-var defaults for srun …

### LINKED ISSUE #279: Support SRUN_*/SALLOC_* environment variable defaults for srun/salloc flags
## Problem

Like sbatch, Slurm reads `SRUN_*` and `SALLOC_*` env vars as defaults for srun and salloc CLI flags respectively. Spur ignores these, breaking workflows that rely on environment-based defaults.

Currently no `#[arg]` in `SrunArgs` or `SallocArgs` has an `env` setting except `--controller` which reads `SPUR_CONTROLLER_ADDR`.

## Expected Behavior

Each srun flag should read from `SRUN_*` env vars, each salloc flag from `SALLOC_*` env vars:
- `SRUN_PARTITION`, `SRUN_ACCOUNT`, `SRUN_MEM`, `SRUN_GRES`, `SRUN_CONSTRAINT`, etc.
- `SALLOC_PARTITION`, `SALLOC_ACCOUNT`, `SALLOC_GRES`, `SALLOC_CONSTRAINT`, etc.

CLI flags override env vars.

## Acceptance Criteria

- [ ] Each relevant `#[arg]` in `SrunArgs` has `env = "SRUN_<FLAG>"` set
- [ ] Each relevant `#[arg]` in `SallocArgs` has `env = "SALLOC_<FLAG>"` set
- [ ] Explicit CLI flags override env var values
- [ ] Existing behavior unchanged when env vars are not set

biluriuday:
…and salloc

## Summary

Adds environment-variable defaults for `srun` and `salloc`, aligning Spur with Slurm's precedence model (closes #279). Also fixes `--gres` CLI parsing to match Slurm semantics.

Precedence for every affected flag is now: **CLI arg > `SPUR_*` native twin > `SLURM_*`/`SRUN_*`/`SALLOC_*` > built-in default**.

## What changed

- **New shared helper module (`env_defaults.rs`)** — `apply_str`/`apply_string`/`apply_flag`/`apply_csv` resolve multi-alias env vars with correct CLI precedence, detected via `ArgMatches::value_source`. Includes a shared `EnvGuard` test fixture and unit tests.
- **`srun`** — parses into `ArgMatches`, then applies `SLURM_*` (plus genuine `SRUN_*` where Slurm uses them) and `SPUR_*` twins as defaults. This lets `srun` inside an allocation inherit `SLURM_NTASKS`/`SLURM_CPUS_PER_TASK`/`SLURM_NNODES`, so a step sizes to the allocation instead of the bare `-n1 -c1` defaults.
- **`salloc`** — same pattern for `SALLOC_*`/`SPUR_*`. Stays Slurm-faithful: `--nodes`, `--ntasks`, `--cpus-per-task`, and `--job-name` have **no** env var in real Slurm, so Spur ignores them too.
- **`--gres` fix (`srun` + `salloc`)** — added `value_delimiter = ','` + `overrides_with = "gres"` so a comma-list (`--gres=gpu:1,fpga:2`) expands to multiple GRES and a repeated `--gres` replaces (last wins), matching Slurm and Spur's own `sbatch`. Previously the CLI kept comma-lists as one un-parseable element and accumulated repeated flags.

## Design notes

- Clap's `#[arg(env=...)]` supports only one env var per flag, so multi-alias resolution is done in a post-parse step keyed off `value_source` (keeps CLI-set detection accurate even with `overrides_with`).
- Env-dependent tests run serially (`serial_test`) and use `EnvGuard`, which clears all `SPUR_/SLURM_/SALLOC_/SRUN_`-prefixed vars on construction/drop, keeping them independent of the runner's environment and of each other.

## Testing

- `cargo clippy -p spur-cli --all-targets` — clean.
- `cargo test -p spur-cli` — all 151 tests pass (new unit tests cover defaults, CLI override, alias precedence, numeric/bool/CSV parsing, gres comma-split, and last-wins).
- **Bare-metal cluster e2e** (3 nodes, non-Raft): verified on the deployed binaries:
  - `salloc`: `SALLOC_PARTITION` env default applied; CLI `-p` overrides it.
  - `srun`: `SLURM_NTASKS` env default applied to submitted job; CLI `-n` overrides it; `SPUR_NTASKS` takes precedence over the `SLURM_NTASKS` twin.

## Known gap / follow-up 

sbatch still resolves env defaults via clap's single SBATCH_* attribute per flag and does not yet honor SPUR_* twins (e.g. SPUR_PARTITION), unlike srun/salloc. Unifying sbatch onto env_defaults is deferred to a follow-up PR since it must preserve sbatch's CLI > env > #SBATCH directive precedence.

### DIFF
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -407,13 +407,23 @@ fn resolve_srun_env(matches: &ArgMatches, args: &mut SrunArgs) -> Result<()> {
     apply_str(
         matches,
         "partition",
-        &["SPUR_PARTITION", "SLURM_PARTITION"],
+        &[
+            "SPUR_PARTITION",
+            "SPUR_JOB_PARTITION",
+            "SLURM_PARTITION",
+            "SLURM_JOB_PARTITION",
+        ],
         &mut args.partition,
     );
     apply_str(
         matches,
         "account",
-        &["SPUR_ACCOUNT", "SLURM_ACCOUNT"],
+        &[
+            "SPUR_ACCOUNT",
+            "SPUR_JOB_ACCOUNT",
+            "SLURM_ACCOUNT",
+            "SLURM_JOB_ACCOUNT",
+        ],
         &mut args.account,
     );
     apply_num(
@@ -1206,6 +1216,14 @@ mod tests {
 
     #[test]
     #[serial(env_injection)]
+    fn numeric_env_trims_whitespace() {
+        let env = EnvGuard::new();
+        env.set("SLURM_NTASKS", " 4 ");
+        assert_eq!(resolve_from(&["srun", "hostname"]).ntasks, 4);
+    }
+
+    #[test]
+    #[serial(env_injection)]
     fn invalid_numeric_env_errors() {
         let env = EnvGuard::new();
         env.set("SLURM_NNODES", "abc");
@@ -1259,6 +1277,15 @@ mod tests {
     }
 
     #[test]
+    #[serial(env_injection)]
+    fn gres_env_trims_whitespace_and_drops_empties() {
+        let env = EnvGuard::new();
+        env.set("SLURM_GRES", "gpu:1, fpga:2 ,");
+        let args = resolve_from(&["srun", "hostname"]);
+        assert_eq!(args.gres, vec!["gpu:1", "fpga:2"]);
+    }
+
+    #[test]
     fn cli_gres_comma_list_splits_into_multiple() {
         let args = SrunArgs::try_parse_from(["srun", "--gres=gpu:1,fpga:2", "hostname"])
             .expect("parse failed");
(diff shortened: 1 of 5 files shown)


crates/spur-cli/src/srun.rs
```diff
@@ -388,6 +393,173 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
...
+        matches,
+        "job_name",
+        &["SPUR_JOB_NAME", "SLURM_JOB_NAME"],
+        &mut args.job_name,
+    );
+    apply_str(
+        matches,
+        "partition",
+        &[
+            "SPUR_PARTITION",
```
shiv-tyagi, line 411: Q: `sbatch` still uses clap's single `env=` attribute (one `SBATCH_*` var per flag, no `SPUR_*` twin), while `srun`/`salloc` now honor `SPUR_*` twins and multi-alias `SLURM_*` fallbacks through `env_defaults`. That is user-visible: `SPUR_PARTITION` works for `srun`/`salloc` but is silently ignored by `sbatch`. Is unifying `sbatch` onto `env_defaults` a planned follow-up, or intentionally out of scope? Worth a line in the PR description either way.

biluriuday: SBATCH is intentionally left out-of-scope of this PR. we can handle it in a follow up PR

biluriuday: opened a follow-up issue to track this https://github.com/ROCm/spur/issues/448

shiv-tyagi (approved):

PR #423, opened 2026-07-10
feat(spurctld,spur-cli): runtime partition CRUD with HA persistence

yansun1996:
## What

Adds runtime partition management via `scontrol`, persisted through the Raft WAL so changes survive controller restarts and HA failover.

**New commands (spur-native `--flag` style):**
- `scontrol create-partition --name=<name> [--nodes=...] [--max-time=...] [--state=...] [--allow-accounts=...] [--deny-accounts=...] [--allow-qos=...] [--deny-qos=...] ...`
- `scontrol update-partition --name=<name> [fields to patch]`
- `scontrol delete-partition --name=<name>`

**Slurm-compatible inline syntax (full parity):**
- `scontrol create PartitionName=<name> Nodes=<list> MaxTime=<limit> ...`
- `scontrol update PartitionName=<name> MaxTime=<new> ...`
- `scontrol delete PartitionName=<name>`

Key=value parsing is case-insensitive. Slurm-only keys (AllocNodes, DisableRootJobs, etc.) are silently ignored so existing Slurm scripts work unchanged.

`scontrol show partition` updated to display AllowAccounts, DenyAccounts, AllowQos, DenyQos, State, MinNodes, MaxNodes, PreemptMode matching Slurm output format.

AllowQos / DenyQos are enforced at job submission — `validate_partition()` rejects jobs whose QoS is excluded or denied.

## Approach

Raft WAL is the single source of truth. `spur.conf` is a cold-start seed only and can no longer shadow runtime changes.

Three new `WalOperation` variants (`PartitionCreate`, `PartitionUpdate`, `PartitionDelete`) are applied identically on all Raft peers via `apply_operation()`. On restart, snapshot + WAL replay produces correct final state regardless of spur.conf contents.

To prevent config-file partitions from re-seeding after runtime deletion, `ClusterSnapshot` carries a `deleted_partition_names: HashSet<String>` tombstone set. On `restore_from_snapshot()`: tombstoned names are removed from the config-seeded baseline first, then snapshot entries overlay the remainder.

The leader also writes spur.conf back atomically (temp-file + rename) after each partition operation so operators don't see stale config files — but this is cosmetic; Raft WAL is authoritative.

AllowAccounts / DenyAccounts / AllowGroups are stored and displayed. AllowGroups is not enforced (Unix group is not carried in `JobSpec`) — noted as a known gap.

## Testing

**Unit tests (19):** tombstone round-trip, WAL replay across restart, spur.conf writeback, `is_default` promotion, AllowQos/DenyQos enforcement, create/update/delete state machine.

**E2E pytest tests (43 across 7 test classes):**
- `TestPartitionCRUD` — basic create/update/delete and persistence across restart
- `TestPartitionDisplay` — `scontrol show partition` output format
- `TestPartitionScheduling` — jobs route to correct partition, reject wrong partition
- `TestPartitionDefaultFlag` — `--default` promotion clears other defaults
- `TestSlurmSyntax` (12 tests) — Slurm-compatible `scontrol create/update/delete PartitionName=...` inline syntax
- `TestPartitionAllowDenyAccounts` — AllowAccounts / DenyAccounts enforcement
- `TestPartitionAllowDenyQos` — AllowQos / DenyQos enforcement (requires `accounting_cluster` fixture)

**Live HA cluster (10.11.96.92 + 10.11.98.147):**
- Modified a config-defined partition via new CLI, restarted spurctld, verified changes persisted
- Forced leader re-election; new leader served correct updated state from WAL despite stale spur.conf
- 31/31 E2E tests passed (5 AllowQos/DenyQos tests skipped — need accounting cluster with Postgres)

## Known gaps

- AllowGroups is stored and displayed but not enforced — `JobSpec` does not carry the submitting user's Unix group
- spur.conf writeback is leader-only; followers keep stale config files until cold start

### DIFF
--- a/crates/spur-cli/src/scontrol.rs
+++ b/crates/spur-cli/src/scontrol.rs
@@ -66,6 +66,126 @@ pub enum ScontrolCommand {
         /// Job ID
         job_id: u32,
     },
+    /// Create a partition
+    #[command(name = "create-partition")]
+    CreatePartition {
+        /// Partition name
+        #[arg(long)]
+        name: String,
+        /// Hostlist of nodes (mutually exclusive with --selector)
+        #[arg(long, default_value = "")]
+        nodes: String,
+        /// Label selector as KEY=VALUE pairs, comma-separated (mutually exclusive with --nodes)
+        #[arg(long, default_value = "")]
+        selector: String,
+        /// Partition state: UP (default), DOWN, DRAIN, INACTIVE
+        #[arg(long, default_value = "UP")]
+        state: String,
+        /// Mark as the cluster default partition
+        #[arg(long)]
+        default: bool,
+        /// Maximum job wall-clock time (e.g. "24:00:00" or "INFINITE")
+        #[arg(long, default_value = "")]
+        max_time: String,
+        /// Default job wall-clock time (e.g. "01:00:00")
+        #[arg(long, default_value = "")]
+        default_time: String,
+        /// Maximum number of nodes per job
+        #[arg(long)]
+        max_nodes: Option<u32>,
+        /// Minimum number of nodes per job
+        #[arg(long, default_value = "1")]
+        min_nodes: u32,
+        /// Comma-separated accounts allowed (empty = all)
+        #[arg(long, default_value = "")]
+        allow_accounts: String,
+        /// Comma-separated groups allowed (empty = all)
+        #[arg(long, default_value = "")]
+        allow_groups: String,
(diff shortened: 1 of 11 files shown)


yansun1996:
Closing temporarily — the partition CRUD design needs to be completed with `scontrol reconfigure` (live spur.conf reload) before merging. Without it, operators have no way to apply bulk config changes or roll back to a known-good state, which breaks the Slurm operational model. Will reopen once reconfigure is implemented.

PR #424, opened 2026-07-10
feat(config): add production security baseline

yusingh-amd:
## Summary
- Adds `security.mode = dev | production` and production config validation that rejects auth `none`, open node admission, missing/empty JWT secrets when token auth is used, and the development fallback key.
- Gates the existing `spur-default-key` node-admission JWT fallback to dev mode so production no longer silently uses it.
- Labels example configs as dev-only and adds production security baseline guidance for native-host and Kubernetes deployments.

## Scope
This is PR 1 of the Fleet/Spur integration security foundation. It is intentionally limited to Spur production security groundwork and does not implement full mTLS, full authorization, Raft peer authentication, or Fleet compatibility.

## Test plan
- [x] IDE diagnostics on edited Rust files: no linter errors reported.
- [ ] `cargo fmt` (not run: `cargo` is not available in this Windows shell).
- [ ] Targeted Rust tests for `spur-core` and `spurctld` (not run: `cargo` is not available in this Windows shell).

### DIFF
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -41,6 +41,9 @@ pub struct SlurmConfig {
     pub auth: AuthConfig,
 
     #[serde(default)]
+    pub security: SecurityConfig,
+
+    #[serde(default)]
     pub partitions: Vec<PartitionConfig>,
 
     #[serde(default)]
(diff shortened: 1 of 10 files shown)


shiv-tyagi:
A config that validates other config is circular. If the admin knows to set `security.mode = production`, they know to set `auth = "token"` directly. The real fix is making secure values the default and requiring explicit opt-out for insecure settings (with a loud startup warning), not a meta-config layer on top. We could do this when we have TLS work in place.

yusingh-amd:
Random Agent generated PR ignore please! ;/

PR #427, opened 2026-07-13
fix(spur-core): enforce node and memory limits in QOS per-user checks

yansun1996:
## What

Fixes SPUR-19 and its follow-up gaps, fully closing the loop:
1. `check_qos_limits()`'s per-user TRES cap (`max_tres_per_user`) only checked CPU, silently ignoring configured Node and Memory limits.
2. Association/account-level `AccountLimits` (max_running_jobs, max_submit_jobs, max_tres_per_job, grp_tres, max_wall_minutes) had zero enforcement anywhere.
3. `sacctmgr show qos` was missing most limit columns (MaxTRES, MaxTRESPU, GrpTRES, MaxSubmitPU, GrpWall) that the proto already carried but the CLI never rendered.
4. A submit-count off-by-one bug that could permanently self-block every job for a user once their submit quota was reached.
5. No CLI/proto write path existed for any `AccountLimits` field — now fixed end-to-end (proto → CLI parsing → gRPC handler → Postgres upsert).
6. Memory limit checks (QOS and account level) only counted `--mem`/`memory_per_node_mb`, silently ignoring `--mem-per-cpu` — a job submitted that way bypassed every memory-based limit entirely.
7. `sacctmgr` silently coerced invalid numeric limit values (typos, negatives, overflow) to 0 ("unlimited") instead of erroring.
8. `TresRecord::parse` silently dropped unparseable TRES tokens (e.g. unit-suffixed `mem=1G`) instead of erroring, producing a no-op limit with zero admin feedback.

## Approach

- Added Node/Memory checks to the QOS per-user TRES block, mirroring the existing `grp_tres` sibling block.
- Implemented `check_account_limits`, mirroring the QOS-level check pattern, wired independently alongside it in the admission path.
- Added the missing QOS display columns, plus a new `grp_wall_minutes` field (proto + DB + CLI).
- Fixed the submit-count off-by-one (present in both the account and QOS checks) by excluding the job being evaluated from its own count.
- Extended `AddUserRequest` with 5 new fields, wired through the existing `associations` table upsert.
- Added a shared `effective_memory_mb()` helper (computing from either `--mem` or `--mem-per-cpu × cpus`) and used it everywhere a job's memory footprint is checked.
- `sacctmgr` limit-field parsing now returns a hard error on invalid input instead of silently zeroing.
- `TresRecord::parse` now returns a `Result`, rejecting malformed TRES strings at the gRPC boundary instead of silently dropping them.

All new `PendingReason` variants verified byte-exact against SchedMD's documented Slurm reason-code vocabulary.

## Known limitations, documented not fixed (pre-existing conventions this PR extends, not new bugs)

- `maxjobs=0` (or any unparseable value, now rejected outright rather than silently accepted) is indistinguishable from "unset" once validly set to 0 — no way to express Slurm's real "hard zero jobs allowed" semantic. Same `0=clear` convention already established elsewhere; redesigning it is out of scope here.
- `AddUserRequest` is a full resend, not a partial patch — any `modify user` call that doesn't specify all 5 limit fields will clear previously-set values on that association. Consistent with `modify account`/`modify qos`'s existing behavior. Worth a future tracked issue about a partial-patch model.
- `accounts.max_running_jobs` (a separate schema column) is a different cap from `associations.max_running_jobs` (the one enforced here) — setting the account-level one currently has no effect.
- `create_account`'s `max_running_jobs as i32` cast is inconsistent with the `i32::try_from` pattern used elsewhere in this PR — but `accounts.max_running_jobs` isn't wired to any enforcement path today, so impact is nil.
- `TresRecord::parse` still has no K/M/G unit-suffix support (Slurm CLI compatibility) — it now errors cleanly on such input rather than silently dropping it, but adding suffix parsing itself is separate, larger scope.

## Testing

- Unit tests for all new limit checks, isolated so each only exercises its own limit type, including `--mem-per-cpu` equivalence tests.
- E2E tests for QOS-level checks, the CLI write path, and the new hard-error behavior on invalid input.
- DB-level round-trip tests covering set / change-to-new-value / clear for the new association limit fields.
- Live-verified on the real 2-node cluster across 3 rounds, including: the off-by-one fix, QOS display columns, the full add/modify-user write path, `--mem-per-cpu` limit enforcement, and CLI/parser hard-error behavior on invalid input.

## Review

4 rounds of independent adversarial review (including GitHub Copilot's automated review, cross-validated each time) + cross-validation passes, which caught: a broken e2e test input format, a missing Slurm-parity table entry, the submit-count off-by-one deadlock, the missing-write-path scope gap, a UX inconsistency in `modify user`'s confirmation output, a test coverage gap, the `--mem-per-cpu` bypass, and the silent-invalid-input-acceptance issues in both the CLI and `TresRecord::parse`.


crates/spur-core/src/qos.rs
```diff
@@ -85,6 +85,21 @@ pub fn check_qos_limits(
...
+        }
+
+        if let Some(mem) = job.spec.memory_per_node_mb {
+            let job_mem = mem * job.spec.num_nodes as u64;
+            let new_total_mem = user_running_tres.get(TresType::Memory) + job_mem;
+            if max_tres.get(TresType::Memory) > 0 && new_total_mem > max_tres.get(TresType::Memory)
+            {
+                return QosCheckResult::Blocked(PendingReason::QosMaxMemoryPerUser);
+            }
+        }
```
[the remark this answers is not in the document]
yansun1996: Confirmed real via 2 independent reviewer cross-validation, and it's a practical bypass, not theoretical — `--mem-per-cpu` is a fully wired sbatch flag, and no normalization step exists before this check runs. The one place that computes effective memory from `memory_per_cpu_mb` (`spur_sched::backfill::job_resource_request`) is structurally unreachable from `spur-core` (wrong dependency direction). Fixed by adding a shared `effective_memory_mb()` helper in `spur-core` and using it at all 5 affected sites (qos.rs's 3 memory checks, account_limits.rs's 2, plus `cluster.rs::sum_running_tres`, which had the same gap). Live-verified: a job with `--mem-per-cpu=600 --cpus-per-task=2` now correctly trips the same QOS memory cap as an equivalent `--mem`-based job. Fixed in deaa3c8.

crates/spur-core/src/qos.rs
```diff
@@ -255,6 +270,53 @@ mod tests {
...
+                max_tres_per_user: Some(tres),
+                ..Default::default()
+            },
+            ..Default::default()
+        };
+        let mut job = make_test_job();
+        job.spec.num_nodes = 1;
+        job.spec.memory_per_node_mb = Some(3000); // job needs 3 GiB
+        let mut running = TresRecord::new();
+        running.set(TresType::Memory, 2000); // already using 2 GiB; 2000 + 3000 > 4096
```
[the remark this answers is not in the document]
yansun1996: Confirmed the labeling is inaccurate (3 GiB would be 3072 MB, not 3000). Verified it's purely cosmetic though — the test only asserts `2000 + 3000 > 4096`, which holds regardless of what unit label is on the comment, and the identical mislabeling already exists in a pre-existing test in this same file predating this PR (copy-pasted forward). Leaving as-is for now rather than touching unrelated test comments in an already-large PR; happy to clean up in a follow-up pass.

yansun1996:
## Live testing performed

Verified on a real 2-node cluster (not simulated/mocked), across several rounds as fixes were added:

- Configured QOS/account entities with Node and Memory limits; confirmed jobs are blocked with the correct pending reason when a limit is exceeded, and allowed to run when within limits.
- Verified the submit-count fix: the first in-quota job runs normally; only a subsequent over-quota job gets blocked (previously, hitting the quota incorrectly self-blocked every job for that user, including ones within limits).
- Verified `sacctmgr show qos` now displays the previously-missing limit columns with correct values.
- Verified the new CLI/proto write path end-to-end: creating and modifying limits via `sacctmgr add/modify user`, and confirming a changed value actually takes effect on a subsequent job submission.
- Verified a job specifying memory per-CPU (rather than per-node) is now correctly counted against memory-based limits — previously it silently bypassed them entirely.
- Verified `sacctmgr` now rejects invalid/non-numeric limit input with a clear error, instead of silently treating it as "unlimited".

crates/spur-core/src/qos.rs
```diff
@@ -123,13 +133,11 @@ pub fn check_qos_limits(
...
-            {
-                return QosCheckResult::Blocked(PendingReason::QosGrpMemLimit);
-            }
+        let job_mem = effective_memory_mb(&job.spec, job.spec.num_nodes);
+        if grp.get(TresType::Memory) > 0
+            && qos_running_tres.get(TresType::Memory) + job_mem > grp.get(TresType::Memory)
+        {
+            return QosCheckResult::Blocked(PendingReason::QosGrpMemLimit);
         }
     }
```
shiv-tyagi, line 163: `grp_wall_minutes` is settable, stored, validated, and displayed, but `check_qos_limits()` never reads it, so the limit has no runtime effect. Either enforce it here alongside the other Grp* checks, or add it to the PR's "Known limitations" list so admins aren't misled into thinking it's active. (The description says the proto "already carried" GrpWall, but field 12 is new in this PR.)

crates/spur-core/src/job.rs
```diff
@@ -510,6 +535,20 @@ impl Default for JobSpec {
...
+/// Total memory (MB) a job of `num_nodes` nodes requests, derived from
+/// either an explicit per-node request or a per-CPU request applied across
+/// the job's total CPU count. Falls back to 0 (unconstrained) if neither is
+/// set.
+pub fn effective_memory_mb(spec: &JobSpec, num_nodes: u32) -> u64 {
+    spec.memory_per_node_mb
+        .map(|mem| mem * num_nodes as u64)
+        .or_else(|| {
+            spec.memory_per_cpu_mb
+                .map(|mem| mem * (spec.num_tasks * spec.cpus_per_task) as u64)
```
shiv-tyagi: Nit: `num_tasks * cpus_per_task` multiplies in u32 before the `as u64` cast, so it can overflow before widening. Widen first for consistency with the per-node branch above:
```suggestion
                .map(|mem| mem * spec.num_tasks as u64 * spec.cpus_per_task as u64)
```

shiv-tyagi (commented):
Two non-blocking notes: `grp_wall_minutes` is wired end-to-end but not enforced, and a minor u32 overflow nit in the new memory helper. Core fixes look solid and well-tested.

crates/spur-core/src/qos.rs
yansun1996: You're right on both points. `grp_wall_minutes` is stored/validated/displayed but `check_qos_limits()` never reads it — the check only has instantaneous running-TRES snapshots, whereas GrpWall is cumulative consumed wall-time over the QOS's reset period, which needs a persisted, period-resetting accumulator (parallel to the fairshare machinery) that doesn't exist yet. Faking it against point-in-time data would mislead admins, so that's a follow-up. Documented it as not-enforced at `check_qos_limits` and corrected the description — field 12 is indeed new in this PR, not pre-existing. Fixed in fbebdbf.

crates/spur-core/src/job.rs
yansun1996: Agreed — the u32 product can wrap before the cast, and the per-node branch already widens first. Applied your suggestion in fbebdbf.

shiv-tyagi (approved):

Issue #294, opened 2026-06-15
Fix SPANK ABI: rename spur_spank_* symbols to spank_*


shiv-tyagi:
## Problem

The SPANK plugin host exports functions with the `spur_spank_*` prefix instead of Slurm's `spank_*` prefix. Existing SPANK plugins compiled against Slurm headers call `spank_get_item()`, `spank_setenv()`, etc. They get unresolved symbol errors when loaded into Spur's SPANK host because the exported symbols use the wrong prefix.

For example, `spur_spank_get_item` is exported instead of `spank_get_item`.

## Expected Behavior

Slurm's SPANK API exports symbols with the `spank_` prefix:
- `spank_get_item` — retrieve job context items
- `spank_setenv` — set environment variables
- `spank_getenv` — get environment variables
- `spank_job_control_setenv` — set env in job control context
- `spank_err` / `spank_log` / `spank_debug` — logging functions

Existing SPANK plugins expect these exact symbol names.

## Acceptance Criteria

- [ ] All `spur_spank_*` exports renamed to `spank_*` (e.g., `spur_spank_get_item` → `spank_get_item`)
- [ ] Function names also match Slurm's naming (e.g., `spur_spank_set_var` → `spank_setenv`)
- [ ] Exported symbol `spank_get_item` is resolvable via `dlsym`
- [ ] Existing Slurm SPANK plugins can resolve all required symbols when loaded

biluriuday:
this is addressed by PR https://github.com/ROCm/spur/pull/429

PR #429, opened 2026-07-13
fix(spur-spank): rename SPANK exports to Slurm-compatible symbols

biluriuday:
## Fix SPANK ABI and wire up plugin env/args (#294)

Existing Slurm SPANK plugins failed against Spur's host: the API symbols used a `spur_spank_*` prefix, and even after loading, plugins couldn't affect the job. This makes the SPANK host drop-in compatible with Slurm's `spank.h`.

### What changed

**ABI / symbol compatibility**
- Renamed exports to Slurm's names and aligned signatures to `spank.h`: `spank_get_item`, `spank_setenv` (added the `overwrite` arg), plus new `spank_getenv`, `spank_unsetenv`, `spank_job_control_{set,get,unset}env`, `spank_strerror`, and the `slurm_error/info/verbose/debug/debug2/debug3` + `slurm_spank_log` loggers.
- Fixed `SpankItem` discriminants to match Slurm's `enum spank_item` (previously mis-ordered, so `S_JOB_ID` returned the gid).
- Added `ESPANK_*` return codes; `spank_get_item` now returns `ESPANK_NOT_AVAIL` for known-but-unsupported items vs `ESPANK_BAD_ARG` for unknown ids.
- Made `spurd` export these symbols in its dynamic symbol table so `dlopen`'d plugins resolve them, via linker flags (`--undefined` + `--export-dynamic`, guarded on the target OS) in `spurd/build.rs`.

**Functional wiring**
- `invoke_hook` now threads a caller-owned `SpankHandle`, so plugin env edits persist and are visible across hooks/plugins.
- The executor seeds the handle from the job's env + context, runs `Init`/`TaskInit` after the env is assembled, and folds plugin env changes back into the launched job (both container and command paths).
- Plugstack arguments are now delivered to hooks as `ac`/`argv` (previously always `0`/`NULL`).

### Testing
- Unit/integration tests updated and extended (`spur-spank` + `t58_spank`): item-ID ordering, `setenv` overwrite semantics, `getenv` round-trip/errors, env-in seeding, `NOT_AVAIL` vs `BAD_ARG`.
- `cargo clippy --workspace --exclude spur-ffi --all-targets` clean.
- **End-to-end on the bare-metal cluster**: built a real SPANK plugin (`slurm_spank_task_init`) using `spank_setenv`/`spank_get_item` and a plugstack arg. Confirmed it loads (`SPANK plugins loaded count=1`) and a job sees the injected env (`SPUR_SPANK_TEST=hello_from_spank`, `SPUR_SPANK_ARG=arg_from_plugstack`).

### Known limitations / follow-ups
- Logging functions are non-variadic (stable Rust can't define C-variadic fns), so `printf`-style format specifiers aren't expanded — the format string is logged verbatim.
- `spank_get_item` supports scalar items only (no `S_JOB_ARGV`/`S_JOB_ENV` or pid-to-id lookups).
- Hooks run in the `spurd` process, not the forked/privilege-dropped task.
- `job_control_env` isn't yet wired into prolog/epilog.
- Option-registration symbols (`spank_option_register`, etc.) are not yet provided.

### DIFF
--- a/crates/spur-spank/src/lib.rs
+++ b/crates/spur-spank/src/lib.rs
@@ -4,14 +4,14 @@
 //! SPANK plugin host.
 //!
 //! Loads existing Slurm SPANK plugins (.so files) via dlopen and provides
-//! the spank_* callback API (11 hooks, ~12 API functions).
+//! the `spank_*`/`slurm_*` callback API expected by `spank.h`.
 
 use std::collections::HashMap;
 use std::ffi::CStr;
-use std::os::raw::{c_char, c_int};
+use std::os::raw::{c_char, c_int, c_void};
 use std::path::{Path, PathBuf};
 
-use tracing::{debug, info, warn};
+use tracing::{debug, error, info, warn};
 
 /// SPANK callback hook points (matches Slurm's spank.h).
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
@@ -48,20 +48,44 @@ impl SpankHook {
     }
 }
 
-/// SPANK item IDs for spank_get_item (10 common items).
+/// SPANK item IDs for `spank_get_item`.
+///
+/// Discriminants match Slurm's `enum spank_item` in `spank.h`; plugins pass
+/// these numeric values, so the ordering must stay identical.
 #[repr(C)]
 pub enum SpankItem {
-    JobId = 0,
-    JobUid = 1,
-    JobGid = 2,
+    JobUid = 0,
+    JobGid = 1,
+    JobId = 2,
     JobStepId = 3,
     JobNnodes = 4,
     JobNodeid = 5,
     JobLocalTaskCount = 6,
     JobTotalTaskCount = 7,
-    JobArgv = 8,
-    TaskPid = 9,
+    JobNcpus = 8,
+    JobArgv = 9,
+    JobEnv = 10,
+    TaskId = 11,
+    TaskGlobalId = 12,
+    TaskExitStatus = 13,
+    TaskPid = 14,
 }
+
+/// `spank_err_t` return codes (subset of `spank.h`).
+///
+/// Plugins primarily check for `ESPANK_SUCCESS`; the remaining codes let
+/// callers distinguish common failures.
+pub const ESPANK_SUCCESS: c_int = 0;
+pub const ESPANK_ERROR: c_int = 1;
+pub const ESPANK_BAD_ARG: c_int = 2;
+pub const ESPANK_NOT_TASK: c_int = 3;
+pub const ESPANK_ENV_EXISTS: c_int = 4;
+pub const ESPANK_ENV_NOEXIST: c_int = 5;
+pub const ESPANK_NOSPACE: c_int = 6;
+pub const ESPANK_NOT_REMOTE: c_int = 7;
+pub const ESPANK_NOEXIST: c_int = 8;
+pub const ESPANK_NOT_AVAIL: c_int = 10;
+pub const ESPANK_NOT_LOCAL: c_int = 11;
 
 /// A loaded SPANK plugin.
 struct SpankPlugin {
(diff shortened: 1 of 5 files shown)


crates/spur-spank/src/lib.rs
```diff
@@ -212,111 +240,250 @@ impl SpankHost {
...
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
+pub extern "C" fn spank_getenv(
+    handle: *mut SpankHandle,
+    var: *const c_char,
+    buf: *mut c_char,
+    len: c_int,
+) -> c_int {
+    if handle.is_null() || var.is_null() || buf.is_null() || len < 0 {
+        return ESPANK_BAD_ARG;
+    }
```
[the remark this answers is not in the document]
biluriuday: this mirrors slurm's behaviour

crates/spurd/build.rs
```diff
@@ -0,0 +1,38 @@
...
+        return;
+    }
+
+    // --undefined forces the linker to pull each symbol from the spur-spank rlib
+    // despite having no caller; --export-dynamic then places them in the dynamic
+    // symbol table so plugin dlsym resolves them.
+    for sym in SPANK_SYMBOLS {
+        println!("cargo::rustc-link-arg-bins=-Wl,--undefined={sym}");
+    }
+    println!("cargo::rustc-link-arg-bins=-Wl,--export-dynamic");
```
shiv-tyagi: Q: `--export-dynamic` exports the whole binary's globals into the dynamic symbol table, not just the SPANK API. Since `SPANK_SYMBOLS` is already enumerated, a scoped `--dynamic-list`/version script would export just those symbols and avoid exposing all Rust/std globals (and keep the dynamic table small). `--undefined` still forces inclusion; the dynamic-list only handles visibility. Worth tightening?

crates/spurd/src/agent_server.rs
```diff
@@ -249,10 +249,15 @@ impl AgentService {
                 // Invoke SPANK TaskExit and JobEpilog hooks for completed jobs
                 if let Some(ref spank_host) = *spank {
                     for c in &completed {
-                        if let Err(e) = spank_host.invoke_hook(SpankHook::TaskExit) {
+                        let context = SpankContext {
+                            job_id: c.job_id,
```
shiv-tyagi, line 253: The handle for TaskExit/JobEpilog only sets `job_id`, so `uid`/`gid` stay 0. An epilog plugin calling `spank_get_item(S_JOB_UID/S_JOB_GID)` gets 0 here, while the executor path seeds uid/gid. Intentional, or should completed-job context carry uid/gid too?

crates/spur-tests/src/t58_spank.rs
```diff
@@ -93,18 +94,87 @@ mod tests {
...
-        let result = host.load_plugin(Path::new("/nonexistent/plugin.so"));
+        let result = host.load_plugin(Path::new("/nonexistent/plugin.so"), &[]);
         assert!(result.is_err());
         assert_eq!(host.plugin_count(), 0);
     }
+
+    // -- T58.8: Item IDs match Slurm's enum spank_item --------------------
+
+    #[test]
+    fn t58_8_item_ids_match_slurm() {
```
shiv-tyagi: Nit: this duplicates a subset of `test_spank_item_ids_match_slurm` in `spur-spank/src/lib.rs`, which already asserts the full ordering. Could drop this or defer the ordering assertion to the unit test.

shiv-tyagi (commented):
A few suggestions, all minor, nothing blocking. Symbol/enum ordering against `spank.h` looks correct and the tests are thorough. PTAL.

crates/spur-tests/src/t58_spank.rs
biluriuday: dropped the test

crates/spurd/src/agent_server.rs
biluriuday: yes, added uid/gid to completed-job context

crates/spurd/build.rs
biluriuday: changed the scope to `--dynamic-list` to only export SPANK_SYMBOLS instead of earlier `--export-dynamic`

shiv-tyagi (approved):

PR #430, opened 2026-07-13
fix(spurctld): wire QOS priority into scheduling and preemption

yansun1996:
## What

Fixes SPUR-40 completely: a job's QOS priority was never factored into scheduling/preemption, and a QOS's `preempt_mode` setting was never consulted either (only partition-level `preempt_mode` determined how a running job got preempted).

## Approach

QOS priority is applied as a **weighted-additive** term on top of the existing multiplicative `fairshare * age_factor * partition_tier` pipeline (an earlier draft fed it into the multiplicative base directly, which let unrelated fairshare/age amplify or invert the QOS delta by up to ~20x — caught in review, confirmed numerically, fixed).

QOS `preempt_mode` (Off/Cancel/Requeue/Suspend) now overrides the partition-level preempt mode for a running job when explicitly set to something other than `Off` — letting a QOS assert "jobs under me get preempted this specific way" regardless of partition config. `Off` is treated as "no override" since the wire format can't distinguish "explicitly Off" from "never configured."

Also fixes, found during review:
- `try_preempt`'s priority comparison previously compared a running job's raw stored priority against a pending job's fully-adjusted priority — now recomputed consistently.
- A deterministic tie-break (`job_id`) added to pending-job sorting, since the priority floor can now realistically collapse an entire low-priority QOS tier to the same value.
- A wrap-instead-of-saturate cast bug in the priority floor calculation.
- A redundant/inconsistent QOS cache lookup: the candidate job's QOS was being resolved twice per scheduling pass (once for priority, once for preempt-mode, in two temporally-separate reads that could theoretically observe different cache snapshots) — consolidated into a single resolve-once-per-candidate lookup, closing both a performance issue (was O(pending × running) instead of O(running)) and the snapshot-consistency gap.

## Known limitation, documented not fixed

A QOS explicitly set to `preempt_mode=Off` cannot express "never preempt regardless of partition" — the most protective use case is the one this feature can't express, since the wire format can't distinguish that from "never configured." Fixing this would need a wire-format change (e.g. `optional string` in the proto) — out of scope here, documented in code.

## Testing

- Unit tests including one with non-neutral fairshare/age proving QOS-tier priority ordering holds under realistic conditions.
- Unit tests for QOS preempt-mode override vs. partition fallback, plus regression coverage for the 4 pre-existing partition-only tests.
- An integration test through the real `try_preempt` → cache wiring (not just the pure decision function) confirming a QOS's `Suspend` override actually results in the job being suspended, not cancelled.
- Live-verified on the real 2-node cluster: a high-priority-QOS job correctly preempts a low-priority-QOS job, and a QOS with `preempt_mode=suspend` correctly overrides a partition's `cancel` default (confirmed via SIGSTOP at the agent level, not a kill).

## Review

3 rounds of independent adversarial review + cross-validation, which caught: the multiplicative-amplification bug (required a real redesign, not a patch), the raw-vs-adjusted priority mismatch, the missing tie-break, and the redundant/inconsistent QOS lookup described above.

### DIFF
--- a/crates/spur-core/src/qos.rs
+++ b/crates/spur-core/src/qos.rs
@@ -115,10 +115,16 @@ pub fn check_qos_limits(
     QosCheckResult::Allowed
 }
 
-/// Calculate effective priority including QOS priority adjustment.
+/// Add a QOS's priority delta on top of an already-computed priority.
+///
+/// Callers should pass the fully weighted priority (fairshare/age/partition
+/// tier already applied) as `base_priority`, not a job's raw submitted
+/// priority: the QOS delta is a flat addition, so applying it before those
+/// multiplicative factors would let them amplify or dilute it in proportion
+/// to values that have nothing to do with QOS.
 pub fn qos_adjusted_priority(base_priority: u32, qos: &Qos) -> u32 {
     let adjusted = base_priority as i64 + qos.priority as i64;
-    adjusted.max(1) as u32
+    adjusted.clamp(1, u32::MAX as i64) as u32
 }
 
 #[cfg(test)]
--- a/crates/spur-core/src/qos.rs
+++ b/crates/spur-core/src/qos.rs
@@ -27,6 +27,12 @@ impl From<QosPreemptMode> for PreemptMode {
 /// legitimate explicit choice — there is no way to tell them apart — so
 /// `Off` is treated as "no override" here, deferring to whatever the
 /// caller would otherwise resolve (e.g. partition-level `PreemptMode`).
+///
+/// This means a QOS can't express "never preempt this job regardless of
+/// partition" — the most protective use case is exactly the one this
+/// override can't reach. Doing so would need a wire-format change (e.g.
+/// `optional string preempt_mode` in the proto) to distinguish "explicitly
+/// Off" from "unset"; out of scope here.
 pub fn qos_preempt_override(qos: &Qos) -> Option<PreemptMode> {
     match qos.preempt_mode {
         QosPreemptMode::Off => None,
--- a/crates/spur-core/src/qos.rs
+++ b/crates/spur-core/src/qos.rs
@@ -5,8 +5,34 @@
 //!
 //! Checks per-QOS limits before allowing a job to be scheduled.
 
-use crate::accounting::{Qos, TresRecord, TresType};
+use crate::accounting::{Qos, QosPreemptMode, TresRecord, TresType};
 use crate::job::{Job, PendingReason};
+use crate::partition::PreemptMode;
+
+impl From<QosPreemptMode> for PreemptMode {
+    fn from(mode: QosPreemptMode) -> Self {
+        match mode {
+            QosPreemptMode::Off => PreemptMode::Off,
+            QosPreemptMode::Cancel => PreemptMode::Cancel,
+            QosPreemptMode::Requeue => PreemptMode::Requeue,
+            QosPreemptMode::Suspend => PreemptMode::Suspend,
+        }
+    }
+}
+
+/// A QOS-level preempt mode override for a running job, or `None` if the
+/// QOS carries no override.
+///
+/// `QosPreemptMode::Off` is both the wire format's "unset" default and a
+/// legitimate explicit choice — there is no way to tell them apart — so
+/// `Off` is treated as "no override" here, deferring to whatever the
+/// caller would otherwise resolve (e.g. partition-level `PreemptMode`).
+pub fn qos_preempt_override(qos: &Qos) -> Option<PreemptMode> {
+    match qos.preempt_mode {
+        QosPreemptMode::Off => None,
+        other => Some(other.into()),
+    }
+}
 
 /// Result of QOS limit check.
 #[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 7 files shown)


yansun1996:
## Live testing performed

Verified on a real 2-node cluster (not simulated/mocked), across three rounds as the design was refined:

- Verified a job under a higher-priority QOS correctly preempts a job under a lower-priority QOS holding the same resource.
- Verified a QOS-level preemption-mode override (e.g. suspend) correctly takes effect instead of the partition's default preemption behavior — confirmed at the process level (suspend signal, not a kill) on the affected node.
- Verified a job specifying multiple candidate partitions correctly picks up the highest-priority tier among them, instead of incorrectly defaulting to the lowest tier (a bug caught by review and fixed in this branch).

crates/spurctld/src/cluster.rs
```diff
@@ -1655,19 +1663,26 @@ impl ClusterManager {
             let fair_share = self
                 .fairshare_cache
                 .get(&job.spec.user, job.spec.account.as_deref().unwrap_or(""));
-            job.priority = spur_sched::priority::effective_priority(
+            job.priority = compute_effective_priority(
                 job.priority,
                 fair_share,
                 age_minutes,
                 partition_tier,
+                &qos_by_job[&job.job_id],
```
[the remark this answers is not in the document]
yansun1996: Confirmed real via independent cross-validation. `job_preempt_mode`'s `partition_for` closure already correctly split `job.spec.partition` on `,` and matched each name against `partitions` — but this call site (and `current_effective_priority_with_qos`) did a naive single-name lookup, silently falling back to tier 1 for any multi-partition job. Fixed by extracting the split/match/max-by-tier logic once into `spur_core::partition::matched_partitions()`/`max_priority_tier()`, and switching both flagged call sites (plus `job_preempt_mode` and `try_preempt`'s own `partition_for`, which had been duplicating the same logic independently) to use it. Added regression tests confirming a job spanning multiple partitions with different tiers correctly resolves to the highest tier. Fixed in f9c72e3, live-verified on a real cluster.

crates/spurctld/src/cluster.rs
```diff
@@ -2656,13 +2671,45 @@ impl ClusterManager {
...
+        let age_minutes = (now - job.submit_time).num_minutes().max(0);
+        let partition_tier = {
+            let partitions = self.partitions.read();
+            job.spec
+                .partition
+                .as_ref()
+                .and_then(|pname| partitions.iter().find(|p| p.name == *pname))
+                .map(|p| p.priority_tier)
+                .unwrap_or(1)
+        };
```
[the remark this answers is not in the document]
yansun1996: Same root cause as the sibling comment above — fixed together in f9c72e3 via the shared `matched_partitions()`/`max_priority_tier()` helper. `current_effective_priority_with_qos()` now correctly resolves the highest tier among a job's comma-separated partition list instead of defaulting to 1.

crates/spur-core/src/qos.rs
```diff
@@ -115,10 +136,11 @@ pub fn check_qos_limits(
     QosCheckResult::Allowed
 }
 
-/// Calculate effective priority including QOS priority adjustment.
+/// Add a QOS's flat priority delta on top of an already fairshare/age/tier
+/// weighted priority; applying it earlier would let those factors amplify it.
 pub fn qos_adjusted_priority(base_priority: u32, qos: &Qos) -> u32 {
     let adjusted = base_priority as i64 + qos.priority as i64;
-    adjusted.max(1) as u32
+    adjusted.clamp(1, u32::MAX as i64) as u32
```
shiv-tyagi, line 143: The floor case has a regression test, but the new upper clamp doesn't. `base_priority` (u32) + `qos.priority` (i32) can exceed `u32::MAX`, which is exactly the wrap-instead-of-saturate bug this line fixes. Add a test proving it saturates:

```rust
#[test]
fn qos_adjusted_priority_saturates_instead_of_wrapping() {
    let qos = Qos { priority: i32::MAX, ..Default::default() };
    assert_eq!(qos_adjusted_priority(u32::MAX, &qos), u32::MAX);
}
```

crates/spurctld/src/cluster.rs
```diff
@@ -2656,13 +2664,28 @@ impl ClusterManager {
...
+        let age_minutes = (now - job.submit_time).num_minutes().max(0);
+        let partition_tier = {
+            let partitions = self.partitions.read();
+            spur_core::partition::max_priority_tier(job.spec.partition.as_deref(), &partitions)
+        };
+        let fair_share = self
+            .fairshare_cache
+            .get(&job.spec.user, job.spec.account.as_deref().unwrap_or(""));
+        compute_effective_priority(job.priority, fair_share, age_minutes, partition_tier, qos)
+    }
```
shiv-tyagi: `current_effective_priority_with_qos` re-acquires `self.partitions.read()` once per running job. `try_preempt` calls it in a loop while already holding a `partitions: &[Partition]` slice, and uses that same slice for `partition_for` and `job_preempt_mode`. So a running job's priority tier is read from a different snapshot than the preempt-mode decision made against it (a config reload between the two reads would diverge them), plus it's N lock acquisitions per pass. Take partitions as a parameter so the whole pass uses one snapshot:

```rust
pub(crate) fn current_effective_priority_with_qos(
    &self,
    job: &Job,
    qos: &Qos,
    partitions: &[Partition],
) -> u32 {
    let now = Utc::now();
    let age_minutes = (now - job.submit_time).num_minutes().max(0);
    let partition_tier =
        spur_core::partition::max_priority_tier(job.spec.partition.as_deref(), partitions);
    let fair_share = self
        .fairshare_cache
        .get(&job.spec.user, job.spec.account.as_deref().unwrap_or(""));
    compute_effective_priority(job.priority, fair_share, age_minutes, partition_tier, qos)
}
```

crates/spur-core/src/qos.rs
yansun1996: Confirmed and added — the function already correctly saturates (computes in i64, clamps to [1, u32::MAX] before the final cast), it just had no regression test for the upper bound. Added `qos_adjusted_priority_saturates_instead_of_wrapping` using exactly the test you suggested. Fixed in 77b6b74.

crates/spurctld/src/cluster.rs
yansun1996: Confirmed — the snapshot-divergence concern is currently latent (no runtime partition-mutation path exists outside tests today), but the redundant lock acquisition (N per scheduling pass instead of 1) is real regardless. Applied your suggested fix: `current_effective_priority_with_qos` now takes `partitions: &[Partition]` as a parameter instead of acquiring the lock internally, using the existing `max_priority_tier` helper. Updated the one caller (`try_preempt`, which already had the slice in scope) and the existing unit test call site. Fixed in 77b6b74.

shiv-tyagi (approved):

PR #431, opened 2026-07-13
fix(spurctld): stop counting preemption requeues against the failure-requeue hold limit

yansun1996:
## What

Fixes SPUR-41: a job could get permanently stuck in Pending (`JobHoldMaxRequeue`) purely from being preempted repeatedly — a normal, expected occurrence in a busy multi-tenant cluster — because preemption-driven requeues and genuine failure/timeout-driven requeues shared a single `requeue_count` counter and cap.

## Approach

Split the counter into two independently tracked fields: `requeue_count` (failure/timeout/dispatch-failure requeues, capped at `max_batch_requeue` as before) and a new `preempt_requeue_count` (preemption-driven requeues, intentionally uncapped — matching real Slurm semantics, where repeated preemption is rate-limited, not count-limited). Refactored the shared job run-state reset logic into one helper used by both paths, replacing what was previously duplicated inline.

## Testing
- Regression tests proving both halves: a job preempted well past the old shared cap is never held, and a job that genuinely times out/fails repeatedly is still held at the same limit as before (no change to the original intended failure-path behavior).
- Verified WAL/Raft backward compatibility for the new persisted field (old-log replay defaults correctly).
- Live-verified on the real 2-node cluster: a job preempted 5 times in a row (exceeding the configured `max_batch_requeue`) was never held, and completed normally once nothing preempted it.

## Review
2 independent adversarial reviews + cross-validation pass — both came back clean, no blockers.

### DIFF
--- a/crates/spur-core/src/job.rs
+++ b/crates/spur-core/src/job.rs
@@ -544,9 +544,14 @@ pub struct Job {
     #[serde(default)]
     pub derived_exit_code: i32,
 
-    /// Number of times this job has been requeued.
+    /// Number of times this job has been requeued after a dispatch failure
+    /// or Timeout/NodeFail. Capped by `max_batch_requeue`.
     #[serde(default)]
     pub requeue_count: u32,
+    /// Number of times this job has been requeued after preemption; tracked
+    /// separately since it isn't a failure signal and never counts toward `max_batch_requeue`.
+    #[serde(default)]
+    pub preempt_requeue_count: u32,
 
     // Heterogeneous job support
     /// Links het job components to the first component's job ID.
@@ -607,6 +612,7 @@ impl Job {
             exit_signal: 0,
             derived_exit_code: 0,
             requeue_count: 0,
+            preempt_requeue_count: 0,
             het_job_id: None,
             het_group: None,
             node_completions: HashMap::new(),
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -284,8 +284,8 @@ pub struct ControllerConfig {
     #[serde(default)]
     pub heartbeat_timeout_secs: Option<u64>,
 
-    /// Maximum automatic requeues before a job is held with `JobHoldMaxRequeue`.
-    /// Configured in TOML as `[controller] max_batch_requeue` (default: 5).
+    /// Maximum automatic requeues (excluding preemption) before a job is held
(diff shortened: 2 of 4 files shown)


yansun1996:
## Live testing performed

Verified on a real 2-node cluster (not simulated/mocked):

- Repeatedly preempted the same job multiple times in a row — deliberately exceeding the configured failure-requeue hold limit purely through preemption cycling (no actual failures). Confirmed the job was never permanently held; it kept cycling back to pending and eventually ran to completion once nothing preempted it further.

crates/spur-core/src/job.rs
```diff
@@ -544,9 +544,14 @@ pub struct Job {
...
 
-    /// Number of times this job has been requeued.
+    /// Number of times this job has been requeued after a dispatch failure
+    /// or Timeout/NodeFail. Capped by `max_batch_requeue`.
     #[serde(default)]
     pub requeue_count: u32,
+    /// Number of times this job has been requeued after preemption; tracked
+    /// separately since it isn't a failure signal and never counts toward `max_batch_requeue`.
+    #[serde(default)]
+    pub preempt_requeue_count: u32,
```
shiv-tyagi, line 554: Q: `preempt_requeue_count` is write-only. It's incremented in `reset_job_for_preempt_requeue` and persisted, but nothing ever reads it, not the scheduler, and it isn't surfaced via proto/REST/CLI/metrics (unlike `requeue_count`, which at least gates the hold check). Is it kept purely as future observability / to assert separation in tests? If so that's fine, just confirming it's intentionally internal-only for now and not a half-wired feature.

yansun1996: Confirmed by tracing every reference — you're right, it's write-only in production code today (only ever incremented, never read outside the two unit tests that assert the counter-split behavior). But it's not a half-wired feature: it's consistent with the *existing* status quo — `requeue_count` (its pre-existing sibling) isn't proto/REST/CLI/metrics-visible either; `scontrol show job` doesn't surface either counter today. The field exists so tests can assert preemption requeues no longer erode the failure-requeue budget, and needs to be a real persisted `u32` (not a local/ephemeral value) since it must survive WAL replay across a restart mid-preemption-storm. Agreed this would be a reasonable follow-up (surfacing both counters via `scontrol show job` for chronic-preemption diagnostics — arguably exactly the operator-facing signal this whole PR is motivated by), but bundling it into this hold-limit bugfix would be scope creep. Happy to file a tracking issue for it if useful.

shiv-tyagi (approved):

PR #432, opened 2026-07-13
feat(cluster): SPUR-native k0s clusters via spur k8s up/down + mesh-native CNI

powderluv:
## SPUR-native k0s clusters

SPUR provisions and **owns** a Kubernetes (k0s) cluster across its own nodes. This is the inverse of the existing `[kubernetes]` integration (SPUR running *inside* an external k8s): here SPUR is the cluster's lifecycle owner, driven end-to-end by the existing spurctld/spurd control plane.

```
spur k8s up      # assign roles + mesh IPs + pod CIDRs, install pinned k0s,
                 # bring up the control plane, mint tokens, join workers,
                 # generate the k0s config + program the WireGuard mesh (calico)
spur k8s status  # phase + per-node component state
spur k8s kubeconfig > admin.conf
spur k8s down --reset   # tear every node down (k0s reset), mesh preserved
```

k0s runs under a **spurd-owned systemd unit** (never a SPUR job), so it survives spurd restarts; spurd re-adopts a running unit on startup.

### What's in it

| Crate | Adds |
|-------|------|
| **spur-core** | `[cluster]` / `[cluster.arc]` config; shared k0s types (role/phase/join-token/cluster-state); pinned k0s version + `k0s_controller_config_yaml` generator |
| **spur-proto** | SlurmController cluster RPCs (Up/Down/Status/Kubeconfig); SlurmAgent component RPCs (Start/Stop/GetStatus, CreateK0sJoinToken, GetAdminKubeconfig, ApplyMesh) |
| **spurctld** | leader-gated k0s reconcile loop: AddressPool IPAM (mesh IP + pod /24 per node), deterministic control-plane selection, agent fan-out, and **durable** role/token/CIDR/phase state in the raft state machine (survives failover) |
| **spurd** | RPC-driven `K0sAgent` that owns the k0s unit (write token/config, render/reconcile/heal, re-adopt on restart), installs a pinned k0s if missing, mints tokens / reads the admin kubeconfig on the control plane, applies the WireGuard mesh, writes the GPU CDI spec on join |
| **spur-cli** | `spur k8s up\|down\|status\|kubeconfig\|install-k0s`, `spur net mesh` |

### Highlights

- **k0s packaging** — downloads a pinned (or `--latest`) k0s release, SHA-256 verified, atomic install. Bare-metal nodes need no manual k0s prestaging (`spur k8s install-k0s`, and auto-install on first `up`).
- **Mesh-native CNI** — `[cluster] cni = "calico"` generates a k0s config advertising the API on the control-plane's mesh IP with Calico in **bird** mode, and sets each worker's kubelet `--node-ip` to its mesh IP, so pods route over the WireGuard mesh (native routing, no overlay). The default `kuberouter` behavior is unchanged.
- **Mesh reconcile** — the controller pushes the authoritative full-mesh membership (`ApplyMesh`), **level-triggered** so node-local drift, failed pushes, and controller failover all self-heal; the agent prunes departed peers.

### Testing

- Full workspace unit tests pass; no hardcoded infrastructure IPs (test/example addresses use RFC 5737 TEST-NET).
- **HW-validated on a real 5-node cluster** (1 control-plane + 4 GPU workers):
  - `spur k8s up` → 4 workers `Ready` on the pinned k0s `v1.36.2`; token-mint join with retry
  - `ApplyMesh` reconcile (verified a removed peer is re-added on the next tick)
  - **Calico bird BGP full-mesh Established over the WireGuard mesh** (nodes' INTERNAL-IP = mesh IP)
  - `spur k8s down --reset` tears every node down (`k0s reset`, `spur0` preserved); full **up → down → up** lifecycle

### Notes

- Default CNI (`kuberouter`) and non-`[cluster]` behavior are unchanged; the whole path is gated on `[cluster].enabled`.
- Squashed to a single commit for review.

Draft for review.

### DIFF
--- a/crates/spur-cli/src/k8s.rs
+++ b/crates/spur-cli/src/k8s.rs
@@ -35,9 +35,6 @@ pub enum K8sCommand {
         /// Control-plane node (default: picked from inventory / [cluster] config).
         #[arg(long)]
         control_plane_node: Option<String>,
-        /// Also install the ARC CI runner scale sets.
-        #[arg(long)]
-        install_arc: bool,
     },
     /// Tear the k0s cluster down.
     Down {
@@ -72,10 +69,7 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
     let parsed = K8sArgs::try_parse_from(args)?;
     let controller = parsed.controller;
     match parsed.command {
-        K8sCommand::Up {
-            control_plane_node,
-            install_arc,
-        } => cmd_up(&controller, control_plane_node, install_arc).await,
+        K8sCommand::Up { control_plane_node } => cmd_up(&controller, control_plane_node).await,
         K8sCommand::Down { reset } => cmd_down(&controller, reset).await,
         K8sCommand::Status => cmd_status(&controller).await,
         K8sCommand::Kubeconfig => cmd_kubeconfig(&controller).await,
@@ -105,17 +99,10 @@ async fn cmd_install_k0s(version: &str, path: &str, force: bool) -> Result<()> {
     Ok(())
 }
 
-async fn cmd_up(
-    controller: &str,
-    control_plane_node: Option<String>,
-    install_arc: bool,
-) -> Result<()> {
+async fn cmd_up(controller: &str, control_plane_node: Option<String>) -> Result<()> {
     let mut client = SlurmControllerClient::new(spur_client::connect_channel(controller).await?);
     let resp = client
-        .cluster_up(ClusterUpRequest {
-            control_plane_node,
-            install_arc,
-        })
+        .cluster_up(ClusterUpRequest { control_plane_node })
         .await?
         .into_inner();
     if resp.accepted {
@@ -179,21 +166,11 @@ mod tests {
 
     #[test]
     fn parses_up_with_control_plane() {
-        let args = K8sArgs::try_parse_from([
-            "k8s",
-            "up",
-            "--control-plane-node",
-            "head-node",
-            "--install-arc",
-        ])
-        .unwrap();
+        let args =
+            K8sArgs::try_parse_from(["k8s", "up", "--control-plane-node", "head-node"]).unwrap();
         match args.command {
-            K8sCommand::Up {
-                control_plane_node,
-                install_arc,
-            } => {
+            K8sCommand::Up { control_plane_node } => {
                 assert_eq!(control_plane_node.as_deref(), Some("head-node"));
-                assert!(install_arc);
             }
             _ => panic!("wrong command"),
         }
--- a/crates/spur-cli/src/k8s.rs
+++ b/crates/spur-cli/src/k8s.rs
@@ -1,0 +1,215 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! `spur k8s` subcommands: drive the SPUR-managed k0s cluster.
+
+use anyhow::Result;
+use clap::{Parser, Subcommand};
+
+use spur_proto::proto::slurm_controller_client::SlurmControllerClient;
+use spur_proto::proto::{
+    ClusterDownRequest, ClusterKubeconfigRequest, ClusterStatusRequest, ClusterUpRequest,
+};
+
+/// Manage the SPUR-provisioned k0s cluster.
+#[derive(Parser, Debug)]
+#[command(name = "k8s", about = "Manage the SPUR-provisioned k0s cluster")]
+pub struct K8sArgs {
+    /// Controller address
+    #[arg(
+        long,
+        env = "SPUR_CONTROLLER_ADDR",
+        default_value = "http://localhost:6817",
+        global = true
+    )]
+    controller: String,
+
+    #[command(subcommand)]
+    pub command: K8sCommand,
+}
+
+#[derive(Subcommand, Debug)]
+pub enum K8sCommand {
+    /// Bring the k0s cluster up (assign roles/IPs, then start each node's component).
+    Up {
+        /// Control-plane node (default: picked from inventory / [cluster] config).
+        #[arg(long)]
+        control_plane_node: Option<String>,
+        /// Also install the ARC CI runner scale sets.
+        #[arg(long)]
+        install_arc: bool,
+    },
+    /// Tear the k0s cluster down.
+    Down {
+        /// Also `k0s reset` each node (destructive: wipes cluster state).
+        #[arg(long)]
+        reset: bool,
+    },
+    /// Show cluster phase + per-node component status.
+    Status,
+    /// Print the admin kubeconfig to stdout.
+    Kubeconfig,
+    /// Download + install the k0s binary on THIS node (local; no controller needed).
+    /// Run as root for the default /usr/local/bin path.
+    InstallK0s {
+        /// k0s release tag to install, or "latest". Defaults to spur's pinned version.
+        #[arg(long, default_value_t = String::from(spur_core::k0s::K0S_PINNED_VERSION))]
+        version: String,
+        /// Install path for the k0s binary.
+        #[arg(long, default_value_t = String::from(spur_core::k0s::K0S_DEFAULT_BINARY))]
+        path: String,
+        /// Reinstall even if a k0s binary already exists at --path.
+        #[arg(long)]
+        force: bool,
+    },
+}
+
+pub async fn main() -> Result<()> {
+    main_with_args(std::env::args().collect()).await
+}
+
+pub async fn main_with_args(args: Vec<String>) -> Result<()> {
+    let parsed = K8sArgs::try_parse_from(args)?;
+    let controller = parsed.controller;
+    match parsed.command {
+        K8sCommand::Up {
+            control_plane_node,
+            install_arc,
+        } => cmd_up(&controller, control_plane_node, install_arc).await,
+        K8sCommand::Down { reset } => cmd_down(&controller, reset).await,
+        K8sCommand::Status => cmd_status(&controller).await,
+        K8sCommand::Kubeconfig => cmd_kubeconfig(&controller).await,
+        K8sCommand::InstallK0s {
+            version,
+            path,
+            force,
+        } => cmd_install_k0s(&version, &path, force).await,
+    }
+}
+
+async fn cmd_install_k0s(version: &str, path: &str, force: bool) -> Result<()> {
+    let dest = std::path::Path::new(path);
+    if dest.exists() && !force {
+        eprintln!("k0s already present at {path} (use --force to reinstall)");
+        return Ok(());
+    }
+    eprintln!("Installing k0s {version} -> {path} ...");
+    let info = spur_update::k0s::install_k0s(version, dest).await?;
+    let short = &info.sha256[..info.sha256.len().min(16)];
+    eprintln!(
+        "Installed k0s {} to {} (sha256 {}…)",
+        info.version,
+        info.path.display(),
+        short
+    );
+    Ok(())
+}
+
+async fn cmd_up(
+    controller: &str,
+    control_plane_node: Option<String>,
+    install_arc: bool,
+) -> Result<()> {
+    let mut client = SlurmControllerClient::new(spur_client::connect_channel(controller).await?);
+    let resp = client
+        .cluster_up(ClusterUpRequest {
+            control_plane_node,
+            install_arc,
+        })
+        .await?
+        .into_inner();
+    if resp.accepted {
+        println!("k0s cluster up requested: {}", resp.message);
+    } else {
+        eprintln!("k0s cluster up NOT accepted: {}", resp.message);
+    }
+    for n in resp.nodes {
+        println!("  {} [{}] {}", n.node, n.role, n.component_state);
+    }
+    Ok(())
+}
+
+async fn cmd_down(controller: &str, reset: bool) -> Result<()> {
+    let mut client = SlurmControllerClient::new(spur_client::connect_channel(controller).await?);
+    let resp = client
+        .cluster_down(ClusterDownRequest { reset })
+        .await?
+        .into_inner();
+    if resp.accepted {
+        println!("k0s cluster down requested: {}", resp.message);
+    } else {
+        eprintln!("k0s cluster down NOT accepted: {}", resp.message);
+    }
+    Ok(())
+}
+
+async fn cmd_status(controller: &str) -> Result<()> {
+    let mut client = SlurmControllerClient::new(spur_client::connect_channel(controller).await?);
+    let resp = client
+        .cluster_status(ClusterStatusRequest {})
+        .await?
+        .into_inner();
+    println!("phase: {}", resp.phase);
+    if !resp.control_plane_node.is_empty() {
+        println!("control-plane: {}", resp.control_plane_node);
+    }
+    for n in resp.nodes {
+        println!(
+            "  {:<24} {:<11} {:<11} enabled={}",
+            n.node, n.role, n.component_state, n.enabled
+        );
+    }
+    Ok(())
+}
+
+async fn cmd_kubeconfig(controller: &str) -> Result<()> {
+    let mut client = SlurmControllerClient::new(spur_client::connect_channel(controller).await?);
+    let resp = client
+        .cluster_kubeconfig(ClusterKubeconfigRequest {})
+        .await?
+        .into_inner();
+    // stdout = data (the YAML), so it can be redirected to a kubeconfig file.
+    print!("{}", resp.kubeconfig);
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn parses_up_with_control_plane() {
+        let args = K8sArgs::try_parse_from([
+            "k8s",
+            "up",
+            "--control-plane-node",
+            "head-node",
+            "--install-arc",
+        ])
+        .unwrap();
+        match args.command {
+            K8sCommand::Up {
+                control_plane_node,
+                install_arc,
+            } => {
+                assert_eq!(control_plane_node.as_deref(), Some("head-node"));
+                assert!(install_arc);
+            }
+            _ => panic!("wrong command"),
+        }
+    }
+
+    #[test]
+    fn parses_down_reset_and_status() {
+        let args = K8sArgs::try_parse_from(["k8s", "down", "--reset"]).unwrap();
+        assert!(matches!(args.command, K8sCommand::Down { reset: true }));
+        let args = K8sArgs::try_parse_from(["k8s", "status"]).unwrap();
+        assert!(matches!(args.command, K8sCommand::Status));
+    }
+
+    #[test]
+    fn controller_defaults_and_env() {
+        let args = K8sArgs::try_parse_from(["k8s", "status"]).unwrap();
+        assert_eq!(args.controller, "http://localhost:6817");
+    }
+}
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -660,6 +660,15 @@ pub struct ClusterConfig {
     /// worker's kubelet `--node-ip` to its mesh IP.
     #[serde(default = "default_cni")]
     pub cni: String,
+    /// Storage provisioner SPUR ships so PVC workloads work out of the box (k0s bundles none).
+    /// "local-path" (default — RWO node-local, set as the default StorageClass) or "none" (bring your
+    /// own). Applied via k0s's manifest deployer on the control-plane node.
+    #[serde(default = "default_storage_provisioner")]
+    pub storage_provisioner: String,
+    /// On-node directory the local-path provisioner stores PersistentVolumes in. Point this at a
+    /// large scratch disk if PVCs will hold much data — the default lives under `/var/lib` (root fs).
+    #[serde(default = "default_local_path_dir")]
+    pub local_path_dir: String,
 }
 
 fn default_cluster_distro() -> String {
@@ -683,6 +692,12 @@ fn default_service_cidr() -> String {
 fn default_cni_mtu() -> u16 {
     1450
 }
+fn default_storage_provisioner() -> String {
+    "local-path".into()
+}
+fn default_local_path_dir() -> String {
+    crate::k0s::DEFAULT_LOCAL_PATH_DIR.into()
+}
 
 impl Default for ClusterConfig {
     fn default() -> Self {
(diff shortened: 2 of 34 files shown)


crates/spurctld/src/cluster_k8s.rs
```diff
@@ -0,0 +1,688 @@
...
+            pool.allocate()?
+        };
+        let ordinal = next_free_ordinal(&used_ordinals);
+        used_ordinals.insert(ordinal);
+        let pod_cidr = carve_pod_cidr(&net.pod_cidr, ordinal)?;
+        cluster.assign_node_k0s(&node.name, role, &mesh_ip.to_string(), &pod_cidr)?;
+    }
+
+    // Persist the control-plane choice if not already recorded.
+    if state.control_plane_node.as_deref() != Some(cp_node.as_str()) {
```
yansun1996, line 206: **[CRITICAL] Control-plane IP/role collision across a mid-provision failover**

The CP node is re-selected every tick (`state.control_plane_node → config → lexically-first`), but each `assign_node_k0s` is its own raft write while this CP-choice persist happens as a separate write *after* the loop. With the `if node.k0s_role.is_some() { continue }` guard and the CP branch assigning `mesh_ip = controller_ip` directly (bypassing the pool's uniqueness check), there's a window:

1. `assign(A = Controller, .1)` commits; crash before this CP-choice write lands.
2. On restart `state.control_plane_node` is still `None`; a lexically-earlier node registers and is selected as CP next tick — assigned `.1`/Controller too, while `A` is skipped but keeps `.1`/Controller.

Result: two nodes share mesh IP `.1` and both are Controller, silently, surviving failover.

Fix: derive `cp_node` from an already-assigned Controller/Single node before falling back to config/lexical order, or persist the CP choice atomically with the first assignment.

crates/spurd/src/cluster.rs
```diff
@@ -0,0 +1,824 @@
...
+                ),
+            }
+            return; // at most one k0s unit runs per node
+        }
+    }
+
+    /// Start (or update) this node's k0s component. Writes the join token (0600) + k0s config to
+    /// files, then reconciles the spurd-owned unit. Returns the resulting systemd active-state.
+    /// The plaintext `join_token` is never logged.
+    pub async fn start(
```
yansun1996, line 507: **[MAJOR] `[cluster].enabled` not enforced on the agent side**

The doc comment for `enabled` states "When false, spurd never touches systemd/k0s," but `K0sAgent::from_config` doesn't retain `enabled`, and `start_cluster_component` → `start()` installs the k0s binary, writes/enables the unit, and starts k0s regardless. (The `enabled` refs in spurd are all systemd `is-enabled`, not `cfg.enabled`.)

So a `StartClusterComponent` RPC can provision a node where the operator disabled the feature. Fix: short-circuit `start()`/`start_cluster_component` with `failed_precondition` when `!enabled` — or relax the doc comment if gating is meant to live only on the controller.

crates/spur-core/src/config.rs
```diff
@@ -603,6 +607,184 @@ impl Default for KubernetesConfig {
...
+            installation_name: default_arc_installation_name(),
+            github_config_url: String::new(),
+            container_mode: String::new(),
+            runner_image: default_arc_runner_image(),
+        }
+    }
+}
+
+/// Basic dependency-free CIDR sanity check (`<ip>/<prefix>`), used by config validation.
+fn is_valid_cidr(s: &str) -> bool {
```
yansun1996, line 721: **[MAJOR] `is_valid_cidr` is more permissive than the reconcile path can provision**

`is_valid_cidr` accepts IPv6 and any prefix up to `/128`, but `carve_pod_cidr` (`cluster_k8s.rs`) requires IPv4 with prefix `<= 24`. So `pod_cidr = "10.42.0.0/26"` (or any IPv6 CIDR) passes `validate()` at startup, then fails every reconcile tick with only a `warn!` — cluster stuck in `Provisioning` indefinitely, no surfaced error.

Two related gaps for the same pass:
- `network.wg_cidr` isn't validated here at all — a malformed mesh CIDR bypasses `validate()` and fails later in `AddressPool::new` every tick (same silent stuck state).
- `cidrs_overlap` returns `false` for non-IPv4 input, so an IPv6 pod/service CIDR skips the overlap check.

Fix: one IPv4-asserting pass over `wg_cidr`/`pod_cidr`/`service_cidr` when `cluster.enabled` (assert IPv4, pod prefix `<= 24`, mutual non-overlap).

crates/spur-cli/src/k8s.rs
```diff
@@ -0,0 +1,215 @@
...
+#[derive(Subcommand, Debug)]
+pub enum K8sCommand {
+    /// Bring the k0s cluster up (assign roles/IPs, then start each node's component).
+    Up {
+        /// Control-plane node (default: picked from inventory / [cluster] config).
+        #[arg(long)]
+        control_plane_node: Option<String>,
+        /// Also install the ARC CI runner scale sets.
+        #[arg(long)]
+        install_arc: bool,
```
yansun1996: **[MAJOR] `--install-arc` / `[cluster.arc]` advertised but not wired up**

`--install-arc` is surfaced on the CLI and sent as `ClusterUpRequest.install_arc`, but the `cluster_up` handler (`spurctld/src/server.rs`) never reads `req.install_arc`, and the `[cluster.arc]` fields (`github_config_url`, `runner_image`, `installation_name`, `container_mode`) have no consumer outside `config.rs`. Yet `validate()` *requires* `github_config_url` when `arc.enabled` — so users get an enforced config section and a flag that do nothing.

Fix: reject `install_arc == true` with a "not yet implemented" error (and/or gate the config validation) so the surface matches behavior, until ARC lands.

powderluv:
Thanks for the review — all 10 comments addressed in 2e10c02:

- **write_secret_file (token perms):** create the file `0600` from the start (was write-then-chmod) — no world-readable window.
- **create_dir_all ignored:** now propagated with context.
- **stop / supervise hold the mutex across IO:** `ClusterSupervisor` is now `Clone`; `stop`/`status`/`supervise` snapshot it out from under the `active` lock and do the systemctl/k0s IO **without** holding the lock, so concurrent RPCs aren't serialized.
- **mesh sort by string:** sort numerically by `Ipv4Addr` (no more `.10` before `.2` / spurious per-tick diffs).
- **live_node_statuses `enabled: true`:** report the agent's real `enabled` bit (new `fetch_component_status`).
- **spawn_apply_mesh no timeout:** connect + RPC now bounded by `AGENT_TIMEOUT`.
- **install_arc ignored:** not implemented yet — `cluster_up` now warns and says so in the response instead of a silent no-op.
- **is_valid_cidr accepts IPv6:** restricted to IPv4 (the provisioning paths are all `Ipv4Addr`).
- **mesh endpoint = mesh IP:** the controller no longer derives the WG endpoint from `Node.address` (which is the mesh IP when WG is up). It's left empty so `apply_mesh` **preserves** the underlay tunnel `spur net join` established; membership reconciliation only maintains peers + `AllowedIPs`.

Also added an admin+user how-to guide (`docs/deployment/managed-kubernetes.rst`) in 76c0e18. Full workspace tests pass.

crates/spurctld/src/cluster.rs
```diff
@@ -1374,6 +1377,72 @@ impl ClusterManager {
...
+            reset_requested,
+        })?;
+        info!(?phase, "k0s cluster phase set");
+        Ok(())
+    }
+
+    /// record minted join-token METADATA (never the plaintext secret) via Raft.
+    /// Consumer is the deferred worker join-token mint on the control-plane agent.
+    #[allow(dead_code)]
+    pub fn create_k0s_join_token(
```
shiv-tyagi: This whole join-token metadata path is dead in production. `create_k0s_join_token`/`revoke_k0s_join_token` (both `#[allow(dead_code)]`), `K0sClusterState.join_tokens`, the `K0sJoinTokenCreate`/`K0sJoinTokenRevoke` WAL variants, and the snapshot field are only exercised by tests. The real join path (`mint_worker_token` -> agent `CreateK0sJoinToken`) returns the plaintext token and never records metadata.

This is replicated, snapshot-restored Raft state that nothing writes, which is the most expensive kind of dead code: it reads as load-bearing. Spur is pre-1.0 with no back-compat constraint, so drop it and re-add alongside the consumer when token lifecycle/revocation is actually built.

crates/spur-core/src/config.rs
```diff
@@ -603,6 +607,186 @@ impl Default for KubernetesConfig {
...
+    /// Pod network CIDR. Calico bird native routing over the mesh carves per-node /24s from this.
+    #[serde(default = "default_pod_cidr")]
+    pub pod_cidr: String,
+    /// Service network CIDR.
+    #[serde(default = "default_service_cidr")]
+    pub service_cidr: String,
+    /// CNI MTU. Defaults to 1450 to leave headroom for WireGuard's ~50-byte overhead over the mesh;
+    /// set explicitly if your underlay differs.
+    #[serde(default = "default_cni_mtu")]
+    pub cni_mtu: u16,
```
shiv-tyagi, line 646: `cni_mtu` (and `flannel_iface`, `gpu_worker_selector`, `device_plugin` just below) are never read anywhere in the tree. This is a public config surface that does nothing: operators will set it and expect an effect.

`cni_mtu` is also a correctness gap. The doc says it exists for WireGuard's ~50-byte overhead, but `k0s_controller_config_yaml` emits no MTU, so Calico pods run at 1500 over a ~1450 mesh (silent fragmentation / perf loss on the exact `cni="calico"` path this PR headlines).

Wire `cni_mtu` into the generated k0s config at minimum, or drop these four fields from the struct, `examples/spur.conf`, and the docs until they are used.

crates/spur-core/src/config.rs
```diff
@@ -603,6 +607,186 @@ impl Default for KubernetesConfig {
...
+    /// Node label selector that marks GPU workers.
+    #[serde(default = "default_gpu_worker_selector")]
+    pub gpu_worker_selector: String,
+    /// GPU device plugin advertising `amd.com/gpu` ("rocm" = ROCm/k8s-device-plugin;
+    /// "spur" = native spur-device-plugin, a later milestone).
+    #[serde(default = "default_device_plugin")]
+    pub device_plugin: String,
+    /// ARC (actions-runner-controller) settings for CI runner scale sets.
+    #[serde(default)]
+    pub arc: ClusterArcConfig,
```
shiv-tyagi: The `--install-arc` fix (warn + note in the response) handles the flag, but the entire `[cluster.arc]` block (`installation_name`, `github_config_url`, `container_mode`, `runner_image`) is still dead config that only validates itself. Nothing installs ARC.

Same use-it-or-lose-it call: drop `[cluster.arc]` and `--install-arc` and land them with the actual ARC installer, instead of shipping config that does nothing.

crates/spurd/src/main.rs
```diff
@@ -184,6 +185,10 @@ async fn main() -> anyhow::Result<()> {
         })
         .collect();
 
+    // this node's WireGuard public key (best-effort) so the controller can reconcile mesh peers.
+    let wg_iface = std::env::var("SPUR_WG_INTERFACE").unwrap_or_else(|_| "spur0".into());
+    let wg_pubkey = spur_net::wireguard::interface_public_key(&wg_iface).unwrap_or_default();
```
shiv-tyagi: `wg_pubkey` is read once here at startup and reported at registration; it is never refreshed. `build_mesh_membership` skips any node with an empty pubkey, so if the WireGuard mesh (`spur net join`) comes up after spurd has started, or `spur0` is recreated, the controller never learns this node's key and silently excludes it from `ApplyMesh` until spurd restarts.

Re-read the pubkey on each heartbeat (or when it flips from empty to set) so mesh membership self-heals without a spurd restart.

crates/spurctld/src/cluster_k8s.rs
```diff
@@ -0,0 +1,708 @@
...
+            continue;
+        }
+        all_active = false;
+        // For a native-routing CNI, pin the worker's kubelet node-ip to its mesh IP.
+        let node_ip = if net.cni == "calico" {
+            node.k0s_mesh_ip.clone()
+        } else {
+            None
+        };
+        match mint_worker_token(cluster).await {
```
shiv-tyagi: `mint_worker_token` runs on every 30s reconcile tick for any worker not yet `active`, so during the join window each tick mints a fresh k0s token and rewrites the token file. That churns server-side tokens and races the join.

Mint once and cache it (keyed by node) until the worker reports `active`, or gate the mint on an actual join attempt.

crates/spurd/src/cluster.rs
```diff
@@ -0,0 +1,844 @@
...
+
+    /// `systemctl disable --now <unit>` — stop + remove the enable symlink. Idempotent (ignores an
+    /// absent unit).
+    async fn disable_unit(&self) {
+        let unit = self.unit_file_name();
+        let _ = self.systemctl(&["disable", "--now", &unit]).await;
+    }
+
+    /// `k0s reset` (destructive host-wide cleanup). Best-effort — logs, never fails the caller.
+    async fn k0s_reset(&self) {
```
shiv-tyagi: `k0s reset` failures are logged and swallowed here, and `stop` / `down --reset` return `Ok` regardless. So `spur k8s down --reset` reports success even when the reset failed on a node, leaving stale k0s state that a later `up` inherits.

Surface the failure (aggregate per-node errors) so the CLI can report a partial teardown instead of a false success.

crates/spurctld/src/cluster_k8s.rs
```diff
@@ -0,0 +1,708 @@
...
+    let pod_base = cidr_base(&net.pod_cidr)?;
+    let mut used_ordinals: HashSet<u32> = nodes
+        .iter()
+        .filter_map(|n| n.k0s_pod_cidr.as_deref())
+        .filter_map(|c| pod_ordinal(c, pod_base))
+        .collect();
+
+    let single = nodes.len() == 1;
+    for node in &nodes {
+        if node.k0s_role.is_some() {
```
shiv-tyagi, line 181: `provision_assignments` skips any node that already has a role, but `cluster_up` overwrites `state.control_plane_node` from the request. So running `spur k8s up --control-plane-node X` after the cluster was already provisioned with a different control plane updates the recorded CP without reassigning roles: the old controller stays `Controller` and the new node comes up as a `Worker`, giving an inconsistent topology.

Either reject a CP change while assignments exist, or re-derive roles when the CP node changes.

crates/spurd/src/cluster.rs
```diff
@@ -0,0 +1,844 @@
...
+             Documentation=https://docs.k0sproject.io\n\
+             ConditionFileIsExecutable={bin}\n\
+             After=network-online.target\n\
+             Wants=network-online.target\n\
+             \n\
+             [Service]\n\
+             ExecStart={exec}\n\
+             Restart=always\n\
+             RestartSec=10\n\
+             StartLimitIntervalSec=0\n\
```
shiv-tyagi: Nit: `StartLimitIntervalSec=0` disables systemd's restart rate-limit entirely, so a crash-looping k0s restarts every `RestartSec=10` forever with no ceiling, and the 30s supervisor adds no backoff either. Consider a finite `StartLimitIntervalSec`/`StartLimitBurst` so a hard-failing node backs off instead of hot-looping.

crates/spurctld/src/cluster_k8s.rs
```diff
@@ -0,0 +1,708 @@
...
+            K0sPhase::Degraded => warn!("k0s cluster degraded"),
+        }
+    }
+}
+
+/// Assign role + mesh IP + pod /24 to any node that lacks one. Idempotent: a node's persisted
+/// `k0s_role`/`k0s_mesh_ip`/`k0s_pod_cidr` IS the allocation record, so assigned nodes are skipped
+/// and never re-allocated. The (in-memory) AddressPool is re-seeded from persisted inventory on
+/// every call — skipping that would hand out IPs already in use after a controller restart.
+pub(crate) fn provision_assignments(
```
shiv-tyagi, line 127: Nit: mesh IPs and pod /24s are allocated but never released when a node leaves inventory, so the `AddressPool` and pod-ordinal set leak across node churn. Fine for a fixed cluster; worth a follow-up (free on node removal) if nodes come and go.

shiv-tyagi (changes requested):
Posted some comments. PTAL

powderluv:
Thanks @shiv-tyagi — all nine addressed in efea9c0. Verified green (`cargo check --workspace` + full test suite) and deployed + validated on our 5-node cluster (up → `down --reset` → up lifecycle, each fix confirmed on hardware).

**Drops (use-it-or-lose-it):**
- **[1] join-token metadata** — removed the `K0sJoinTokenRecord` / `K0sClusterState.join_tokens` raft state, the `K0sJoinTokenCreate/Revoke` WAL variants + apply arms, and the `create/revoke_k0s_join_token` mutators. The real join path (agent `CreateK0sJoinToken` → plaintext token) is untouched; the metadata layer comes back with a real consumer (revocation/audit).
- **[3] `[cluster.arc]` + `--install-arc`** — removed the config block, the CLI flag, and the proto field (`reserved 2`). spur/k0s doesn't install ARC.
- **[2, part]** — dropped `flannel_iface` / `gpu_worker_selector` / `device_plugin` (no reader; the worker's `--node-ip` already pins its mesh IP, so `flannel_iface` was redundant).

**Fixes:**
- **[2] `cni_mtu`** — wired into the generated Calico config (`spec.network.calico.mtu`); it was dead and left pods at MTU 1500 over the ~1450 mesh. Confirmed `mtu: 1450` in the emitted `k0s.yaml` on the calico path.
- **[4] wg_pubkey** — the reporter re-reads the interface key every heartbeat (new `HeartbeatRequest.wg_pubkey`); the controller learns an empty→set / changed key and includes the node in ApplyMesh without a spurd restart.
- **[5] worker-token churn** — the reconcile loop caches the minted token per node and reuses it until the worker is active (mint once, not every tick).
- **[6] swallowed reset failure** — `k0s reset` now returns a `Result`, surfaced through `stop`/`stop_untracked` → the agent reports `stopped=false` + message → **the controller now inspects that in-band response** (it was only checking transport errors — the actual swallow). `down --reset` no longer reports a false success.
- **[7] CP-change topology** — `cluster_up` rejects `--control-plane-node X` once roles are assigned (`FailedPrecondition`, pointing at `down --reset` to re-elect).
- **[8] restart rate-limit** — the unit sets `StartLimitIntervalSec=300` / `StartLimitBurst=5`; the reconcile supervisor still re-enables it.

**[9]** — the mesh IP + pod /24 are already reclaimed on node removal: `provision_assignments` rebuilds the `AddressPool` + pod-ordinal set from live inventory every tick, so a removed node's allocation simply isn't re-marked. A `down`-but-present node keeps its assignment by design (it may return). Happy to add an explicit release hook if you'd prefer belt-and-suspenders.

yansun1996:
Thanks for the quick turnaround here — this is looking mostly good. Most of the major review comments have been addressed:

- **`--install-arc` / `[cluster.arc]`** — fully removed (proto field reserved). ✅
- **CIDR validation** — `is_valid_cidr` is now IPv4-only. ✅
- **Token file perms, `create_dir_all` propagation, mutex-across-IO, mesh sort, `apply_mesh` timeout, `cni_mtu`, token churn, `k0s reset` failure surfacing, restart rate-limit** — all addressed. ✅

CI is fully green (build, clippy, fmt, test, coverage, deny, Trivy, SPDX, and both E2E suites).

A couple of things before this is good to merge:

1. **Please rebase — the PR currently conflicts with `main`.** There's a real conflict in `crates/spurd/src/main.rs`: `main` added memlock rlimit handling to the `AgentService` constructor (#449) while this PR swapped in `with_cluster_config(...)`. The merged constructor needs both the cluster config and the memlock limit.

2. **Two open items from the earlier review** (fine as follow-ups if you'd prefer, just flagging they're still open):
   - **`[cluster].enabled` isn't enforced on the agent** — `K0sAgent` has no `enabled` field and `start_cluster_component` runs unconditionally, so the config doc's "when false, spurd never touches systemd/k0s" guarantee doesn't hold.
   - **Control-plane assignment crash window** — in `provision_assignments`, the Controller/`.1` assignment and the CP-choice persist are two separate Raft writes; a crash between them combined with lexical re-election can still produce a duplicate `.1`/Controller. The config-edit trigger is now guarded, but this path remains.

Overall really nice work — just the rebase + a call on the two items above.

yansun1996 (dismissed):
Approving — all review comments have been addressed. Thanks for iterating on this.

One thing before merge: the branch currently conflicts with `main` in `crates/spurd/src/main.rs` and `docs/deployment/index.rst`, so it'll need a rebase to be mergeable.

powderluv:
Rebased onto `main` (was 24 commits behind + conflicting → now `MERGEABLE`) and addressed the outstanding review feedback. CI is green across all checks, including both `E2E / k8s` and `E2E / native-host`.

## Addressed in this round (the open threads)

@yansun1996 — the three CRITICAL/MAJOR items:

- **[CRITICAL] Control-plane IP/role collision across a mid-provision failover** — `provision_assignments` now derives the control-plane node from an **already-assigned `Controller`/`Single` node first**, before the recorded choice / config override / lexical fallback. Since the persisted role assignment is the durable source of truth, a crash between the first `assign(CP, .1)` and the separate CP-choice persist can no longer let a restart pick a different (lexically-earlier) node as CP and hand it the same `.1`/`Controller` — the two-controllers-sharing-`.1` window is closed.
- **[MAJOR] `[cluster].enabled` not enforced agent-side** — `K0sAgent` now retains `enabled` and `start()` refuses to install/write/enable/start k0s when it's false (surfaced as `started=false` with a clear message), honoring the documented "spurd never touches systemd/k0s" contract even if the controller sends a `StartClusterComponent` RPC.
- **[MAJOR] `is_valid_cidr` more permissive than the reconcile path can provision** — `validate()` now (when `cluster.enabled`) asserts `network.wg_cidr` is a valid IPv4 CIDR (it feeds `AddressPool`), the pod CIDR prefix is `<= /24` (`carve_pod_cidr` carves per-node /24s), and mesh/pod/service are mutually non-overlapping. No more configs that pass `validate()` at startup and then fail every reconcile tick with only a `warn!` (silently stuck in `Provisioning`).

## Already addressed in earlier rounds (why several threads show as outdated)

- `--install-arc` / `[cluster.arc]` dead config — the request field and the config block were removed entirely.
- `spawn_apply_mesh` now has a per-agent timeout; `write_secret_file` creates the token `0600` atomically (no write-then-chmod window); `create_dir_all` errors are handled; `StartLimitIntervalSec` is `300` (not `0`); `is_valid_cidr` is IPv4-only; `cni_mtu` is wired into the generated Calico config; and `cluster_up` already rejects a control-plane change once roles are assigned.

The `cluster_k8s.rs` address-pool/pod-ordinal leak on node removal is noted as a follow-up (harmless for a fixed cluster; free-on-removal can come later).

Verified locally on Linux before pushing: full workspace `cargo check`, all unit tests (spur-core / spurctld / spurd / spur-k8s), `cargo fmt --check`, and `cargo clippy --workspace -D warnings`. Ready for another look 🙏

crates/spurd/src/cluster.rs
powderluv: This should be addressed now.

yansun1996 (approved):
LGTM

PR #437, opened 2026-07-14
feat(spur-cli): add -V/--version support across all binaries and CLI entry points

yansun1996:
## What

Fixes SPUR-61: none of Spur's Slurm-compatible CLI tools (`squeue`, `sinfo`, `sbatch`, `srun`, `scontrol`, etc.) or daemons (`spurctld`, `spurd`) supported `-V`/`--version`, unlike every real Slurm binary (`squeue -V` → `slurm 25.11.6`). Only the native `spur` command itself had it.

## Approach

- Added a `crates/spur-core/build.rs` that captures `git describe --always --dirty` at compile time, exposed via `spur_core::version::version_string()`. A clean, exact release build shows the plain `spur 0.4.1` (matching Slurm's format); a build with commits past the last release tag or an uncommitted change shows `spur 0.4.1 (v0.4.1-12-gabc1234[-dirty])`, so nightly/dirty/in-development builds are unambiguously distinguishable from a real release, without any CI/workflow redesign.
- Wired `-V`/`--version` into every Slurm-compatible entry point (the multi-call `spur` binary and all its symlinks) via a single early check scoped to the first CLI argument, plus both daemons.
- Daemon startup logs now include the same richer version string, not just the bare semver, per the ticket's third requirement.

## Bugs found and fixed during review (before this ever reached the point of being reviewable as "done")

- **A naive first draft of the version-flag check scanned the entire argument list**, which silently hijacked commands like `srun python -V` or `spur exec <id> python -V` — spur would print its own version and exit instead of ever running the wrapped command. Fixed by scoping the check to only the first argument.
- **Docker-built nightly/release binaries would never have shown any git metadata at all** — `.dockerignore` excludes `.git` from the build context entirely, so `build.rs` had no way to compute a git descriptor inside a container build. Fixed by threading the descriptor through as a Docker build-arg, computed on the CI runner (which has full git history) and passed to the Docker build step in both `nightly.yml` and `release.yml`.
- **The `--dirty` flag could go stale** on ordinary incremental rebuilds, since the build script was only re-triggered by `.git` metadata changes, not working-tree file edits. Fixed by making the build script always rerun (a standard idiom for this class of tool, matching `vergen`/`git-version`).

## Testing

- Unit tests for the version-string formatting logic across all cases (clean tag, commits-past-tag, dirty, no-git-available).
- Live-verified on the real 2-node cluster: daemon `--version` flags, daemon startup log output, CLI symlinks (`squeue`, `sinfo`, `sbatch`, `scontrol`), the argv-hijack regression fix (confirmed `srun echo hello -V` actually submits and runs the job, passing `-V` through as a literal argument), and a completely ordinary job submission to confirm nothing else broke.
- Manually verified the Docker build-arg wiring with a real `docker build --target dist --build-arg SPUR_GIT_DESCRIBE=...` run.

## Review

2 rounds of independent adversarial review + cross-validation, which caught all 3 bugs described above before this was mergeable.

### DIFF
--- a/crates/spur-cli/src/main.rs
+++ b/crates/spur-cli/src/main.rs
@@ -65,6 +65,14 @@ fn main() -> anyhow::Result<()> {
         let _ = signal(Signal::SIGPIPE, SigHandler::SigDfl);
     }
 
+    // Applies to every entry point (native `spur` and all Slurm-compatible
+    // symlinks alike), ahead of per-subcommand clap parsing that doesn't
+    // otherwise register -V/--version.
+    if std::env::args().any(|a| a == "-V" || a == "--version") {
+        println!("{}", spur_core::version::version_string());
+        std::process::exit(0);
+    }
+
     load_controller_addr_from_config();
 
     // Multi-call binary: dispatch based on argv[0] (symlink name).
@@ -197,7 +205,7 @@ fn main() -> anyhow::Result<()> {
 
     match args[1].as_str() {
         "version" | "--version" | "-V" => {
-            println!("spur {}", env!("CARGO_PKG_VERSION"));
+            println!("{}", spur_core::version::version_string());
             if args.len() > 2 && args[2] == "--check" {
                 runtime.block_on(async {
                     print!("Checking for updates... ");
--- a/crates/spur-cli/src/main.rs
+++ b/crates/spur-cli/src/main.rs
@@ -70,7 +70,7 @@ fn main() -> anyhow::Result<()> {
     // otherwise register -V/--version. Only the first argument is checked so
     // that trailing args forwarded to a user program (srun, exec, scontrol
     // update, sacctmgr, ...) can't be mistaken for this flag.
-    if matches!(std::env::args().nth(1).as_deref(), Some("-V" | "--version")) {
+    if matches!(std::env::args_os().nth(1).as_deref(), Some(a) if a == "-V" || a == "--version") {
         println!("{}", spur_core::version::version_string());
         std::process::exit(0);
     }
--- a/crates/spur-cli/src/main.rs
+++ b/crates/spur-cli/src/main.rs
@@ -69,8 +69,12 @@ fn main() -> anyhow::Result<()> {
     // symlinks alike), ahead of per-subcommand clap parsing that doesn't
     // otherwise register -V/--version. Only the first argument is checked so
     // that trailing args forwarded to a user program (srun, exec, scontrol
-    // update, sacctmgr, ...) can't be mistaken for this flag.
-    if matches!(std::env::args_os().nth(1).as_deref(), Some(a) if a == "-V" || a == "--version") {
+    // update, sacctmgr, ...) can't be mistaken for this flag. Requiring no
+    // further arguments lets `spur -V --check`/`spur --version --check` fall
+    // through to the native dispatch below, where `--check` is handled.
+    if std::env::args_os().len() <= 2
+        && matches!(std::env::args_os().nth(1).as_deref(), Some(a) if a == "-V" || a == "--version")
+    {
         println!("{}", spur_core::version::version_string());
         std::process::exit(0);
     }
--- a/crates/spur-cli/src/main.rs
+++ b/crates/spur-cli/src/main.rs
@@ -67,8 +67,10 @@ fn main() -> anyhow::Result<()> {
 
     // Applies to every entry point (native `spur` and all Slurm-compatible
     // symlinks alike), ahead of per-subcommand clap parsing that doesn't
-    // otherwise register -V/--version.
-    if std::env::args().any(|a| a == "-V" || a == "--version") {
+    // otherwise register -V/--version. Only the first argument is checked so
+    // that trailing args forwarded to a user program (srun, exec, scontrol
+    // update, sacctmgr, ...) can't be mistaken for this flag.
+    if matches!(std::env::args().nth(1).as_deref(), Some("-V" | "--version")) {
         println!("{}", spur_core::version::version_string());
         std::process::exit(0);
     }
--- a/crates/spur-core/build.rs
+++ b/crates/spur-core/build.rs
@@ -1,51 +1,8 @@
 // Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::path::{Path, PathBuf};
+use std::path::Path;
 use std::process::Command;
-
-/// Walk up from `start` looking for a `.git` entry (directory for a normal
-/// clone, file for a worktree checkout).
-fn find_git_path(start: &Path) -> Option<PathBuf> {
-    let mut dir = start;
-    loop {
-        let candidate = dir.join(".git");
-        if candidate.exists() {
-            return Some(candidate);
-        }
-        dir = dir.parent()?;
-    }
-}
-
-/// Paths whose mtime should invalidate the cached git descriptor: the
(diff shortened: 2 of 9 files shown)


crates/spurd/src/main.rs
```diff
@@ -79,6 +79,11 @@ struct Args {
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
+    if matches!(std::env::args().nth(1).as_deref(), Some("-V" | "--version")) {
+        println!("{}", spur_core::version::version_string());
+        return Ok(());
+    }
```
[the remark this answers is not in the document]
yansun1996: Confirmed real, and verified live — `spurd -f /home/vm/spur/etc/spur.conf -V` did fail with clap's "unexpected argument" error before this fix. Also independently reproduced the `args()` panic-on-non-UTF-8 claim with a raw byte injection test. Since `spurd::Args` has no trailing-arg-forwarding field (unlike spur-cli's `srun`/`exec`, which forward arbitrary args to a wrapped user command), it's safe to scan the full argv here — no position could ever legitimately hold `-V` as a value. Fixed in 64a08fc: switched to `args_os().skip(1).any(...)`, both fixing the position limitation and avoiding the panic. Live re-verified: `spurd -f <config> -V` now prints the version, and `spurd -V` (position 1) still works.

crates/spurctld/src/main.rs
```diff
@@ -62,6 +62,11 @@ struct Args {
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
+    if matches!(std::env::args().nth(1).as_deref(), Some("-V" | "--version")) {
+        println!("{}", spur_core::version::version_string());
+        return Ok(());
+    }
```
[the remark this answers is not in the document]
yansun1996: Same finding and same fix as the sibling comment on spurd — confirmed real, same reasoning applies (no trailing-arg-forwarding field in `spurctld::Args`, so full-argv scanning is safe). Fixed in 64a08fc. Also applied the `args_os()` fix to spur-cli's own version check for consistency, while deliberately keeping its argv[1]-only scoping unchanged (that scoping is a separate, already-verified fix for a different bug — a full-argv scan there would hijack commands like `srun <program> -V`). All live re-verified, including a regression check that the srun-hijack fix still holds.

crates/spur-cli/src/main.rs
```diff
@@ -65,6 +65,16 @@ fn main() -> anyhow::Result<()> {
...
 
+    // Applies to every entry point (native `spur` and all Slurm-compatible
+    // symlinks alike), ahead of per-subcommand clap parsing that doesn't
+    // otherwise register -V/--version. Only the first argument is checked so
+    // that trailing args forwarded to a user program (srun, exec, scontrol
+    // update, sacctmgr, ...) can't be mistaken for this flag.
+    if matches!(std::env::args_os().nth(1).as_deref(), Some(a) if a == "-V" || a == "--version") {
+        println!("{}", spur_core::version::version_string());
+        std::process::exit(0);
+    }
```
shiv-tyagi: This early exit runs before the native `version` handler, so `spur --version --check` and `spur -V --check` now print the version and exit without checking for updates. Pre-PR both reached the `--check` branch. Only `spur version --check` (no dashes) still works.

Either exclude the native `spur` entry point when `--check` follows, or accept the regression (in which case see the note on L209).

crates/spur-cli/src/main.rs
```diff
@@ -197,7 +207,7 @@ fn main() -> anyhow::Result<()> {
 
     match args[1].as_str() {
         "version" | "--version" | "-V" => {
```
shiv-tyagi, line 213: With the early exit on L73, `args[1]` can never be `"--version"` or `"-V"` here, so those two patterns are dead. If that regression is intentional, drop them:

```suggestion
        "version" => {
```

If `--check` should keep working for all forms instead, the early exit needs to defer to this arm.

crates/spur-core/build.rs
```diff
@@ -0,0 +1,35 @@
...
+        .current_dir(manifest_dir)
+        .output()
+        .ok()
+        .filter(|output| output.status.success())
+        .and_then(|output| String::from_utf8(output.stdout).ok())
+        .map(|s| s.trim().to_string())
+        .unwrap_or_default()
+}
+
+// Not unit tested; verified manually and in CI.
```
shiv-tyagi: This narrates test status rather than a non-obvious "why". Git history and the PR already record how it was validated.

```suggestion
```

crates/spurctld/src/main.rs
```diff
@@ -62,6 +62,14 @@ struct Args {
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
+    if std::env::args_os()
+        .skip(1)
+        .any(|a| a == "-V" || a == "--version")
+    {
+        println!("{}", spur_core::version::version_string());
+        return Ok(());
+    }
```
shiv-tyagi, line 71: The daemons scan the whole argv with `.skip(1).any(...)` while the CLI deliberately checks only `nth(1)` to avoid hijacking forwarded args. Harmless today since daemons don't forward args, but the two entry points reading the same flag differently invites confusion. Consider matching the CLI's `nth(1)` approach.

shiv-tyagi (commented):
A few things from review. Mostly minor, plus one behavior regression worth a decision.

crates/spur-cli/src/main.rs
yansun1996: Confirmed real — you're right, and this was actually flagged by my own earlier review pass too, but the fix I applied then didn't actually address it (I mistakenly reasoned that the argv[1] scoping alone would restore --check, which it doesn't since the early exit calls process::exit(0) unconditionally). Fixed properly now: the early exit additionally requires `args_os().len() <= 2` (i.e. -V/--version must be the *only* argument), so `spur -V --check`/`spur --version --check` fall through to the native dispatch's existing --check handling instead of being swallowed. Verified: `spur -V --check` now prints the version and attempts the update check. Fixed in 55effa5.

crates/spur-cli/src/main.rs
yansun1996: Resolved as a byproduct of the fix on the sibling comment (55effa5) — with the early exit now deferring when a trailing argument follows -V/--version, this arm is reachable again (verified: `spur -V --check` reaches this exact match arm and runs the --check logic). Not dead code anymore, no change needed here.

crates/spur-core/build.rs
yansun1996: Agreed, removed — that comment was narrating test/validation status rather than explaining non-obvious behavior. Fixed in 55effa5.

crates/spurctld/src/main.rs
yansun1996: This is intentional, not an oversight — the two entry points need different scoping precisely because they have different hazards. `spurd`/`spurctld` have no trailing-argument-forwarding fields (no subcommand forwards arbitrary args to a wrapped program), so a full-argv scan is safe and was specifically required to fix a real bug: `spurd -f <config> -V` needs to work with `-V` appearing after another flag, not just in position 1 (this was Copilot's finding on this same PR, cross-validated and fixed earlier). `spur-cli`, by contrast, has `srun`/`exec`/`scontrol update`/`sacctmgr` forwarding trailing args to a wrapped command, so it MUST stay scoped to argv[1] only, or it hijacks things like `srun python -V` (a real bug from an earlier round). Matching the CLI's nth(1) approach on the daemons would reintroduce the bug Copilot's comment fixed. Not changing this, but happy to add a short cross-referencing comment on both sites if that would help future readers avoid re-litigating it.

shiv-tyagi (approved):

PR #438, opened 2026-07-14
fix(spur-net): honor requested image architecture

### LINKED ISSUE #343: fix(spur-net): `--arch` flag accepted but silently ignored during registry image import
## Problem

`spur image import --arch arm64 ubuntu:22.04` prints `(arch: arm64)` but always pulls amd64 layers. The `arch` parameter in `cmd_import()` is only used in a log line — it is never passed to `spur_net::pull_image()`, which has no arch parameter. `resolve_manifest_list()` in `oci.rs` hardcodes `linux/amd64`.

On arm64 hosts, users get x86_64 binaries and jobs fail with "Exec format error". For arm64-only multi-arch images, the pull fails with "no linux/amd64 manifest found" even when `--arch arm64` is specified.

## Expected Behavior

- `pull_image()` and `pull_and_extract()` accept an `arch` parameter
- `resolve_manifest_list()` uses the requested architecture instead of hardcoding amd64
- Default arch is detected from the host (`std::env::consts::ARCH`)

## Acceptance Criteria

- [ ] `--arch` value is threaded through to `resolve_manifest_list()`
- [ ] `--arch arm64` pulls arm64 layers from multi-arch images
- [ ] Default arch matches the host architecture
- [ ] Error message on missing arch variant includes the requested architecture

hnotshe:
Closes #343.

Thread the requested image architecture through the native OCI pull path and use it when resolving multi-architecture manifest lists. Host architecture names are normalized to their OCI equivalents, and missing variants now report the requested architecture.

Imported squashfs artifacts record their architecture so an existing image for another architecture is rebuilt instead of silently reused. Replacement is staged so a failed pull does not destroy the existing artifact.

Tested with workspace Clippy and the full locked test suite on Linux.


shiv-tyagi (dismissed):
LGTM

hnotshe:
@shiv-tyagi  Thanks for the review! All checks are green -- could I please get some help merging?

yansun1996:
Solid change. The staged-replacement design is careful — same-filesystem rename, the existing artifact is left intact on a failed pull, and sidecar-less artifacts are safely rebuilt once — and the new pure functions are well-tested and network-free. No blocking issues; notes inline.

Two items are pre-existing (not regressions from this PR), flagged just for context: the layer-extraction gate `media_type.contains("gzip") || digest.starts_with("sha256:")` effectively forces gzip on every layer, so a non-gzip/zstd layer would silently produce a partial rootfs; and arm v6/v7 variants aren't disambiguated since `Platform` doesn't deserialize `variant`.

crates/spur-net/src/oci.rs
```diff
@@ -165,8 +194,20 @@ pub async fn pull_image(image: &str, output_dir: &Path) -> anyhow::Result<PathBu
         }
     }
 
-    // Clean up temp dir
+    let finalize_result = (|| -> anyhow::Result<()> {
+        std::fs::write(&staged_arch_path, oci_architecture(arch))?;
+        std::fs::rename(&staged_sqsh_path, &sqsh_path)
```
yansun1996, line 199: The install is two separate renames (`.sqsh` then `.sqsh.arch`). A crash between them leaves the new payload with a stale-or-missing arch sidecar. It is self-healing (the next pull rebuilds), but since the sidecar is the source of truth for the "rebuild on arch change" decision, writing the metadata first and renaming the payload in last (payload as the commit point) would make "sqsh present implies arch recorded" hold.

crates/spur-net/src/oci.rs
```diff
@@ -165,8 +194,20 @@ pub async fn pull_image(image: &str, output_dir: &Path) -> anyhow::Result<PathBu
         }
     }
 
-    // Clean up temp dir
+    let finalize_result = (|| -> anyhow::Result<()> {
+        std::fs::write(&staged_arch_path, oci_architecture(arch))?;
```
yansun1996, line 198: The sidecar is written as the requested arch unconditionally. On the single-manifest path (a registry returning an image manifest directly rather than an index), no platform verification runs, so the recorded arch could be wrong and defeat the rebuild-on-arch-change guard. Verifying the fetched manifest's platform (or documenting the limitation) would close this.

crates/spurd/src/container.rs
```diff
@@ -1109,7 +1109,7 @@ pub fn container_init(
 /// umoci, or enroot. Only needs mksquashfs (squashfs-tools).
 pub async fn import_image(uri: &str) -> anyhow::Result<PathBuf> {
     let dir = image_dir();
-    spur_net::pull_image(uri, &dir).await
+    spur_net::pull_image(uri, &dir, std::env::consts::ARCH).await
```
yansun1996, line 1112: Requested arch is honored on the pull path, but `import_image` here hardcodes host arch and `resolve_image` matches purely by filename (never reading the `.sqsh.arch` sidecar) — so a cross-arch artifact could be launched on the wrong host unchecked. Likely out of scope for this PR's title, but worth a follow-up; the sidecar added here is the enabling primitive for that check.

crates/spur-net/src/oci.rs
```diff
@@ -101,31 +119,42 @@ pub fn parse_image_ref(image: &str) -> ImageRef {
...
+            return Ok(sqsh_path);
+        }
+        info!(path = %sqsh_path.display(), architecture = arch, "replacing image for requested architecture");
     }
 
     std::fs::create_dir_all(output_dir)?;
 
     // Create temp directory for rootfs assembly
     let tmp_dir = output_dir.join(format!(".pulling_{}", sanitized));
+    if tmp_dir.exists() {
```
yansun1996, line 149: Minor: the `.pulling_<name>` temp dir is deterministic per image, so two concurrent pulls of the same image race — and this unconditional `remove_dir_all` can wipe an in-flight pull's working tree. A PID/uuid suffix would make it safe. Low likelihood for interactive imports.

crates/spur-net/src/oci.rs
```diff
@@ -515,50 +561,51 @@ async fn get_auth_token(
...
-    #[derive(Deserialize)]
-    struct Platform {
-        architecture: String,
-        os: String,
+fn oci_architecture(arch: &str) -> &str {
+    match arch {
+        "x86_64" => "amd64",
+        "aarch64" => "arm64",
+        "x86" => "386",
+        arch => arch,
```
yansun1996, line 569: Minor: the normalization fallthrough passes unknown arches through unchanged; combined with `Platform` not deserializing `variant`, 32-bit arm (v6/v7) cannot be disambiguated and the first `arm` entry wins. Fine for the amd64/arm64 common case — flagging for completeness.

crates/spur-net/src/oci.rs
```diff
@@ -136,7 +165,7 @@ pub async fn pull_image(image: &str, output_dir: &Path) -> anyhow::Result<PathBu
     let mksquashfs_result = std::process::Command::new("mksquashfs")
         .args([
             rootfs_dir.to_str().unwrap(),
-            sqsh_path.to_str().unwrap(),
+            staged_sqsh_path.to_str().unwrap(),
```
yansun1996, line 168: Nit: `to_str().unwrap()` in a library crate trips the no-unwrap-in-lib guideline (AGENTS.md). Pre-existing, but since this line is in the diff, a `.context("non-UTF-8 image path")?` would be a cheap tidy-up.

shiv-tyagi:
@hnotshe Can you please address Yan's comments?

yansun1996 (changes requested):

crates/spur-net/src/oci.rs
hnotshe: Cleaned this up by passing the filesystem paths directly to `Command`, removing both UTF-8 conversions and the library `unwrap()` calls.

crates/spur-net/src/oci.rs
hnotshe: Fixed by validating direct manifests through their OCI config blob before layer extraction. The config `os` and normalized `architecture` must match the requested Linux platform before the sidecar can be recorded, with regression coverage for matching and mismatched platforms.

crates/spur-net/src/oci.rs
hnotshe: The shared staging directory is gone. Each pull now uses a UUID-suffixed directory, and same-image pulls are serialized through the cache recheck and final publication, so concurrent imports cannot delete or interleave one another's artifacts.

crates/spur-net/src/oci.rs
hnotshe: Good catch. First-time installs now publish the sidecar before exposing the payload. Replacements invalidate the old sidecar before publishing the new payload, then publish the new sidecar last, so an interruption forces a cache miss rather than pairing a payload with stale metadata. Both paths have filesystem-level regression coverage.

crates/spur-net/src/oci.rs
hnotshe: That limitation remains intentional here. The issue contract is architecture selection, while arm v6/v7 support needs a variant input and normalization contract rather than silently choosing the first `arm` entry. I am keeping variant handling out of this patch.

crates/spurd/src/container.rs
hnotshe: This is a separate execution-time validation boundary. This PR keeps daemon-side imports host-native and stays focused on honoring the requested architecture during import; changing `resolve_image` and launch behavior would expand beyond issue #343 acceptance criteria. I am leaving that follow-up out of this patch.

PR #442, opened 2026-07-14
fix(spurctld): raise grpc message-size limits and cap job submission size

shiv-tyagi:
## Problem

Large Raft `InstallSnapshot` RPCs failed to decode on bigger clusters. openraft serializes snapshots as JSON (expanding `Vec<u8>` ~3.4x), pushing messages past tonic's default 4 MiB decode limit. The follower rejects the message, so a lagging or rejoining node can never install a snapshot and catch up.

## Fix

**Transport limits.**
- Pin `RAFT_MAX_MESSAGE_SIZE` (32 MiB) for the Raft-internal service. Bound installed snapshot chunks to 1 MiB in the openraft `Config`.
- Add asymmetric limits for the client-facing surface: inbound requests capped at `MAX_GRPC_REQUEST_SIZE` (8 MiB, sized to the 4 MiB `JobSpec` cap + headroom), outbound responses up to `MAX_GRPC_MESSAGE_SIZE` (32 MiB) for large `GetJobs`/`GetNodes` payloads. Route all construction through helpers so the caps stay symmetric across crates.

**AppendEntries size guard.**
- After serializing an `AppendEntries` batch, if it exceeds `RAFT_MAX_MESSAGE_SIZE`, return `PayloadTooLarge` with a halved entries hint. openraft retries immediately with a smaller batch, preventing replication wedge during follower catch-up with large entries.

**Submission cap (defense in depth).**
- Cap the serialized `JobSpec` at 4 MiB at the single chokepoint (`ClusterManager::submit_job`), checked after defaults are applied. Measured with a zero-allocation counting writer. Clear user-facing error on rejection.

## Testing

- `append_entries` PayloadTooLarge unit tests (oversized batch rejected with correct hint, small batch passes through).
- Transport regression tests: >4 MiB `install_snapshot` accepted on raft-internal, 12 MiB raft-internal accepted above client-facing cap, constant invariant assertions.
- Submission size-cap unit tests.
- `clippy` clean; `cargo test` green across changed crates.
- Reproduced the decode failure on a 3-node bare-metal Raft cluster, then confirmed snapshots ship and end-to-end submission works after the fix.

## Rolling upgrade note

An un-upgraded `spur-cli`/`spurd`/`spur-k8s` binary still defaults to tonic's 4 MiB limits. Once `spurctld` is upgraded and starts returning >4 MiB responses (e.g. large `GetJobs`), an old client will fail with an opaque `OutOfRange` decode error. Upgrade clients alongside the controller to avoid this.

Note: the two failing `spur-devices` CDI tests are pre-existing and environment-dependent (they read host `/etc/cdi`, from #262), unrelated to this change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3706,6 +3706,7 @@ dependencies = [
  "sqlx",
  "tempfile",
  "tokio",
+ "tokio-stream",
  "tonic",
  "tower",
  "tower-http 0.7.0",
--- a/crates/spur-cli/src/exec.rs
+++ b/crates/spur-cli/src/exec.rs
@@ -5,7 +5,6 @@
 
 use anyhow::{Context, Result};
 use clap::Parser;
-use spur_proto::proto::slurm_controller_client::SlurmControllerClient;
 use spur_proto::proto::ExecInJobRequest;
 
 /// Execute a command inside a running containerized job.
@@ -41,7 +40,7 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
     let channel = spur_client::connect_channel(&args.controller)
         .await
         .context("failed to connect to controller")?;
-    let mut client = SlurmControllerClient::new(channel);
+    let mut client = spur_proto::controller_client(channel);
 
     let resp = client
         .exec_in_job(ExecInJobRequest {
--- a/crates/spur-cli/src/node.rs
+++ b/crates/spur-cli/src/node.rs
@@ -8,7 +8,6 @@ use std::collections::HashMap;
 use anyhow::{bail, Result};
 use clap::{Parser, Subcommand};
 
-use spur_proto::proto::slurm_controller_client::SlurmControllerClient;
 use spur_proto::proto::UpdateNodeRequest;
 
 /// Node management commands.
@@ -104,7 +103,7 @@ fn parse_label_args(label_args: &[String]) -> Result<(HashMap<String, String>, V
 }
 
 async fn cmd_label(controller: &str, node: String, label_args: Vec<String>) -> Result<()> {
-    let mut client = SlurmControllerClient::new(spur_client::connect_channel(controller).await?);
+    let mut client = spur_proto::controller_client(spur_client::connect_channel(controller).await?);
     let (set_labels, remove_labels) = parse_label_args(&label_args)?;
 
     client
@@ -128,7 +127,7 @@ async fn cmd_label(controller: &str, node: String, label_args: Vec<String>) -> R
 }
 
 async fn cmd_drain(controller: &str, node: String, reason: Option<String>) -> Result<()> {
-    let mut client = SlurmControllerClient::new(spur_client::connect_channel(controller).await?);
+    let mut client = spur_proto::controller_client(spur_client::connect_channel(controller).await?);
     let resp = client
         .drain_node(spur_proto::proto::DrainNodeRequest {
             name: node.clone(),
@@ -158,7 +157,7 @@ async fn cmd_remove(
     force: bool,
     reason: Option<String>,
 ) -> Result<()> {
-    let mut client = SlurmControllerClient::new(spur_client::connect_channel(controller).await?);
+    let mut client = spur_proto::controller_client(spur_client::connect_channel(controller).await?);
     let resp = client
         .deregister_node(spur_proto::proto::DeregisterNodeRequest {
             name: node.clone(),
--- a/crates/spur-cli/src/sacct.rs
+++ b/crates/spur-cli/src/sacct.rs
@@ -3,7 +3,6 @@
 
 use anyhow::{Context, Result};
 use clap::Parser;
-use spur_proto::proto::slurm_accounting_client::SlurmAccountingClient;
 use spur_proto::proto::GetJobHistoryRequest;
 
 use crate::exit_fmt::format_exit;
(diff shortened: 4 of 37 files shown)


yansun1996 (commented):
## Summary

The InstallSnapshot fix itself is correct, well-motivated, and well-tested — I'd merge that part today. But this PR bundles in a much larger, unscoped change (a uniform 8x message-size increase across the *entire* client-facing gRPC surface) that introduces a new latent DoS class the description doesn't mention or address. I'd like to see that split out or explicitly addressed before merging.

## What's good

- The core bug (large InstallSnapshot RPCs failing past tonic's 4 MiB default) is real and this fixes it cleanly. `raft_server.rs` has a genuine round-trip test proving it (8 MiB payload accepted with raised limits, rejected with defaults as a negative control) — exactly the right test for a wire-format fix.
- `check_submission_size`'s serialize-and-count approach is the right way to bound a `JobSpec` — measures the whole serialized value via a zero-alloc counting writer, so it can't be bypassed by a field the check forgot to sum.
- The wide diff across `spur-cli`/`spur-ffi`/`spur-k8s` is legitimate, correctly-scoped symmetric propagation of the same client-construction change (verified `health.rs` was correctly left alone — it's a probe-only connect, never decodes an RPC response).

## Two things I think are blocking, not follow-ups

**1. The job-spec cap doesn't protect what it looks like it protects.** `check_submission_size` runs inside `ClusterManager::submit_job`, which only executes on the leader. On a follower, `submit_job` in `server.rs` forwards the fully-decoded request to the leader *before* that check ever runs. So repeatedly hitting any follower with payloads up to the new 32 MiB cap costs it a full decode + forward on every request, with the leader doing the actual rejection — that's amplification, not protection. Separately, `MAX_GRPC_MESSAGE_SIZE` is applied to the whole `SlurmController`/`SlurmAgent` trait via `controller_server`/`agent_server`, so `UpdateNode`, `RegisterAgent`, `ExecInJob`, etc. — none of which had size validation before — silently accept 8x larger payloads now. Given the stated goal here was fixing Raft snapshot transfer specifically, could the Raft-internal service get its own limit (it already nearly does, via `RAFT_MAX_MESSAGE_SIZE`) while the client-facing default stays closer to what it was — or is the scope increase a deliberate choice worth documenting?

**2. AppendEntries (log replication, not snapshot transfer) has no size bound at all**, and this PR makes hitting one more likely: `max_payload_entries` (openraft default 300) isn't overridden in `raft.rs`, and the new 4 MiB per-job cap means 300 batched entries could serialize past 1 GB in a single `AppendEntries` RPC — ~37x over the 32 MiB `RAFT_MAX_MESSAGE_SIZE` this PR introduces in the same file. A job array (`sbatch --array=1-50`) with multi-MB scripts, submitted while a follower briefly lags, could plausibly hit this. Failures currently map to generic `Unreachable`/backoff rather than openraft's `PayloadTooLarge`, so the built-in "retry with fewer entries" mechanism never engages. Worth either capping `max_payload_entries` proportionally, or wiring `append_entries` to return `PayloadTooLarge` on oversized batches.

## Suggestion

Land the InstallSnapshot/chunk-sizing fix as-is (it's correct and needed now), and open a fast-follow for the client-facing gRPC limit increase + AppendEntries bound — these are pre-existing risk classes this PR happens to widen, not regressions of the thing being fixed, but I don't think they should ship silently as part of a PR titled as a snapshot-size fix.

One more minor note: an old (pre-this-PR) `spur-cli`/`spurd`/`spur-k8s` binary still defaults to tonic's 4 MiB limits, so once `spurctld` is upgraded and starts returning >4 MiB responses, an un-upgraded client will fail with an opaque `OutOfRange` error with no indication it's a version-skew issue — worth a callout in the PR description for anyone doing a rolling upgrade.

(Reviewed with 2 independent fresh-context passes + a cross-validation pass before posting, given the size of the diff — happy to dig into any of this further.)

shiv-tyagi:
Thanks Yan, both blockers addressed in 2c33749:

**Blocker 1 (uniform 32 MiB on client-facing surface):**
Split into asymmetric limits. Client-facing servers now decode at `MAX_GRPC_REQUEST_SIZE` (8 MiB, sized to the 4 MiB job-spec cap + headroom) and encode responses at `MAX_GRPC_MESSAGE_SIZE` (32 MiB). All client helpers and inline callsites updated to match. Raft-internal stays at 32 MiB both directions. The inbound 8x amplification is gone.

**Blocker 2 (AppendEntries unbounded):**
After serializing the batch, if `payload.len() > RAFT_MAX_MESSAGE_SIZE`, we return `RPCError::PayloadTooLarge(PayloadTooLarge::new_entries_hint(entries.len() / 2))`. openraft's replication layer honors this immediately and for the next 10 RPCs, halving the batch until it fits. No static `max_payload_entries` cap needed; a single entry (capped at 4 MiB JSON, ~13.6 MiB inflated) always fits under 32 MiB, so convergence is guaranteed.

**Copilot inline (check_submission_size placement):**
Moved to after `apply_default_qos` so the measured spec reflects the final state. Array expansion only adds bounded integer metadata per task, so a single pre-expansion check still bounds each Raft log entry.

**Rolling-upgrade version skew:**
Called out in the PR description: un-upgraded clients still default to 4 MiB and will fail once `spurctld` returns larger responses.

New tests: `append_entries_rejects_oversized_batch_with_payload_too_large`, `append_entries_accepts_small_batch`, `raft_server_accepts_above_client_cap`, `request_size_tighter_than_raft_size`. All green.

yansun1996 (approved):

PR #446, opened 2026-07-14
feat(spurctld): show valid accounts in association errors

### LINKED ISSUE #420: User-account association error should show valid associations
When a user submits a job with an account they're not associated with:

```
code: 'Client specified an invalid argument', message: "user 'akasanag' is not associated with account 'amd-spur'"
```

The error tells you what's wrong but gives no guidance on how to fix it.

**Expected behavior:**

Include actionable information in the error:

```
user 'akasanag' is not associated with account 'amd-spur'. Valid accounts for this user: [amd-aifw-dev, amd-silo-tiger]. Use --account=<account> to specify.
```

Or if the user has no associations at all:

```
user 'akasanag' has no account associations. Contact your cluster admin to run: sacctmgr add user name=akasanag account=<account>
```

hnotshe:
## Summary

- Show a user's valid account associations when job submission specifies an invalid account.
- Direct users with no associations to the appropriate `sacctmgr add user` command.
- Read membership status and alternatives from one cache snapshot while preserving the constant-time valid-membership path.

## Testing

- `git spur-check`

Closes #420

### DIFF
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -4031,7 +4031,7 @@ fn validate_user_account(
             )))
         }
         AccountMembership::NotMember(valid_accounts) => Err(SubmitError::invalid(format!(
-            "user '{}' is not associated with account '{account}'. Valid accounts for this user: [{}]. Use --account=<account> to specify.",
+            "user '{}' is not associated with account '{account}'. Accounts associated with this user: [{}].",
             spec.user,
             valid_accounts.join(", ")
         ))),
@@ -6421,7 +6421,7 @@ mod tests {
         assert_eq!(
             err,
             SubmitError::invalid(
-                "user 'testuser' is not associated with account 'research'. Valid accounts for this user: [student-a, student-z]. Use --account=<account> to specify."
+                "user 'testuser' is not associated with account 'research'. Accounts associated with this user: [student-a, student-z]."
             )
         );
     }
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -31,7 +31,7 @@ use spur_metrics::partition::PartitionMetricsSnapshot;
 use spur_metrics::user_acct::UserAcctMetricsSnapshot;
 
 use crate::accounting::{AccountingNotifier, JobStartRecord};
-use crate::association_cache::AssociationCache;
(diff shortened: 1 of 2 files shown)


yansun1996:
Well-built. The valid (`Member`) path early-returns before building the alternatives list so the hot path stays constant-time, the membership check and alternatives are read under a single snapshot guard (no torn message under concurrent refresh), only the caller's own accounts are listed (no cross-user disclosure), there is no `unwrap()` in library code, and the tests assert sorted message content deterministically. Two small wording notes inline.

crates/spurctld/src/cluster.rs
```diff
@@ -4022,16 +4022,20 @@ fn validate_user_account(
...
+    match assoc_cache.account_membership(&spec.user, account) {
+        AccountMembership::CacheUnavailable | AccountMembership::Member => Ok(()),
+        AccountMembership::NotMember(valid_accounts) if valid_accounts.is_empty() => {
+            Err(SubmitError::invalid(format!(
+                "user '{}' has no account associations. Contact your cluster admin to run: sacctmgr add user name={} account=<account>",
+                spec.user, spec.user
+            )))
+        }
+        AccountMembership::NotMember(valid_accounts) => Err(SubmitError::invalid(format!(
+            "user '{}' is not associated with account '{account}'. Valid accounts for this user: [{}]. Use --account=<account> to specify.",
```
yansun1996: Two small wording notes on this message: (a) `validate_user_account` runs before `validate_partition`, so a listed account can still be rejected by the partition ACL (`allow_accounts`/`deny_accounts`) — a user who follows the hint may hit a second, different error. Softening to something like "Accounts associated with this user:" would avoid implying they are guaranteed acceptable. (b) The `Use --account=<account>` hint is CLI-specific, but this path is shared by the REST/FFI/gRPC submission surfaces where `--account` is not the mechanism. Both cosmetic.

hnotshe: Thanks — agreed on both points. Addressed in the latest commit: the message now says “Accounts associated with this user” so it does not imply partition eligibility, and the CLI-specific `--account` hint has been removed so the shared submission error remains transport-neutral. I also updated the regression test, and `git spur-check` passes. Would appreciate another look when convenient.

yansun1996 (approved):
Thanks @hnotshe  !

PR #447, opened 2026-07-14
fix(accounting): filter users by name

### LINKED ISSUE #418: sacctmgr show user name=X silently ignores the name filter
`sacctmgr show user name=akasanag` parses the `name=` parameter into the params map but the `show user` handler only reads `p.get("account")`. The `name` filter is silently discarded and all users are returned.

**Current behavior:**
- `sacctmgr show user name=akasanag` returns the entire user list (hundreds of entries)
- No warning that the filter was ignored
- `ListUsersRequest` proto only has an `account` field, no `user`/`name` field

**Expected behavior:**
- Either implement server-side filtering by user name (add `string user = 2` to `ListUsersRequest` and filter in the SQL query)
- Or filter client-side after fetching all users
- At minimum, warn when an unsupported filter parameter is provided

hnotshe:
## Summary

Fix `sacctmgr show user name=X` returning every configured user.

The CLI now forwards `name=` and `user=` through the accounting RPC. `ListUsersRequest` gains an additive user field, and the accounting query applies exact user and account filters server-side.

Closes #418.

## Testing

- Full Linux validation: formatting, SPDX headers, cargo-deny, clippy, all-target build, and complete test suite
- Focused sacctmgr parameter regression test
- PostgreSQL integration test covering user-only, account-plus-user, and no-match filtering
- Independent reviewer: PASS

### DIFF
--- a/crates/spur-cli/src/sacctmgr.rs
+++ b/crates/spur-cli/src/sacctmgr.rs
@@ -1219,4 +1219,14 @@ mod tests {
         assert_eq!(account, "testacct");
         assert_eq!(user, "testuser");
     }
+
+    #[test]
+    fn list_user_filters_include_user_alias() {
+        let p = parse_params(&["user=testuser".into()]);
+
+        let (account, user) = list_user_filters(&p);
+
+        assert!(account.is_empty());
+        assert_eq!(user, "testuser");
+    }
 }
--- a/crates/spur-cli/src/sacctmgr.rs
+++ b/crates/spur-cli/src/sacctmgr.rs
@@ -234,6 +234,16 @@ fn build_add_user_request(
     })
 }
 
+fn list_user_filters(p: &std::collections::HashMap<String, String>) -> (String, String) {
+    (
+        p.get("account").cloned().unwrap_or_default(),
+        p.get("name")
+            .or_else(|| p.get("user"))
+            .cloned()
+            .unwrap_or_default(),
+    )
+}
+
 async fn add(entity: &str, params: &[String], addr: &str) -> Result<()> {
     let p = parse_params(params);
 
@@ -642,12 +652,13 @@ async fn show(entity: &str, params: &[String], addr: &str) -> Result<()> {
             Ok(())
         }
         "user" | "users" => {
-            let account_filter = p.get("account").cloned().unwrap_or_default();
+            let (account_filter, user_filter) = list_user_filters(&p);
 
             let mut client = connect(addr).await?;
             let resp = client
                 .list_users(ListUsersRequest {
                     account: account_filter,
+                    user: user_filter,
                 })
                 .await
                 .context("ListUsers RPC failed")?;
@@ -1197,5 +1208,15 @@ mod tests {
     fn qos_empty_format_string_produces_no_fields() {
         let fields = format_engine::parse_named_format("", &qos_field_spec, &qos_header);
         assert!(fields.is_empty());
+    }
+
+    #[test]
+    fn list_user_filters_include_name() {
+        let p = parse_params(&["name=testuser".into(), "account=testacct".into()]);
+
+        let (account, user) = list_user_filters(&p);
+
+        assert_eq!(account, "testacct");
+        assert_eq!(user, "testuser");
     }
 }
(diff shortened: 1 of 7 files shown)


yansun1996:
Clean, correct, and backward-compatible. The new proto field `user = 2` is a genuinely free tag (additive, no wire-compat break), the `list_users` query rewrite preserves the columns, `DISTINCT ON`, and `ORDER BY`, empty-string from existing callers maps to `None` so list-all is preserved, and the filters are parameter-bound. A couple of minor Slurm-parity notes inline.

crates/spur-cli/src/sacctmgr.rs
```diff
@@ -140,6 +140,16 @@ fn build_add_user_request(
     Ok((name, account, admin, default_qos))
 }
 
+fn list_user_filters(p: &std::collections::HashMap<String, String>) -> (String, String) {
+    (
+        p.get("account").cloned().unwrap_or_default(),
+        p.get("name")
```
yansun1996, line 240: Minor: `name=`/`account=` take a single exact value, so `name=alice,bob` binds the literal string and matches zero rows (silently, no warning). Real Slurm treats it as a comma-separated list. Could be a follow-up if list support is desired.

crates/spurctld/src/accounting/db.rs
```diff
@@ -575,38 +575,28 @@ pub async fn remove_user(pool: &PgPool, user: &str, account: &str) -> anyhow::Re
...
+    let rows = sqlx::query(
+        r#"
+        SELECT DISTINCT ON (u.name, u.account)
+            u.name, u.account, u.admin_level, u.default_account, a.default_qos
+        FROM users u
+        LEFT JOIN associations a
+            ON a.user_name = u.name AND a.account = u.account
+                AND (a.partition_name IS NULL OR a.partition_name = '')
+        WHERE ($1::TEXT IS NULL OR u.account = $1)
+            AND ($2::TEXT IS NULL OR u.name = $2)
```
yansun1996, line 669: Minor: the match is case-sensitive (`=`, values not lowercased), so `name=Alice` will not match a stored `alice`. Slurm user names are effectively case-sensitive too, so this is defensible — just worth a one-line mention.

crates/spur-cli/src/sacctmgr.rs
```diff
@@ -678,4 +689,14 @@ mod tests {
             build_add_user_request(&modify_params).unwrap()
         );
     }
+
+    #[test]
+    fn list_user_filters_include_name() {
```
yansun1996, line 1214: Minor: this test covers the `name=` path; the `user=` alias branch (`get("name").or_else(|| get("user"))`) — the more Slurm-idiomatic `sacctmgr show user user=X` form — is not asserted anywhere. A small case would close that gap.

hnotshe: The `user=` path is covered now in `503b6cf`; I added a focused test alongside the existing `name=` case. Both the targeted test and the full Linux validation suite are green.

crates/spurctld/src/accounting/db.rs
hnotshe: Case-sensitive matching is intentional here. Usernames are treated as exact identifiers, which also lines up with Slurm’s behavior, so I’ve left the query unchanged.

crates/spur-cli/src/sacctmgr.rs
hnotshe: That makes sense. For this fix I’m keeping the filter to one exact value, matching the issue as reported. Comma-separated values would be useful Slurm-parity work, but are better handled separately.

yansun1996 (dismissed):

yansun1996:
Rebased onto the latest `main` — there were a few conflicts in the accounting tests where recent changes to `add_user` (association limits) and `list_users` overlapped with this PR's signature changes; resolved by keeping both sets of tests and threading the new filter argument through.

Also verified the fix end-to-end on a 2-node cluster with accounting enabled:

- `sacctmgr show user name=<user>` returns only that user
- `sacctmgr show user user=<user>` (alias form) returns only that user
- `sacctmgr show user name=<user> account=<acct>` applies both filters
- `sacctmgr show user account=<acct>` returns just that account's users
- a non-existent name returns an empty result (header only), no error

All green — build, clippy, fmt, and the unit tests pass on the rebased branch.

yansun1996 (dismissed):

yansun1996 (approved):

PR #449, opened 2026-07-15
fix(spurd): raise RLIMIT_MEMLOCK for job steps (default unlimited)

shiv-tyagi:
## Summary

- spurd never called `setrlimit(RLIMIT_MEMLOCK)`, so job steps inherited the systemd default (8 MiB). NCCL/InfiniBand RDMA failed with `ibv_create_cq: Cannot allocate memory`.
- Adds a `[rlimits]` config section with a `memlock` knob (default `"unlimited"`). spurd raises the limit as root before dropping privilege, on both launch paths (non-container `pre_exec` and container `fork`).
- With the default, NCCL jobs work out of the box. No operator action needed.

## Approach

The fix mirrors how Slurm's `slurmstepd` sets rlimits: a privileged actor (`spurd`, running as root) calls `setrlimit` in the child process before the uid/gid drop. The limit persists across `setpriv`/`setuid`/`unshare`/`exec`.

Three config values are supported:
- `"unlimited"` (default) — `RLIM_INFINITY`, fixes RDMA/NCCL
- `"inherit"` — do not call setrlimit (old behavior)
- `"<bytes>"` — fixed cap

The `[rlimits]` section is scoped to POSIX per-process resource limits only (not QoS caps, not cgroup limits). Tier 2 (per-submission propagation from the submitter) will extend this section later.

## Test plan

- [x] Unit tests for parser: unlimited, inherit, bytes, invalid input, zero (treated as unlimited)
- [x] `cargo test --locked` all passing
- [x] Validated on bare-metal cluster: root spurd -> job gets `unlimited` memlock; non-root spurd -> raises to system hard limit (best-effort)

### DIFF
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -103,6 +103,10 @@ pub struct SlurmConfig {
     /// Node admission control.
     #[serde(default)]
     pub admission: AdmissionConfig,
+
+    /// POSIX per-process resource limits applied to job steps at launch.
+    #[serde(default)]
+    pub rlimits: RlimitsConfig,
 }
 
 /// Configuration for auto-update checking and self-update.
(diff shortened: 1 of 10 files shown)


crates/spur-core/src/config.rs
```diff
@@ -790,6 +794,56 @@ pub enum AdmissionMode {
...
+
+impl RlimitsConfig {
+    pub fn memlock_limit(&self) -> MemlockLimit {
+        match self.memlock.trim().to_lowercase().as_str() {
+            "unlimited" | "" => MemlockLimit::Unlimited,
+            "inherit" => MemlockLimit::Inherit,
+            other => match other.parse::<u64>() {
+                Ok(0) => MemlockLimit::Unlimited,
+                Ok(n) => MemlockLimit::Bytes(n),
+                Err(_) => MemlockLimit::Unlimited,
```
biluriuday: should we fail here in case the parsing fails instead of setting to unlimited?

crates/spurd/src/main.rs
```diff
@@ -212,8 +212,12 @@ async fn main() -> anyhow::Result<()> {
     });
 
     // Start agent gRPC server (receives job launches from spurctld)
+    let memlock = config
+        .as_ref()
+        .map(|c| c.rlimits.memlock_limit())
+        .unwrap_or(spur_core::config::MemlockLimit::Unlimited);
```
biluriuday: log the effective memlock limit spurd can actually grant, and emit a warning if it can't reach the configured value. Consider a metric/counter. This can help usability.

crates/spurd/src/executor.rs
```diff
@@ -113,6 +116,26 @@ pub fn decode_wait_status(status: nix::sys::wait::WaitStatus) -> (i32, i32) {
...
+        rlim_cur: v,
+        rlim_max: v,
+    };
+    if unsafe { libc::setrlimit(libc::RLIMIT_MEMLOCK, &rl) } != 0 {
+        eprintln!(
+            "spur: setrlimit(MEMLOCK) failed: {}",
+            std::io::Error::last_os_error()
+        );
+    }
+}
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed both. Removed eprintln (and all allocating calls) from the pre_exec path. The fallback now does getrlimit then raises soft to the inherited hard limit when setting RLIM_INFINITY fails. Validated on bare-metal: root spurd yields unlimited, non-root raises soft to hard (~993 MB on the test nodes).

crates/spur-core/src/config.rs
shiv-tyagi: Done. memlock_limit() now returns Result<MemlockLimit, ConfigError> and spurd refuses to start on invalid input. Error message includes the bad value and valid options.

crates/spurd/src/main.rs
shiv-tyagi: Done. At startup spurd logs an INFO with configured and effective_hard values, and emits a WARN if configured=unlimited but the process hard limit is finite (only when non-root, since root can raise it in the child). Extracted into a log_memlock_status() helper to keep main clean.

sgopinath1 (approved):
LGTM

biluriuday (approved):

Issue #364, opened 2026-06-29
spurd: user jobs fail with "Permission denied" when job script staged into a dir the executing user cannot read


yansun1996:
## What happened

Non-root batch jobs fail immediately with:

```
/bin/bash: /root/.spur_job_<jobid>.sh: Permission denied
```

The job is marked `FAILED` (NonZeroExitCode) and the job body never executes. The job's stdout is empty; the error only appears in the job's stderr file.

## Root cause (observed)

`spurd` runs as root and stages the job script as `.spur_job_<jobid>.sh` inside the job's working directory (`crates/spurd/src/executor.rs`, ~line 287). The job process itself is then launched as the **submitting user**.

The working directory is taken from the job spec (submitter CWD, or `--chdir`). When that directory is writable by root but **not readable/traversable by the executing user**, root successfully writes the script but the user's job process cannot open it.

Two real-world directory situations both trigger this:

1. **work_dir resolves to `/root`** (mode `700`). Root writes `/root/.spur_job_N.sh` (the file is even `chmod 755`), but the non-root user cannot traverse `/root` to reach it.
2. **work_dir is an NFS path under root_squash** (e.g. a managed NFS mount with `root_squash`). Here root is squashed to `nobody` and cannot create the script in the user-owned directory at all.

In `executor.rs` (~line 263) `work_dir` is accepted if `create_dir_all(work_dir)` succeeds for root, otherwise it falls back to `/tmp`. The decision is based purely on whether **root** can write — it never checks whether the **executing user** can read the resulting script.

## Expected

Any user able to submit a job should have that job's script staged somewhere the executing user is guaranteed to be able to read, so the job body runs regardless of the working directory's ownership/permissions.

## Repro

1. Cluster with a non-root user (uid present on the compute node), spurd running as root.
2. Submit a trivial batch job whose effective working directory is `/root` or an NFS root_squash path, with `#SBATCH --output=/tmp/spur-%j.out` and `--error=/tmp/spur-%j.err` (local, so the error is observable):

   ```bash
   #!/bin/bash
   #SBATCH --job-name=stage-repro
   #SBATCH --nodes=1
   #SBATCH --output=/tmp/spur-%j.out
   #SBATCH --error=/tmp/spur-%j.err
   echo "hello from $(id -un)"
   ```
3. Job ends `FAILED`; `/tmp/spur-<id>.err` contains `/bin/bash: /root/.spur_job_<id>.sh: Permission denied`.

[...]

PR #450, opened 2026-07-15
fix(spurd): release GPU allocation on completion and post-record launch failure

yansun1996:
## What this fixes

SPUR-65 (P2): a `spurd` agent could permanently retain a completed/failed job's resource reservation in its in-memory allocation table. Once stranded, every subsequent dispatch to that node was rejected with `controller-allocated GPUs unavailable on this node`, even though the resources were physically free and the controller still saw the node `IDLE`. The controller kept selecting the node, so jobs bounced there and ended in `JobHoldMaxRequeue` — a silent black hole until `spurd` restart.

This is the agent-side counterpart of SPUR-16 ("nodes not released after job completion"), whose controller-side release path was previously audited and found correct. This PR addresses the agent side; it resolves the SPUR-16 symptom for the allocation-leak case. (SPUR-16's separately-reported `exit_code=126` aspect traces to GH #364, an unrelated work_dir-permission bug, and is out of scope here.)

## Root cause & fix

The production-reachable leak was a **missing release on the launch path**. `allocate_local_resources` reserved the job's resources, but the job was only inserted into `running` — whose completion path is the only place release happened — after several more fallible steps. If device-registry resolution failed, the `?` returned early and the reservation was never released, with no tracked job to ever free it. The node then rejected all future dispatches while the controller believed it was idle.

The fix reworks agent-side allocation to be **per-job and self-healing**:

- **Per-job ownership**: `NodeAllocation` now tracks what each job holds by job id (`owners` map), so an allocation is released by id on any teardown path rather than by replaying a specific `AllocationResult` on one path.
- **Eager release on every launch-failure exit** (device-injection failure, missing container config, executor error): the reservation is freed immediately.
- **Self-healing reconcile**: the monitor loop reclaims any allocation whose job is no longer tracked and is not mid-launch, so the node recovers on the next tick instead of stranding until restart. A `launching` guard marks reservations between allocate and commit (`commit_job` runs under the `running` lock before insertion) so an in-flight launch — slow image pull, fork — is never mistaken for an orphan, and a launch that never commits can never leave a permanent phantom reservation.
- **device_id unification** (`spur-sched`): `AllocationResult.gpu_ids` previously meant vec indices in one path and device_ids in another; release/record interpreted them differently. Unified to device_id everywhere. Today `assign_device_ids` makes device_id a dense `0..N` counter equal to vec position, so the mismatch does not currently misfire — this is hardening against that latent trap, honestly labeled as such.
- **Memory-accounting fix**: memory was only added to the used pool when `cpus_per_task > 0` but always subtracted on release, so a 0-cpu job would drift the used count downward over time. Memory is now accounted symmetrically.

The old `try_allocate` / `record_gpus` had no remaining callers after this rework and were removed; `release` is now a private helper of `release_job`.

## Not in this PR

- Controller-driven heartbeat reconciliation was considered but is unnecessary: the agent's own `running` set is authoritative, so agent-local reconcile covers the orphan class without a proto change or controller cooperation.

## Testing

- Unit (`cons_tres.rs`): device-id release with non-contiguous ids; all-or-nothing GPU reservation rollback; release-by-id idempotency; reconcile spares live + launching and reclaims only true orphans; memory symmetry with 0 cpus. Each new invariant confirmed to fail against the pre-fix code.
- Agent-level (`agent_server.rs`): a launch that fails after reservation releases it; reconcile reclaims an orphan while sparing a tracked job (exercises the real monitor-loop wiring without driving the timed loop).
- `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` — clean; `cargo test --locked` — all pass.
- Live on the 2-node lab cluster (CPU-only, so the GPU-specific rejection path can't run directly, but the leak/strand and accounting mechanisms are resource-agnostic): deployed to both nodes and verified (a) bursts of concurrent jobs all complete with the node returning to `IDLE`/`CPUAlloc=0` and **no spurious reconcile warnings** (launch race does not misfire under real load); (b) a job whose launch fails after reservation requeues to `JobHoldMaxRequeue` while the node stays fully allocatable — the next job launches and completes on it; (c) memory returns exactly to baseline across repeated `--mem` job cycles (no drift).

### DIFF
--- a/crates/spur-sched/src/cons_tres.rs
+++ b/crates/spur-sched/src/cons_tres.rs
@@ -27,15 +27,12 @@ pub struct NodeAllocation {
     pub gpu_allocated: Vec<bool>,
     /// GPU info for type matching.
     pub gpus: Vec<GpuResource>,
-    /// Per-job ownership of what each job holds, so an allocation can be
-    /// released by job id and orphaned allocations can be reconciled away.
-    /// This is the single source of truth for what is allocated — the bitmaps
-    /// above are a derived index for fast free-count queries.
+    /// Per-job ownership, so an allocation can be released by job id and
+    /// orphans reconciled. Source of truth; the bitmaps above are a derived
+    /// index for fast free-count queries.
     owners: HashMap<u32, AllocationResult>,
-    /// Jobs whose resources are reserved but that have not yet been committed
-    /// to the agent's running set (still mid-launch: image pull, fork, etc.).
-    /// Reconcile must never reclaim these — they have no tracked process yet
-    /// but are not orphaned.
+    /// Reserved but not yet committed to the running set (mid-launch: image
+    /// pull, fork). Reconcile spares these — not yet tracked, but not orphaned.
     launching: HashSet<u32>,
 }
 
@@ -115,16 +112,11 @@ impl NodeAllocation {
         }
     }
 
-    /// Reserve resources for a job, tracked by job id. GPU device ids are the
-    /// hard gate: if any is unknown or already in use, nothing is reserved and
-    /// None is returned (the caller must reject the dispatch). CPU/memory are
-    /// recorded as usage bookkeeping — the controller is authoritative for
-    /// placement, so an oversubscribed CPU count records what is available
-    /// rather than failing. Memory is always accounted, independent of CPU
-    /// count, so release stays symmetric.
-    ///
-    /// The job is marked `launching` until `commit_job` or `release_job` is
-    /// called, so reconcile never reclaims an in-flight launch.
+    /// Reserve resources for a job, keyed by job id. GPU device ids are the
+    /// hard gate (unknown or in-use → None, caller rejects the dispatch); CPU
+    /// is best-effort since the controller owns placement. Memory is always
+    /// accounted so release stays symmetric. Marked `launching` until
+    /// `commit_job`/`release_job` so reconcile spares an in-flight launch.
     pub fn allocate_for_job(
         &mut self,
         job_id: u32,
@@ -182,10 +174,9 @@ impl NodeAllocation {
     }
 
     /// Release every owned allocation whose job is neither in `live` nor still
-    /// launching. Returns the reclaimed job ids. This is the self-healing path:
-    /// if a job's teardown ever fails to release (a completion path that dropped
-    /// the job without releasing), its resources are recovered here instead of
-    /// stranding the node until spurd restart (SPUR-65).
+    /// launching, returning the reclaimed ids. Self-healing backstop: if a
+    /// teardown ever fails to release, resources are recovered here instead of
+    /// stranding the node until spurd restart.
     pub fn reconcile(&mut self, live: &HashSet<u32>) -> Vec<u32> {
         let orphaned: Vec<u32> = self
             .owners
@@ -293,12 +284,9 @@ mod tests {
 
     #[test]
     fn test_record_then_release_noncontiguous_device_ids() {
-        // Real nodes expose GPUs whose device_id is not its position in the
-        // vec (e.g. DRM render ids 128..135). allocate_for_job/release_job key
-        // by device_id, so a released device must return to the free pool —
-        // otherwise the slot is never cleared and the controller's next
-        // legitimate allocation of that device is rejected forever (SPUR-65:
-        // node becomes a black hole -> JobHoldMaxRequeue).
+        // Real nodes expose GPUs whose device_id != vec position (DRM render
+        // ids 128..135). Release keys by device_id, so a released device must
+        // return to the free pool or it is rejected forever.
         let mut node = make_node_with_ids(64, 256_000, vec![128, 129, 130, 131], "mi350x");
 
         assert!(node.allocate_for_job(1, 0, 0, &[129, 131]).is_some());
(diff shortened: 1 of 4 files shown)


crates/spur-sched/src/cons_tres.rs
```diff
@@ -73,23 +97,50 @@ impl NodeAllocation {
...
-        {
-            return None;
+        let mut gpu_indices = Vec::with_capacity(gpu_device_ids.len());
+        for &id in gpu_device_ids {
+            let idx = self.gpus.iter().position(|g| g.device_id == id)?;
+            if self.gpu_allocated[idx] || gpu_indices.contains(&idx) {
+                return None;
+            }
+            gpu_indices.push(idx);
         }
```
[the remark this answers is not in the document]
yansun1996: Good catch — fixed in f6ed4fe. `allocate_for_job` now rejects a reservation whose `job_id` is already owned (returns `None`, which the caller maps to a clean dispatch rejection), so a retried LaunchJob RPC can't double-count CPU/memory or orphan the prior owner entry. Added a unit test (`test_allocate_for_job_rejects_duplicate_job_id`).

crates/spurd/src/agent_server.rs
```diff
@@ -1888,6 +1932,135 @@ mod tests {
...
+        );
+
+        assert_eq!(
+            svc.free_gpu_count().await,
+            1,
+            "GPU allocation must be released after a post-record launch failure"
+        );
+    }
+
+    // The monitor loop.s reconcile step must reclaim an
```
shiv-tyagi: Nit: typo, `loop.s` -> `loop's`.

crates/spur-sched/src/cons_tres.rs
```diff
@@ -221,59 +275,152 @@ mod tests {
...
+        assert!(node.release_job(1));
+        assert!(node.release_job(3));
+        assert_eq!(node.free_gpus(None), 4);
+    }
+
+    #[test]
+    fn test_memory_released_symmetrically_with_zero_cpus() {
+        // A job with 0 cpus must still have its memory reserved and released
+        // symmetrically — otherwise release drives allocated_memory_mb below
+        // what was added (the pre-fix drift).
```
shiv-tyagi: Nit: drop the fix-narration. The invariant is the point, not that it regressed before.
```suggestion
        // A job with 0 cpus must still have its memory reserved and released
        // symmetrically, or release drives allocated_memory_mb below what was
        // added.
```

crates/spurd/src/agent_server.rs
```diff
@@ -1433,25 +1472,27 @@ impl AgentService {
...
-        } else {
-            None
-        };
-
-        let gpu_ids = controller_gpu_ids.clone();
-        let result = AllocationResult {
-            cpu_ids: cpu_alloc.map(|a| a.cpu_ids).unwrap_or_default(),
-            gpu_ids: controller_gpu_ids,
-            memory_mb: spec.memory_per_node_mb,
         };
```
shiv-tyagi: Q: `allocate_for_job` returns None on both GPU-in-use and duplicate-job-id (retried LaunchJob). Both surface as `resource_exhausted("GPUs unavailable")`, and the warn claims GPUs are in use even for a CPU-only duplicate. Should a duplicate job_id be distinguished (e.g. treat an already-owned reservation as idempotent success, or return a distinct status) so a retried RPC isn't reported as GPU exhaustion?

crates/spurd/src/agent_server.rs
```diff
@@ -1413,6 +1451,7 @@ impl AgentService {
     /// Record controller-allocated GPUs and allocate local CPU/memory resources.
     async fn allocate_local_resources(
         &self,
+        job_id: u32,
         spec: &JobSpec,
         allocated: Option<&ResourceAllocations>,
     ) -> Result<(Option<AllocationResult>, Vec<u32>), Status> {
```
shiv-tyagi: Nit: the Option is always Some on the Ok path now. Consider returning `Result<(AllocationResult, Vec<u32>), Status>` and simplifying the call site's `.as_ref().map(...).unwrap_or_default()` to `alloc_result.cpu_ids.clone()`.

shiv-tyagi (dismissed):
LGTM

crates/spurd/src/agent_server.rs
yansun1996: Fixed to `loop's` in 3859df2.

crates/spur-sched/src/cons_tres.rs
yansun1996: Agreed, the invariant is the point. Applied your wording in 3859df2.

crates/spurd/src/agent_server.rs
yansun1996: Good catch — real API-clarity bug. `allocate_for_job` now returns `Result<AllocationResult, AllocError>` with distinct `GpusUnavailable` vs `DuplicateJob` variants. The caller maps GpusUnavailable → `resource_exhausted` and DuplicateJob → `already_exists` (the RPC was already accepted). I did not make it idempotent-success: on the Ok path we spawn a process and commit_job, and launch isn't idempotent, so returning the existing reservation would spawn a second process for a running job. Fixed in 3859df2.

crates/spurd/src/agent_server.rs
yansun1996: Correct, the Option was dead weight. `allocate_local_resources` now returns `Result<(AllocationResult, Vec<u32>), Status>` and the call site is `alloc_result.cpu_ids.clone()`. Done alongside the comment-3 refactor in 3859df2.

yansun1996:
**Live re-test (2-node cluster) — JobHoldMaxRequeue confirmed fixed.**

Ran the two canonical SPUR-65 e2e tests against the PR head on a real 2-node cluster:

- `test_chronic_preemption_never_holds_job` — **PASSED** (low-pri job preempted+requeued more times than `max_batch_requeue`; never held with `JobHoldMaxRequeue`, returns to RUNNING each cycle).
- `test_failed_launch_releases_reservation` — **PASSED** (node recovers after a launch failure instead of black-holing all future dispatches).

```
2 passed in 59.83s
```

Confirmed: this PR fixes the JobHoldMaxRequeue issue.

shiv-tyagi (approved):

PR #451, opened 2026-07-16
fix(spurctld): delete sacctmgr users atomically

### LINKED ISSUE #435: sacctmgr delete user reports success but does not remove the user/association
## What happened

Running `sacctmgr delete user name=<user> account=<account>` prints a success confirmation ("Deleting user ... / Done."), but the underlying row is not actually removed from the `users`/`associations` tables.

## Expected

The user/association row should be removed from the database after a successful `sacctmgr delete user`, matching the reported outcome.

## How to reproduce

1. `sacctmgr add user name=<user> account=<account>`
2. `sacctmgr delete user name=<user> account=<account>` — reports success
3. Query the database directly (or `sacctmgr show user`) — the row is still present

Found incidentally while live-verifying #427 on a real cluster.

hnotshe:
Fixes #435.

`sacctmgr delete user` could report `Done.` without deleting anything because the all-accounts form sent an empty account while the database required an exact account match. The RPC also ignored zero affected rows.

This change deletes matching `users` and `associations` rows in one transaction, treats an omitted account as all accounts, and returns `NotFound` when no row was removed instead of reporting false success.

Testing:
- focused PostgreSQL regression covering one-account deletion, all-account deletion, association cleanup, and repeated deletion
- `cargo fmt --all --check`
- `cargo clippy --workspace --exclude spur-ffi --all-targets --locked`
- `cargo test --locked`

### DIFF
--- a/crates/spurctld/src/accounting/db.rs
+++ b/crates/spurctld/src/accounting/db.rs
@@ -855,19 +855,22 @@ pub async fn add_user<'a>(
     Ok(())
 }
 
-/// Remove a user from an account.
-pub async fn remove_user(pool: &PgPool, user: &str, account: &str) -> anyhow::Result<()> {
-    sqlx::query("DELETE FROM users WHERE name = $1 AND account = $2")
+/// Remove a user from one account, or every account when `account` is empty.
+pub async fn remove_user(pool: &PgPool, user: &str, account: &str) -> anyhow::Result<u64> {
+    let mut tx = pool.begin().await?;
+    let associations =
+        sqlx::query("DELETE FROM associations WHERE user_name = $1 AND ($2 = '' OR account = $2)")
+            .bind(user)
+            .bind(account)
+            .execute(&mut *tx)
+            .await?;
+    let users = sqlx::query("DELETE FROM users WHERE name = $1 AND ($2 = '' OR account = $2)")
         .bind(user)
         .bind(account)
-        .execute(pool)
+        .execute(&mut *tx)
         .await?;
-    sqlx::query("DELETE FROM associations WHERE user_name = $1 AND account = $2")
-        .bind(user)
-        .bind(account)
-        .execute(pool)
-        .await?;
-    Ok(())
+    tx.commit().await?;
+    Ok(associations.rows_affected() + users.rows_affected())
 }
 
 /// List users, joining each one's own association row for `default_qos`/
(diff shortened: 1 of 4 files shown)


yansun1996:
The core fix looks correct: a single transaction with proper rollback on partial failure (`?` drops the tx before commit), `($2 = '' OR account = $2)` correctly meaning all-accounts, the summed `rows_affected()` mapping zero → `NotFound`, and the test asserting row absence rather than just return codes. A couple of the notes inline are pre-existing rather than introduced here — flagged just for context.

One additional note: the new regression test is `#[ignore]`'d (it needs `DATABASE_URL`), so `cargo test --locked` won't run it in CI. That's consistent with the other DB tests, but it does mean the guard for this fix isn't exercised by default CI.

crates/spurctld/src/accounting/grpc.rs
```diff
@@ -360,9 +360,17 @@ impl SlurmAccounting for AccountingService {
         request: Request<RemoveUserRequest>,
     ) -> Result<Response<()>, Status> {
         let req = request.into_inner();
-        db::remove_user(&self.pool, &req.user, &req.account)
+        let deleted = db::remove_user(&self.pool, &req.user, &req.account)
```
yansun1996, line 463: Minor / pre-existing: the in-memory `AssociationCache` is not invalidated on delete (the refresh loop is the only update path), so a just-deleted association can still validate a job submission until the next refresh. This is not introduced here (add/delete-account share the same lag) — flagging for context, not a blocker.

crates/spurctld/src/accounting/db.rs
```diff
@@ -557,19 +557,22 @@ pub async fn add_user(
...
-        .bind(user)
-        .bind(account)
-        .execute(pool)
-        .await?;
-    sqlx::query("DELETE FROM associations WHERE user_name = $1 AND account = $2")
+/// Remove a user from one account, or every account when `account` is empty.
+pub async fn remove_user(pool: &PgPool, user: &str, account: &str) -> anyhow::Result<u64> {
+    let mut tx = pool.begin().await?;
+    let associations =
+        sqlx::query("DELETE FROM associations WHERE user_name = $1 AND ($2 = '' OR account = $2)")
```
yansun1996, line 862: Minor: `usage` rows are retained after a user is removed (no FK, not touched here). This looks intentional — Slurm keeps sacct/usage history after user deletion — just confirming it is the intended behavior.

crates/spurctld/src/accounting/grpc.rs
```diff
@@ -360,9 +360,17 @@ impl SlurmAccounting for AccountingService {
...
             .await
             .map_err(|e| Status::internal(e.to_string()))?;
+        if deleted == 0 {
+            let target = if req.account.is_empty() {
+                format!("user '{}'", req.user)
+            } else {
+                format!("user '{}' in account '{}'", req.user, req.account)
+            };
+            return Err(Status::not_found(format!("{target} does not exist")));
```
yansun1996, line 472: Correct server-side behavior. Worth noting the downstream effect: the `sacctmgr delete user` CLI path surfaces this `NotFound` as an error with a non-zero exit, whereas real Slurm prints "Nothing deleted" and exits 0. This appears to be the intended stricter behavior (it is the point of the fix) — flagging the change for any scripts that expect exit 0 on a no-op delete.

crates/spurctld/src/accounting/grpc.rs
yansun1996: Confirmed pre-existing — add and delete both rely on the refresh loop, neither invalidates immediately. Tracking as a separate follow-up rather than expanding this PR.

crates/spurctld/src/accounting/db.rs
yansun1996: Confirmed intended — matches Slurm keeping usage/sacct history after user deletion, and is consistent with how job rows are retained. No change needed.

crates/spurctld/src/accounting/grpc.rs
yansun1996: Confirmed intended — erroring on a no-op delete is the point of the fix. Matching Slurm exit-0 "Nothing deleted" would be a separate CLI change if we decide we want that parity later.

yansun1996: Went ahead and added the CLI parity in a follow-up commit: a no-op `delete user` now prints "Nothing deleted." and exits 0, matching Slurm. Real errors still propagate non-zero, and the server still returns NotFound so the no-op stays truthful over the wire.

yansun1996 (dismissed):

crates/spurctld/src/accounting/grpc.rs
yansun1996: Filed as ROCm/spur#543 for the follow-up.

yansun1996 (dismissed):

yansun1996:
Added some e2e test cases to improve the test coverage on the atomic operation

yansun1996 (approved):

PR #453, opened 2026-07-16
fix(spur-sched): spread equal-weight jobs across idle nodes

### LINKED ISSUE #404: spurctld: backfill scheduler concentrates rapid small jobs on one node instead of spreading across idle nodes
## Problem

On an idle or lightly-loaded cluster, small jobs submitted in quick succession all land on the **same node** instead of spreading across available idle nodes. A node keeps being selected until it is completely full, even when other nodes sit at zero allocation.

## Reproduced live (2-node cluster: node A = 24 CPU, node B = 10 CPU, both idle)

Submit 8 single-node single-CPU jobs back-to-back:

```
for i in $(seq 1 8); do spur sbatch -N1 -n1 -t 10:00 small.sh; done
spur squeue -o "%.6i %.2t %.4C %R"
```

Result — all 8 on one node, the other untouched:

```
 JOBID ST CPUS NODELIST(REASON)
     1  R    1 node-B
     2  R    1 node-B
     3  R    1 node-B
     4  R    1 node-B
     5  R    1 node-B
     6  R    1 node-B
     7  R    1 node-B
     8  R    1 node-B
```

```
node-B  CPUAlloc=8   (received every job)
node-A  CPUAlloc=0   (idle the whole time)
```

The same 8 jobs submitted with `--spread-job` distribute evenly (4 CPUs on each node), which confirms the concentration is a node-selection choice, not a capacity or availability constraint.

## Expected

Back-to-back small jobs on an otherwise-idle cluster should distribute across idle nodes, not stack onto a single node until it saturates. A node that already holds a job should lose a placement tie to a completely empty node.

## Notes / observations

- Both nodes were `idle` with default (equal) weight, so the placement tie between them is never broken toward the less-loaded node on the default submission path.
- The behavior is independent of the periodic scheduler tick: each `sbatch` wakes the scheduler immediately (event-driven), and a just-started job's allocation *is* applied before the next job is scheduled — but an updated-yet-not-full node is still selected again.
- `--spread-job` avoids it, but it is opt-in per job (and only on `sbatch`, not `srun`/`salloc`), so it does not address cluster-wide behavior for users who don't set it.
- Real-world impact reported on a production cluster: multiple users' `sbatch` submissions all pinned to the same two nodes while other partition nodes were free, causing contention/oversubscription on those nodes.

## Environment

- Spur controller (`spurctld`), backfill scheduler, single-controller (1-member Raft).
- Verified on the `main`-equivalent controller deployed to the lab testbed.

hnotshe:
## Summary

- Prefer the least-loaded node when candidate start time and scheduling weight are equal.
- Use timeline allocation at the candidate start time, including current node allocations and assignments made earlier in the same scheduler cycle.
- Apply the same timeline-aware load metric to the `--spread-job` secondary ordering.
- Cover both same-cycle reservations and allocations persisted across separate scheduler cycles.

The previous ordering left equal-weight candidates in fixed cluster order, causing rapid small jobs to concentrate on one node.

Closes #404

## Testing

- `cargo test -p spur-sched --locked`
- Full x86_64 Linux CI-equivalent validation: formatting, SPDX headers, dependency policy, warning-free clippy, all-target build, and locked workspace tests

### DIFF
--- a/crates/spur-sched/src/backfill.rs
+++ b/crates/spur-sched/src/backfill.rs
@@ -190,30 +190,36 @@ impl Scheduler for BackfillScheduler {
                 )
             });
 
+            let allocated_cpus_at_start: HashMap<usize, u32> = node_starts
+                .iter()
+                .map(|(ni, start)| (*ni, self.timelines[*ni].accumulated_at(*start).cpus))
+                .collect();
+
             // For --spread-job, sort by least-loaded (ascending alloc) so we
             // prefer nodes with the most available resources. For normal jobs,
-            // sort by earliest start time. For weighted nodes, prefer higher
-            // weight (descending).
+            // sort by earliest start time, then higher weight, then least load.
             if job.spec.spread_job {
                 node_starts.sort_by(|(a_ni, a_t), (b_ni, b_t)| {
                     // Primary: earliest start time
                     a_t.cmp(b_t)
                         // Secondary: least allocated CPUs (ascending = most free)
                         .then_with(|| {
-                            cluster.nodes[*a_ni]
-                                .alloc_resources
-                                .cpus
-                                .cmp(&cluster.nodes[*b_ni].alloc_resources.cpus)
+                            allocated_cpus_at_start[a_ni].cmp(&allocated_cpus_at_start[b_ni])
                         })
                 });
             } else {
-                // Default: sort by time, then prefer higher-weight nodes
+                // Default: sort by time, then prefer higher-weight nodes. Equal
+                // weights favor the least-loaded node.
                 node_starts.sort_by(|(a_ni, a_t), (b_ni, b_t)| {
-                    a_t.cmp(b_t).then_with(|| {
-                        cluster.nodes[*b_ni]
-                            .weight
-                            .cmp(&cluster.nodes[*a_ni].weight)
-                    })
+                    a_t.cmp(b_t)
+                        .then_with(|| {
+                            cluster.nodes[*b_ni]
+                                .weight
+                                .cmp(&cluster.nodes[*a_ni].weight)
+                        })
+                        .then_with(|| {
+                            allocated_cpus_at_start[a_ni].cmp(&allocated_cpus_at_start[b_ni])
+                        })
                 });
             }
 
@@ -1018,6 +1024,39 @@ mod tests {
         assert_eq!(assignments.len(), 1);
         // Should prefer least-loaded node (node003)
         assert_eq!(assignments[0].nodes[0], "node003");
+    }
+
+    #[test]
+    fn test_equal_weight_jobs_spread_across_least_loaded_nodes() {
+        let mut sched = BackfillScheduler::new(100);
+        let mut nodes = make_nodes(2);
+        nodes[0].total_resources.cpus = 24;
+        nodes[1].total_resources.cpus = 10;
+
+        let partitions = vec![Partition {
+            name: "default".into(),
+            ..Default::default()
+        }];
+        let pending: Vec<Job> = (1..=8).map(|id| make_job(id, 1, 1)).collect();
+        let cluster = ClusterState {
+            nodes: &nodes,
+            partitions: &partitions,
+            reservations: &[],
+            topology: None,
+        };
+
+        let assignments = sched.schedule(&pending, &cluster);
+        let node001_jobs = assignments
+            .iter()
+            .filter(|assignment| assignment.nodes == ["node001"])
+            .count();
+        let node002_jobs = assignments
+            .iter()
+            .filter(|assignment| assignment.nodes == ["node002"])
+            .count();
+
+        assert_eq!(node001_jobs, 4);
+        assert_eq!(node002_jobs, 4);
     }
 
     #[test]
--- a/crates/spur-sched/src/backfill.rs
+++ b/crates/spur-sched/src/backfill.rs
@@ -1060,6 +1060,45 @@ mod tests {
     }
 
     #[test]
+    fn test_equal_weight_jobs_spread_across_scheduler_cycles() {
+        let mut sched = BackfillScheduler::new(100);
+        let mut nodes = make_nodes(2);
+        nodes[0].total_resources.cpus = 24;
+        nodes[1].total_resources.cpus = 10;
+
+        let partitions = vec![Partition {
+            name: "default".into(),
+            ..Default::default()
+        }];
+
+        for id in 1..=8 {
+            let pending = vec![make_job(id, 1, 1)];
+            let cluster = ClusterState {
+                nodes: &nodes,
+                partitions: &partitions,
+                reservations: &[],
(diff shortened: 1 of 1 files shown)


yansun1996:
Nice fix — the tie-break is deterministic (stable sort over `u32` weights, no float fragility) and the added regression genuinely fails under the pre-fix behavior. No blocking issues; a few non-blocking notes inline.

One small doc suggestion: the `--spread-job` path's secondary key also changed (`alloc_resources.cpus` → `accumulated_at(start).cpus`) — a nice improvement, but might be worth a line in the description so that path gets re-validated too.

crates/spur-sched/src/backfill.rs
```diff
@@ -1020,6 +1026,39 @@ mod tests {
         assert_eq!(assignments[0].nodes[0], "node003");
     }
 
+    #[test]
+    fn test_equal_weight_jobs_spread_across_least_loaded_nodes() {
```
yansun1996, line 1030: The test covers the same-cycle path (all 8 jobs in one `schedule()` call, load accumulating via same-cycle `reserve`). The scenario in #404 — load arriving across separate scheduler cycles via `node.alloc_resources` seeding (lines 102-109) — goes through a different mechanism that this test does not exercise. Since that seeding path is what manifested the original bug, a future refactor there could regress it without failing tests. Might be worth a case that pre-seeds `alloc_resources`.

crates/spur-sched/src/backfill.rs
```diff
@@ -190,30 +190,36 @@ impl Scheduler for BackfillScheduler {
...
-                            .cmp(&cluster.nodes[*a_ni].weight)
-                    })
+                    a_t.cmp(b_t)
+                        .then_with(|| {
+                            cluster.nodes[*b_ni]
+                                .weight
+                                .cmp(&cluster.nodes[*a_ni].weight)
+                        })
+                        .then_with(|| {
+                            allocated_cpus_at_start[a_ni].cmp(&allocated_cpus_at_start[b_ni])
```
yansun1996, line 221: Minor: the tie-break ranks by absolute allocated CPUs rather than fraction-of-capacity. On heterogeneous nodes (e.g. the 24 vs 10 CPU nodes in the test) the smaller node reaches proportional saturation sooner. This is a reasonable simple default — just flagging that "least-loaded" is capacity-blind on mixed-capacity clusters.

crates/spur-sched/src/backfill.rs
```diff
@@ -190,30 +190,36 @@ impl Scheduler for BackfillScheduler {
                 )
             });
 
+            let allocated_cpus_at_start: HashMap<usize, u32> = node_starts
+                .iter()
+                .map(|(ni, start)| (*ni, self.timelines[*ni].accumulated_at(*start).cpus))
```
yansun1996, line 195: Minor: the load metric here is CPU-only. For a GPU-first scheduler, two nodes with equal CPU but differing GPU occupancy tie and fall through to insertion order, so GPU-heavy jobs can still concentrate. Could be a follow-up if GPU spreading is in scope.

hnotshe: Agreed. This PR is scoped to the CPU-only reproduction in #404. GPU-aware tie-breaking would need a broader resource-aware load policy, so I left it out rather than expanding this targeted fix.

crates/spur-sched/src/backfill.rs
hnotshe: Agreed. I kept absolute allocated CPU count here because this PR is scoped to breaking the empty-vs-loaded tie described in #404. Capacity-normalized placement would be a broader scheduling-policy change on heterogeneous clusters, so I left it out of this fix.

crates/spur-sched/src/backfill.rs
hnotshe: Good catch. I added `test_equal_weight_jobs_spread_across_scheduler_cycles`, which schedules one job per `schedule()` call and persists each assignment into `node.alloc_resources` before the next call. That covers the seeding path from the live reproduction, while the existing test continues to cover same-cycle reservations.

crates/spur-sched/src/backfill.rs
```diff
@@ -190,30 +190,36 @@ impl Scheduler for BackfillScheduler {
...
                 node_starts.sort_by(|(a_ni, a_t), (b_ni, b_t)| {
                     // Primary: earliest start time
                     a_t.cmp(b_t)
                         // Secondary: least allocated CPUs (ascending = most free)
                         .then_with(|| {
-                            cluster.nodes[*a_ni]
-                                .alloc_resources
-                                .cpus
-                                .cmp(&cluster.nodes[*b_ni].alloc_resources.cpus)
+                            allocated_cpus_at_start[a_ni].cmp(&allocated_cpus_at_start[b_ni])
```
yansun1996, line 207: Optional: `allocated_cpus_at_start[a_ni]` is safe today since the map is built from `node_starts` after the reservation `retain`, so every sorted index is a key. If a future change ever built this map before the filtering (or over a superset), the index would panic. `\.get(a_ni).copied().unwrap_or(0)` would make it robust to that without changing current behavior — feel free to leave as-is.

crates/spur-sched/src/backfill.rs
```diff
@@ -190,30 +190,36 @@ impl Scheduler for BackfillScheduler {
...
                         })
                 });
             } else {
-                // Default: sort by time, then prefer higher-weight nodes
+                // Default: sort by time, then prefer higher-weight nodes. Equal
+                // weights favor the least-loaded node.
```
yansun1996, line 212: Small doc nit: for multi-node `tree`/`block` jobs the topology reorder below rebuilds `node_starts` by locality, which supersedes this least-loaded key. Might be worth a half-line noting the tie-break applies to the single-node / no-topology path, so a future reader doesn't assume spread holds for gang jobs too.

yansun1996 (approved):
Re-reviewed after the latest changes — thanks for adding the cross-cycle coverage. `test_equal_weight_jobs_spread_across_scheduler_cycles` exercises exactly the seeding path from #404: because the pre-fix normal path ordered only by (start, weight) with no load key, equal-weight jobs fell through to cluster order and all landed on the larger node — so this test genuinely fails without the fix and passes with it. That fully addresses the earlier concern.

The tie-break reads cleanly and is deterministic (stable sort, integer keys), the map is built after the reservation `retain` so the sort indices are always present, and the load key only decides when start times are equal. Looks good to merge.

A couple of purely optional nits below — nothing blocking.

crates/spur-sched/src/backfill.rs
yansun1996: Confirmed — the new test drives one job per scheduler cycle and persists each allocation back into the node before the next cycle, so it goes through the seeding path and would place all jobs on the larger node under the old (start, weight)-only ordering. Genuine regression coverage. Thanks!

PR #455, opened 2026-07-16
fix(spurctld): reset stale node features/weight when NodeConfig stops…

biluriuday:
… matching

When a node's labels change so it no longer matches any [[nodes]] selector, the previously-applied features and weight persisted silently, causing the scheduler to act on stale policy. Reset features/weight to node defaults when no NodeConfig matches, applied consistently across node registration, label updates, and snapshot restore.

Introduce Node::DEFAULT_WEIGHT and Node::reset_config_policy so the reset path shares a single source of truth with Node::new.

## Submission Checklist

### DIFF
--- a/crates/spur-core/src/node.rs
+++ b/crates/spur-core/src/node.rs
@@ -268,10 +268,13 @@ pub struct Node {
 }
 
 fn default_weight() -> u32 {
-    1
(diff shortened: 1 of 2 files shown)


shiv-tyagi (approved):
LGTM, couldn't find anything other than a few nits. Ignoring those.

PR #456, opened 2026-07-16
docs: add partitioning guide for hostlist and label-based node matching

shiv-tyagi:
## Summary

Adds a Partitioning page to the deployment docs explaining how nodes join partitions. There was no doc covering node-to-partition matching; `native-host.rst` only showed a hostlist example in passing.

Covers:
- Label-based membership via `selector` (recommended), including how nodes advertise labels with `spurd --label`.
- Hostlist-based membership via `nodes`, noting it is Slurm-style bracket expansion (not a regex) plus the `ALL` short-circuit.
- The union semantics when both are set, and that the same dual-path applies to `[[nodes]]` blocks.
- Verification with `sinfo`.

## Approach

Label selectors are framed as the recommended approach because membership is decoupled from hostnames, which suits dynamic/elastic and heterogeneous clusters.

## Testing

- Every claim was verified live on the 4-node bare-metal cluster (`ALL`, bracket ranges/lists, comma-joined patterns, literal-not-regex, single- and multi-key selector AND, hostlist-OR-selector union, `--label` multi vs `SPUR_NODE_LABELS` single-label behavior).
- RST underlines, toctree wiring, and cross-references checked.

### DIFF
--- a/docs/deployment/index.rst
+++ b/docs/deployment/index.rst
@@ -8,3 +8,4 @@ Guides for deploying Spur in production across multiple nodes.
 
    native-host
    kubernetes
+   partitioning
--- a/docs/deployment/partitioning.rst
+++ b/docs/deployment/partitioning.rst
@@ -1,0 +1,94 @@
+Partitioning
+============
+
+Partitions (Slurm's queues) group nodes so jobs can target a subset of the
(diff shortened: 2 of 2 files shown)


biluriuday (approved):

sgopinath1 (approved):
LGTM

PR #457, opened 2026-07-16
fix(spur-cli): wire format engine into sacctmgr show qos

shiv-tyagi:
## Summary

- `sacctmgr show qos` printed a fixed 6-column table that omitted GrpTRES, MaxTRES, MaxTRESPU, Description, MaxSubmitPU, and GrpWall. The backend already stores and returns these fields via `ListQos`/`QosInfo`, so this was purely a CLI display gap.
- Wire the existing format engine (`parse_named_format` + `format_row`) into the `show qos` handler, matching the pattern already used by `sacct`. The default output now includes 11 columns (up from 6), adding MaxJobsPU, MaxSubmitPU, MaxWall, GrpWall, MaxTRES, MaxTRESPU, and GrpTRES. The `Prio` header is renamed to `Priority` for Slurm parity.
- Support `format=Name,GrpTRES,...` for field selection, `format=all` for every field, and `where name=` for filtering to specific QOS entries. Unrecognized field names produce a clear error listing valid options.
- Extract `print_header()` and `resolve_format()` into `format_engine` to deduplicate the header+separator and format-resolution patterns shared by `sacct` and `sacctmgr`.

**Breaking change:** The default `sacctmgr show qos` output now has 11 columns instead of 6, and the `Prio` header is renamed to `Priority`. Anything parsing fixed column positions will need updating.

## Test plan

- [x] 9 new unit tests in `sacctmgr.rs`: TRES rendering, field-spec aliases, default/all format coverage, zero-field blanking (including GrpWall), name filtering with whitespace trimming, unknown-field rejection, empty-format handling
- [x] 4 E2E tests in `test_accounting.py` (using existing `accounting_cluster` fixture with Docker Postgres): default output shows TRES columns, `format=` selects specific fields, `where name=` filters correctly, all limit columns render with correct values
- [x] Full smoke test on 4-node bare-metal cluster: `sacctmgr show qos` (default, format=all, format=Name,GrpTRES,MaxTRES, format=Name,GrpWall, where name=, where name=comma,list, where name=nonexistent, format=Bogus), `sacctmgr show account/user`, `sacct`, `sinfo`, `squeue`, `sbatch` job lifecycle
- [x] `cargo test -p spur-cli` passes (182 tests)

Closes SPUR-60

### DIFF
--- a/crates/spur-cli/src/format_engine.rs
+++ b/crates/spur-cli/src/format_engine.rs
@@ -254,7 +254,8 @@ pub fn resolve_format(
         Some(fmt) if fmt.eq_ignore_ascii_case("all") => Ok(parse_format(all_format, header_map)),
         Some(fmt) => {
             let f = parse_named_format(fmt, name_to_spec, header_map);
-            if f.is_empty() {
+            let has_fields = f.iter().any(|t| matches!(t, FormatToken::Field(_)));
+            if !has_fields {
                 anyhow::bail!("no recognized fields in format='{fmt}'. Valid fields: {valid_hint}");
             }
             Ok(f)
--- a/crates/spur-cli/src/format_engine.rs
+++ b/crates/spur-cli/src/format_engine.rs
@@ -225,6 +225,44 @@ fn format_field(value: &str, field: &FormatField) -> String {
     }
 }
 
+/// Print a header row followed by a separator line (dashes under text, spaces under spaces).
+pub fn print_header(tokens: &[FormatToken]) {
+    let header = format_header(tokens);
+    let sep: String = header
+        .chars()
+        .map(|c| if c == ' ' { ' ' } else { '-' })
+        .collect();
+    println!("{header}");
+    println!("{sep}");
+}
+
+/// Resolve a `format=` parameter into tokens.
+///
+/// - `"all"` (case-insensitive) expands `all_format` via `parse_format`.
+/// - Any other value is parsed as comma-separated field names via `parse_named_format`.
+/// - Returns an error if the value contains no recognized field names (with `valid_hint`).
+/// - `None` falls back to `default_format` via `parse_format`.
+pub fn resolve_format(
+    format_param: Option<&str>,
+    default_format: &str,
+    all_format: &str,
+    name_to_spec: &dyn Fn(&str) -> Option<char>,
+    header_map: &dyn Fn(char) -> &'static str,
+    valid_hint: &str,
+) -> anyhow::Result<Vec<FormatToken>> {
+    match format_param {
+        Some(fmt) if fmt.eq_ignore_ascii_case("all") => Ok(parse_format(all_format, header_map)),
+        Some(fmt) => {
+            let f = parse_named_format(fmt, name_to_spec, header_map);
+            if f.is_empty() {
+                anyhow::bail!("no recognized fields in format='{fmt}'. Valid fields: {valid_hint}");
+            }
+            Ok(f)
+        }
+        None => Ok(parse_format(default_format, header_map)),
+    }
+}
+
 /// Default squeue format string (matches Slurm default).
 pub const SQUEUE_DEFAULT_FORMAT: &str = "%.18i %.9P %.8j %.8u %.2t %10M %6D %R";
 
--- a/crates/spur-cli/src/sacctmgr.rs
+++ b/crates/spur-cli/src/sacctmgr.rs
@@ -687,20 +687,15 @@ async fn show(entity: &str, params: &[String], addr: &str) -> Result<()> {
                 .into_inner()
                 .qos_list;
 
+            let has_name_filter = p.contains_key("name");
             if let Some(name_filter) = p.get("name") {
-                let names: Vec<&str> = name_filter.split(',').collect();
+                let names: Vec<&str> = name_filter.split(',').map(str::trim).collect();
                 qos_list.retain(|q| names.iter().any(|n| n.eq_ignore_ascii_case(&q.name)));
             }
 
             format_engine::print_header(&fields);
 
-            let rows: Vec<&QosInfo> = if qos_list.is_empty() {
-                vec![]
-            } else {
-                qos_list.iter().collect()
-            };
-
-            if rows.is_empty() {
+            if qos_list.is_empty() && !has_name_filter {
                 let default_qos = QosInfo {
                     name: "normal".into(),
                     preempt_mode: "off".into(),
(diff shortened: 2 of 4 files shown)


crates/spur-cli/src/sacctmgr.rs
```diff
@@ -512,46 +513,56 @@ async fn show(entity: &str, params: &[String], addr: &str) -> Result<()> {
...
-                .context("ListQos RPC failed")?;
+                .context("ListQos RPC failed")?
+                .into_inner()
+                .qos_list;
 
-            let qos_list = resp.into_inner().qos_list;
+            if let Some(name_filter) = p.get("name") {
+                let names: Vec<&str> = name_filter.split(',').collect();
+                qos_list.retain(|q| names.iter().any(|n| n.eq_ignore_ascii_case(&q.name)));
+            }
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed. Now trims whitespace after splitting on commas, so `name=alpha, beta` matches correctly.

crates/spur-cli/src/sacctmgr.rs
```diff
@@ -512,46 +513,56 @@ async fn show(entity: &str, params: &[String], addr: &str) -> Result<()> {
...
+                qos_list.iter().collect()
+            };
+
+            if rows.is_empty() {
+                let default_qos = QosInfo {
+                    name: "normal".into(),
+                    preempt_mode: "off".into(),
+                    usage_factor: 1.0,
+                    ..Default::default()
+                };
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed. The synthetic "normal" row now only appears when no name filter is active and the backend genuinely returned an empty list. With `where name=nonexistent`, the output is just the header with no rows.

crates/spur-cli/src/format_engine.rs
```diff
@@ -225,6 +225,44 @@ fn format_field(value: &str, field: &FormatField) -> String {
...
+) -> anyhow::Result<Vec<FormatToken>> {
+    match format_param {
+        Some(fmt) if fmt.eq_ignore_ascii_case("all") => Ok(parse_format(all_format, header_map)),
+        Some(fmt) => {
+            let f = parse_named_format(fmt, name_to_spec, header_map);
+            if f.is_empty() {
+                anyhow::bail!("no recognized fields in format='{fmt}'. Valid fields: {valid_hint}");
+            }
+            Ok(f)
+        }
```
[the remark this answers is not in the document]
shiv-tyagi: This matches Slurm's behavior. Slurm's `sacctmgr` silently drops unknown field names as long as at least one field is recognized. Erroring on partial typos would break compatibility for users migrating from Slurm.

crates/spur-cli/src/sacctmgr.rs
```diff
@@ -678,4 +757,111 @@ mod tests {
...
+                ..Default::default()
+            },
+            QosInfo {
+                name: "gamma".into(),
+                ..Default::default()
+            },
+        ];
+        let filter = "alpha, gamma";
+        let names: Vec<&str> = filter.split(',').map(str::trim).collect();
+        list.retain(|q| names.iter().any(|n| n.eq_ignore_ascii_case(&q.name)));
```
yansun1996, line 1181: Nice coverage here. One small thing: this test re-implements the split/trim/case-insensitive `retain` filter inline rather than calling the production show path, so it wouldn't catch a regression if the real filter changes. Might be worth extracting a small helper (e.g. `filter_qos_by_name(&mut Vec<QosInfo>, &str)`) used by both the show path and this test. Very minor — feel free to defer.

crates/spur-cli/src/sacctmgr.rs
```diff
@@ -583,6 +589,79 @@ async fn show(entity: &str, params: &[String], addr: &str) -> Result<()> {
     }
 }
 
+const QOS_DEFAULT_FORMAT: &str = "%-15N %-8p %-10P %-12U %-20G %-20T %-20V %-10J %-10W";
```
yansun1996: Worth a note in the PR description: this changes the default `sacctmgr show qos` columns from 6 to 9 (adds GrpTRES/MaxTRES/MaxTRESPU) and renames the `Prio` header to `Priority`. Nice move toward Slurm parity and it surfaces the TRES data — just flagging since anything parsing fixed columns would be affected.

yansun1996:
Thanks for wiring this through the shared format engine — the extraction of `print_header`/`resolve_format` is clean, and sacct/squeue are untouched. Build/clippy/tests all pass locally and the e2e assertions look solid.

No blockers from my side. A couple of optional parity fast-follows (not for this PR): `MaxWall` still renders as raw minutes vs Slurm's `HH:MM:SS`, and `usage_factor` shows `1` instead of `1.000000`.

One thing to flag separately (not this PR's fault): `GrpWall`/`grp_wall_minutes` isn't shown, but that's because the QOS create/list server path doesn't carry the field yet (proto field 12 is dropped in `create_qos`/`list_qos`), so there's nothing to display. Probably worth a standalone issue on the server side rather than blocking this display change.

yansun1996:
Heads-up on a significant overlap: my open PR #427 (SPUR-19) also rewrites `sacctmgr show qos` — it widens the same table with `GrpTRES`/`MaxTRES`/`MaxTRESPU` and the `Prio`→`Priority` rename, so our two changes land on exactly the same lines and will conflict on merge.

Your approach here is the better one — routing through the shared format engine is a superset of what #427 does (it also exposes `Descr` and `MaxSubmitPU` via `format=`, which the hardcoded table can't). So rather than both of us touching this, I'd suggest this PR **wait for #427 to merge first**, then rebase on top — I'll drop the now-redundant `show qos` display changes from #427 so there's a clean base for your format-engine version. That keeps #427 focused on the enforcement/write-path work and lets your PR own the display cleanly.

Happy to coordinate on timing. Holding my detailed review notes until the rebase so they're against the final shape.

shiv-tyagi:
Rebased on #427.

yansun1996 (approved):

crates/spur-cli/src/sacctmgr.rs
shiv-tyagi: Agree, extracting a helper is cleaner. Let us get this merged for today's release and I'll follow up with the refactor.

crates/spur-cli/src/sacctmgr.rs
shiv-tyagi: Done. Updated the PR description to call out the column count change (6 to 11) and the Prio to Priority rename.

PR #458, opened 2026-07-16
fix(spur-core): derive version from cargo version plus git sha

shiv-tyagi:
## What

`--version` showed a stale version (`spur 0.5.0 (v0.3.0-262-gfd995b2-dirty)` or similar) because the string was built from `git describe --always --dirty`.

## Why

`git describe` anchors to the most recent *annotated* tag. `v0.3.0` is annotated; `v0.4.x`/`v0.5.0` are lightweight and not ancestors of `main`'s HEAD, so `describe` walks back to `v0.3.0`. The result always drifts from the declared package version.

## Approach

Make `Cargo.toml` authoritative and reduce git to build provenance (the Kubernetes `gitVersion`/`gitCommit`/`gitTreeState` model):

- `build.rs` captures the short commit hash (`SPUR_GIT_SHA`) and dirty state (`SPUR_GIT_DIRTY`). It honors env overrides first (CI/Docker, where `.git` is absent) and falls back to shelling out to git for local builds.
- `version_string()` formats `spur <cargo-version> (<sha>[-dirty])`, or plain `spur <version>` when no git info is available.
- Docker `ARG`/`ENV` and the release/nightly workflows updated to pass sha/dirty. Both release and nightly pass the short sha, so release and nightly binaries/images carry commit provenance.

## Output

| Build | Version |
|---|---|
| tagged release | `spur 0.5.0 (<sha>)` |
| nightly | `spur 0.5.0 (<sha>)` |
| local, clean | `spur 0.5.0 (<sha>)` |
| local, dirty | `spur 0.5.0 (<sha>-dirty)` |
| no git (tarball) | `spur 0.5.0` |

## Testing

- `cargo test -p spur-core version`: 4/4 pass.
- Verified all five override/fallback scenarios produce the expected `--version` output across `spur`, `spurctld`, `spurd`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,8 +42,6 @@ jobs:
           target: dist
           push: false
           outputs: type=local,dest=./dist
-          build-args: |
-            SPUR_GIT_DESCRIBE=${{ env.TAG }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
@@ -156,8 +154,6 @@ jobs:
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
-          build-args: |
-            SPUR_GIT_DESCRIBE=${{ env.TAG }}
           cache-from: type=gha,scope=release-docker
           cache-to: type=gha,scope=release-docker,mode=max
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,6 +34,10 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
 
+      - name: Set metadata
+        id: gitmeta
+        run: echo "short_sha=$(git rev-parse --short=8 HEAD)" >> "$GITHUB_OUTPUT"
+
       - name: Build portable binaries (Ubuntu 22.04+ compatible)
         uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
         with:
@@ -42,6 +46,9 @@ jobs:
           target: dist
           push: false
           outputs: type=local,dest=./dist
+          build-args: |
+            SPUR_GIT_SHA=${{ steps.gitmeta.outputs.short_sha }}
+            SPUR_GIT_DIRTY=false
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -42,8 +42,6 @@ jobs:
           target: dist
           push: false
           outputs: type=local,dest=./dist
-          build-args: |
-            SPUR_GIT_DESCRIBE=${{ env.TAG }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
```
yansun1996, line 53: I think this release build (and the runtime image build further down at line 149) needs to pass the git SHA as a build-arg — the nightly workflow does this now, but release doesn't, so the two paths ended up asymmetric. Since `.git/` is in `.dockerignore`, `build.rs` can't recover the SHA inside the container, so release binaries/images would report a bare `spur X.Y.Z` with no commit — a step back from the old scheme that embedded the tag via `SPUR_GIT_DESCRIBE`. The Dockerfile already declares `ARG SPUR_GIT_SHA`, so it should just need something like:
```yaml
          build-args: |
            SPUR_GIT_SHA=${{ github.sha }}
            SPUR_GIT_DIRTY=false
```
on both build-push steps (use a short-sha meta step if you'd prefer the abbreviated form).

yansun1996:
Nice cleanup — decoupling the authoritative version (Cargo.toml) from git provenance and dropping the fragile `git describe`/tag-equality logic is the right call, and `build.rs` degrades gracefully outside a git repo (tarball/vendored/crates.io) without hard-failing. version.rs is panic-free and the format stays script-parseable.

One thing I'd flag before merge (left an inline note on release.yml): the release workflow no longer passes any git info to the Docker build, while nightly does — so released binaries/images would report a bare `spur X.Y.Z` with no SHA (`.git/` is dockerignored, so build.rs can't recover it in-container). That's the PR's own goal, so worth closing the gap by passing `SPUR_GIT_SHA` as a build-arg in both release build steps.

Two smaller, optional notes (non-blocking): `git status --porcelain` marks local dev builds `-dirty` on untracked files (broader than the old `git describe --dirty`) — `--untracked-files=no` would tighten it if you care; and `SPUR_GIT_DIRTY` handling / `env!` usage all checked out fine. Thanks for tackling this.

yansun1996 (changes requested):

.github/workflows/release.yml
shiv-tyagi: Fixed. Both release build steps now pass the short sha:
```yaml
build-args: |
  SPUR_GIT_SHA=${{ steps.gitmeta.outputs.short_sha }}
  SPUR_GIT_DIRTY=false
```
A `gitmeta` step computes `git rev-parse --short=8 HEAD` (8-char, matching nightly). Release binaries and images now report `spur X.Y.Z (<sha>)` instead of a bare version.

shiv-tyagi:
Thanks Yan. Addressed the release gap in the release.yml thread: both build steps now pass the short sha, so releases report `spur X.Y.Z (<sha>)`, symmetric with nightly.

On `--untracked-files=no`: keeping the porcelain default. Untracked files mean the tree isn't a clean checkout, so `-dirty` is accurate as build provenance. It only affects local dev builds; CI sets `SPUR_GIT_DIRTY` explicitly, so nightly/release are unaffected.

yansun1996 (approved):
LGTM

PR #459, opened 2026-07-16
feat(spurctld): auto-derive raft node_id from controller.peers position

biluriuday:
Derive each controller's Raft node_id from its position in controller.peers by matching the system hostname against each entry's host part (full name or first DNS label), so operators no longer need to set controller.node_id. This also supports hostnames the ordinal path could not, e.g. spur-node1.

Resolution precedence is explicit controller.node_id -> peers position -> hostname ordinal, extracted into a testable resolve_node_id(). Every path is validated to fall within 1..=peers.len(), ambiguous multi-matches are rejected, and an unmatched host fails fast at startup with a descriptive error instead of silently splitting the cluster. Startup logs the derivation source.

## Submission Checklist

### DIFF
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -269,33 +269,17 @@ pub struct ControllerConfig {
     #[serde(default = "default_one")]
     pub first_job_id: u32,
 
-    /// Raft peers for HA consensus. Each entry is "host:port" (Raft gRPC address).
-    /// If empty, single-node mode (no Raft, no replication).
+    /// Raft peers for HA consensus, "host:port" each. Empty = single-node mode.
+    /// Must be identically ordered on every controller: node ids derive from
+    /// list position, so a reordered list can form an inconsistent voter set.
     /// Example: ["node1:6821", "node2:6821", "node3:6821"]
-    ///
-    /// INVARIANT: every controller must be given a byte-identical,
-    /// identically-ordered `peers` list. Node ids derive from list position, so
-    /// a reordered list makes a host derive a different id for itself than its
-    /// peers map to it — at initial bootstrap this can form an inconsistent
-    /// voter set.
     #[serde(default)]
     pub peers: Vec<String>,
 
-    /// This node's Raft ID. Normally left unset. Only consulted in multi-node
-    /// mode (`peers` non-empty); single-node mode always uses id 1.
-    ///
-    /// In multi-node mode the id is resolved by precedence:
-    /// 1. this explicit value, if set;
-    /// 2. this host's position in `peers` (its hostname matched against each
-    ///    entry's host part, on the full name or first DNS label);
-    /// 3. the hostname ordinal (e.g. spurctld-2 → node_id 3), only for IP-only
-    ///    peer lists where hostname matching cannot work.
-    ///
-    /// The resolved id must fall within `1..=peers.len()`, else startup fails.
-    /// Only needed when peers are IP-only and cannot match a hostname. An
-    /// explicit id that is in range but disagrees with this host's positional
-    /// index in `peers` is accepted silently, so when set it must match that
-    /// position (id == index + 1) or peers will map the host to a different id.
+    /// This node's Raft ID. Normally unset; single-node mode always uses 1.
+    /// Otherwise resolved as: explicit value, else position in `peers` (by
+    /// hostname), else hostname ordinal (IP-only peers). Must be in
+    /// `1..=peers.len()` and, if set, equal its position (index + 1).
     pub node_id: Option<u64>,
 
     /// Listen address for Raft internal gRPC traffic (separate from client API).
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -275,11 +275,15 @@ pub struct ControllerConfig {
     #[serde(default)]
     pub peers: Vec<String>,
 
-    /// This node's Raft ID. Normally left unset. Resolution precedence is:
+    /// This node's Raft ID. Normally left unset. Only consulted in multi-node
+    /// mode (`peers` non-empty); single-node mode always uses id 1.
+    ///
+    /// In multi-node mode the id is resolved by precedence:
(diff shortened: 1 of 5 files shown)


crates/spurctld/src/raft.rs
```diff
@@ -759,10 +759,76 @@ impl RaftHandle {
...
+            "hostname {hostname:?} yields Raft node_id {id} via ordinal fallback, \
+             out of range for {n} configured controller.peers; fix the hostname \
+             to match its controller.peers entry or set controller.node_id"
+        );
+    }
+
+    anyhow::bail!(
+        "this host ({hostname:?}) matched no entry in controller.peers; fix the \
+         hostname to match its peers entry or set controller.node_id in spur.conf"
+    )
```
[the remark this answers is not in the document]
biluriuday: done

crates/spurctld/src/raft.rs
```diff
@@ -759,10 +759,94 @@ impl RaftHandle {
...
+    }
+}
+
+/// Resolve this node's Raft id for a multi-node cluster (`peers` non-empty).
+///
+/// Precedence: explicit `controller.node_id` -> position in `peers` -> hostname
+/// ordinal. Every path is validated to fall in `1..=peers.len()` so a
+/// mismatched id fails fast at startup instead of silently splitting the
+/// cluster (an id outside that range exists in no peer's `build_peer_map`).
+pub fn resolve_node_id(
```
yansun1996, line 790: Really nicely guarded derivation — the ambiguity error, range check, and the deliberate no-ordinal-fallback for hostname peers are all the right instincts, and the tests cover them well.

One case worth closing: the node_id is re-derived from live hostname + current `peers` on every boot, but the persisted Raft state (`{state_dir}/raft/vote.json`, log) belongs to whatever id this node ran under before. If someone inserts/removes a peer *before* this host in the list (or the host is renamed), the derived id shifts while the on-disk vote/log stays under the old id — openraft then runs with a mismatched identity and silently orphans committed state rather than failing cleanly. It's low-probability under a k8s StatefulSet (stable ordinals, append-only peers), but silent + correctness-fatal when it hits. A small guard would close it: persist the resolved id once (e.g. `{state_dir}/raft/node_id`) and `bail!` at startup if the newly-derived id differs. ~15 lines and it turns a silent corruption into a clear error.

crates/spur-core/src/config.rs
```diff
@@ -275,8 +275,18 @@ pub struct ControllerConfig {
...
     pub peers: Vec<String>,
 
-    /// This node's Raft ID. If not set, auto-derived from hostname ordinal
-    /// (e.g. spurctld-2 → node_id 3) or defaults to 1.
+    /// This node's Raft ID. Normally left unset. Only consulted in multi-node
+    /// mode (`peers` non-empty); single-node mode always uses id 1.
+    ///
+    /// In multi-node mode the id is resolved by precedence:
+    /// 1. this explicit value, if set;
+    /// 2. this host's position in `peers` (its hostname matched against each
```
yansun1996: Since the id derives from position in `peers`, it'd help to document the hard invariant that every controller must have a byte-identical, identically-ordered `peers` list — if one node's list is reordered, it derives a different id for the same host, which at initial bootstrap can form an inconsistent voter set. And a note that an explicit `node_id` which is in range but disagrees with this host's positional index is accepted silently (so explicit ids should match positional order). A couple of lines here could save an operator a confusing HA incident.

yansun1996:
Thanks for this — the self-identification logic is more careful than a naive index match: validating the id into `1..=peers.len()`, erroring on ambiguous multi-matches, and deliberately refusing the ordinal fallback for hostname peers are all the right calls, and the tests are genuinely adversarial (reorder/ambiguity/IPv6/out-of-range/no-fallback-for-hostnames) rather than happy-path.

Nothing here blocks under the intended k8s StatefulSet model (stable ordinals, append-only peers). The one item I'd suggest addressing in this PR is a restart guard against a silently-changed node_id — details in the inline note on `resolve_node_id`; it's ~15 lines and converts the only silent state-corruption path into a clear startup error.

The rest are document-the-invariant items (left an inline note on the config `node_id` doc): the peers list must be identical and identically ordered across nodes, explicit ids should match positional order, and the IP-only ordinal fallback is range-checked but not identity-checked. A migration note in the PR description for existing clusters with explicit node_ids would also help. Solid contribution overall.

yansun1996 (changes requested):
Thanks for this — the self-identification logic is well-built and the tests are genuinely adversarial. Requesting changes on one item before merge; the rest are document-the-invariant suggestions (left inline).

Please address:
- **Restart guard against a silently-changed node_id** (inline on `resolve_node_id`): the id is re-derived from hostname + current `peers` every boot, but persisted Raft state (`{state_dir}/raft/vote.json`, log) belongs to the previous id. If a peer is inserted/removed before this host, or the host is renamed, the derived id shifts and the node silently runs under a mismatched identity, orphaning committed state. Persisting the resolved id once (e.g. `{state_dir}/raft/node_id`) and `bail!`-ing on mismatch turns this silent corruption into a clear startup error (~15 lines).

Please also consider (non-blocking, docs-only):
- Document that `peers` must be byte-identical and identically ordered on every controller, and that an explicit `node_id` should match this host's positional index (an in-range-but-mismatched explicit id is currently accepted silently).
- A migration note for existing clusters that used explicit node_ids.

Nothing else here blocks under the k8s StatefulSet deployment model. Nice work overall.

crates/spurctld/src/raft.rs
biluriuday: addressed this

crates/spur-core/src/config.rs
biluriuday: done

yansun1996 (commented):
lgtm, pls avoid the large sections of comments in the code, then we can merge the PR

yansun1996 (approved):
LGTM

shiv-tyagi (approved):
LGTM

PR #465, opened 2026-07-17
fix(spur-cli): hide unregistered nodes from sinfo

### LINKED ISSUE #317: sinfo shows placeholder nodes from partition config before agent registration
## Problem

When a partition config explicitly lists node hostnames (e.g. `nodes = "gpu[001-008]"`), sinfo shows those nodes as "idle" immediately — even before any agent has registered. This is misleading in dynamic-first deployments where nodes only exist after an agent connects with a token.

## Current behavior

```toml
[[partitions]]
name = "gpu"
nodes = "gpu-node1,gpu-node2"
```

```
$ sinfo
PARTITION AVAIL  TIMELIMIT  NODES  STATE NODELIST
gpu*        up   infinite      2   idle gpu-node1,gpu-node2
```

No agent has registered. The nodes don't actually exist.

## Expected behavior

Nodes should only appear in sinfo after an agent successfully registers via `RegisterAgent` RPC. The partition's `nodes` field should act as an allowlist (which hostnames are permitted in this partition) rather than an inventory declaration.

Workaround: use `nodes = "ALL"` which only matches actually-registered nodes.

joshkmartinez:
## Summary

Configured partition hostlists were treated as live inventory by `sinfo` whenever no agents matched, causing nodes that had never registered to appear idle. Partition-oriented output now derives its node count, state, and node list exclusively from registered nodes.

Empty partitions remain visible, matching Slurm's behavior: `NODES=0`, `STATE=n/a`, and an empty `NODELIST`. The configured `nodes` value continues to act as the partition membership allowlist.

Regression coverage verifies both sides of the transition: configured nodes stay hidden before registration, and a matching node appears after registration.

Closes #317.

## Testing

- `cargo clippy -p spur-cli --all-targets --locked`
- `cargo test -p spur-cli --locked` (161 unit tests and 1 integration test)
- Local controller smoke test using the real `spurctld`, `sinfo`, and `RegisterAgent` RPC:
  - before registration: `gpu*|0|n/a|`
  - after registering only `gpu-node1`: `gpu*|1|idle|gpu-node1`

The workspace-wide Clippy and test commands were also attempted on macOS, but the existing Linux-only `spurd` mount, namespace, Landlock, and seccomp code does not compile on that platform. All checks for the affected crate pass.

### DIFF
--- a/crates/spur-cli/src/sinfo.rs
+++ b/crates/spur-cli/src/sinfo.rs
@@ -241,44 +241,19 @@ fn resolve_partition_field(
                 "infinite".into()
             }
         }
-        'D' => {
-            // part.total_nodes may be 0 (not populated by server),
-            // so fall back to the actual node count from the query.
-            // If node filtering returned 0 matches (e.g. nodes lack partition
-            // metadata), fall back to counting entries in the partition's
-            // nodelist string, then to part.total_nodes.
-            if !nodes.is_empty() {
-                nodes.len().to_string()
-            } else if !part.nodes.is_empty() {
-                part.nodes
-                    .split(',')
-                    .filter(|s| !s.trim().is_empty())
-                    .count()
-                    .to_string()
-            } else if part.total_nodes > 0 {
-                part.total_nodes.to_string()
-            } else {
-                "0".into()
-            }
-        }
+        'D' => nodes.len().to_string(),
         't' | 'T' => {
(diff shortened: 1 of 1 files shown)


powderluv:
Thanks for the PR. spurd on macOS has been bothering me maybe time to make it work.

shiv-tyagi (approved):
LGTM

PR #467, opened 2026-07-17
feat(spur-cli): support nodefile-based nodelists

### LINKED ISSUE #378: feat(spur-cli): support --nodefile and file-path detection in -w/--nodelist
## Problem

Slurm interprets a `-w/--nodelist` value containing `/` as a filename, reading node names from that file (one per line or comma-separated). It also provides `-F/--nodefile` as an explicit alternative. Spur treats the value as a literal string in all cases.

Slurm also supports `node1*2` repetition syntax for `--distribution=arbitrary`. Spur does not.

## Expected behavior

- `-w /path/to/nodes.txt` reads node names from the file
- `-F/--nodefile` accepted as an alias
- `node1*2` expands to `node1,node1` (only relevant with `--distribution=arbitrary`)

## Notes

Both features are rarely used in practice. The file-path detection is the more useful of the two. The repetition syntax is only meaningful with distribution modes Spur doesn't yet implement.

## Acceptance criteria

- [ ] `-w /tmp/nodes.txt` reads the file and expands to node names
- [ ] `-F /tmp/nodes.txt` works identically
- [ ] `node1*2` expands correctly (if distribution=arbitrary is supported)

joshkmartinez:
## Summary

- add `-F/--nodefile` support to `sbatch`, `salloc`, and `srun`
- interpret `-w/--nodelist` values containing `/` as node-list files
- normalize comma- and whitespace-separated node names into the existing nodelist submission path
- preserve `sbatch` CLI, environment, and directive precedence across both option forms

Nodefile resolution stays in `spur-cli`, so the controller and scheduler continue using the existing `JobSpec.nodelist` behavior. Repetition syntax remains unchanged because arbitrary task distribution is not implemented by Spur yet.

## Testing

- `cargo test -p spur-cli --locked`
- `cargo clippy -p spur-cli --all-targets --locked`
- local compiled-CLI smoke tests for `-F`, slash-containing `-w`, and missing-file errors

Workspace-wide tests and clippy were attempted but are currently blocked on macOS by pre-existing Linux-only `spurd` mount, namespace, Landlock, and seccomp APIs.

Closes #378

### DIFF
--- a/crates/spur-cli/src/nodelist.rs
+++ b/crates/spur-cli/src/nodelist.rs
@@ -1,0 +1,102 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::path::Path;
+
+use anyhow::{Context, Result};
+
+pub fn resolve(nodelist: Option<String>, nodefile: Option<String>) -> Result<Option<String>> {
+    if let Some(path) = nodefile {
+        return read(&path).map(Some);
+    }
+
+    match nodelist {
+        Some(value) if value.contains('/') => read(&value).map(Some),
+        value => Ok(value),
+    }
+}
+
+fn read(path: &str) -> Result<String> {
+    let contents = std::fs::read_to_string(path).with_context(|| {
+        format!(
+            "failed to read node list file: {}",
+            Path::new(path).display()
+        )
+    })?;
+
+    Ok(contents
+        .split(|c: char| c == ',' || c.is_whitespace())
+        .filter(|name| !name.is_empty())
+        .collect::<Vec<_>>()
+        .join(","))
+}
+
+#[cfg(test)]
+mod tests {
+    use std::path::PathBuf;
+    use std::sync::atomic::{AtomicU64, Ordering};
+
+    use super::*;
+
+    static NEXT_FIXTURE: AtomicU64 = AtomicU64::new(0);
+
+    struct NodeFileFixture {
+        directory: PathBuf,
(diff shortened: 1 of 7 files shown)


crates/spur-cli/src/nodelist.rs
```diff
@@ -0,0 +1,102 @@
...
+        fn path(&self) -> String {
+            self.path.to_string_lossy().into_owned()
+        }
+    }
+
+    impl Drop for NodeFileFixture {
+        fn drop(&mut self) {
+            std::fs::remove_dir_all(&self.directory).expect("remove fixture directory");
+        }
+    }
```
shiv-tyagi: The test fixture hand-rolls temp-dir creation, PID + atomic-counter naming, and a `Drop` that calls `.expect()`. `tempfile = "3"` is already used in `spur-core` and `spurd`. Reusing it drops ~30 lines and avoids the `expect()` in `Drop`, which aborts (double panic) if cleanup runs while a test assertion is already unwinding.

Add `tempfile = "3"` to `spur-cli` dev-dependencies and replace `NodeFileFixture` with `tempfile::tempdir()`.

shiv-tyagi (commented):
One minor item, non-blocking but good to do. Feature and tests look correct.

crates/spur-cli/src/nodelist.rs
joshkmartinez: ah good find, fixing!

shiv-tyagi:
Performed testing on a 3 node cluster, looking good.

shiv-tyagi (approved):

PR #468, opened 2026-07-17
fix(spurctld): schedule jobs across partition OR-lists

### LINKED ISSUE #436: Jobs submitted with a comma-separated multi-partition spec cannot reach scheduling
## What happened

`validate_partition`/`partition_block()` in `crates/spurctld/src/cluster.rs` do single-exact-name partition matching against `job.spec.partition`. This is inconsistent with other partition-resolution call sites in the scheduler (e.g. preemption-mode and priority-tier resolution), which correctly split a comma-separated value and match it as an OR-list of partition names.

As a result, a job submitted with `--partition=partA,partB` (a valid multi-partition OR-list spec) is either rejected outright at submit time (partition not found), or — if it slipped past validation some other way — would get permanently stuck `Pending (PartitionConfig)` and never reach the normal scheduling path.

## Expected

A job submitted with a comma-separated multi-partition spec should be accepted and scheduled against whichever of the listed partitions is eligible, consistent with how the rest of the scheduler already treats `job.spec.partition` as an OR-list.

## How to reproduce

1. Configure two partitions.
2. Submit a job with `--partition=<partitionA>,<partitionB>`.
3. Observe the submission is rejected, or the job never leaves `Pending`.

Found while live-verifying the multi-partition priority-tier fix in #430.

joshkmartinez:
## What this fixes

Jobs submitted with a comma-separated partition OR-list were treated as if the entire value were one partition name during admission and pending-state classification. A valid request such as `--partition=partA,partB` was therefore rejected or permanently held as `Pending (PartitionConfig)` before it could reach the scheduler, even though node placement already supported OR-list semantics.

## Approach

- Centralize parsing of trimmed, non-empty partition names and reuse it when resolving configured partitions.
- Validate every requested partition name and preserve account ACL enforcement across the full list.
- Consider a pending job eligible when any requested partition is up and satisfies its node/time limits.
- Preserve the existing behavior for internal jobs without an assigned partition.

This keeps admission, pending-state classification, priority/preemption resolution, and scheduler node matching consistent without changing the stored partition string.

## Testing

- `cargo test -p spur-core -p spurctld --locked` — 610 passed, 9 database-backed tests ignored.
- `cargo test -p spur-sched test_multi_partition_or_matching --locked` — passed.
- `cargo clippy -p spur-core -p spurctld --all-targets --locked` — clean.
- `cargo fmt --all -- --check` and `git diff --check` — clean.
- Local CLI-to-gRPC-to-controller smoke test: with `partA` down and limited to one node and `partB` up and allowing two nodes, a two-node `--partition=partA,partB` job was accepted and remained schedulable rather than receiving `PartitionConfig`; a list containing an unknown partition was rejected.

Workspace-wide test and clippy commands were also attempted on macOS, but the existing Linux-only `spurd` container/syscall implementation does not compile on that platform.

Closes #436.


shiv-tyagi (commented):
Three nits, all in service of this PR's own goal of centralizing OR-list parsing in `requested_partition_names`. None block.

**Nit: `crates/spurctld/src/cluster.rs` (`self_partitions_for_job`, ~L3811)**
This reimplements exactly what `requested_partition_names` now does (split, trim, drop empties). Reuse the helper:
```rust
fn self_partitions_for_job(job: &Job) -> Vec<String> {
    requested_partition_names(job.spec.partition.as_deref())
        .map(String::from)
        .collect()
}
```

**Nit: `crates/spurctld/src/scheduler_loop.rs` (~L503)**
Same hand-rolled `split(',').map(str::trim)` here, and this one doesn't filter empty entries. Use the helper for consistency:
```rust
let partitions: Vec<&str> =
    spur_core::partition::requested_partition_names(pending.spec.partition.as_deref()).collect();
```

**Nit: `crates/spurctld/src/cluster.rs` (`partition_block` doc comment, ~L3964)**
The doc still describes single-partition semantics ("`PartitionInactive` if the partition is not Up"). With OR-lists it's now: `PartitionInactive` only if no requested partition is Up, `PartitionConfig` if an Up partition exists but none satisfy the limits. Update to match.

shiv-tyagi:
Performed testing on a 3 node cluster. Looking good.

shiv-tyagi:
@joshkmartinez Can you please check my last comment and address it? This PR is very close.

joshkmartinez:
> @joshkmartinez Can you please check my last comment and address it? This PR is very close.

addressed! could you take another look?

shiv-tyagi (approved):
LGTM

PR #469, opened 2026-07-17
refactor(spur-cli): extract filter_qos_by_name helper

shiv-tyagi:
## Summary

- Extract the QoS name filter logic into a `filter_qos_by_name` helper so the unit test exercises the production code path instead of re-implementing it inline.

Follow-up from #457 review feedback.


biluriuday (approved):
LGTM

PR #472, opened 2026-07-18
feat(spur-sched): make nodelist selection additive

### LINKED ISSUE #377: feat(spur-sched): make -w/--nodelist additive (guarantee listed nodes + fill extras)
## Problem

Slurm's `-w/--nodelist` guarantees the listed nodes are included in the allocation and fills additional nodes from the pool when `-N` exceeds the list size. Spur treats `-w` as a strict allow-list filter, so listed nodes are the *only* candidates.

Confirmed on the bare-metal testbed:

```
$ spur sbatch -N3 -w spur-node1 job.sh
Submitted batch job 15
$ spur squeue
 JOBID  NODES      STATE     NODELIST(REASON)
    15      3    PENDING           (Priority)
```

Real Slurm would allocate spur-node1 + 2 scheduler-chosen nodes.

From the [srun docs](https://slurm.schedmd.com/srun.html):

> If you specify a minimum node or processor count larger than can be satisfied by the supplied node list, additional resources will be allocated on other nodes as needed.

## Root cause

`find_suitable_nodes()` in `crates/spur-sched/src/backfill.rs` applies `-w` as a `filter()` that discards all nodes not in the list. There is no mechanism to guarantee listed nodes are selected first, then fill the remainder from the general pool.

## Proposed approach

Split node selection into two phases:
1. **Required**: all `-w` nodes that satisfy resource/partition/reservation checks are included unconditionally.
2. **Fill**: if `num_nodes` exceeds the required set, pick additional nodes from the general pool (excluding `-x` nodes and already-selected required nodes).

This preserves the current pool-filter behavior when `-N` <= list size, while adding Slurm-compatible additive semantics when `-N` > list size.

## Acceptance criteria

- [ ] `sbatch -N3 -w node1` allocates node1 + 2 others (not PENDING)
- [ ] `srun -N1 -w node1,node2,node3` picks one from the pool (existing behavior preserved)
- [ ] `-w` nodes are guaranteed in the allocation when available
- [ ] `-x` is still honored for the fill phase
- [ ] Existing backfill tests continue to pass

joshkmartinez:
## Summary

- treat `-w/--nodelist` as a candidate pool when its unique expanded size is at least `-N`
- retain every eligible listed node and fill remaining slots from normal eligible candidates when `-N` is larger
- keep exclusion, partition, constraint, reservation, ordering, and pending-reason behavior aligned

## Design

`NodePlacement` owns hostlist expansion, deduplication, and candidate-versus-additive mode. Backfill applies normal hard eligibility first, preserves eligible listed nodes through availability, weight, topology, and prospective-reservation ordering, then fills from the remaining candidates. Controller pending reasons use the same placement semantics, including count-aware reservation fences.

## Testing

- `CARGO_PROFILE_TEST_DEBUG=0 CARGO_INCREMENTAL=0 cargo test -p spur-sched --locked` (75 passed)
- `CARGO_PROFILE_TEST_DEBUG=0 CARGO_INCREMENTAL=0 cargo test -p spurctld --locked` (402 passed, 13 PostgreSQL-gated tests ignored)
- `CARGO_PROFILE_DEV_DEBUG=0 CARGO_INCREMENTAL=0 cargo clippy -p spur-sched --all-targets --locked -- -D warnings`
- `CARGO_PROFILE_DEV_DEBUG=0 CARGO_INCREMENTAL=0 cargo clippy -p spurctld --all-targets --locked -- -D warnings`
- local controller/RPC/CLI smoke: registered four nodes, submitted `-N3 -w node001` through the compiled CLI, observed the job running on `node[001-003]`, and observed launch RPCs for node001, node002, and node003

I tried the repo -wide test command but I can't compile the Linux-only `spurd` container/syscall code on macOS; the concurrent build also exhausted the remaining local disk space. => `spurd` E2E tests were not run locally.

Closes #377


crates/spur-sched/src/node_match.rs
```diff
@@ -77,13 +78,27 @@ impl<'a> NodePlacement<'a> {
...
         !self.exclude.contains(name)
     }
 
+pub fn nodelist_is_additive(&self) -> bool {
+    let Some(ref nodelist) = self.nodelist else {
+        return false;
+    };
+    let listed = nodelist.iter().filter(|s| !s.is_empty()).count();
+    (self.job.spec.num_nodes as usize).max(1) > listed
+}
```
shiv-tyagi: `cargo fmt --all -- --check` fails here. `nodelist_is_additive` is indented at column 0 instead of inside the `impl` block:

```suggestion
    pub fn nodelist_is_additive(&self) -> bool {
        let Some(ref nodelist) = self.nodelist else {
            return false;
        };
        let listed = nodelist.iter().filter(|s| !s.is_empty()).count();
        (self.job.spec.num_nodes as usize).max(1) > listed
    }
```

The PR description lists fmt as passing, so this crept in after the last fmt run (the autofix commit). CI will reject it. Run `cargo fmt` and recommit.

shiv-tyagi (commented):
all looks okay. just one formatting issue

crates/spur-sched/src/node_match.rs
joshkmartinez: addressed!

yansun1996 (approved):

Issue #460, opened 2026-07-16
[Issue]: srun does not fan out tasks across nodes in native (non-k8s) runs — only rank 0 executes


mkuznet1:
### Environment
- spur version: 0.4.1
- Deployment: native host (spurctld + spurd), NOT the K8s operator
- OS: NAME= Ubuntu 24.04.4 LTS (Noble Numbat)
### Summary
On a native (spurctld/spurd, non-k8s) deployment, `srun` does not launch one task
per node with a distinct `SLURM_PROCID` the way real Slurm does. A command that
expects N parallel tasks across N nodes executes only once, on a single node, so
only "rank 0" runs. This breaks Slurm scripts that use `srun` as a parallel /
MPI-style task launcher (e.g. `srun bash task.sh` inside an `sbatch` job, where
each node reads its own `SLURM_PROCID`).
Searched existing issues — closest are #146, #69, #312, but none tracks this
native-CLI multi-node fan-out gap specifically (see "Related" below).
### Steps to reproduce
```bash
# Expect one line PER task (ranks 0..N-1), as in real Slurm:
srun -N 2 -n 2 bash -c 'echo "rank=$SLURM_PROCID node=$(hostname)"'
```
Or, closer to a real workload, inside an sbatch job:
```bash
#SBATCH --nodes=2
#SBATCH --ntasks=2
srun bash -c 'echo "rank=$SLURM_PROCID node=$(hostname)"'
```
### Expected behavior
Two tasks launched (one per node), each with a distinct `SLURM_PROCID` (0 and 1)
running on a different node — matching Slurm's srun step launcher.
### Actual behavior
The tasks are not distributed across the allocated nodes — execution stays on a single node (peers on other nodes never start), and per-node SLURM_PROCID is not assigned as real Slurm's srun step launcher would.
### Likely cause (from source)
In `crates/spur-cli/src/srun.rs`:
- Outside an allocation (`SPUR_JOB_ID` unset), `srun` wraps the command via
  `build_command_script` and calls `submit_job` — i.e. it behaves like `sbatch`,
  and the batch script runs once on the first node.
- Inside an allocation (`SPUR_JOB_ID` set), `srun` -> `run_as_step` ->
  `RunStepRequest`/`run_step`, whose own doc-comment says it "dispatches the
  command to **one** of the allocation's nodes." `num_tasks` is recorded on the
  step but a single `{node, stdout, stderr}` is returned.
[...]

PR #473, opened 2026-07-19
fix(srun): fan out tasks across nodes in standalone and step mode

sgopinath1:
## Summary

Fixes multi-node / multi-task `srun` fan-out (#460).

- **Standalone `srun` (native nodes):** submit an `srun_job` allocation, register it on agents via `RegisterJobAllocation` (no batch script launch), dispatch the user command as a job step with per-node `RunCommand` fan-out, then release the allocation with `CompleteJob`.
- **Standalone `srun` (K8s nodes):** fall back to the existing batch launch path when any allocated node is non-native; the CLI reads `JobInfo.srun_step_dispatch` and polls for batch completion instead of using the step-dispatch path.
- **Step mode (`SPUR_JOB_ID`):** `RunStep` fans out across allocated nodes using `build_step_task_plan`, with partial-failure aggregation instead of fail-fast on the first node error.
- **Shared launch helpers:** `spur-core::task_launch` centralizes step task planning, multi-task bash wrappers, and GPU/CPU bind env for batch and step paths.
- **Durability:** job steps are recorded via `WalOperation::JobStepCreate`; `srun_step_dispatch` is persisted on `JobStart` and exposed on `JobInfo`.

## Approach

Native standalone srun splits allocation from execution:

1. Scheduler registers the allocation on each native agent (`RegisterJobAllocation`) **before** transitioning the job to `Running`.
2. CLI creates a step, calls `RunStep`, and on completion calls `CompleteJob` (with `cancel_job` fallback if release fails).
3. Agents track allocation-only jobs (`RunningJob::AllocationOnly`) until cleanup.

`CompleteJob` is rejected unless `srun_step_dispatch` is true, so K8s batch-fallback jobs cannot be completed while batch processes are still running.

Ctrl+C in standalone native srun cancels the allocation job; Ctrl+C in step mode (inside sbatch/salloc) interrupts the step without cancelling the parent allocation.

GPU bind (`map_gpu` / `mask_gpu`) and explicit CPU bind (`rank`, `map_cpu`, `mask_cpu`) are applied in step mode; topology CPU bind modes (`cores`, `threads`, etc.) warn and are skipped. `map_cpu` validates list length against task count before dispatch.

## Test plan

- [x] `cargo test --locked`
- [x] Unit: `build_step_task_plan`, `finish_srun_job_*`, `JobStepCreate` WAL round-trip, CPU/GPU bind helpers
- [x] Native e2e: `tests/native_host/e2e/test_srun_multi_node.py` (standalone 2-node fan-out, allocation hold, sbatch+srun step)

### DIFF
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -599,7 +599,9 @@ async fn dispatch_step(
 
     let environment = srun_dispatch_environment(args);
     warn_unsupported_cpu_bind(&environment);
-    if let Some(err) = spur_core::task_launch::map_cpu_bind_error(&environment, args.ntasks) {
+    if let Some(err) = spur_core::task_launch::map_cpu_bind_error(&environment, args.ntasks)
+        .or_else(|| spur_core::task_launch::mask_cpu_bind_error(&environment, args.ntasks))
+    {
         anyhow::bail!("{err}");
     }
     let resp = client
@@ -666,26 +668,18 @@ async fn dispatch_step_cancellable(
     client: &mut SlurmControllerClient<tonic::transport::Channel>,
     args: &SrunArgs,
     job_id: u32,
-    user: &str,
+    _user: &str,
     work_dir: &str,
     io: &ResolvedIoPaths,
     cancel_job_on_interrupt: bool,
 ) -> Result<StepDispatchResult> {
+    if cancel_job_on_interrupt {
+        return dispatch_step(client, args, job_id, work_dir, io).await;
+    }
     tokio::select! {
         result = dispatch_step(client, args, job_id, work_dir, io) => result,
         _ = tokio::signal::ctrl_c() => {
-            if cancel_job_on_interrupt {
-                eprintln!("\nsrun: cancelling job {}...", job_id);
-                let _ = client
-                    .cancel_job(CancelJobRequest {
-                        job_id,
-                        signal: 2,
-                        user: user.to_string(),
-                    })
-                    .await;
-            } else {
-                eprintln!("\nsrun: step interrupted");
-            }
+            eprintln!("\nsrun: step interrupted");
             std::process::exit(130);
         }
     }
@@ -710,6 +704,8 @@ async fn run_standalone_srun(args: &SrunArgs, hooks: &HooksConfig, work_dir: &st
         .job_id;
 
     eprintln!("srun: Pending job allocation {}...", job_id);
+
+    install_ctrl_c_cancel(client.clone(), job_id, user.clone());
 
     let nodelist = wait_for_job_running(&mut client, job_id).await?;
     if !nodelist.is_empty() {
@@ -747,8 +743,6 @@ async fn run_standalone_srun(args: &SrunArgs, hooks: &HooksConfig, work_dir: &st
         )
         .await;
     }
-
-    install_ctrl_c_cancel(client.clone(), job_id, user.clone());
 
     let output_streamed = if io.stdout.is_empty() {
         try_stream_output(&mut client, &nodelist, job_id).await
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -8,11 +8,11 @@ use spur_core::config::HooksConfig;
 use spur_proto::proto::slurm_agent_client::SlurmAgentClient;
 use spur_proto::proto::slurm_controller_client::SlurmControllerClient;
 use spur_proto::proto::{
-    CancelJobRequest, CreateJobStepRequest, GetJobRequest, GetNodeRequest, JobSpec, JobState,
-    StreamJobOutputRequest, SubmitJobRequest,
+    CancelJobRequest, CompleteJobRequest, CreateJobStepRequest, GetJobRequest, GetNodeRequest,
+    JobSpec, JobState, RunStepRequest, StreamJobOutputRequest, SubmitJobRequest,
 };
 use std::collections::HashMap;
-use std::io::Write;
+use std::io::Write as _;
 
 /// Run a parallel job (interactive or allocation-based).
 #[derive(Parser, Debug)]
@@ -68,7 +68,7 @@ pub struct SrunArgs {
     #[arg(short = 'D', long)]
     pub chdir: Option<String>,
 
-    /// CPU binding (none, cores, threads, sockets, ldoms, rank, map_cpu, mask_cpu)
+    /// CPU binding (none, rank, map_cpu, mask_cpu; topology modes cores/threads/sockets/ldoms are not applied in step mode)
     #[arg(long)]
     pub cpu_bind: Option<String>,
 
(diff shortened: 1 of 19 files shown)


tests/native_host/e2e/cluster.py
```diff
@@ -349,6 +349,18 @@ def cli_allow_fail(self, args: list[str], controller_addr: str | None = None) ->
...
 
+    def srun_with_exit(self, args: list[str]) -> tuple[int, str]:
+        """Run srun and return (exit_code, combined stdout+stderr)."""
+        cmd_parts = [
+            f"SPUR_CONTROLLER_ADDR='{self.controller_addr}'",
+            f"PATH='{self.bin_dir}':$PATH",
+            f"'{self.bin_dir}/srun'",
+        ]
+        cmd_parts.extend(f"'{a}'" for a in args)
+        _, stdout, stderr = self.nodes[0].client.exec_command(" ".join(cmd_parts))
```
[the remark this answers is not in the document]
sgopinath1: Fixed.

crates/spurd/src/agent_server.rs
```diff
@@ -369,6 +371,16 @@ mod completion_report_tests {
...
 
+/// Build a bash script that execs a command vector without shell interpretation.
+fn build_one_shot_command_script(command: &[String]) -> String {
+    let quoted: Vec<String> = command
+        .iter()
+        .map(|arg| shlex::try_quote(arg).unwrap_or_else(|_| std::borrow::Cow::Borrowed(arg)))
+        .map(|c| c.into_owned())
+        .collect();
+    format!("#!/bin/bash\nexec {}\n", quoted.join(" "))
+}
```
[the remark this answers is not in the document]
sgopinath1: Fixed.

crates/spurd/src/agent_server.rs
```diff
@@ -1107,9 +1192,86 @@ impl SlurmAgent for AgentService {
...
+            senv.set("PMIX_NAMESPACE", format!("spur.{job_id}"));
+            senv.set("OMPI_COMM_WORLD_SIZE", num_tasks_total);
+            senv.set("OMPI_COMM_WORLD_NODE_RANK", node_id);
+            senv.set("OMPI_COMM_WORLD_LOCAL_SIZE", tasks_per_node);
+        }
+
+        let (program, program_args) = if num_tasks > 1 {
+            let user_script_path = format!("{work_dir}/.spur_step_{job_id}_{step_id}_{node_id}.sh");
+            let user_script = build_one_shot_command_script(&req.command);
+            std::fs::write(&user_script_path, &user_script)
```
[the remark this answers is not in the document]
sgopinath1: Fixed.

crates/spurd/src/agent_server.rs
```diff
@@ -1081,9 +1157,18 @@ impl SlurmAgent for AgentService {
...
             )
         };
 
+        let hostname = hostname::get()
+            .map(|h| h.to_string_lossy().to_string())
+            .unwrap_or_else(|_| "localhost".into());
+        let node_names: Vec<&str> = nodelist.split(',').filter(|s| !s.is_empty()).collect();
+        let num_nodes = node_names.len().max(1) as u32;
+        let node_id = node_names.iter().position(|n| *n == hostname).unwrap_or(0) as u32;
+
```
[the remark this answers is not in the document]
sgopinath1: Fixed.

crates/spur-core/src/task_launch.rs
```diff
@@ -0,0 +1,411 @@
...
+                 _CPU_IDX=$((SPUR_TASK_OFFSET + LOCAL_RANK))\n  \
+                 if [ \"$_CPU_IDX\" -ge ${{#_CPU_MAP[@]}} ]; then\n    \
+                   echo \"map_cpu: rank $_CPU_IDX exceeds CPU map (len ${{#_CPU_MAP[@]}})\" >&2\n    \
+                   exit 1\n  \
+                 fi\n  \
+                 taskset -c ${{_CPU_MAP[$_CPU_IDX]}} "
+            )
+        }
+        CpuBind::Map(_) => String::new(),
+        CpuBind::Mask(mask) => format!("taskset -m {mask} "),
```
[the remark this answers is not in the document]
sgopinath1: Fixed.

crates/spur-core/src/task_launch.rs
```diff
@@ -0,0 +1,411 @@
...
+            )
+        }
+        CpuBind::Mask(mask) => (
+            "taskset".into(),
+            std::iter::once("-m".to_string())
+                .chain(std::iter::once(mask))
+                .chain(std::iter::once(program.to_string()))
+                .chain(args.iter().cloned())
+                .collect(),
+        ),
```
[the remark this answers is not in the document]
sgopinath1: Fixed.

crates/spurd/src/agent_server.rs
```diff
@@ -1107,9 +1192,86 @@ impl SlurmAgent for AgentService {
...
+        senv.set("PMI_APPNUM", "0");
+        if mpi == "pmix" {
+            senv.set("PMIX_SIZE", num_tasks_total);
+            senv.set("PMIX_NAMESPACE", format!("spur.{job_id}"));
+            senv.set("OMPI_COMM_WORLD_SIZE", num_tasks_total);
+            senv.set("OMPI_COMM_WORLD_NODE_RANK", node_id);
+            senv.set("OMPI_COMM_WORLD_LOCAL_SIZE", tasks_per_node);
+        }
+
+        let (program, program_args) = if num_tasks > 1 {
```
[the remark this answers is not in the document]
sgopinath1: Fixed.

crates/spurd/src/executor.rs
```diff
@@ -193,6 +200,7 @@ impl RunningJob {
                     Err(e) => Err(e.into()),
                 }
             }
+            RunningJob::AllocationOnly => Ok(None),
```
yansun1996, line 203: **Possible allocation leak (higher-impact).** Since `AllocationOnly::try_wait()` always returns `Ok(None)`, the agent never reaps these jobs, so the only thing that releases the allocation is the srun CLI cooperatively calling `CompleteJob`/`CancelJob`. If the CLI dies after the job reaches `Running` but before it releases (crash, SIGKILL, network partition, closed laptop), the job stays `Running` and the node's CPU/mem/GPU stay reserved indefinitely.

There's also no time-based backstop: `enforce_time_limits` skips jobs whose `time_limit` is `None`, srun doesn't set one unless `-t` is passed, and the partition `default_time_minutes` isn't injected into the srun spec at submit. This is the same class of stranding as the earlier agent GPU-leak issue.

Could we add a server-side liveness/lease for allocation-only jobs (e.g. a CLI heartbeat with a controller-side reaper), or at minimum inject the partition default time limit into srun specs so there's a bound? Happy to help think this through.

crates/spur-cli/src/srun.rs
```diff
@@ -560,8 +371,436 @@ fn resolve_srun_env(matches: &ArgMatches, args: &mut SrunArgs) -> Result<()> {
...
+            spec: Some(job_spec),
+        })
+        .await
+        .context("job submission failed")?
+        .into_inner()
+        .job_id;
+
+    eprintln!("srun: Pending job allocation {}...", job_id);
+
+    let nodelist = wait_for_job_running(&mut client, job_id).await?;
```
yansun1996, line 710: **Ctrl+C coverage gap (ties into the allocation-leak note).** By the time we're waiting here the scheduler has already registered the allocation and flipped the job to `Running`, but the cancel-on-interrupt handler (`install_ctrl_c_cancel`) isn't installed until further down — and on the native step-dispatch path `handle_terminal_state` exits the process before we ever reach that install. So a SIGINT during `wait_for_job_running` kills srun with the allocation live and nothing cancels it.

Would it be possible to install the cancel handler right after `submit_job` returns a `job_id`, so the whole wait window is covered?

crates/spurctld/src/server.rs
```diff
@@ -1425,58 +1484,140 @@ impl SlurmController for ControllerService {
...
-                uid: req.uid,
-                gid: req.gid,
-                work_dir: req.work_dir,
-                environment: req.environment,
-                job_id: req.job_id,
-            })
-            .await
-            .map_err(|e| Status::internal(format!("run_command failed: {}", e)))?
-            .into_inner();
+        let mut max_exit = 0i32;
```
yansun1996, line 1592: **Signal-killed tasks can be reported as success (higher-impact).** Aggregation seeds `max_exit = 0` and then takes `max_exit.max(agent_resp.exit_code)`. For a task killed by a signal the agent returns `exit_code = -1` (`output.status.code().unwrap_or(-1)` in `agent_server.rs`, since `code()` is `None` on signal death). `max(0, -1)` is `0`, so the step — and therefore `srun` — reports success for a task that was actually killed.

This isn't just cosmetic: the value flows through `CompleteJob` → `finish_srun_job` into the job's final exit code and the accounting record. Slurm reports `128 + signal` here. Could we map signal death to `128 + signal` at the agent and avoid clamping the aggregate to a `0` floor (seed from the real task results)?

crates/spurctld/src/server.rs
```diff
@@ -1425,58 +1484,140 @@ impl SlurmController for ControllerService {
...
+                        job.allocated_nodes.len()
+                    ))
+                })?
+                .clone();
+            let node = self
+                .cluster
+                .get_node(&node_name)
+                .ok_or_else(|| Status::not_found(format!("node {} not found", node_name)))?;
+            let addr = node.address.as_ref().ok_or_else(|| {
+                Status::internal(format!("node {} has no agent address", node_name))
```
yansun1996: **Node-resolution failure can orphan already-dispatched tasks.** These `?`-based resolutions happen inside the per-node loop before `set.spawn`. If node K fails to resolve after nodes 0..K-1 have already been spawned, the early return drops the `JoinSet` (aborting the local futures), but the `run_command` RPCs already sent to the earlier agents keep executing remotely — leaving tasks the controller no longer tracks. This is a bit inconsistent with the nice partial-failure aggregation used for RPC errors lower in the same loop. Could resolution errors be folded into that same `dispatch_errors`/`max_exit` path (or all addresses resolved up front before any spawn)?

crates/spur-core/src/task_launch.rs
```diff
@@ -0,0 +1,445 @@
...
+                 _CPU_IDX=$((SPUR_TASK_OFFSET + LOCAL_RANK))\n  \
+                 if [ \"$_CPU_IDX\" -ge ${{#_CPU_MAP[@]}} ]; then\n    \
+                   echo \"map_cpu: rank $_CPU_IDX exceeds CPU map (len ${{#_CPU_MAP[@]}})\" >&2\n    \
+                   exit 1\n  \
+                 fi\n  \
+                 taskset -c ${{_CPU_MAP[$_CPU_IDX]}} "
+            )
+        }
+        CpuBind::Map(_) => String::new(),
+        CpuBind::Mask(mask) => format!("taskset {mask} "),
```
yansun1996: **`mask_cpu` doesn't bind per-task for multi-entry masks.** `CpuBind::Mask` holds the entire string after `mask_cpu:` (e.g. `"0x3,0xc"`), so this emits `taskset 0x3,0xc ...`, passing the whole comma list as a single mask applied identically to every task (and `taskset` won't parse a comma list as one mask). By contrast `map_cpu` correctly indexes per rank. Could `mask_cpu` split the list and index per rank the same way, so per-task masks work? The single-mask case (`mask_cpu:0x3`) is fine, so this only affects the multi-entry form.

crates/spurd/src/agent_server.rs
```diff
@@ -1107,9 +1190,97 @@ impl SlurmAgent for AgentService {
...
+            senv.set("PMIX_NAMESPACE", format!("spur.{job_id}"));
+            senv.set("OMPI_COMM_WORLD_SIZE", num_tasks_total);
+            senv.set("OMPI_COMM_WORLD_NODE_RANK", node_id);
+            senv.set("OMPI_COMM_WORLD_LOCAL_SIZE", tasks_per_node);
+        }
+
+        let (program, program_args) = if num_tasks > 1 || req.label {
+            let user_script_path = format!("{work_dir}/.spur_step_{job_id}_{step_id}_{node_id}.sh");
+            let user_script = build_one_shot_command_script(&req.command)?;
+            std::fs::write(&user_script_path, &user_script)
```
yansun1996: **Temp step-script cleanup + `/tmp` safety (minor security).** These wrapper scripts are written with a predictable name (`.spur_step_{job}_{step}_{node}.sh`) into `work_dir`, which defaults to `/tmp`, at mode `0755`, and I don't see them removed afterward. In a shared `/tmp` a predictable name in a world-writable dir is a symlink/TOCTOU risk, and the files also accumulate over time. Could we create them in a per-job private directory (e.g. mode `0700`) or with `O_EXCL`, and unlink them once the command finishes (best-effort, like the cgroup cleanup)?

yansun1996 (commented):
Thanks for taking on the srun multi-node fan-out — the allocation/register/step/complete split is a clean design and the partial-failure aggregation in `RunStep` is a nice touch. I did a careful pass (with a couple of independent cross-checks) and wanted to share a few things I think are worth addressing before merge. I've left inline notes on the specific lines; grouping them here for context.

**Higher-impact (would be great to resolve before merge):**
1. Allocation-only jobs can strand/leak if the srun CLI dies (no server-side liveness + no time-limit backstop).
2. A Ctrl+C during the allocation-wait window isn't covered by the cancel handler, which feeds into (1).
3. Signal-killed step tasks currently aggregate to exit 0, so a killed task can be reported as success and propagate into the job's final exit code + accounting.

**Worth considering:**
4. Mid-plan node-resolution failure in `run_step` can orphan tasks already dispatched to earlier nodes.
5. `mask_cpu` with multiple/per-task masks doesn't bind per rank the way `map_cpu` does.
6. The temporary step-script files are predictable-named under `/tmp` at mode 0755 and aren't cleaned up (minor security + accumulation).

None of these are blockers to the overall approach — happy to discuss any of them, and please push back if I've misread anything. Thanks again!

crates/spurd/src/executor.rs
sgopinath1: Fixed. submit_job now calls apply_default_time_limit, so standalone srun jobs without -t inherit the partition's default_time_minutes and enforce_time_limits can reap them. A CLI heartbeat/lease reaper for crash-after-Running is still a good follow-up when partitions have no default time.

crates/spur-cli/src/srun.rs
sgopinath1: Fixed. install_ctrl_c_cancel runs right after submit_job returns, so SIGINT during the pending/running wait cancels the job. Standalone step dispatch uses that single handler (no duplicate listener in dispatch_step_cancellable).

crates/spurctld/src/server.rs
sgopinath1: Fixed. The agent uses shell_exit_code() (128 + signal on Unix) instead of code().unwrap_or(-1), and resolution/dispatch errors bump max_exit so we no longer report success when tasks fail.

crates/spurctld/src/server.rs
sgopinath1: Fixed. All node names/addresses are resolved upfront into a dispatch list before any JoinSet spawn. Resolution failures go into dispatch_errors (same path as RPC failures) instead of aborting mid-loop after partial fan-out.

crates/spur-core/src/task_launch.rs
sgopinath1: Fixed. Comma-separated mask_cpu masks are indexed per rank in both wrap_command_with_cpu_bind and the bash wrapper prefix, matching map_cpu. Added mask_cpu_bind_error for fail-fast validation at dispatch time.

crates/spurd/src/agent_server.rs
sgopinath1: Fixed. Step scripts live under {work_dir}/.spur_step_{job_id}/ (mode 0700, chowned to the job user via write_job_scratch), scripts are 0700, and a StepScriptCleanup guard removes them on all exit paths after the command finishes.

sgopinath1:
> Thanks for taking on the srun multi-node fan-out — the allocation/register/step/complete split is a clean design and the partial-failure aggregation in `RunStep` is a nice touch. I did a careful pass (with a couple of independent cross-checks) and wanted to share a few things I think are worth addressing before merge. I've left inline notes on the specific lines; grouping them here for context.
> 
> **Higher-impact (would be great to resolve before merge):**
> 
> 1. Allocation-only jobs can strand/leak if the srun CLI dies (no server-side liveness + no time-limit backstop).
> 2. A Ctrl+C during the allocation-wait window isn't covered by the cancel handler, which feeds into (1).
> 3. Signal-killed step tasks currently aggregate to exit 0, so a killed task can be reported as success and propagate into the job's final exit code + accounting.
> 
> **Worth considering:** 4. Mid-plan node-resolution failure in `run_step` can orphan tasks already dispatched to earlier nodes. 5. `mask_cpu` with multiple/per-task masks doesn't bind per rank the way `map_cpu` does. 6. The temporary step-script files are predictable-named under `/tmp` at mode 0755 and aren't cleaned up (minor security + accumulation).
> 
> None of these are blockers to the overall approach — happy to discuss any of them, and please push back if I've misread anything. Thanks again!

Thanks for reviewing @yansun1996. I have addressed your comments. PTAL.

yansun1996 (approved):
Thanks for the quick turnaround on the review feedback — the fixes look good and I'm approving.

**Verified fixed:**
- Ctrl+C now cancels the job across the pending/running wait window (handler installed right after `submit_job`).
- Signal-killed step tasks now report `128+signal` via the new `shell_exit_code()` helper, and resolution/dispatch errors bump the aggregated exit code — no more false success.
- `run_step` resolves all node addresses up front and folds resolution failures into `dispatch_errors`, so a mid-plan failure no longer orphans already-dispatched tasks.
- `mask_cpu` now splits comma-separated masks and binds per rank (matching `map_cpu`), with `mask_cpu_bind_error` validation and unit coverage.
- Step scripts moved to a per-job `0700` directory with a `StepScriptCleanup` RAII guard that removes them on all exit paths.

**Two items I'd suggest handling as follow-up PRs rather than blocking this one:**

1. **Allocation reaping when a partition has no default time (from the allocation-leak note).** `apply_default_time_limit` nicely closes the common case, but when a partition has no `default_time_minutes` the spec's `time_limit` stays `None`, so `enforce_time_limits` won't reap an allocation-only job whose srun CLI died (crash/SIGKILL/partition). Could we track a follow-up to add a server-side liveness/lease reaper (or a documented default) for that case? Not a blocker since it only affects partitions with no default time.

2. **Symlink/TOCTOU hardening on the step-script directory (please treat as a near-term security follow-up).** The per-job dir + RAII cleanup is a solid improvement, but the underlying race isn't fully closed: `prepare_step_script_dir` uses `create_dir_all` (which succeeds on a pre-existing dir/symlink — no `O_EXCL`) and then does path-based `set_permissions` + `chown` as root, while `work_dir` still defaults to `/tmp` with a predictable `job_id`. On a shared node this is a genuine root-privileged chown/write-through-symlink concern (local priv-esc class), so I'd flag it as higher priority than a normal nit. Suggested direction for the follow-up: create the dir with `create_dir`/`O_EXCL` (fail if it already exists) or fd-based `openat(O_NOFOLLOW)` + `fchown`/`fchmod`, and ideally root step scratch under the root-owned spool tree like batch jobs instead of `/tmp`. Happy to help review that change.

Approving so this isn't blocked — thanks again for the solid work on the fan-out.

PR #474, opened 2026-07-20
fix(spurctld): drop stale completion reports from preempted-and-requeued runs

yansun1996:
## What this fixes

A batch job in a `preempt_mode=requeue` partition could land in `FAILED` / `RaisedSignal:9` after preemption instead of returning to `RUNNING`, intermittently. This is the product bug behind the flaky `test_chronic_preemption_never_holds_job` e2e (see the recurring native-host E2E failure: https://github.com/ROCm/spur/actions/runs/29723193165/job/88290371845). A prior change (#466) only raised the test's reschedule timeout; because the job ends terminal, no timeout can make it pass — the failure kept recurring.

## Root cause

The requeue **eligibility hold** is `interval_secs*2+3` (≥ 5s) and the agent's **graceful-cancel grace period** is a fixed 5s — they coincide. On preemption:

1. The controller requeues the low-priority job (clears its allocation, sets a ~5s `begin_time` hold) and tells the agent to graceful-cancel.
2. The job's process dies on SIGTERM; the agent's completion report is correctly rejected (allocation already cleared).
3. ~5s later the hold expires, the job is re-dispatched onto the **same** node (it's `--exclusive` and the other node is busy), and goes `Running` again.
4. At the same ~5s mark the agent's grace timer fires a **SIGKILL** and reports completion **signal=9** — but the report is keyed only by `job_id` with no notion of *which run*, so it is applied to the freshly re-dispatched run and flips it to `FAILED`.

Two gaps combine: the agent's grace-timer SIGKILL matched on `job_id` alone, and completion reports carried no run identity, so the controller couldn't tell a prior run's report from the current one.

## The fix

Introduce a monotonic per-dispatch **run epoch** (`run_attempt`) threaded `JobStart` → `LaunchJobRequest` → `ReportJobStatusRequest`:

- The controller assigns `run_attempt = prev + 1` at each dispatch, persists it on the job (and in the `JobStart` WAL op), and sends it to the agent.
- The agent stores it per tracked job, echoes it on the completion report, and its grace-period SIGKILL now fires **only if the still-tracked run's epoch matches** the one it SIGTERM'd — so it can't kill a re-dispatched run that reused the `job_id`.
- The controller **drops a node completion whose epoch is older** than the job's current run (new `NodeCompleteResult::StaleReport`), so a late report from a superseded run can't fail the current one.
- On same-node re-dispatch, `launch_job` SIGKILLs an older-epoch tracked run before overwriting it, closing a pre-existing gap where a SIGTERM-ignoring prior process could be orphaned.

`run_attempt == 0` (legacy agent or controller, or a pre-upgrade WAL entry) disables both checks, so the change is backward compatible. Proto additions are new fields only (no renumbering, package unchanged); the new WAL field uses `#[serde(default)]`.

## Compatibility note

The race is only fully closed when **both** the controller and agents are upgraded — a legacy component reports/sees epoch 0, which disables the guard. Mixed-version clusters retain the prior behavior until both sides are updated.

## Testing

- `cargo fmt --all --check`, `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` — clean.
- `cargo test --locked` — all pass, including two new unit tests: the controller dropping a stale-epoch completion (`node_complete_drops_stale_run_report`) and the agent grace timer skipping SIGKILL for a reused `job_id` (`graceful_cancel_skips_sigkill_for_reused_job_id`).
- Reviewed by independent passes plus an adversarial cross-validation; findings addressed (superseded-run kill on re-dispatch; hardened the epoch test against a zombie false-pass).
- Manually reproduced on a 2-node cluster: the pre-fix build hit `FAILED` on the first preemption cycle; the fixed build ran repeated preemption cycles with the low-priority job returning to `RUNNING` every time and no grace-period SIGKILL escalations.

## Follow-ups

- The reschedule after requeue currently waits for the next scheduler tick because job completion doesn't wake the scheduler; a completion-triggered wake would reduce latency (out of scope here).

### DIFF
--- a/crates/spurd/src/agent_server.rs
+++ b/crates/spurd/src/agent_server.rs
@@ -44,12 +44,15 @@ struct TrackedJob {
     memory_mb: u64,
     nodelist: String,
     mpi: String,
+    /// Run epoch; echoed on completion and guards the grace-period SIGKILL.
+    run_attempt: u32,
 }
 
 struct CompletedJob {
     job_id: u32,
     exit_code: i32,
     signal: i32,
+    run_attempt: u32,
     rootfs_mode: crate::container::RootfsMode,
     cgroup: Option<std::path::PathBuf>,
     work_dir: String,
@@ -205,6 +208,7 @@ impl AgentService {
                                 job_id: *job_id,
                                 exit_code,
                                 signal,
+                                run_attempt: tracked.run_attempt,
                                 rootfs_mode: tracked.rootfs_mode.clone(),
                                 cgroup,
                                 work_dir: tracked.work_dir.clone(),
@@ -313,6 +317,7 @@ impl AgentService {
                         c.job_id,
                         c.exit_code,
                         c.signal,
+                        c.run_attempt,
                         &local_hostname,
                         drain.as_ref(),
                     )
@@ -453,6 +458,7 @@ async fn report_completion(
     job_id: u32,
     exit_code: i32,
     signal: i32,
+    run_attempt: u32,
     reporting_node: &str,
     drain: Option<&DrainRequest>,
 ) {
@@ -477,6 +483,7 @@ async fn report_completion(
                     drain_node: drain.is_some(),
                     drain_reason: drain.as_ref().map(|d| d.reason.clone()).unwrap_or_default(),
                     reporting_node: reporting_node.to_string(),
+                    run_attempt,
                 };
                 match client.report_job_status(req).await {
                     Ok(_) => {
@@ -544,6 +551,7 @@ impl SlurmAgent for AgentService {
         // not part of the (user-supplied) job spec.
         let array_job_id = req.array_job_id;
         let array_task_id = req.array_task_id;
+        let run_attempt = req.run_attempt;
         let spec = req
             .spec
             .ok_or_else(|| Status::invalid_argument("missing job spec"))?;
@@ -898,6 +906,15 @@ impl SlurmAgent for AgentService {
                 // `launching` (which would let reconcile reclaim it).
                 self.allocation.lock().await.commit_job(job_id);
                 info!(job_id, gpus = ?launch_cfg.gpu_devices, "job launched successfully");
+                // Re-dispatch onto the same node reuses job_id. If an older run
+                // is still tracked (its process ignored SIGTERM and outlived the
+                // requeue), the insert would drop its Child without killing it —
+                // an orphaned runaway. SIGKILL it before overwriting.
+                if let Some(old) = jobs.get(&job_id) {
+                    if old.run_attempt < run_attempt {
+                        let _ = old.job.kill_signal(nix::sys::signal::Signal::SIGKILL);
+                    }
+                }
                 jobs.insert(
                     job_id,
                     TrackedJob {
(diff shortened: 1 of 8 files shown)


crates/spurd/src/agent_server.rs
```diff
@@ -897,6 +905,15 @@ impl SlurmAgent for AgentService {
...
                 info!(job_id, gpus = ?launch_cfg.gpu_devices, "job launched successfully");
+                // Re-dispatch onto the same node reuses job_id. If an older run
+                // is still tracked (its process ignored SIGTERM and outlived the
+                // requeue), the insert would drop its Child without killing it —
+                // an orphaned runaway. SIGKILL it before overwriting.
+                if let Some(old) = jobs.get(&job_id) {
+                    if old.run_attempt < run_attempt {
+                        let _ = old.job.kill_signal(nix::sys::signal::Signal::SIGKILL);
+                    }
+                }
```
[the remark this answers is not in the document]
yansun1996: Good catch — fixed in d425052. The insert now returns the displaced TrackedJob; after SIGKILL we reap it in a background task via child.wait(), so no zombie accumulates across repeat preempt/requeue cycles.

crates/spurd/src/agent_server.rs
```diff
@@ -915,8 +923,24 @@ impl SlurmAgent for AgentService {
...
+                // requeue, kill and reap it here — the monitor loop no longer
+                // tracks it, so without this it would leak as an orphan/zombie.
+                if let Some(mut old) = displaced {
+                    if old.run_attempt < run_attempt {
+                        let _ = old.job.kill_signal(nix::sys::signal::Signal::SIGKILL);
+                        tokio::spawn(async move {
+                            if let executor::RunningJob::Managed { child, .. } = &mut old.job {
+                                let _ = child.wait().await;
+                            }
+                        });
```
shiv-tyagi: The displaced-run cleanup only reaps the `Managed` variant. On root/container clusters `executor::launch_job` returns `RunningJob::Forked`, which spurd reaps itself via `waitpid` from the monitor loop's `try_wait`. Once the old run is displaced out of `running`, nothing calls `try_wait` on it, so a `Forked` job killed here is never reaped and leaks as a zombie until spurd exits. The comment above says "kill and reap it here", but the reap is Managed-only.

`try_wait()` already handles both variants, so poll it instead:

```suggestion
                        tokio::spawn(async move {
                            loop {
                                match old.job.try_wait() {
                                    Ok(Some(_)) | Err(_) => break,
                                    Ok(None) => {
                                        tokio::time::sleep(
                                            std::time::Duration::from_millis(50),
                                        )
                                        .await
                                    }
                                }
                            }
                        });
```

shiv-tyagi (commented):
One comment. PTAL

crates/spurd/src/agent_server.rs
yansun1996: Good catch — fixed in f817e1f. Extracted the displaced-run reap into a helper that polls `try_wait()`, so both variants are collected: the `Managed` child via tokio and the `Forked` container's raw pid via `waitpid`. Added a unit test that forks a child, lets it become a zombie, and asserts it's reaped. Thanks!

yansun1996:
merge this to fix the issue uncovered by the intermittent e2e test failure

PR #478, opened 2026-07-21
fix(spurctld): prevent single scheduling pass from over-subscribing QOS/account aggregate limits

yansun1996:
## What this fixes
QOS and account **aggregate** limits — group `grp_tres`, per-user TRES, and per-user running-job-count caps — were checked against a static snapshot of running jobs with no accumulation across a scheduling pass. Several pending jobs sharing a QOS/account could each fit against that snapshot and all be admitted in a single cycle, collectively exceeding the cap. The visible symptom: jobs correctly pending on a group node limit would all start at once when the scheduler next ran.

## Approach
Enforce these aggregate caps **after** the priority sort and reserve each kept job's footprint highest-priority-first, so lower-priority jobs later in the same pass see the reduced headroom. This mirrors the existing license and burst-buffer reservation passes. A shared `PassReservations` accumulator and a `job_tres()` helper (extracted from the running-usage sum) keep the accounting consistent between running jobs and same-pass reservations.

## Known limitations
A job deferred *only* because higher-priority same-pass siblings consumed the headroom is picked up on the next scheduling cycle once those siblings are running; in the brief interim its displayed `Reason=` may lag. This matches the existing behavior of the license and burst-buffer reservation passes (their tagging path is likewise snapshot-based).

## Testing
- Unit tests covering group `grp_tres`, per-user TRES, and per-user job-count over-subscription within one pass, plus priority-ordered reservation. Verified these fail without the fix and pass with it.
- Extended the accounting e2e suite with a group-cap conformance test, validated end-to-end on an isolated 2-node deployment.

### DIFF
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -1914,11 +1914,6 @@ impl ClusterManager {
             .map(|job| (job.job_id, self.resolve_qos(job)))
             .collect();
 
-        // Account/association and QoS aggregate enforcement is applied after the
-        // priority sort below, so group (`grp_tres`) and per-user caps are
-        // reserved highest-priority-first and a single pass cannot over-subscribe
-        // them. License enforcement follows for the same reason.
-
         // Reservation validation: reject jobs targeting expired/nonexistent reservations
         {
             let reservations = self.get_reservations();
@@ -8904,7 +8899,7 @@ mod tests {
     async fn pending_jobs_does_not_oversubscribe_qos_grp_node_within_one_pass() {
         // Three pending 1-node jobs share a QOS capped at grp_tres node=2. Each
         // fits against the empty running total, but a single pass must admit at
-        // most 2 or the QOSGrpNodeLimit is over-subscribed (SPUR-84).
+        // most 2 or the QOSGrpNodeLimit is over-subscribed.
         let dir = TempDir::new().unwrap();
         let cm = test_cluster(&dir).await;
         register_node(&cm, "n1", 64, 128000);
@@ -9010,6 +9005,39 @@ mod tests {
         assert_eq!(
             granted, 2,
             "pending_jobs() returned {granted} account jobs but grp_tres node=2 allows 2"
+        );
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn pending_jobs_does_not_oversubscribe_account_max_running_jobs_within_one_pass() {
+        // Same guard as the QOS max_jobs_per_user test, one layer up: an account
+        // association capped at max_running_jobs=2 must not admit 3 pending jobs
+        // from one user in a single pass even though none is running yet.
+        let dir = TempDir::new().unwrap();
+        let cm = test_cluster(&dir).await;
+        register_node(&cm, "n1", 64, 128000);
+        cm.association_cache().insert_limits(
+            "testuser",
+            "research",
+            spur_core::accounting::AccountLimits {
+                max_running_jobs: Some(2),
+                ..Default::default()
+            },
+        );
+
+        let ids: Vec<JobId> = (0..3)
+            .map(|i| {
+                let mut s = basic_spec(&format!("m{i}"));
+                s.account = Some("research".into());
+                submit_and_wait(&cm, s)
+            })
+            .collect();
+
+        let pending: Vec<JobId> = cm.pending_jobs().iter().map(|j| j.job_id).collect();
+        let granted = ids.iter().filter(|id| pending.contains(id)).count();
+        assert_eq!(
+            granted, 2,
+            "pending_jobs() returned {granted} account jobs but max_running_jobs=2 allows 2"
         );
     }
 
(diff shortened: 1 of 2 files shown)


crates/spurctld/src/cluster.rs
```diff
@@ -4297,23 +4355,76 @@ fn partition_block(job: &Job, partitions: &[Partition]) -> Option<spur_core::job
...
+        if let Some(account) = job.spec.account.as_deref().filter(|a| !a.is_empty()) {
+            self.account_grp
+                .entry(account.to_string())
+                .or_default()
+                .add(&tres);
+            *self
+                .account_user_count
+                .entry((user, account.to_string()))
+                .or_insert(0) += 1;
+        }
```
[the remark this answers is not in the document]
yansun1996: Good catch — bound `account` once and reuse it for both the group and per-user entries, so it's a single allocation now.

crates/spurctld/src/cluster.rs
```diff
@@ -8791,6 +8900,188 @@ mod tests {
         );
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn pending_jobs_does_not_oversubscribe_qos_grp_node_within_one_pass() {
+        // Three pending 1-node jobs share a QOS capped at grp_tres node=2. Each
+        // fits against the empty running total, but a single pass must admit at
+        // most 2 or the QOSGrpNodeLimit is over-subscribed (SPUR-84).
```
shiv-tyagi: Drop the SPUR-84 reference. AGENTS.md: "Do not reference issue numbers, PR numbers, or task IDs in code comments. That context belongs in git history."

```suggestion
        // most 2 or the QOSGrpNodeLimit is over-subscribed.
```

tests/native_host/e2e/test_accounting.py
```diff
@@ -314,6 +314,48 @@ def test_gpu_cap_sets_qos_pending_reason(self, accounting_cluster):
             f"expected QOSMaxGRESPerUser, got {reason!r}"
         )
 
+    def test_grp_cpu_cap_not_oversubscribed_by_concurrent_jobs(self, accounting_cluster):
+        # SPUR-84: a QOS group cap must hold across a burst of submissions, not
```
shiv-tyagi: Drop the SPUR-84 prefix here too (same AGENTS.md rule). The rest of the comment already explains the why well.

```suggestion
        # A QOS group cap must hold across a burst of submissions, not
```

crates/spurctld/src/cluster.rs
```diff
@@ -4297,23 +4355,74 @@ fn partition_block(job: &Job, partitions: &[Partition]) -> Option<spur_core::job
...
+/// pass, folded into the running aggregates so aggregate QOS/account caps
+/// (`grp_tres`, per-user TRES, per-user running-job counts) can't be
+/// over-subscribed within one pass — the same guarantee licenses and
+/// burst-buffer capacity already get via priority-ordered reservation.
+#[derive(Default)]
+struct PassReservations {
+    qos_grp: HashMap<String, TresRecord>,
+    qos_user_tres: HashMap<(String, String), TresRecord>,
+    qos_user_count: HashMap<(String, String), u32>,
+    account_grp: HashMap<String, TresRecord>,
```
shiv-tyagi, line 4390: Q: `account_user_count` (the per-user running-job reservation for account associations) has no same-pass test. The QOS side is covered by `pending_jobs_does_not_oversubscribe_qos_max_jobs_per_user_within_one_pass`, and the account side covers grp_tres but not `max_running_jobs`. Worth a symmetric account test since it's the one reserved field not exercised?

crates/spurctld/src/cluster.rs
```diff
@@ -1907,26 +1907,17 @@ impl ClusterManager {
...
-        // qos_block_for() so the drop decision and the displayed reason agree.
-        pending.retain(|job| qos_block_for(job, &qos_by_job[&job.job_id], &jobs).is_none());
-
-        // License enforcement is applied after the priority sort below, so scarce
-        // licenses are reserved highest-priority-first and a single pass cannot
-        // over-subscribe the pool.
+        // Account/association and QoS aggregate enforcement is applied after the
+        // priority sort below, so group (`grp_tres`) and per-user caps are
+        // reserved highest-priority-first and a single pass cannot over-subscribe
+        // them. License enforcement follows for the same reason.
```
shiv-tyagi: Nit: this signpost comment restates what the detailed block at the actual enforcement site (below the sort) already says. One of the two is enough.

shiv-tyagi (commented):
some comments PTAL

crates/spurctld/src/cluster.rs
yansun1996: Good catch, dropped.

tests/native_host/e2e/test_accounting.py
yansun1996: Dropped here too.

crates/spurctld/src/cluster.rs
yansun1996: Added `pending_jobs_does_not_oversubscribe_account_max_running_jobs_within_one_pass`, mirroring the QoS analog — 3 pending jobs from one user under an account capped at `max_running_jobs=2`, asserts exactly 2 admitted in the pass.

crates/spurctld/src/cluster.rs
yansun1996: Agreed, trimmed — kept the detailed comment at the enforcement site and dropped the earlier signpost.

shiv-tyagi (approved):

PR #481, opened 2026-07-21
feat(spur-cli): accept --qos for srun

yansun1996:
## What

`srun` did not accept a `--qos` flag, even though `sbatch` does and `JobSpec` already carries a `qos` field. This adds `--qos`/`-q` support to `srun`, on par with the other job-submission commands.

## Approach

Mirrors how `srun` already handles `--partition` and `--account`:
- New `-q`/`--qos` clap arg on `SrunArgs`.
- Environment-default resolution over `SPUR_QOS`, `SPUR_JOB_QOS`, `SLURM_QOS`, `SLURM_JOB_QOS` (SPUR-native alias precedes the SLURM twin; the direct input var precedes the allocation `*_JOB_QOS` twin). This lets an `srun` launched inside a Spur allocation inherit the allocation's QoS, while an explicit CLI value still wins.
- The resolved QoS is carried into the submitted `JobSpec.qos`, so it flows through the existing submit path (the same one `sbatch --qos` uses, including the controller's "QOS does not exist" validation and the QoS's effect on priority, limits, and preemption).

## Behavior in step mode (matches Slurm)

When `srun` runs inside an existing allocation it dispatches a *step* rather than submitting a new job. QoS is a job-level attribute fixed by the parent allocation, so the step inherits the allocation's QoS and an explicit step-level `--qos` has no effect. This matches Slurm exactly: Slurm's step-create request carries no QoS field, and its `srun` likewise accepts `--qos` in an allocation but the step still runs under the allocation's QoS (independent of whether accounting is enabled). The `--qos` flag takes effect on the standalone path, where `srun` submits a new job.

## Tests

- Unit tests for CLI parsing (short/long), env-var precedence (default, CLI override, SPUR-before-SLURM, allocation-twin inheritance), and `JobSpec` wiring (set + empty default).
- A native-host e2e that defines a QoS, submits an `srun --qos` job, and asserts the job reports that QoS.
- Validated end-to-end on an isolated 2-node deployment: an `srun --qos <known>` job runs and the QoS is recorded on the job; an unknown QoS is rejected identically to `sbatch`.

### DIFF
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -30,6 +30,10 @@ pub struct SrunArgs {
     #[arg(short = 'A', long)]
     pub account: Option<String>,
 
+    /// QoS
+    #[arg(short = 'q', long)]
+    pub qos: Option<String>,
+
     /// Number of nodes
     #[arg(short = 'N', long, default_value = "1")]
     pub nodes: u32,
@@ -250,6 +254,12 @@ fn resolve_srun_env(matches: &ArgMatches, args: &mut SrunArgs) -> Result<()> {
         ],
         &mut args.account,
     );
+    apply_str(
+        matches,
+        "qos",
+        &["SPUR_QOS", "SPUR_JOB_QOS", "SLURM_QOS", "SLURM_JOB_QOS"],
+        &mut args.qos,
+    );
     apply_num(
         matches,
         "nodes",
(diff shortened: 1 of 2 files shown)


yansun1996:
Rebased onto the latest upstream/main (no conflicts) and force-pushed.

shiv-tyagi (approved):
LGTM

PR #482, opened 2026-07-21
feat(spur-cli): interactive PTY sessions via srun --pty and sattach

shiv-tyagi:
## Summary

- Add `srun --pty` support: submits a `sleep infinity` holder job for resource allocation, then creates an interactive PTY step for the user's command via the `InteractiveSession` RPC. The holder is cancelled on session exit and the PTY exit code is propagated.
- Add retry loop (up to 5 attempts, 500ms backoff) around `CreateJobStep` + `InteractiveSession` to handle the race between the controller reporting a job as Running and the agent finishing setup.
- Split `interactive.rs` into `open_interactive_session` / `drive_interactive_session` so retry logic can match on `tonic::Status` codes instead of error message text.
- Deduplicate agent connection logic into a shared `connect_agent` helper (used by both `srun` and `sattach`).
- Fall back to `/bin/sh` when `/bin/bash` is missing in container rootfs (fixes container job failures on BusyBox/Alpine).
- Remove unused `rootfs` field from `TrackedJob`/`JobEntry`, replace unreachable priv-drop branch with `debug_assert`, and bound `read_proc_environ` to 1 MiB.

## Test plan

- [x] `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` (clean)
- [x] `cargo test --locked` (all pass)
- [x] Full 25-test bare-metal smoke suite across all 4 nodes:
  - Batch: non-container, container (BusyBox, Alpine), node selection
  - srun non-interactive: non-container, container
  - srun --pty: short commands, interactive shells, container (BusyBox, Alpine), all 4 nodes, exit code propagation, SIGINT cleanup, stderr passthrough, env vars, command-not-found, 5x rapid-fire stress test

### DIFF
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -486,11 +486,7 @@ fn build_srun_job_spec(args: &SrunArgs, work_dir: &str, io: &ResolvedIoPaths) ->
         cpus_per_task: args.cpus_per_task,
         memory_per_node_mb: memory_mb,
         gres,
-        script: if args.pty {
-            "#!/bin/sh\nsleep infinity\n".to_string()
-        } else {
-            build_command_script(&args.command)?
-        },
+        script: build_command_script(&args.command)?,
         work_dir: work_dir.to_string(),
         stdout_path: io.stdout.clone(),
         stderr_path: io.stderr.clone(),
@@ -526,7 +522,7 @@ fn install_ctrl_c_cancel(
     client: SlurmControllerClient<tonic::transport::Channel>,
     job_id: u32,
     user: String,
-) {
+) -> tokio::task::JoinHandle<()> {
     tokio::spawn(async move {
         let mut client = client;
         if tokio::signal::ctrl_c().await.is_ok() {
@@ -540,7 +536,7 @@ fn install_ctrl_c_cancel(
                 .await;
             std::process::exit(130);
         }
-    });
+    })
 }
 
 async fn wait_for_job_running(
@@ -746,7 +742,7 @@ async fn run_standalone_srun(args: &SrunArgs, hooks: &HooksConfig, work_dir: &st
 
     eprintln!("srun: Pending job allocation {}...", job_id);
 
-    install_ctrl_c_cancel(client.clone(), job_id, user.clone());
+    let ctrl_c_handle = install_ctrl_c_cancel(client.clone(), job_id, user.clone());
 
     let nodelist = wait_for_job_running(&mut client, job_id).await?;
     if !nodelist.is_empty() {
@@ -754,6 +750,7 @@ async fn run_standalone_srun(args: &SrunArgs, hooks: &HooksConfig, work_dir: &st
     }
 
     if args.pty {
+        ctrl_c_handle.abort();
         eprintln!("srun: opening interactive session on {}", nodelist);
         let result = run_interactive_pty(&args.controller, job_id, args.command.clone()).await;
         let _ = client
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -5,7 +5,6 @@ use crate::env_defaults::{apply_csv, apply_flag, apply_num, apply_str, apply_str
 use anyhow::{Context, Result};
 use clap::{ArgMatches, CommandFactory, FromArgMatches, Parser};
 use spur_core::config::HooksConfig;
-use spur_proto::proto::slurm_agent_client::SlurmAgentClient;
 use spur_proto::proto::slurm_controller_client::SlurmControllerClient;
 use spur_proto::proto::{
     CancelJobRequest, CompleteJobRequest, CreateJobStepRequest, GetJobRequest, GetNodeRequest,
@@ -197,7 +196,7 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
         if !args.overlap {
             anyhow::bail!("--jobid requires --overlap");
         }
-        let exit_code = run_interactive_overlap(&args, job_id).await?;
+        let exit_code = run_interactive_pty(&args.controller, job_id, args.command.clone()).await?;
         std::process::exit(exit_code);
     }
 
@@ -487,7 +486,11 @@ fn build_srun_job_spec(args: &SrunArgs, work_dir: &str, io: &ResolvedIoPaths) ->
         cpus_per_task: args.cpus_per_task,
         memory_per_node_mb: memory_mb,
         gres,
-        script: build_command_script(&args.command)?,
+        script: if args.pty {
+            "#!/bin/sh\nsleep infinity\n".to_string()
+        } else {
+            build_command_script(&args.command)?
+        },
         work_dir: work_dir.to_string(),
         stdout_path: io.stdout.clone(),
         stderr_path: io.stderr.clone(),
@@ -514,6 +517,7 @@ fn build_srun_job_spec(args: &SrunArgs, work_dir: &str, io: &ResolvedIoPaths) ->
             .collect(),
         container_remap_root: args.container_remap_root,
         srun_job: true,
+        pty: args.pty,
         ..Default::default()
     })
 }
@@ -749,6 +753,19 @@ async fn run_standalone_srun(args: &SrunArgs, hooks: &HooksConfig, work_dir: &st
         eprintln!("srun: job {} running on {}", job_id, nodelist);
     }
 
+    if args.pty {
+        eprintln!("srun: opening interactive session on {}", nodelist);
+        let result = run_interactive_pty(&args.controller, job_id, args.command.clone()).await;
+        let _ = client
+            .cancel_job(CancelJobRequest {
+                job_id,
+                signal: 0,
+                user: user.clone(),
+            })
+            .await;
+        std::process::exit(result?);
+    }
+
     let job = client
         .get_job(GetJobRequest { job_id })
         .await
@@ -854,10 +871,8 @@ async fn try_stream_output(
 
     let agent_addr = format!("http://{first_node}:6818");
 
-    let mut agent = match SlurmAgentClient::connect(agent_addr).await {
-        Ok(c) => c
-            .max_decoding_message_size(spur_proto::MAX_GRPC_MESSAGE_SIZE)
-            .max_encoding_message_size(spur_proto::MAX_GRPC_REQUEST_SIZE),
+    let mut agent = match crate::interactive::connect_agent(&agent_addr).await {
+        Ok(c) => c,
         Err(_) => return false,
     };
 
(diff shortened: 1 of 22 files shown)


crates/spurd/src/pty.rs
```diff
@@ -0,0 +1,181 @@
...
+/// Open a PTY pair and optionally set the initial window size on the slave.
+/// Returns `(master, slave)`.
+pub fn openpty_with_winsize(ws: Option<&WindowSize>) -> Result<(OwnedFd, OwnedFd)> {
+    let pty = openpty(None, None).context("openpty")?;
+
+    if let Some(ws) = ws {
+        unsafe {
+            libc::ioctl(pty.slave.as_raw_fd(), libc::TIOCSWINSZ, &ws.to_libc());
+        }
+    }
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed — now checks the return value and logs a warning. Not fatal since the client resends window size on first interactive message.

crates/spurd/src/executor.rs
```diff
@@ -324,30 +434,73 @@ async fn spawn_job_process(
...
-            );
+    // Build resolved output paths (empty for PTY mode since output goes to the terminal).
+    let (stdout_resolved, stderr_resolved) = if cfg.io_mode == LaunchIo::Pty {
+        (String::new(), String::new())
+    } else {
+        (
+            resolve_output_path(stdout_path, job_id, work_dir),
+            resolve_output_path(stderr_path, job_id, work_dir),
+        )
+    };
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed — PTY jobs now use `/dev/null` for stdout/stderr paths.

crates/spurd/src/executor.rs
```diff
@@ -1083,14 +1222,13 @@ async fn launch_container_job(
...
 
-            // Redirect stdout/stderr to the user-opened output files (inherited
-            // fds), then let close_inherited_fds reap the now-redundant originals.
+            // wire() must run before close_inherited_fds: it dup2's the
+            // job's fds onto 0/1/2, then close_inherited_fds reaps
+            // everything else > 2 (except ready_w).
             unsafe {
-                libc::dup2(stdout_fd.as_raw_fd(), libc::STDOUT_FILENO);
-                libc::dup2(stderr_fd.as_raw_fd(), libc::STDERR_FILENO);
+                let _ = raw_io.wire();
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed — `wire()` failure now writes the error to the ready pipe and calls `_exit(1)`.

crates/spurd/src/agent_server.rs
```diff
@@ -1769,20 +1638,302 @@ impl AgentService {
...
+        use std::process::Stdio;
+
+        let (master, slave) = crate::pty::openpty_with_winsize(winsize)
+            .map_err(|e| Status::internal(format!("openpty: {e}")))?;
+
+        let shell = if argv.is_empty() {
+            vec!["/bin/bash".to_string()]
+        } else {
+            argv.to_vec()
+        };
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed — probes `/proc/<pid>/root/bin/bash` for container jobs, falls back to `/bin/sh` when missing.

crates/spurd/src/agent_server.rs
```diff
@@ -1992,20 +1864,311 @@ impl AgentService {
         });
     }
 
+    /// Extract a `JobEntry` from a tracked running job for namespace entry.
+    async fn job_entry(&self, job_id: u32) -> Result<crate::job_entry::JobEntry, Status> {
```
yansun1996, line 1932: `job_entry` looks up the job by `job_id` with no check that the caller is the job's owner (or admin). Combined with `create_job_step` having the same gap, and `srun --jobid <id> --overlap` now making arbitrary-job attach directly reachable, this means any user could attach an interactive shell into another user's running job. `cancel_job` already has an owner check (`check_job_owner`) — would it be worth adding an equivalent check here (and threading a caller identity through `InitSession`/`CreateJobStepRequest`, which don't carry one today)?

crates/spurctld/src/server.rs
```diff
@@ -1237,7 +1237,17 @@ impl SlurmController for ControllerService {
...
+            .allocated_nodes
+            .first()
+            .and_then(|name| self.cluster.get_node(name))
+            .map(|n| {
+                let host = n.address.as_deref().unwrap_or(&n.name);
+                format!("{}:{}", host, n.port)
+            })
+            .unwrap_or_default();
+
+        Ok(Response::new(CreateJobStepResponse { step_id, node_addr }))
```
yansun1996, line 1260: Similarly to the note on `job_entry` in `agent_server.rs`, this doesn't check that the caller owns `job_id` before creating a step for it (and now also returns the node address, making direct agent access easier to reach). Worth the same ownership check `cancel_job` already applies elsewhere.

crates/spur-cli/src/srun.rs
```diff
@@ -725,6 +753,19 @@ async fn run_standalone_srun(args: &SrunArgs, hooks: &HooksConfig, work_dir: &st
         eprintln!("srun: job {} running on {}", job_id, nodelist);
     }
 
+    if args.pty {
```
yansun1996, line 755: This interactive-session block runs under the existing `install_ctrl_c_cancel` handler installed earlier in this function, which calls `std::process::exit(130)` directly on Ctrl-C — skipping destructors, including `RawModeGuard::drop` in `interactive.rs`. If a SIGINT arrives while `run_interactive_pty` has the terminal in raw mode, the user's terminal is left stuck afterward. The `--jobid --overlap` path (no handler installed at all) has the same failure mode via a different route. Worth funneling both through a path that drops the raw-mode guard before exiting.

crates/spurd/src/agent_server.rs
```diff
@@ -1992,20 +1864,311 @@ impl AgentService {
...
+            }) {
+                Ok(Ok(n)) => written += n,
+                Ok(Err(e)) => return Err(e),
+                Err(_would_block) => continue,
+            }
+        }
+        Ok(())
+    }
+
+    fn spawn_pty_in_job(
```
yansun1996, line 2126: The PID captured in `job_entry` is used here for `nsenter` without re-validating it still refers to the same process — if the job exits and the PID gets recycled before `nsenter` runs, the session could attach into an unrelated process's namespaces. `executor.rs` already uses a `pidfd`-based pattern for this exact class of bug elsewhere in this file family; might be worth reusing it here.

crates/spur-cli/src/srun.rs
```diff
@@ -1019,6 +1054,96 @@ fn warn_unsupported_cpu_bind(environment: &HashMap<String, String>) {
...
+async fn run_interactive_pty(controller: &str, job_id: u32, command: Vec<String>) -> Result<i32> {
+    let channel = spur_client::connect_channel(controller)
+        .await
+        .context("cannot connect to controller")?;
+    let mut ctrl = SlurmControllerClient::new(channel);
+
+    let winsize = crate::interactive::get_terminal_size();
+
+    let mut last_err: Option<anyhow::Error> = None;
+    for attempt in 0..5 {
```
yansun1996, line 1072: Each retry attempt calls `create_job_step` again without cancelling the step from the previous attempt, so a session that needs a few retries before `InteractiveSession` succeeds leaks orphaned `Running` steps (there's no cancel-step RPC to clean them up). Worth tracking the step from a failed attempt and cancelling/reusing it before retrying.

crates/spurd/src/agent_server.rs
```diff
@@ -1992,20 +1864,311 @@ impl AgentService {
...
+                                Some(interactive_input::Msg::Signal(sig)) => {
+                                    let _ = crate::pty::signal_foreground(
+                                        master_raw, child_pid, sig,
+                                    );
+                                }
+                                Some(interactive_input::Msg::Init(_)) | None => {}
+                            }
+                        }
+                        Some(Err(_)) | None => {
+                            unsafe { libc::kill(child_pid, libc::SIGHUP); }
```
yansun1996: This sends `SIGHUP` straight to `child_pid`, while the normal signal path a few lines up goes through `pty::signal_foreground` (process-group aware, via `tcgetpgrp`/`killpg`). If the child has spawned its own foreground process group, this direct kill won't reach it. Worth using `signal_foreground` here too for consistency.

crates/spurd/src/agent_server.rs
```diff
@@ -2705,4 +2874,112 @@ mod tests {
...
+                msg: Some(interactive_input::Msg::Stdin(b"hello\n".to_vec())),
+            }))
+            .await
+            .unwrap();
+
+        let mut collected = Vec::new();
+        let deadline = tokio::time::Instant::now() + std::time::Duration::from_secs(5);
+        loop {
+            let remaining = deadline.duration_since(tokio::time::Instant::now());
+            match tokio::time::timeout(remaining, out_rx.recv()).await {
```
yansun1996: This wraps `out_rx.recv()` in `tokio::time::timeout`, which conflicts with the repo's test convention of no added timeouts (tests should fail deterministically, not via a wall-clock race). Would a bounded polling loop without the timeout work here instead?

yansun1996 (changes requested):
Thanks for this — a substantial and well-tested feature (the manual smoke-test matrix in the description is thorough). Two things worth resolving before merge, left as inline comments:

- `job_entry`/`create_job_step` don't check that the caller owns the target job. `cancel_job` already has this check (`check_job_owner`) — worth mirroring it here, especially now that `srun --jobid --overlap` makes arbitrary-job attach directly reachable.
- The Ctrl-C handling around the new interactive-session path can leave a user's terminal stuck in raw mode if an external SIGINT arrives mid-session (see inline comment for details).

A few smaller should-fix items, also inline: the retry loop around `CreateJobStep`+`InteractiveSession` creates a new step on each attempt without cancelling the previous one (leaks orphaned "Running" steps on a flaky attach); the disconnect path sends `SIGHUP` straight to the child PID instead of through the process-group-aware signal path used elsewhere; the PID used for `nsenter` isn't re-validated against reuse (a `pidfd`-based pattern already exists in `executor.rs` for this); and one new test uses `tokio::time::timeout`, which AGENTS.md asks us not to do in tests.

Non-blocking: no idle timeout on the PTY bridge connection (pre-existing pattern, could be a follow-up), and the new proto tests (`t50_92*`) are fairly shallow (construct-and-assert, no round-trip) — worth strengthening opportunistically.

crates/spur-cli/src/srun.rs
shiv-tyagi: Fixed this and removed exit(130). 

Anything more than that is a scope creep for this PR. This is standard terminal program behavior (ssh, vim, tmux all have the same property). Raw mode is restored on normal return via `RawModeGuard::drop`, and the `exit(130)` handler is aborted before entering the PTY session. No controller or agent side impact.

crates/spurd/src/agent_server.rs
shiv-tyagi: `signal_foreground` already uses `tcgetpgrp` + `killpg` for process-group aware delivery (pty.rs:97-104). Falls back to direct `kill(child_pid)` only when `tcgetpgrp` returns <= 0 (no foreground pgrp). This is the correct behavior.

crates/spur-cli/src/srun.rs
shiv-tyagi: Fixed. The retry loop now caches `(step_id, node_addr)` from the first successful `CreateJobStep` and reuses it on subsequent retries of the `InteractiveSession` RPC. Only re-creates the step if `CreateJobStep` itself failed.

crates/spurd/src/agent_server.rs
shiv-tyagi: Fixed. Switched to `#[tokio::test(start_paused = true)]` and replaced the real 6s `sleep` with `tokio::time::advance(Duration::from_secs(6))`. Test now completes in 0.00s instead of 6s.

crates/spurd/src/agent_server.rs
shiv-tyagi: Legitimate concern. This is pre-existing behavior though, not introduced by this PR. The overlap/attach path existed before; this PR just centralized privilege dropping and PTY wiring. Threading caller identity through gRPC metadata and adding ownership checks to `job_entry` + `CreateJobStep` is a separate effort (needs schema changes to the proto, auth middleware on the agent, etc.). Will track separately.

crates/spurctld/src/server.rs
shiv-tyagi: Same as above — valid point, pre-existing gap. `CreateJobStep` didn't have an ownership check before this PR either. Adding one requires propagating caller identity through the gRPC request (today it's unauthenticated). Out of scope for a privilege-drop refactor, but worth a follow-up issue.

crates/spurd/src/agent_server.rs
shiv-tyagi: Valid race window. This is pre-existing — the PID-based nsenter pattern was already in place before this PR. Using `pidfd_open` + `/proc/<pidfd>/ns/` would close the TOCTOU gap. However, it requires kernel 5.3+ pidfd support and reworking the namespace entry path, which is orthogonal to privilege dropping. Out of scope here, will track as a hardening follow-up.

yansun1996 (approved):
LGTM

PR #483, opened 2026-07-21
fix(spur-cli): limit user cancellation to active jobs

jamesETsmith:
## Motivation

  Running `scancel -u USER` can query every retained job for that user, including completed, failed, and cancelled jobs.
This causes unnecessary work and, with older clients, misleading errors for each historical job.

## Technical Details

  Default filter-based cancellation now requests only cancellable job states from the controller. The existing client-
side terminal-state check remains as a defensive safeguard against unexpected responses.

  Regression coverage verifies that active states are included and terminal states are excluded from the default request

## Test Plan

  - Run focused `scancel` unit tests.
  - Run the complete `spur-cli` test suite.
  - Check formatting and linting.

## Test Result

- `cargo test -p spur-cli --locked`: 203 passed.
- `cargo fmt --check`: passed.
- `cargo clippy -p spur-cli --all-targets --locked -- -D warnings`: passed.
- `git diff --check`: passed.

## Submission Checklist

  - [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests


yansun1996 (approved):

crates/spur-cli/src/scancel.rs
```diff
@@ -168,6 +159,33 @@ fn is_cancellable(proto_state: i32) -> bool {
...
+            states
+                .split(',')
+                .filter_map(|state| parse_state(state.trim()))
+                .map(|state| state as i32)
+                .collect()
+        })
+        .unwrap_or_else(cancellable_states)
+}
+
+fn cancellable_states() -> Vec<i32> {
```
yansun1996, line 186: This hardcodes the same "non-terminal" set that `is_terminal()`/`is_cancellable()` just above already derive dynamically. Could this be generated instead — e.g. `spur_core::job::JobState::ALL.iter().filter(|s| !s.is_terminal()).map(|s| s.to_proto_i32()).collect()` — the same idiom `sdiag.rs` already uses for deriving state subsets? That way if a new `JobState` variant is ever added, this list can't silently drift out of sync with `is_terminal()`; right now it'd need a matching manual update here that nothing would enforce.

yansun1996 (changes requested):
Thanks for the fix — the change to request only cancellable job states by default is a real improvement (avoids fetching a user's entire job history for `scancel -u`).

Requesting a small change before merge, left as an inline comment: `cancellable_states()` hardcodes the same "non-terminal" state set that `is_terminal()`/`is_cancellable()` in this same file already derive dynamically. Generating it instead (see inline comment for the suggested form) avoids a second, easy-to-forget source of truth for which states are cancellable.

crates/spur-cli/src/scancel.rs
jamesETsmith: Updated the PR with this, thanks for the suggestion! I also added a catch to the filter_states function so it catches invalid/malformed states passed on the cli

shiv-tyagi (approved):
LGTM. 

We can merge after @yansun1996's approval.

yansun1996 (approved):

jamesETsmith:
Thanks @shiv-tyagi and @yansun1996!

PR #488, opened 2026-07-22
feat(spurctld): report node-resolved job output path in scontrol

biluriuday:
## Summary

`scontrol show job` previously showed the controller's *computed* `StdOut`/`StdErr`, which could be relative and — when a node couldn't use the job's `work_dir` and fell back to `/tmp` — pointed at the wrong place entirely. The agent already resolves the real absolute path at launch, but the controller never learned it.

This PR reports the real path back and makes controller/agent resolution provably consistent:

- **Report the resolved path.** `LaunchJobResponse` carries the agent's resolved `stdout_path`/`stderr_path`. The controller stores the primary node's (`task_offset == 0`) paths, selected by node identity since the dispatch `JoinSet` completes out of order. `scontrol` prefers the reported path and falls back to the computed one when unset (pending jobs, mixed-version agents). No file contents are copied — only the location.
- **Single shared resolver.** The agent previously expanded only `%j`/`%J`, so a job using `%x`/`%u`/`%N` in `-o`/`-e` wrote (and would report) an unexpanded filename that diverged from the controller. Both sides now use one resolver in `spur-core` (`resolve_output_pattern` / `OutputPathContext`), so reported and computed paths always agree.
- **Absolute paths, matching Slurm.** Relative patterns anchor to `work_dir`, or to `DEFAULT_WORK_DIR` (`/tmp`) when unset — mirroring where the agent actually launches the job.
- **No stale paths.** The reported path is cleared on requeue/evict so `scontrol` never shows a location on a node the job has left; the next dispatch records the new primary's path.

## Design notes / trade-offs

- The reported path is **best-effort, leader-local advisory metadata** — written to the in-memory job, not replicated via Raft/WAL. After a failover or restart-from-WAL, a leader that never observed the launch falls back to the computed (now absolute) path. This is acceptable because the field is display-only.
- Out of scope: copying back file contents, streaming/`--wait`. The batch script still fans out to all allocated nodes (pre-existing behavior); only the primary node's path is reported.

## Test plan

- [x] `spur-core`: resolver produces absolute paths for relative/`%j`/`%x`/`%u`/`%N` patterns and anchors empty `work_dir` to `/tmp`.
- [x] `spurd`: agent expands the same codes as the controller; a successful launch reports absolute paths.
- [x] `spurctld`: primary-node path stored regardless of `JoinSet` completion order; failed primary leaves paths unset; `job_to_proto` prefers the reported path else the absolute computed one; requeue clears the stale path.
- [x] Updated `t28`/`t50` path assertions to the new absolute form.
- [x] Native e2e: submit with a relative `-o` and assert `scontrol show job` reports the absolute path.
- [x] `cargo fmt --check` + SPDX header pre-commit hook pass.

### DIFF
--- a/crates/spur-core/src/job.rs
+++ b/crates/spur-core/src/job.rs
@@ -664,8 +664,9 @@ pub struct Job {
     pub bb_stage_state: BbStageState,
 
     /// Absolute path the primary agent resolved at launch (incl. `/tmp` fallback).
-    /// Best-effort, not Raft-replicated: `None` before launch and after a failover/
-    /// restart that missed it, when queries use `resolved_stdout`/`resolved_stderr`.
+    /// Best-effort advisory: set post-launch, not via WAL replay (may ride along in a
+    /// snapshot). `None` before launch or after a failover that missed it, when
+    /// queries fall back to `resolved_stdout`/`resolved_stderr`.
     #[serde(default)]
     pub actual_stdout_path: Option<String>,
     #[serde(default)]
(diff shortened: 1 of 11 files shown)


crates/spurctld/src/scheduler_loop.rs
```diff
@@ -1099,6 +1117,15 @@ async fn dispatch_job_to_nodes(
...
     }
 
+    // Only surface the primary's paths on a clean launch; a partial/total
+    // failure requeues or evicts the job, leaving it to fall back to the
+    // computed path on the next attempt.
+    if failures == 0 {
+        if let Some(outcome) = primary_outcome {
+            cluster.set_job_output_paths(job_id, outcome.stdout_path, outcome.stderr_path);
+        }
+    }
```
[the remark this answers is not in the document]
biluriuday: handled it inside set_job_output_paths method

yansun1996 (commented):
Thanks for centralizing the output-path resolution — this cleanly fixes the `/tmp`-fallback display bug and keeps agent/controller resolution in sync.

One thing worth addressing before merge, left as an inline note: in `resolve_output_pattern`, the relative/absolute check runs after `%x`/`%u`/`%N` are substituted into the pattern. Since job names aren't validated, a name that itself looks like an absolute path (or contains a literal `%u`/`%N`/`%a`/`%A`) can bypass the `work_dir` anchor or get re-expanded unexpectedly. See the inline comment for details.

A few smaller suggestions, non-blocking:
- The `set_job_output_paths` doc comment says this state is "not Raft-replicated," but it does travel via `ClusterSnapshot`. Worth tightening the wording (e.g. "not applied via WAL op replay, though it may ride along in periodic snapshots") so the failover behavior is described accurately.
- A couple of test gaps worth closing if convenient:
  - Cancelling a `Running` job doesn't clear `actual_stdout_path` (unlike requeue) — could use an explicit assertion either way so the behavior is intentional rather than incidental.
  - The `if failures == 0` gate in `dispatch_job_to_nodes` isn't exercised by a test where the *primary* succeeds but a *secondary* node fails (existing test only covers the primary being unreachable).
  - Might be worth a test asserting `resolve_output_pattern` stays absolute even when `work_dir` itself is relative, since that's the core guarantee this PR establishes.

Nice work overall — the shared resolver approach is a good fix for the agent/controller divergence.

crates/spur-core/src/job.rs
```diff
@@ -788,25 +799,73 @@ impl Job {
...
+    result = result.replace("%x", ctx.name);
+    if let Some(tid) = ctx.array_task_id {
+        result = result.replace("%a", &tid.to_string());
+        result = result.replace("%A", &ctx.array_job_id.unwrap_or(ctx.job_id).to_string());
+    }
+    if let Some(node) = ctx.node {
+        result = result.replace("%N", node);
+    }
+    result = result.replace("%u", ctx.user);
+    if std::path::Path::new(&result).is_relative() {
```
yansun1996: This check runs after `%x`/`%u`/`%N` have already been substituted into `result`. Since job names (`ctx.name`, from `-J`) aren't validated anywhere, a name like `/abs/evil` turns an apparently-relative pattern (e.g. `-o %x.out`) into an absolute path that skips the `work_dir` anchor entirely — the exact inconsistency this PR is trying to close. It also affects the agent's real output file, not just `scontrol` display (via `executor.rs`'s `resolve_output_path`), though not a privilege-escalation risk since the agent opens the file after dropping to the job's own uid.

Separately, since substitution is a chain of `.replace()` calls, a name containing a literal `%u`/`%N`/`%a`/`%A` gets re-expanded by a later replace call.

Would it be worth checking relativity against the original `pattern` (before substitution) and rejecting/escaping `/`, `..`, and `%` in `name`/`user` before splicing them in? A single-pass template scan instead of sequential replaces would also close the re-expansion gap.

biluriuday: addressed the issues

yansun1996:
rebased on top of main branch to resolve conflicts

yansun1996 (approved):
LGTM

PR #493, opened 2026-07-23
fix(spur-cli): display node features in sinfo

### LINKED ISSUE #355: sinfo/spur nodes: node features (%f) not displayable — missing from NodeInfo proto and sinfo resolver
## What happened

Node `features` configured in `spur.conf` (e.g. `features = ["mi350x", "atl", "<hostname>"]`) cannot be displayed via the CLI. `sinfo`/`spur nodes` with the `%f` (AVAIL_FEATURES) format specifier always prints `?`:

```
$ sinfo -N -o "%n %f"
HOSTNAMES AVAIL_FEATURES
ml-ai-...-11 ?
ml-ai-...-03 ?
...
```

## What was expected

`%f` should print each node's configured features, the same way Slurm's `sinfo -o "%f"` does — so users can see which labels are available to target with `-C/--constraint`.

## Root cause (two gaps)

1. **`NodeInfo` proto has no `features` field.** `proto/slurm.proto` (`message NodeInfo`, ~line 201) carries `name`, `state`, `total_resources`, `arch`, `os`, `switch_name`, `active_reservation`, and `labels` (map, field 51) — but not the `features` list that `NodeConfig` (`crates/spur-core/src/config.rs:497`) and the scheduler use. So feature data never reaches the CLI.

2. **`resolve_node_field` has no `'f'` case.** `crates/spur-cli/src/sinfo.rs` (~lines 175-219) handles `N n P R t T c m G D a O e l` but not `f`, so it falls through to `_ => "?"`. The header map (`format_engine.rs`) does know `'f' => "AVAIL_FEATURES"`, which is why the column header renders but every value is `?`.

## Impact

- Constraint *matching* works correctly — the scheduler reads `features` from `spur.conf` directly. Verified: `salloc -N 1 -C <hostname>` allocates exactly that node, and `-C mi350x` matches by GPU type.
- Only the *display* is broken: there is no user-facing way to discover what features a node has. Users must read `spur.conf` (admin-only) to know what `-C` values are valid.

## Suggested fix

1. Add `repeated string features` to `NodeInfo` in `proto/slurm.proto`.
2. Populate it from `NodeConfig.features` where the controller builds `NodeInfo` (`crates/spurctld/src/server.rs`).
3. Add a `'f' => node.features.join(",")` case to `resolve_node_field` in `crates/spur-cli/src/sinfo.rs`.

## Secondary note

`%n` (HOSTNAMES) only resolves in per-node mode (`-N`); without `-N` the grouped/partition view resolver (`resolve_partition_field`) returns `?` for `%n`. This matches Slurm behavior reasonably (use `%N` in grouped view), but may be worth documenting.

## Environment

- Spur 0.3.0
- 3-controller HA + 12 compute nodes

joshkmartinez:
## Summary

- expose each node's resolved features through `NodeInfo`
- render `%f` in node-oriented `sinfo`, using a comma-separated list or `(null)` when empty
- cover both the controller conversion and populated/empty CLI output

## Design

Dynamic registration already maps node labels through `NodeConfig` into the controller's `Node.features`. this adds the missing transport and rendering path w/o changing feature resolution

Partition-oriented `%f` is out of scope b/c those rows are currently grouped only by node state; reporting heterogeneous feature sets correctly there requires feature-aware grouping.

## Testing

- `PROTOC=/opt/homebrew/bin/protoc cargo test -p spur-cli --locked`
- `PROTOC=/opt/homebrew/bin/protoc cargo test -p spurctld --locked`
- `PROTOC=/opt/homebrew/bin/protoc cargo clippy -p spur-cli -p spurctld --all-targets --locked -- -D warnings`
- local release-build smoke test using a real `spurctld`, synthetic `RegisterAgent` calls, and `sinfo -N`:

  ```text
  gpu-node1|mi350x,atl
  gpu-node2|(null)
  ```

Attempted the full workspace test and Clippy commands, but the untouched  `spurd` crate does not compile on macOS because it uses Linux mount namespace, Landlock, and seccomp APIs.
Both changed crates pass their complete test suites and Clippy checks.

Closes #355

### DIFF
--- a/crates/spur-cli/src/sinfo.rs
+++ b/crates/spur-cli/src/sinfo.rs
@@ -215,6 +215,13 @@ fn resolve_node_field(
         }
         'O' => node.cpu_load.to_string(),
         'e' => node.free_memory_mb.to_string(),
+        'f' => {
+            if node.features.is_empty() {
+                "(null)".into()
+            } else {
+                node.features.join(",")
+            }
+        }
         'l' => "UNLIMITED".into(), // Would need partition context
         _ => "?".into(),
     }
@@ -458,6 +465,29 @@ mod tests {
         assert!(lines[0].contains("idle"));
         assert!(lines[1].contains("n2"));
         assert!(lines[1].contains("down"));
+    }
+
+    #[test]
+    fn node_oriented_output_displays_available_features() {
+        let fields = format_engine::parse_format("%n|%f", &format_engine::sinfo_header);
+        let partitions = vec![make_partition("gpu", true)];
+        let mut node = make_node("gpu-node1", NodeState::NodeIdle, "gpu");
+        node.features = vec!["mi350x".into(), "atl".into()];
+
+        let lines = render_sinfo_output(&fields, &partitions, &[node], true);
+
+        assert_eq!(lines, ["gpu-node1|mi350x,atl"]);
+    }
+
+    #[test]
+    fn node_oriented_output_displays_null_when_features_are_empty() {
+        let fields = format_engine::parse_format("%n|%f", &format_engine::sinfo_header);
+        let partitions = vec![make_partition("cpu", true)];
+        let node = make_node("cpu-node1", NodeState::NodeIdle, "cpu");
+
+        let lines = render_sinfo_output(&fields, &partitions, &[node], true);
+
+        assert_eq!(lines, ["cpu-node1|(null)"]);
     }
 
     // --- effective_state_str tests ---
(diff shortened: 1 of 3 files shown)


yansun1996 (approved):
Reviewed the diff, proto field usage, and rendering logic — looks correct and well-scoped. The `%f` implementation matches the existing `'G'` field's convention, the proto field number doesn't collide, old/new client-server compatibility is safe, and the tests exercise the real code paths. Good call scoping out partition-oriented output given the grouping constraints there. LGTM.

PR #495, opened 2026-07-23
fix(spurctld): unify pending job classification

### LINKED ISSUE #349: spurctld: pending_jobs() and tag_blocked_pending_reasons() duplicate every block evaluation with a must-match ordering contract
## Problem

In `crates/spurctld/src/cluster.rs`, job blocking is evaluated twice:

- `pending_jobs()` holds a read lock on jobs and **drops** structurally/transiently blocked jobs from the scheduling set, but cannot write back the reason.
- `tag_blocked_pending_reasons()` re-evaluates every block type (partition, dependency, QoS, reservation, license) to set the displayed `pending_reason`.

Because the two passes are independent, every block type must be evaluated twice, and the two evaluations must apply the blocks in the **same order** (currently `partition → dep → QoS → resv → lic`). This ordering is an implicit contract: if the two passes drift, a job can display one reason while being dropped for another, or be tagged with a reason while still re-entering scheduling each cycle.

## Expected

A single evaluation should determine both whether a job is dropped from scheduling and the reason shown, eliminating the duplicated scan and the ordering coupling.

## Notes

Raised by @shiv-tyagi in review of #326. One direction: have `pending_jobs()` return the drop reason alongside the filtered list so the caller applies reasons in one write pass. Out of scope for #326, which only added the partition-block paths.

maybeharshit:
## Summary

- Replace the separate `tag_blocked_pending_reasons()` + `pending_jobs()` scheduler passes with one `classify_pending_jobs()` evaluation.
- Add `pending_jobs_and_tag_reasons()` for the scheduler loop; keep `pending_jobs()` read-only for callers that only need the filtered list.
- Apply block reasons in a single write pass after the read snapshot, with the same re-checks for holds, deadlines, and active `BeginTime`.

## Behavior

Block order is unchanged: partition → dependency → account → QoS → reservation → (priority sort) → license → burst buffer.

Greedy license/BB drops still exclude lower-priority jobs without tagging them as absolute shortages. Scheduling holds, `DeadLine`, and active `BeginTime` holds are not overwritten.

Fixes #349

## Test plan

- [x] `cargo test --locked`
- [x] 3-node LXD cluster review for pending-reason edge cases

### DIFF
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -249,6 +249,17 @@ pub struct ClusterManager {
     sched_stats: OnceLock<Arc<SchedStatsCollector>>,
 }
 
+struct PendingJobClassification {
+    jobs: Vec<Job>,
+    blocked: Vec<(JobId, PendingReason)>,
+}
+
+struct PendingJobCandidate {
+    job: Job,
+    scheduling_eligible: bool,
+    tag_reason: bool,
+}
+
 impl ClusterManager {
     pub fn new(config: SlurmConfig, _state_dir: &Path) -> anyhow::Result<Self> {
         let partitions = config.build_partitions();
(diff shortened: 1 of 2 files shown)


crates/spurctld/src/cluster.rs
```diff
@@ -1970,26 +1971,27 @@ impl ClusterManager {
...
-            pending.retain(|job| {
+            let available = self.available_licenses_with(&jobs);
+            let mut remaining = available.clone();
+            candidates.retain(|candidate| {
+                let job = &candidate.job;
+                if let Some(reason) = license_block(job, &available) {
+                    record_blocked(&mut blocked, candidate, reason);
+                    return false;
+                }
+                if !candidate.scheduling_eligible {
```
yansun1996, line 1992: This and the burst-buffer gate below handle `!scheduling_eligible` with opposite outcomes (keep here, drop there). It's correct today only because burst-buffer runs last and checks eligibility before its own short-circuits — would you consider a dedicated `retain(|c| c.scheduling_eligible)` after both gates' shortage checks instead, so exclusion doesn't depend on ordering? Also seems worth locking in with a couple of tests: an array job at `array_max_concurrent` capacity with a running sibling (asserting exclusion from `pending_jobs_and_tag_reasons()`), and the same combined with a scarce license/BB request (asserting the pool isn't consumed and it's still excluded).

crates/spurctld/src/cluster.rs
```diff
@@ -2006,24 +2008,27 @@ impl ClusterManager {
...
                 let req = extract_bb_requirement(&job.spec);
                 if req == 0 {
-                    return true; // no BB -> unaffected by the gate
+                    return true;
                 }
                 match job.bb_stage_state {
-                    // Capacity already reserved in bb_capacity_in_use(), so
-                    // don't double-count `remaining`.
                     BbStageState::Ready => true,
                     BbStageState::Staging => false,
```
yansun1996: Now that `Staging` is handled explicitly above, this arm looks unreachable — could probably drop it and just check `Ready` before the `None` branch.

crates/spurctld/src/cluster.rs
```diff
@@ -2304,86 +2315,13 @@ impl ClusterManager {
         cancelled
     }
 
-    /// Set `pending_reason` for jobs `pending_jobs()` drops from scheduling
-    /// (dependency/QoS/license/reservation), which never reach
-    /// `update_pending_reasons()` and would otherwise show a stale reason.
-    /// Leader-only; mirrors `cancel_unsatisfiable_dependency_jobs()`.
+    /// Reclassify pending jobs and apply their displayed block reasons.
```
yansun1996, line 2317: The previous doc noted this is leader-only (mirroring `cancel_unsatisfiable_dependency_jobs()`), enforced only by the scheduler-loop caller, not this function itself. Worth keeping that note so a future caller doesn't invoke it off-leader?

yansun1996 (changes requested):
Nice consolidation — cutting the duplicate pass and reusing one classification for both scheduling and the displayed reason removes a real drift risk.

One concern before merging: the license and burst-buffer gates handle the new `scheduling_eligible` flag with opposite polarity, and the correct result today depends on gate ordering rather than being explicit — see inline comment for a suggested fix, plus two test cases worth adding to lock in the behavior.

A couple of smaller notes inline. Thanks for the cleanup!

PR #497, opened 2026-07-23
fix(oci): key container images by canonical reference

shiv-tyagi:
## What

`spur image import` and the agent-side image resolution (used by `srun --container-image`) keyed stored squashfs files off the raw reference string via `sanitize_name`. Equivalent references for the same image therefore mapped to different files:

| Reference | Stored file (before) |
|---|---|
| `busybox` | `busybox.sqsh` |
| `busybox:latest` | `busybox+latest.sqsh` |
| `docker://busybox:latest` | `busybox+latest.sqsh` |
| `docker.io/library/busybox:latest` | `docker.io+library+busybox+latest.sqsh` |

Import one form and reference another (e.g. import `busybox`, run `srun --container-image busybox:latest`) and the job failed with a spurious "not found, import it first", even though `parse_image_ref` already knew all four forms are the same image.

## Approach

Add `ImageRef::canonical()` (→ `registry/repository:tag`) and `image_file_stem()` in `spur-net`, and derive filenames from the normalized reference instead of the raw string. Applied consistently in:

- the native OCI puller (`pull_image`)
- the `dockerd://` / `podman://` importers
- `spur image remove`
- the agent's `resolve_image` / `remove_image`

All equivalent references now collapse to a single stored image (`docker.io+library+busybox+latest.sqsh`). The local `sanitize_name` in `spurd` is retained only for `--container-name`, a separate namespace.

## Compatibility

Not backward compatible: images stored under the old raw-string names no longer resolve and must be re-imported. Acceptable pre-release.

## Testing

- New unit tests in `spur-net` verifying equivalent refs collapse to one canonical name/stem, plus a custom-registry case.
- Updated the `spurd` user-home resolution test to use the canonical stem.
- Verified end-to-end: all four import forms produce one file, and `spur image remove busybox` deletes an image imported as `busybox:latest`.
- `cargo test -p spur-net -p spurd` and `cargo clippy -p spur-net -p spurd -p spur-cli --all-targets` pass clean.

## Note

`spur image list` renders `+` as `/`, so canonical stems display as `docker.io/library/busybox/latest` (tag separator shown as `/`). Pre-existing display quirk, left untouched to keep this focused.

### DIFF
--- a/crates/spur-net/src/oci.rs
+++ b/crates/spur-net/src/oci.rs
@@ -116,6 +116,18 @@ pub fn image_file_stem(image: &str) -> String {
     sanitize_name(&parse_image_ref(image).canonical())
 }
 
+/// Render a stored filename stem back to a canonical image reference for display.
+///
+/// The last `+` is always the tag separator (canonical form guarantees a tag),
+/// and remaining `+` map back to `/`. Port-bearing registries lose the port
+/// colon (shown as `/`) since `sanitize_name` maps both `:` and `/` to `+`.
+pub fn display_name(stem: &str) -> String {
+    match stem.rsplit_once('+') {
+        Some((path, tag)) => format!("{}:{}", path.replace('+', "/"), tag),
+        None => stem.to_string(),
+    }
+}
+
 /// Pull an image from a registry and create a squashfs file.
 ///
 /// Returns the path to the squashfs file.
@@ -128,7 +140,7 @@ pub async fn pull_image(image: &str, output_dir: &Path) -> anyhow::Result<PathBu
         "pulling image"
     );
 
-    let sanitized = image_file_stem(image);
+    let sanitized = sanitize_name(&image_ref.canonical());
     let sqsh_path = output_dir.join(format!("{}.sqsh", sanitized));
 
     if sqsh_path.exists() {
@@ -800,6 +812,29 @@ mod tests {
             image_file_stem("nvcr.io/nvidia/pytorch:24.01"),
             "nvcr.io+nvidia+pytorch+24.01"
         );
+    }
+
+    #[test]
+    fn test_canonical_port_bearing_registry() {
+        let r = parse_image_ref("localhost:5000/myimage:dev");
+        assert_eq!(r.canonical(), "localhost:5000/myimage:dev");
+        assert_eq!(
+            image_file_stem("localhost:5000/myimage:dev"),
+            "localhost+5000+myimage+dev"
+        );
+    }
+
+    #[test]
+    fn test_display_name() {
+        assert_eq!(
+            display_name("docker.io+library+busybox+latest"),
+            "docker.io/library/busybox:latest"
+        );
+        assert_eq!(
+            display_name("nvcr.io+nvidia+pytorch+24.01"),
+            "nvcr.io/nvidia/pytorch:24.01"
+        );
+        assert_eq!(display_name("alpine"), "alpine");
     }
 
     #[test]
--- a/crates/spur-net/src/oci.rs
+++ b/crates/spur-net/src/oci.rs
@@ -98,6 +98,24 @@ pub fn parse_image_ref(image: &str) -> ImageRef {
     }
 }
 
+impl ImageRef {
+    /// Canonical `registry/repository:tag` form.
+    ///
+    /// Equivalent references (`busybox`, `busybox:latest`, `docker://busybox`,
+    /// `docker.io/library/busybox:latest`) all normalize to the same string.
+    pub fn canonical(&self) -> String {
+        format!("{}/{}:{}", self.registry, self.repository, self.tag)
+    }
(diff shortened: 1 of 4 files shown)


crates/spurd/src/container.rs
```diff
@@ -1469,7 +1469,12 @@ mod tests {
         // test_resolve_job_user_home_for_current_uid +
         // test_compose_image_dirs_includes_existing_user_home_dir.)
         let images_dir = tempfile::tempdir().unwrap();
-        let image_path = images_dir.path().join("myimage.sqsh");
+        // Stored under the canonical stem, matching how `spur image import`
+        // names files (registry+repository+tag).
+        let image_path = images_dir.path().join(format!(
```
yansun1996, line 1474: This pins down that the stored filename matches `spur image import`'s naming convention, but doesn't yet cover the actual scenario from the bug report — resolving an image under a *different* equivalent reference than the one it was imported under. Might be worth adding a case that writes the fixture under `image_file_stem("docker://busybox:latest")` and then calls `resolve_image("busybox", ...)`, asserting it still resolves.

crates/spur-net/src/oci.rs
```diff
@@ -750,6 +768,40 @@ mod tests {
...
+                image_file_stem(r),
+                "docker.io+library+busybox+latest",
+                "ref: {}",
+                r
+            );
+        }
+    }
+
+    #[test]
+    fn test_canonical_custom_registry() {
```
yansun1996, line 806: Might be worth adding a case for a port-bearing registry too, e.g. `localhost:5000/myimage:dev` — `sanitize_name` maps both `/` and `:` to `+`, so it'd be good to have a test pinning down that the port's colon and the tag's colon don't produce anything unexpected.

yansun1996 (changes requested):
Solid fix — the canonicalization approach (`ImageRef::canonical()` + `image_file_stem()`) is clean, and the new unit tests clearly pin down the collapse behavior for equivalent references. Left inline notes on a missed call site and a couple of smaller follow-ups. The backward-incompatibility for pre-fix images is already called out in the description and seems like a fair trade-off pre-release.

yansun1996:
Two more notes that fall outside this diff's changed lines, so they can't be anchored as inline review comments:

**`crates/spur-cli/src/sbatch.rs:621`** (`resolve_container_image`) — this still calls the old `spur_net::oci::sanitize_name(image)` instead of `image_file_stem(image)`. Since storage now uses the canonical stem everywhere else in this PR, this submit-time fast path (resolving `--container-image` to an absolute path for shared-filesystem setups) will no longer find images imported under any non-bare-string reference, and will silently fall through to raw-string resolution — the same class of bug this PR fixes, just at a different call site. Worth switching this to `image_file_stem` too.

**`crates/spur-cli/src/image.rs:314`** (`cmd_list`) — since storage now canonicalizes before sanitizing, the `+` → `/` reverse-mapping here shows every Docker Hub image with the full `docker.io/library/...` prefix (e.g. `busybox` → `docker.io/library/busybox/latest`), which is more verbose than before for the common case, and the tag separator now looks identical to a path separator. Might be worth a friendlier display transform (e.g. only replace the last `+` with `:` for the tag) as a fast follow-up — cosmetic only, not blocking.

shiv-tyagi:
All review comments addressed in 4121615:

**sbatch.rs:621 (missed call site)** — switched `resolve_container_image` to `image_file_stem`. Added `test_resolve_container_image_cross_form` verifying that an image imported as `docker://busybox:latest` resolves via bare `busybox`.

**container.rs:1474 (cross-form test)** — added `test_resolve_image_cross_form_reference` exercising the actual bug scenario (import under one form, resolve under another).

**oci.rs:794 (port-bearing registry)** — added `test_canonical_port_bearing_registry` pinning `localhost:5000/myimage:dev`.

**oci.rs:131 (redundant reparse)** — now derives the stem from the already-parsed `image_ref` instead of reparsing the raw string.

**image list display** — added `display_name()` in oci.rs that maps the last `+` back to `:` (the tag separator), and uses `/` for the rest. `spur image list` now shows `docker.io/library/busybox:latest` instead of `docker.io/library/busybox/latest`. Known limitation: port-bearing registries render the port colon as `/` (`localhost/5000/myimage:dev`), inherent to `sanitize_name` mapping both `:` and `/` to `+`. Fixing this would require a reversible on-disk separator scheme, out of scope for this PR.

crates/spur-net/src/oci.rs
shiv-tyagi: Fixed

crates/spurd/src/container.rs
shiv-tyagi: Fixed

yansun1996 (approved):
Thanks for the quick turnaround — all the follow-ups look good. LGTM.

PR #500, opened 2026-07-24
fix(spurctld): unify pending job classification

### LINKED ISSUE #349: spurctld: pending_jobs() and tag_blocked_pending_reasons() duplicate every block evaluation with a must-match ordering contract

maybeharshit:
## Summary

- Replace the separate scheduler calls to `tag_blocked_pending_reasons()` and `pending_jobs()` with one priority-ordered classification that returns schedulable jobs and their displayed block reasons.
- Select burst-buffer stage-in candidates from that same classification so blocked or array-throttled jobs cannot reserve capacity before eligibility is known.
- Reserve array concurrency slots within each scheduling pass, preventing a fresh `%N` array from admitting multiple siblings before the first reaches `Running`.

## Approach

`classify_pending_jobs()` evaluates each candidate once using the existing precedence:

partition → dependency → reservation → priority ordering → account/QoS pass reservations → license → burst buffer

The scheduler applies collected reason updates after releasing the read-side job snapshot, stages only classifier-selected BB jobs, and dispatches the returned jobs. `pending_jobs()` remains a read-only wrapper for callers that only need the filtered list.

Absolute license and BB shortages retain their pending reasons. Greedy same-pass contention remains untagged, and scheduling holds, deadlines, and active `BeginTime` holds are protected from reason overwrites.

## Edge cases

Regression coverage includes:

- dependency reason precedence over a missing reservation;
- same-pass account, QoS, license, BB, and array concurrency reservations;
- array-throttled tasks not consuming scarce license or BB capacity;
- BB stage-in transitioning only classifier-approved jobs;
- protected held, deadline, and active `BeginTime` reasons.

## Test plan

- [x] `cargo test --locked`
- [x] `cargo clippy --workspace --exclude spur-ffi --all-targets --locked -- -D warnings`
- [x] Three-node Ubuntu 24.04 LXD cluster
- [x] LXD array `%1` + 160 GB BB pool: one 60 GB array task running, sibling throttled, independent 60 GB job running
- [x] LXD absolute shortage, dependency precedence, held-job, and greedy license contention cases

Fixes #349

## Prior review

This supersedes PR #495 and incorporates all requested review changes from that attempt:

- make scheduling eligibility explicit after the license and BB gates instead of relying on opposite gate polarity or ordering;
- remove the unreachable `BbStageState::Staging` match arm;
- preserve the leader-only contract documentation for pending-reason updates;
- add array-throttle regression coverage for scheduling exclusion and scarce license/BB pools;
- extend the BB case end-to-end so stage-in itself only reserves capacity for classifier-approved candidates.

### DIFF
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -252,6 +252,18 @@ pub struct ClusterManager {
     sched_stats: OnceLock<Arc<SchedStatsCollector>>,
 }
 
+struct PendingJobClassification {
+    jobs: Vec<Job>,
+    blocked: Vec<(JobId, PendingReason)>,
+    bb_stage_candidates: Vec<JobId>,
+}
+
+struct PendingJobCandidate {
+    job: Job,
+    scheduling_eligible: bool,
+    tag_reason: bool,
+}
+
 impl ClusterManager {
     pub fn new(config: SlurmConfig, _state_dir: &Path) -> anyhow::Result<Self> {
         let partitions = config.build_partitions();
@@ -1881,24 +1893,59 @@ impl ClusterManager {
             .collect()
     }
 
-    /// Get pending jobs sorted by priority, filtering out held and dependency-blocked jobs.
+    /// Get pending jobs sorted by priority, filtering out blocked jobs.
     /// Recomputes effective priority using QoS, age, and partition tier before sorting.
     pub fn pending_jobs(&self) -> Vec<Job> {
-        let jobs = self.jobs.read();
-        let mut pending: Vec<Job> = jobs
-            .values()
-            .filter(|j| j.state == JobState::Pending && !j.pending_reason.is_scheduling_hold())
-            .cloned()
-            .collect();
+        self.classify_pending_jobs().jobs
+    }
 
-        // Drop jobs blocked by partition state/config (shares partition_block()
-        // with tag_blocked_pending_reasons() so drop and shown reason agree).
+    /// Classify pending jobs once, apply displayed block reasons, and return the
+    /// jobs eligible for scheduling.
+    pub fn pending_jobs_and_tag_reasons(&self) -> Vec<Job> {
+        let classification = self.classify_pending_jobs();
+        self.apply_blocked_pending_reasons(classification.blocked);
+        self.advance_bb_staging_for(&classification.bb_stage_candidates);
+        classification.jobs
+    }
+
+    fn classify_pending_jobs(&self) -> PendingJobClassification {
+        let jobs = self.jobs.read();
+        let now = Utc::now();
+        let running_array_counts: HashMap<JobId, u32> = jobs
+            .values()
+            .filter(|job| job.state == JobState::Running)
+            .filter_map(|job| job.spec.array_job_id)
+            .fold(HashMap::new(), |mut counts, array_id| {
+                *counts.entry(array_id).or_insert(0) += 1;
+                counts
+            });
+        let mut candidates: Vec<PendingJobCandidate> = jobs
+            .values()
+            .filter(|job| job.state == JobState::Pending)
+            .filter(|job| !job.pending_reason.is_scheduling_hold())
+            .filter_map(|job| {
+                let active_begin_hold = job.pending_reason == PendingReason::BeginTime
+                    && job.spec.begin_time.is_some_and(|begin| now < begin);
+                if active_begin_hold {
+                    return None;
+                }
+                let before_begin_time = job.spec.begin_time.is_some_and(|begin| now < begin);
+                Some(PendingJobCandidate {
+                    job: job.clone(),
+                    scheduling_eligible: !before_begin_time,
+                    tag_reason: job.pending_reason != PendingReason::DeadLine,
+                })
+            })
+            .collect();
+        let mut blocked = Vec::new();
+
         {
             let partitions = self.partitions.read();
-            pending.retain(|job| partition_block(job, &partitions).is_none());
+            retain_unblocked(&mut candidates, &mut blocked, |job| {
+                partition_block(job, &partitions)
+            });
         }
 
-        // Check dependencies
         let get_job = |id: JobId| -> Option<Job> { jobs.get(&id).cloned() };
         let get_array_tasks = |id: JobId| -> Vec<Job> {
             jobs.values()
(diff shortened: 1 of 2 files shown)


crates/spurctld/src/cluster.rs
```diff
@@ -1998,51 +2002,74 @@ impl ClusterManager {
...
-        // An absolute shortage is also reported as `Licenses` by
-        // tag_blocked_pending_reasons() via license_block().
         {
-            let mut remaining = self.available_licenses_with(&jobs);
-            pending.retain(|job| {
+            let available = self.available_licenses_with(&jobs);
+            let mut remaining = available.clone();
+            candidates.retain(|candidate| {
+                let job = &candidate.job;
+                if let Some(reason) = license_block(job, &available) {
```
yansun1996, line 2069: The account/QoS gate above short-circuits `if !candidate.scheduling_eligible { return true; }` before evaluating any block, but the license gate here (and the burst-buffer gate below) run `*_block` + `record_blocked` first and only check eligibility afterward. So a job with `scheduling_eligible == false` — a future-`begin_time` job whose reason isn't yet `BeginTime` (e.g. a freshly submitted `--begin`, reason `Priority`), or an array-`%N`-throttled sibling — that requests more licenses/BB than are currently free gets tagged `Licenses`/`BurstBufferResources`. That reason is then written by `apply_blocked_pending_reasons` and isn't corrected afterward, since `update_pending_reasons` only runs over jobs still in the scheduling set. Before this change those jobs were filtered out ahead of the consumable gates, so they never displayed a consumable reason.

Would it be worth hoisting the `!candidate.scheduling_eligible` check to the top of all three gates (matching the account/QoS one) so an ineligible job is never tagged by a consumable gate?

Also, the new tests all exercise the array-`%1` path — a small regression test for the `begin_time` case (future-begin job + an absolute license/BB shortage, asserting its `pending_reason` is *not* `Licenses`/`BurstBufferResources`) would lock this down, since that path is currently uncovered.

yansun1996 (commented):
Nice cleanup — folding the drop decision and the displayed reason into one `classify_pending_jobs()` pass removes the must-match ordering contract that #349 flagged, and the prior review points (dedicated eligibility retain, dropped `Staging` arm, leader-only note) all look addressed.

One thing worth a look before merge: the three consumable gates don't treat `scheduling_eligible` the same way, which lets an ineligible job show a consumable reason it otherwise couldn't. Inline detail below.

yansun1996 (changes requested):

crates/spurctld/src/cluster.rs
```diff
@@ -2357,86 +2385,16 @@ impl ClusterManager {
...
 
-    /// Set `pending_reason` for jobs `pending_jobs()` drops from scheduling
-    /// (dependency/QoS/license/reservation), which never reach
-    /// `update_pending_reasons()` and would otherwise show a stale reason.
-    /// Leader-only; mirrors `cancel_unsatisfiable_dependency_jobs()`.
+    #[cfg(test)]
+    /// Reclassify pending jobs and apply their displayed block reasons.
+    /// Leader-only; enforced by the scheduler-loop caller, not this function
+    /// itself (mirrors `cancel_unsatisfiable_dependency_jobs()`).
     pub fn tag_blocked_pending_reasons(&self) {
```
shiv-tyagi: Nit: `tag_blocked_pending_reasons` is test-only now, so `pub` is misleading (the test module is in-crate). The doc also reads better before the attribute.
```suggestion
    /// Reclassify pending jobs and apply their displayed block reasons.
    /// Leader-only; enforced by the scheduler-loop caller, not this function
    /// itself (mirrors `cancel_unsatisfiable_dependency_jobs()`).
    #[cfg(test)]
    fn tag_blocked_pending_reasons(&self) {
```

crates/spurctld/src/cluster.rs
```diff
@@ -1881,24 +1893,59 @@ impl ClusterManager {
...
-    /// Get pending jobs sorted by priority, filtering out held and dependency-blocked jobs.
+    /// Get pending jobs sorted by priority, filtering out blocked jobs.
     /// Recomputes effective priority using QoS, age, and partition tier before sorting.
     pub fn pending_jobs(&self) -> Vec<Job> {
+        self.classify_pending_jobs().jobs
+    }
+
+    /// Classify pending jobs once, apply displayed block reasons, and return the
+    /// jobs eligible for scheduling.
+    pub fn pending_jobs_and_tag_reasons(&self) -> Vec<Job> {
```
shiv-tyagi, line 1904: `pending_jobs_and_tag_reasons` also advances BB staging (`advance_bb_staging_for`), moving stage candidates `None -> Staging`. The name and doc only mention tagging reasons. Document the BB staging side effect so it is discoverable at the call site.

crates/spurctld/src/cluster.rs
```diff
@@ -1881,24 +1893,59 @@ impl ClusterManager {
...
+                if active_begin_hold {
+                    return None;
+                }
+                let before_begin_time = job.spec.begin_time.is_some_and(|begin| now < begin);
+                Some(PendingJobCandidate {
+                    job: job.clone(),
+                    scheduling_eligible: !before_begin_time,
+                    tag_reason: job.pending_reason != PendingReason::DeadLine,
+                })
+            })
```
shiv-tyagi, line 1936: Nit: `before_begin_time` recomputes the `begin_time` check already done inside `active_begin_hold`. Compute it once.
```suggestion
            .filter_map(|job| {
                let before_begin_time = job.spec.begin_time.is_some_and(|begin| now < begin);
                if before_begin_time && job.pending_reason == PendingReason::BeginTime {
                    return None;
                }
                Some(PendingJobCandidate {
                    job: job.clone(),
                    scheduling_eligible: !before_begin_time,
                    tag_reason: job.pending_reason != PendingReason::DeadLine,
                })
            })
```

crates/spurctld/src/cluster.rs
```diff
@@ -1881,24 +1893,59 @@ impl ClusterManager {
...
             .collect();
+        let mut blocked = Vec::new();
 
-        // Drop jobs blocked by partition state/config (shares partition_block()
-        // with tag_blocked_pending_reasons() so drop and shown reason agree).
         {
             let partitions = self.partitions.read();
-            pending.retain(|job| partition_block(job, &partitions).is_none());
+            retain_unblocked(&mut candidates, &mut blocked, |job| {
+                partition_block(job, &partitions)
```
shiv-tyagi, line 1945: Q: this partition pass (and the dependency/reservation passes below) run before the eligibility gate, so a not-yet-eligible future `begin_time` job (`scheduling_eligible == false`) can still be tagged Partition/Dependency/Reservation. License and BB tagging are gated by `scheduling_eligible` and covered by `future_begin_jobs_do_not_receive_consumable_block_reasons`, but this asymmetry is not. Intended? If so, a one-line note would help.

shiv-tyagi (changes requested):
Solid refactor unifying pending classification, tagging, and BB stage selection into one pass. Logic and the same-pass vs absolute-shortage tagging distinction are preserved, and array-throttle correctly skips consumable reservations. A few minor items below, mostly nits plus one behavior question.

maybeharshit:
Addressed the latest review feedback in `1ac4f8f`:

- documented that `pending_jobs_and_tag_reasons()` also advances classifier-selected burst-buffer stage-in;
- compute future `begin_time` eligibility once;
- documented why structural partition/dependency/reservation blockers retain precedence before begin-time eligibility, while consumable gates wait until the job can reserve capacity;
- moved the test-only helper's documentation before `#[cfg(test)]` and made the helper private.

Validation:
- all 446 `spurctld` tests
- isolated Ubuntu 24.04 LXD controller + two-node run covering array `%1` with a 160 GB BB pool, absolute license/BB shortages, dependency precedence, held jobs, future-begin jobs, and same-pass license contention

The full workspace test run only failed in six environment-dependent `spurd` cgroup/PTY tests (`/dev/ptmx` and cgroup permissions) in the local sandbox; the relevant controller suite and LXD cases passed.

shiv-tyagi (approved):
All prior review points addressed. Smoke-tested the build on a 4-node bare-metal cluster: basic scheduling, resource contention, array %1/%2 throttling, dependency ordering, future begin_time eligibility, license contention, array-throttle license gating, and cancellation all behave correctly. The within-pass array slot reservation (fix for the #349 case) and the consumable-reason gating for not-yet-eligible jobs both verified end-to-end. No regressions found.

shiv-tyagi:
@yansun1996 Please approve and merge.

yansun1996 (approved):

PR #504, opened 2026-07-25
docs: add user/admin/deployment guides with rocm-docs-core build + CI

yansun1996:
## Summary

Fills major documentation gaps with a Slurm-style user and admin guide, and makes the docs build on the ROCm docs toolchain (`rocm-docs-core`) with a hermetic, containerized build + lint that runs in CI.

**New documentation** (Slurm-familiar structure, since Spur users are largely Slurm users):

- **User guide** — architecture, submitting jobs, interactive/parallel jobs, monitoring & controlling jobs, containers, Slurm compatibility.
- **Admin guide** — a full `spur.conf` configuration reference (every section/field with type + default, including `[mpi]`) and an accounting guide (accounts, users, QOS, associations, TRES, auth).
- **Deployment** — the recommended Ansible toolkit path, the manual systemd path, upgrading, and uninstalling.
- **Migration from Slurm** — what works unchanged, config differences, and current behavioral limitations.

All content is source-derived (CLI crates, `config.rs`, `install.sh`, the spur-toolkit Ansible roles).

**Docs build + CI (`rocm-docs-core`):**

- `docs/conf.py` uses the `rocm_docs` extension and theme; version is tracked from the workspace `Cargo.toml`.
- Navigation uses an external table of contents (`docs/sphinx/_toc.yml.in`) instead of per-page `toctree` directives, per ROCm convention.
- Pinned toolchain (`docs/sphinx/requirements.{in,txt}`) and `.readthedocs.yaml` for hosting.
- `docs/Dockerfile` builds and lints the docs in one hermetic image; `.github/workflows/docs.yml` runs it on docs changes: `sphinx-build -W --keep-going` (warnings are errors) + `markdownlint-cli2` + `pyspelling`.
- Lint config: `.markdownlint.yaml`, `.spellcheck.yaml`, `.wordlist.txt`.

The hermetic build has no ROCm external-projects registry to resolve against, so `rocm-docs-core` emits a single benign "project not found" log line. A small `conf.py` logging filter drops just that message so `-W` stays meaningful without suppressing real warnings.

## Test plan

- [x] Built the docs in the container with warnings-as-errors: `sphinx-build -W --keep-going -b html docs docs/_build/html` — exit 0, all pages render.
- [x] `pyspelling -c .spellcheck.yaml` — passes (build artifacts under `docs/_build/` excluded from the spell globs).
- [x] `markdownlint-cli2 "docs/**/*.md"` — no Markdown sources yet, no errors.
- [ ] Confirm the new **Docs** GitHub Actions workflow is green on this PR.
- [ ] (Optional) Enable the Read the Docs project so `.readthedocs.yaml` publishes a preview.


sajmera-pensando:
A few things to address before merging:

**Must fix**

1. **`docs.yml`: missing `-W` on `sphinx-build`**
   The CI step runs `sphinx-build -b html docs docs/_build/html` without `-W`, so Sphinx warnings silently pass. The PR description and test plan both claim warnings-as-errors. Either add `-W --keep-going` to the CI step and the `Dockerfile` `CMD`, or drop that claim from the description.

2. **`deployment/uninstalling.rst`: `spurdbd` reference is wrong**
   The teardown description says it stops "the legacy `spurdbd` unit if present." `spurdbd` is a Slurm name — Spur has only ever shipped two daemons: `spurctld` and `spurd`. This should be removed. The `spurrestd` entry in `.wordlist.txt` has the same problem.

3. **`admin-guide/configuration.rst`: `heartbeat_timeout_secs` default is wrong**
   The reference documents the default as `90`, but in `config.rs` the field is `Option<u64>` and the `Default` impl returns `None`. The 90-second value is a runtime fallback applied when the field is absent, not the TOML default. Should be documented as `none` (unset; runtime applies a 90 s fallback).

**Nice to fix**

4. **`admin-guide/configuration.rst`: `[mpi]` section is missing**
   `config.rs` has a `MpiConfig` struct (used for `--mpi=pmix` job steps) with no corresponding section in the configuration reference. The doc says it aims to cover every field.

5. **`docs/Dockerfile`: `pyspelling` pinned inline rather than in `requirements.txt`**
   `pyspelling==2.12.1` is installed in a separate `RUN pip install` line instead of being listed in `docs/sphinx/requirements.txt` alongside the rest of the toolchain. Splitting the pins makes version management inconsistent.

docs/Dockerfile.dockerignore
```diff
@@ -0,0 +1,14 @@
...
+
+# Build context for docs/Dockerfile. The root .dockerignore excludes docs/ (it
+# targets the Rust application image); the docs image needs the opposite, so
+# this file — which Docker prefers for docs/Dockerfile — starts from nothing and
+# allows only what the doc build touches.
+*
+!docs
+!.wordlist.txt
+!.markdownlint.yaml
+!.spellcheck.yaml
```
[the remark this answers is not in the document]
yansun1996: `.dockerignore` doesn't follow gitignore's re-include rule — un-ignoring a directory (`!docs`) does re-include its contents in the build context. The only context `COPY` is `docs/sphinx/requirements.txt`, which is available, and `sphinx-build` reads `docs/` at runtime via the volume mount rather than from the build context. Verified the image builds and runs, so leaving this as-is.

yansun1996:
Thanks — all addressed:

1. **`docs.yml` `-W`** — added `-W --keep-going` to both the CI step and the Dockerfile `CMD` (and the example snippets). The hermetic build has no ROCm external-projects registry, so `rocm-docs-core` emits one benign "project not found" line; a small `conf.py` logging filter drops just that message so `-W` stays meaningful. Build now passes clean with warnings-as-errors.
2. **`spurdbd`/`spurrestd`** — removed the `spurdbd` clause in `uninstalling.rst` and both entries from `.wordlist.txt`.
3. **`heartbeat_timeout_secs`** — documented as `none` (unset), with a note that the controller applies a 90 s fallback when absent.
4. **`[mpi]` section** — added, covering `plugin_dir`, `pmix_plugin`, `pmix_tmpdir`, and `pmix_min_version`.
5. **`pyspelling`** — moved into `docs/sphinx/requirements.in` and recompiled the lockfile; dropped the inline pin from the Dockerfile.

Also excluded `docs/_build/**` from the spellcheck globs (it was scanning generated artifacts) and word-listed a few legitimate technical terms.

sajmera-pensando (dismissed):
LGTM

.markdownlint.yaml
```diff
@@ -0,0 +1,11 @@
...
+MD013: false
+MD024:
+  siblings_only: true
+MD026:
+  punctuation: ".,;:!"
+MD029:
+  style: ordered
+MD033: false
+MD034: false
+MD041: false
```
shiv-tyagi: Keeping the markdownlint chain is fine (good to catch future Markdown), but don't pre-loosen it. `MD033` (inline HTML), `MD034` (bare URLs), and `MD041` (first-line heading) aren't disabled in ROCm's canonical config (`rocm-docs-core` and `gpu-operator` both stop at `MD029`). Disabling them before any `.md` file lands means the first Markdown that comes in silently skips these checks.

```suggestion
MD029:
  style: ordered
```

Loosen individual rules later only when a real file needs it.

docs/admin-guide/configuration.rst
```diff
@@ -0,0 +1,928 @@
...
+
+``spur.conf`` is a TOML file describing controller, node, accounting, scheduling,
+and network settings. The default location is ``/etc/spur/spur.conf`` (the Ansible
+layout installs it at ``<spur_home>/etc/spur.conf``). Only ``cluster_name`` is
+required; every section has a default and may be omitted, and unknown keys are
+silently ignored. The controller validates the file on load.
+
+The sections below are grouped by subsystem. Every field lists its type, default,
+and meaning. The authoritative source of truth is ``crates/spur-core/src/config.rs`` —
+keep this reference in sync with it.
```
shiv-tyagi: This points readers at an internal source path (`crates/spur-core/src/config.rs`) and carries a maintainer instruction ("keep this reference in sync with it"). Neither is actionable for an admin reading the config reference. Move the sync note to `CONTRIBUTING` or a code comment and drop the source path here.

docs/admin-guide/configuration.rst
```diff
@@ -0,0 +1,928 @@
...
+----------------------------------------------------
+
+``[federation]``
+   Peer clusters for cross-cluster job routing. Each ``[[federation.clusters]]``
+   entry has ``name`` (string) and ``address`` (string, e.g.
+   ``"http://peer-ctrl:6817"``). Defaults to no peers.
+
+``[topology]``
+   Optional switch-hierarchy configuration for locality-aware scheduling. Its fields
+   are defined in the topology module rather than the core config schema.
```
shiv-tyagi: "Its fields are defined in the topology module rather than the core config schema" describes the code layout, not the config. An admin can't act on "topology module". Document the actual fields, or state the section is reserved, instead of pointing at where they live in the source.

docs/deployment/native-host.rst
```diff
@@ -1,32 +1,127 @@
...
+     - ``[::]:6818``
+     - Agent gRPC listen address.
+   * - ``--log-level <LEVEL>``
+     - ``info``
+     - Log verbosity.
+
+.. note::
+
+   ``spurd`` is configured entirely by CLI flags. It logs ``failed to load spur.conf``
+   at startup — this is harmless; the agent never truly reads a config file.
```
shiv-tyagi: This documents around a product wart rather than fixing it: `spurd` advertises `-f, --config` (see the flag table above) but ignores it and logs `failed to load spur.conf` on every boot. Teaching users to ignore a scary error line is worse than removing the cause. Q: can `spurd` drop the flag (or stop logging an error for expected behavior) so this note isn't needed?

shiv-tyagi (commented):
Posted reviews, PTAL

.markdownlint.yaml
yansun1996: Good call — dropped `MD033`/`MD034`/`MD041` so we don't pre-loosen before any Markdown lands. Kept `MD029: style: ordered`.

docs/admin-guide/configuration.rst
yansun1996: Agreed, that's not admin-facing. Removed the source path and the sync instruction.

docs/admin-guide/configuration.rst
yansun1996: Documented the actual `[topology]` fields (`plugin`, `switches[]` with `name`/`nodes`/`switches`, and `block_size`) instead of pointing at the source.

docs/deployment/native-host.rst
yansun1996: You're right — the note was wrong: `spurd` does read `spur.conf` for local agent settings (`[hooks]`, `[devices]`, `rlimits.memlock`, `[cluster]`, `[mpi]`). Corrected the note and the flag description here and in the config reference. The `-f/--config` flag is load-bearing so I've kept it; the leftover wart is that a missing default file logs at `warn!`, which I've filed as a follow-up: https://github.com/ROCm/spur/issues/586.

shiv-tyagi (approved):
LGTM

PR #505, opened 2026-07-26
feat(mpi): add single-node PMIx MPI via dlopen plugin

sgopinath1:
## Summary

Adds single-node PMIx MPI for Open MPI jobs, Slurm-style via `--mpi=none|pmix|list` on `srun` and `sbatch`.

- **`spur_mpi_pmix.so`** — cdylib loaded by `spurd` at job/step launch (OpenPMIx stays out of `spurd`, controller, and CLI)
- **`[mpi]` config** — `plugin_dir`, `pmix_tmpdir`, `pmix_min_version`
- **Launch path** — controller builds `PmixLaunchPlan`, agents start/stop PMIx with rollback via `PmixLaunchGuard`
- **Step inherit** — inside an allocation, `srun` omits `--mpi` to inherit the job setting; explicit `--mpi=none` overrides
- **Scope guard** — multi-node `--mpi=pmix` rejected at submit; multi-agent PMIx steps rejected at dispatch

## Approach

Plugin model matches Slurm: `spurd` dlopens `spur_mpi_pmix.so`, which calls OpenPMIx server APIs. Batch launch and step dispatch share `PmixLaunchPlan` over gRPC. Single-node multi-task jobs use one `PMIX_SERVER_URI` in the parent env; the multi-task wrapper sets per-rank `PMIX_RANK` via `SPUR_PROCID`.

## Test plan

- [x] `cargo test --locked` (unit tests in `spur-core::mpi`, `spurd::mpi_plugin`)

## Follow-ups (PR2)

- Multi-node PMIx (cross-agent fence, full proc map)
- Step inherit e2e (`#SBATCH --mpi=pmix` + inner `srun` without `--mpi`)
- Per-job PMIx tmpdir subdirs; raise 64-namespace plugin cap if needed


crates/spurd/src/agent_server.rs
```diff
@@ -242,8 +250,10 @@ impl AgentService {
                         crate::executor::cleanup_cgroup(cgroup);
                     }
                     allocation.lock().await.release_job(c.job_id);
-                    if let Some(pmi) = pmi_servers.lock().await.remove(&c.job_id) {
-                        pmi.cleanup();
+                    if c.mpi == MPI_PMIX || mpi_host.has_active_pmix(c.job_id) {
```
yansun1996: This branch replaced the previous `pmi_servers.remove(&c.job_id).cleanup()` with the PMIx stop. Non-pmix multi-task jobs still create a `PmiServer` at launch, but now nothing cleans it on normal completion (`drop_tracked_job` only stops PMIx, and `cleanup_pmi_server` runs only on the abort/reclaim path), so this looks like a leak of the map entry and its socket. Could we keep both here — remove/cleanup the PMI server and conditionally stop PMIx?

crates/spurd/src/agent_server.rs
```diff
@@ -1528,9 +1587,46 @@ impl SlurmAgent for AgentService {
...
+            if let Err(e) = spur_core::hooks::run_hook(task_prolog, &ctx).await {
+                return Err(Status::aborted(format!("TaskProlog failed: {}", e)));
+            }
+        }
+
+        let mut env = senv.into_map();
+        let mut _pmix_step_guard = None;
+        if let Some(proto_plan) = req.pmix_plan.as_ref() {
+            let plan = mpi_plugin::plan_from_proto(proto_plan);
+            if !self.mpi_host.has_active_pmix(plan.job_id) {
```
yansun1996: `has_active_pmix` and `PmixLaunchGuard::start` aren't atomic, so two concurrent `--mpi=pmix` steps in the same allocation can both observe `false`, both start a server, and whichever finishes first deregisters the namespace out from under the other. Batch jobs are safe (single owned guard). Would holding `active_namespaces` across the check-and-start, or refcounting per step, close this?

crates/spurd/src/mpi_plugin.rs
```diff
@@ -0,0 +1,484 @@
...
+            plugin: Mutex::new(None),
+            active_namespaces: Mutex::new(HashMap::new()),
+        }
+    }
+
+    pub fn plugin_path(&self) -> PathBuf {
+        self.config.resolve_pmix_plugin_path()
+    }
+
+    pub fn has_active_pmix(&self, job_id: u32) -> bool {
```
yansun1996, line 138: `.lock().ok().is_some_and(...)` treats a poisoned lock as "no active namespace", so a completion path would silently skip the stop. Might be worth surfacing the poison and treating it conservatively as active, so a stop is still attempted.

proto/slurm.proto
```diff
@@ -662,9 +676,8 @@ message LaunchJobRequest {
   // array_job_id is the parent id (0 = not an array task); exported as SLURM_ARRAY_*.
   uint32 array_job_id = 7;
   uint32 array_task_id = 8;
-  // Run epoch for this dispatch; agent echoes it in ReportJobStatusRequest so
-  // the controller can drop stale-run reports. 0 = unset (legacy), check off.
   uint32 run_attempt = 9;
```
yansun1996, line 703: Looks like the `run_attempt` doc comment (run-epoch / stale-report semantics) was removed when `pmix_plan = 10` was added, though the field itself is unchanged — probably worth restoring.

crates/spurd/src/mpi_plugin.rs
```diff
@@ -0,0 +1,484 @@
...
+    for (idx, proc) in plan.local_procs.iter().take(256).enumerate() {
+        c_plan.local_procs[idx] = SpurMpiProc {
+            rank: proc.rank,
+            local_rank: proc.local_rank,
+        };
+    }
+    c_plan
+}
+
+fn write_c_str(dest: &mut [c_char], value: &str) {
```
yansun1996: `write_c_str` indexes `dest[copy_len]`, which would panic on a zero-length `dest` (`saturating_sub(1)` → 0). Unreachable with the current fixed-size callers, but an early `if dest.is_empty() { return }` would make it safe to reuse.

yansun1996 (commented):
Nice change — the plugin/dlopen isolation and the `PmixLaunchGuard` rollback design read really cleanly. A few things left inline:

- **(blocking)** The reconcile completion branch stops PMIx but no longer cleans the PMI-1 `PmiServer`, which looks like a leak for normally-completing non-pmix jobs.
- A concurrent-step namespace race, a poisoned-lock swallow in `has_active_pmix`, an accidentally dropped `run_attempt` proto comment, and a small `write_c_str` nit.

Worth adding a unit test for the reconcile-completion cleanup path to cover the first item. The existing Copilot threads look already addressed by the latest commit and can be resolved.

crates/spur-mpi-pmix/c/pmix_server.c
```diff
@@ -0,0 +1,344 @@
...
+
+    pmix_proc_t proc;
+    PMIX_PROC_CREATE(&proc, 1);
+    PMIX_LOAD_PROCID(&proc, plan->namespace_, rank);
+
+    pmix_info_t *info = NULL;
+    size_t ninfo = 0;
+    pmix_status_t rc = PMIx_server_setup_fork(&proc, &info, &ninfo);
+    PMIX_PROC_RELEASE(&proc);
+    if (rc != PMIX_SUCCESS) {
```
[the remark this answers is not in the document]
sgopinath1: Fixed. Uses PMIX_LOAD_PROCID only; no heap macros on stack proc.

crates/spur-mpi-pmix/c/pmix_server.c
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    if (find_session(plan->namespace_) != NULL) {
+        spur_mpi_debug(
+            "spur_mpi_pmix: namespace %s already registered\n",
+            plan->namespace_
+        );
+        pthread_mutex_unlock(&g_session_lock);
+        return 0;
+    }
```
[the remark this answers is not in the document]
sgopinath1: Fixed. Rejects mismatched universe_size and local proc layout (local_procs_match); same layout returns success.

docs/developer/building.rst
```diff
@@ -18,6 +18,27 @@ Build
 
 All crates build in one invocation — no separate steps needed.
 
+MPI plugin (optional)
+~~~~~~~~~~~~~~~~~~~~~
+
+Single-node Open MPI jobs use ``--mpi=pmix``, which loads ``spur_mpi_pmix.so`` on
+compute nodes at runtime (``spurd`` itself does not link libpmix). Build the
+plugin when libpmix development packages are installed (``pkg-config libpmix`` must
+succeed):
```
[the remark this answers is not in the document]
sgopinath1: Fixed. Updated to pkg-config pmix.

docs/developer/building.rst
```diff
@@ -18,6 +18,27 @@ Build
...
+
+.. code-block:: bash
+
+   cargo build --release -p spur-mpi-pmix
+   sudo install -D target/release/libspur_mpi_pmix.so /usr/lib/spur/spur_mpi_pmix.so
+
+Without libpmix, the crate still builds a stub plugin that fails at load time with
+an actionable error. Container images can include a functional plugin by building
+with ``BUILD_MPI_PLUGIN=1`` (see ``deploy/Dockerfile``).
```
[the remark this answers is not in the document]
sgopinath1: Fixed. Now references top-level Dockerfile.

docs/deployment/native-host.rst
```diff
@@ -171,6 +171,50 @@ The default can be changed in ``spur.conf``:
...
+
+Build and install the plugin on every agent node:
+
+.. code-block:: bash
+
+   cargo build --release -p spur-mpi-pmix
+   sudo install -D target/release/libspur_mpi_pmix.so /usr/lib/spur/spur_mpi_pmix.so
+
+The plugin requires **libpmix** at runtime (``pkg-config libpmix`` when building).
+Open MPI on the node must be built against a compatible PMIx version.
```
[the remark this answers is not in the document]
sgopinath1: Fixed. Updated to pkg-config pmix.

crates/spurd/src/agent_server.rs
```diff
@@ -1305,9 +1294,26 @@ impl SlurmAgent for AgentService {
...
+                PmixLaunchGuard::start(self.mpi_host.clone(), &plan).map_err(
+                    Status::failed_precondition,
+                )?;
+            // Same single-uri bootstrap as batch launch; per-rank PMIX_RANK comes from
+            // the multi-task wrapper when num_tasks > 1.
+            let pmix_env = self
+                .mpi_host
+                .pmix_env_for_rank(&plan, req.task_offset)
+                .map_err(Status::failed_precondition)?;
+            pmix_guard.disarm();
```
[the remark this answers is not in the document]
sgopinath1: Fixed. Guard kept through cmd.output() when this step starts PMIx; skips start if already active; drop_tracked_job stops PMIx on allocation cancel.

crates/spur-mpi-pmix/c/pmix_server.c
```diff
@@ -0,0 +1,344 @@
...
+        plan->namespace_,
+        plan->universe_size,
+        procs,
+        NULL,
+        0
+    );
+    for (uint32_t i = 0; i < plan->num_local_procs; i++) {
+        PMIX_PROC_RELEASE(&procs[i]);
+    }
+    free(procs);
```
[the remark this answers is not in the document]
sgopinath1: Fixed. Loop uses PMIX_LOAD_PROCID only; free(procs) after register.

crates/spurd/src/agent_server.rs
sgopinath1: Good catch — that was a regression. Restored PMI cleanup via cleanup_completed_job_mpi(), which removes/cleans up the PmiServer and releases the PMIx ref on batch completion. Added completion_cleanup_removes_pmi_server.

crates/spurd/src/agent_server.rs
sgopinath1: Fixed by reserving the namespace under active_namespaces lock before calling into the plugin, with per-job refcounting (ActiveNamespace { refs }). Concurrent steps join the existing namespace instead of racing a second start. run_command always goes through PmixLaunchGuard::start.

crates/spurd/src/mpi_plugin.rs
sgopinath1: Agreed. Now returns true on a poisoned lock so we treat the namespace as still active rather than skipping teardown.

proto/slurm.proto
sgopinath1: Restored the run-epoch comment from main; only the comment was missing, no wire change.

crates/spurd/src/mpi_plugin.rs
sgopinath1: Restored the run-epoch comment from main; only the comment was missing, no wire change.

sgopinath1:
> Nice change — the plugin/dlopen isolation and the `PmixLaunchGuard` rollback design read really cleanly. A few things left inline:
> 
> * **(blocking)** The reconcile completion branch stops PMIx but no longer cleans the PMI-1 `PmiServer`, which looks like a leak for normally-completing non-pmix jobs.
> * A concurrent-step namespace race, a poisoned-lock swallow in `has_active_pmix`, an accidentally dropped `run_attempt` proto comment, and a small `write_c_str` nit.
> 
> Worth adding a unit test for the reconcile-completion cleanup path to cover the first item. The existing Copilot threads look already addressed by the latest commit and can be resolved.

Thanks for reviewing @yansun1996. Addressed all your comments.

yansun1996 (approved):

crates/spur-mpi-pmix/build.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+fn main() {
+    println!("cargo:rerun-if-changed=c/pmix_server.c");
+    println!("cargo:rerun-if-changed=c/stub_server.c");
+    println!("cargo:rerun-if-changed=include/spur_mpi_plugin.h");
+
+    let include = std::path::Path::new("include");
+    let mut build = cc::Build::new();
```
shiv-tyagi, line 10: Blocking: the built cdylib exports none of the `spur_mpi_pmix_*` symbols, so `spurd`'s `dlsym` fails at launch. Verified on a bare-metal node: `MPI plugin missing spur_mpi_pmix_version: dlsym failed`, and `nm -D libspur_mpi_pmix.so` lists zero `spur_mpi_pmix_*` symbols. The C objects compile into a static archive that no Rust code references, so the linker drops them from the `.so`.

Force the C entry points into the dynamic symbol table: link the archive with `+whole-archive`, reference the functions from Rust so they are retained, or drop the C and bind libpmix directly from this crate.

crates/spur-mpi-pmix/build.rs
```diff
@@ -0,0 +1,21 @@
...
+    build.include(include);
+
+    if pkg_config::Config::new().probe("pmix").is_ok() {
+        build.file("c/pmix_server.c");
+        build.compile("spur_mpi_pmix_server");
+        println!("cargo:rustc-cfg=spur_pmix_linked");
+    } else {
```
shiv-tyagi, line 17: Blocking on Debian/Ubuntu: the probe result is discarded, so `cc` never receives the pmix include dirs and `pmix.h` is not found (Ubuntu ships headers under `/usr/lib/x86_64-linux-gnu/pmix2/include`). Feed the probe's include paths into the build.

```suggestion
    if let Ok(lib) = pkg_config::Config::new().probe("pmix") {
        for path in &lib.include_paths {
            build.include(path);
        }
        build.file("c/pmix_server.c");
        build.compile("spur_mpi_pmix_server");
        println!("cargo:rustc-cfg=spur_pmix_linked");
    } else {
```

crates/spur-mpi-pmix/c/pmix_server.c
```diff
@@ -0,0 +1,385 @@
...
+static pmix_status_t spur_client_connected(
+    const pmix_proc_t *proc,
+    void *server_object,
+    pmix_op_cbfunc_t cbfunc,
+    void *cbdata
+) {
+    (void)proc;
+    (void)server_object;
+    if (cbfunc != NULL) {
+        cbfunc(PMIX_SUCCESS, NULL, 0, cbdata, NULL, NULL);
```
shiv-tyagi, line 61: Blocking: `pmix_op_cbfunc_t` is `(pmix_status_t, void *cbdata)`, not 6 args, so this does not compile against OpenPMIx 4.x/5.x. The 6-arg form is `pmix_modex_cbfunc_t`.

```suggestion
        cbfunc(PMIX_SUCCESS, cbdata);
```

crates/spur-mpi-pmix/c/pmix_server.c
```diff
@@ -0,0 +1,385 @@
...
+static pmix_status_t spur_client_finalized(
+    const pmix_proc_t *proc,
+    void *server_object,
+    pmix_op_cbfunc_t cbfunc,
+    void *cbdata
+) {
+    (void)proc;
+    (void)server_object;
+    if (cbfunc != NULL) {
+        cbfunc(PMIX_SUCCESS, NULL, 0, cbdata, NULL, NULL);
```
shiv-tyagi, line 75: Blocking: same `pmix_op_cbfunc_t` arity mismatch as in `spur_client_connected`.

```suggestion
        cbfunc(PMIX_SUCCESS, cbdata);
```

crates/spur-mpi-pmix/c/pmix_server.c
```diff
@@ -0,0 +1,385 @@
...
+    }
+    return PMIX_SUCCESS;
+}
+
+static pmix_server_module_t spur_module = {
+    .client_connected = spur_client_connected,
+    .client_finalized = spur_client_finalized,
+    .fence = spur_fence,
```
shiv-tyagi, line 108: Blocking: `pmix_server_module_t` (v4.0.0 ABI) has no `fence` member; it is `fence_nb`, and `spur_fence` already has the non-blocking `pmix_modex_cbfunc_t` signature.

```suggestion
    .fence_nb = spur_fence,
```

crates/spur-mpi-pmix/c/pmix_server.c
```diff
@@ -0,0 +1,385 @@
...
+        PMIX_LOAD_PROCID(&procs[i], plan->namespace_, plan->local_procs[i].rank);
+    }
+
+    pmix_status_t rc = PMIx_server_register_nspace(
+        plan->namespace_,
+        plan->universe_size,
+        procs,
+        NULL,
+        0
+    );
```
shiv-tyagi, line 236: Blocking: `PMIx_server_register_nspace(nspace, nlocalprocs, info[], ninfo, cbfunc, cbdata)` expects a `pmix_info_t*` array and an op callback. Here it gets a `pmix_proc_t*` as `info[]`, `NULL` as `ninfo`, and no callback. Local procs must be conveyed via info keys (e.g. `PMIX_PROC_DATA` / `PMIX_LOCAL_PEERS`), and the call is asynchronous so it needs a completion callback (or an explicit wait). Won't compile or function as written.

crates/spur-mpi-pmix/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! PMIx MPI plugin (`spur_mpi_pmix.so`) loaded at `--mpi=pmix` job launch.
```
shiv-tyagi, line 4: No Rust here references the C entry points, which is the root of the symbol drop noted on `build.rs`. If you keep the C, add a Rust shim that re-exports the symbols (or use `+whole-archive`). A pure-Rust FFI plugin (bindgen) would avoid this entirely: a cdylib exports `#[no_mangle] extern "C"` symbols by default, compiles no C (so no include-path issue), and generated bindings enforce the real PMIx API at compile time.

crates/spurd/src/mpi_plugin.rs
```diff
@@ -0,0 +1,747 @@
...
+        assert!(host.stop_pmix_server(6).is_err());
+        assert!(host.has_active_pmix(6));
+        assert_eq!(
+            host.active_namespaces.lock().unwrap().get(&6).unwrap().refs,
+            2
+        );
+    }
+
+    #[test]
+    #[ignore = "requires SPUR_TEST_MPI_PLUGIN pointing at a built spur_mpi_pmix.so"]
```
shiv-tyagi, line 725: The only test that loads a real plugin is `#[ignore]` and needs `SPUR_TEST_MPI_PLUGIN`, and CI builds only the stub, so the build/link/API defects above are uncovered. Please add a test that builds this cdylib and asserts its entry points resolve, plus a CI lane that builds against libpmix and runs one real MPI job.

docs/deployment/native-host.rst
```diff
@@ -171,6 +171,50 @@ The default can be changed in ``spur.conf``:
    the ``spurd`` systemd unit is no longer required. The agent raises the limit
    itself while still privileged.
 
+MPI (PMIx)
+----------
+
+Spur supports single-node Open MPI jobs via ``--mpi=pmix``. The controller and
```
shiv-tyagi, line 177: Q: this states single-node `--mpi=pmix` works, but as built the plugin can't load (see the plugin comments). Accurate once the plugin is fixed; until then it overstates current support.

shiv-tyagi (changes requested):
Requesting changes. The control-plane surface is solid: `--mpi` parsing, `--mpi=list`, the single-node and step guards, and `SLURM_MPI_TYPE` inheritance all behave correctly, and the missing-plugin error path is clean and actionable. But the PMIx plugin cannot load or run as built. I deployed this branch to a 4-node bare-metal cluster and `--mpi=pmix` fails at launch even with the plugin installed.

Three independent blockers:

1. The cdylib exports none of its `spur_mpi_pmix_*` entry points, so `spurd`'s `dlsym` fails. Confirmed on-cluster: `MPI plugin missing spur_mpi_pmix_version: dlsym failed`. `nm -D libspur_mpi_pmix.so` lists zero `spur_mpi_pmix_*` symbols. The C objects compile into a static archive that nothing references, so the linker drops them from the `.so`.

2. The C server code does not match the OpenPMIx 4.x/5.x server API and does not compile against OpenPMIx 5.0.1: `op_cbfunc` is called with 6 args (it takes 2), `pmix_server_module_t` has no `.fence` member (it is `fence_nb`), and `PMIx_server_register_nspace` is called with the wrong argument types and arity.

3. `build.rs` probes `pkg-config pmix` but discards the include paths, so `pmix.h` is not found on Debian/Ubuntu (headers live under `/usr/lib/x86_64-linux-gnu/pmix2/include`). It only builds where headers sit in the default path (RHEL `openpmix-devel`, which the Dockerfile uses).

Together these mean the functional plugin has almost certainly never compiled against a recent libpmix. CI builds only the stub, and every e2e MPI test is skipped or `#[ignore]`d when the plugin is absent, so none of this was caught.

Before merge I would like:
- Fixes for the three defects.
- A test that builds this cdylib and asserts its entry points resolve (would have caught [1]).
- A CI lane that builds the functional plugin against libpmix and runs one real MPI job (would have caught [1]-[3]).
- Consider a pure-Rust FFI plugin (bindgen). A Rust cdylib exports `#[no_mangle] extern "C"` symbols by default (no [1]), compiles no C so there is no include-path issue ([3]), and generated bindings enforce the real API at compile time ([2]).

Test evidence (bare-metal, 4 nodes, non-Raft):
- PASS: baseline job, `--mpi=list` (srun and sbatch), `--mpi=pmi2` rejected, `--mpi=pmix -N2` rejected at submit, `ldd spurd` has no libpmix, missing-plugin error.
- FAIL: `--mpi=pmix` with the built plugin present, `dlsym failed`.

PR #506, opened 2026-07-27
fix(spurctld): serve read RPCs from local state when no Raft leader

biluriuday:
Read RPCs forwarded every request to the Raft leader and hard-failed with "no leader elected yet" whenever no leader was elected (startup election window or quorum loss), so squeue/sinfo/etc. returned errors. Since every node applies the committed Raft log locally, reads never actually need the leader.

Read RPCs now forward to the leader for freshness when one is reachable but fall back to local applied state on any forwarding failure instead of erroring. Write RPCs keep strict forward-or-fail behavior.

### DIFF
--- a/crates/spurctld/src/server.rs
+++ b/crates/spurctld/src/server.rs
@@ -128,13 +128,13 @@ impl ControllerService {
         )
     }
 
-    /// Best-effort forward of a read to the leader. `Some` only when the leader
-    /// answered; `None` (skip, unreachable, or forwarded call errored) tells the
-    /// caller to serve local state. `rpc` names the call for the fallback log.
+    /// Best-effort forward of a read to the leader: `Some(payload)` forwards
+    /// (clone lazily via `forward.then(|| ...)`), `None` serves local state. Any
+    /// forward error is swallowed to local — safe only while read handlers return
+    /// Ok/NotFound; a handler with a real error (e.g. InvalidArgument) masks it.
     async fn forward_read_optional<T, R, F, Fut>(
         &self,
-        forward: bool,
-        payload: T,
+        payload: Option<T>,
         rpc: &str,
         call: F,
     ) -> Option<Response<R>>
@@ -142,9 +142,7 @@ impl ControllerService {
         F: FnOnce(SlurmControllerClient<tonic::transport::Channel>, Request<T>) -> Fut,
         Fut: std::future::Future<Output = Result<Response<R>, Status>>,
     {
-        if !forward {
-            return None;
-        }
+        let payload = payload?;
         let client = self.leader_proxy.try_get_leader_client().await?;
         let mut fwd = Request::new(payload);
         *fwd.metadata_mut() = Self::forwarded_metadata();
@@ -285,9 +283,11 @@ impl SlurmController for ControllerService {
         let forward = self.read_should_forward(&request);
         let req = request.into_inner();
         if let Some(resp) = self
-            .forward_read_optional(forward, req.clone(), "get_jobs", |mut c, r| async move {
-                c.get_jobs(r).await
-            })
+            .forward_read_optional(
+                forward.then(|| req.clone()),
+                "get_jobs",
+                |mut c, r| async move { c.get_jobs(r).await },
+            )
             .await
         {
             return Ok(resp);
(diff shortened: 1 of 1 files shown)


crates/spurctld/src/server.rs
```diff
@@ -208,17 +260,16 @@ impl SlurmController for ControllerService {
...
-                let mut fwd = Request::new(request.into_inner());
-                *fwd.metadata_mut() = Self::forwarded_metadata();
-                return client.get_jobs(fwd).await;
-            }
-        }
-
+        let forward = self.read_should_forward(&request);
         let req = request.into_inner();
+        if let Some(resp) = self
+            .forward_read_optional(forward, req.clone(), "get_jobs", |mut c, r| async move {
```
yansun1996: `req.clone()` is evaluated at the call site before `forward_read_optional` checks `!forward`, so the locally-served path (leader, or an already-forwarded request) pays the clone even though the payload is never used. Same in `get_nodes` and `get_node`. Would it be worth only cloning when actually forwarding — e.g. take the payload by value and clone inside the `forward` branch, or pass `forward.then(|| req.clone())` — so the common local read avoids the copy? (`get_job`/`get_job_steps` sidestep this since their request types are `Copy`.)

crates/spurctld/src/server.rs
```diff
@@ -105,6 +118,45 @@ impl ControllerService {
...
+        if !forward {
+            return None;
+        }
+        let client = self.leader_proxy.try_get_leader_client().await?;
+        let mut fwd = Request::new(payload);
+        *fwd.metadata_mut() = Self::forwarded_metadata();
+        match call(client, fwd).await {
+            Ok(resp) => Some(resp),
+            Err(e) => {
+                warn!("forwarding {rpc} to leader failed, serving locally: {e}");
```
yansun1996, line 152: This falls back to local state on *any* forwarded error, which also swallows transport failures (leader mid-crash, deadline) and answers with local applied state rather than surfacing the error. That's harmless today — every read handler's local path only returns `NotFound`/`Ok`, so nothing misleading can result — but it's a deliberate availability-over-freshness choice. Might be worth a one-line note here that the fallback is intentional for all error kinds, so a future read handler that can return a real error (e.g. `InvalidArgument`) doesn't get silently masked by stale-but-Ok local data.

crates/spurctld/src/server.rs
```diff
@@ -2377,6 +2437,84 @@ mod tests {
...
+        // Follower forwards a fresh read to the leader.
+        assert!(read_forwarding_policy(false, false));
+        // An already-forwarded read is served locally to avoid forward loops.
+        assert!(!read_forwarding_policy(false, true));
+    }
+
+    /// A non-leader that can't reach a leader must still answer reads
+    /// from local applied state, not fail with "no leader elected yet".
+    #[tokio::test]
+    async fn get_jobs_serves_locally_when_no_leader_elected() {
```
yansun1996, line 2630: Nice test. To pin the other half of the contract, consider a companion using the same no-leader harness that asserts a write RPC (e.g. `submit_job`) still returns `Unavailable`/"not the Raft leader" — that guards against a future refactor accidentally extending the local-fallback to writes.

yansun1996 (commented):
Nice fix — since every node applies the committed Raft log locally, falling back to local state on reads instead of hard-failing when there's no leader is the right call, and keeping writes strict forward-or-fail is the correct split. The `x-spur-forwarded` loop guard and the leader/follower handling all look sound.

A few small things noted inline. One coverage gap worth considering: there's a good regression test that reads now serve locally with no leader, but nothing pins the other half of the contract — that a write RPC (e.g. `submit_job`) still hard-fails without a leader. The new test's harness makes that a cheap companion to add.

crates/spurctld/src/server.rs
biluriuday: done

crates/spurctld/src/server.rs
biluriuday: added a comment/note stating the same

crates/spurctld/src/server.rs
biluriuday: added a companion test for writes

yansun1996 (approved):

PR #507, opened 2026-07-27
fix(spur-net): cache OCI layers under the resolved image directory

### LINKED ISSUE #345: fix(spur-net): layer cache path does not follow image directory fallback for non-root users
## Problem

The image storage directory (`resolve_image_dir()` in `spur-cli/src/image.rs`) has a writability-aware fallback: system dir → `~/.spur/images/`. The layer cache in `pull_and_extract()` (`spur-net/src/oci.rs`) does not — it always defaults to `/var/spool/spur/images/.layers` regardless of writability.

For non-root users, `create_dir_all` and `fs::write` on the cache path fail silently (`let _ = ...`), so layers are never cached. Every fresh import re-downloads all layers. No warning is printed, unlike the image dir fallback which prints a note.

## Expected Behavior

- Layer cache should fall back alongside the image dir (e.g. `~/.spur/images/.layers/`)
- Or derive cache path from the `output_dir` passed to `pull_image()` (e.g. `output_dir.join(".layers")`)
- Cache write failures should log a warning

## Acceptance Criteria

- [ ] Layer cache falls back to a writable user-local path when system dir is not writable
- [ ] Cache write failures produce a warning log
- [ ] `SPUR_IMAGE_CACHE` override still works
- [ ] Layers are actually cached for non-root users

maybeharshit:
## Motivation

Non-root image imports fall back from `/var/spool/spur/images` to `~/.spur/images`, but the OCI layer cache remained hardcoded under the system directory. Cache writes consequently failed silently, causing every fresh import to download all layers again.

Fixes ROCm/spur#345.

## Technical Details

- Derive the default layer cache from the resolved image output directory using `<output_dir>/.layers`.
- Preserve `SPUR_IMAGE_CACHE` as an explicit override.
- Continue image imports when cache creation or writes fail.
- Emit structured warnings when tracing is configured and visible stderr warnings otherwise.
- Add unit coverage for default cache resolution and environment overrides.

## Test Plan

- Run the `spur-net` unit tests.
- Run Clippy across the workspace.
- Run the complete workspace test suite.
- Validate rootless imports in an Ubuntu 24.04 LXD VM using an unprivileged user and a deterministic local OCI registry.
- Verify default fallback, cache reuse, explicit override, and unwritable-cache behavior.

## Test Result

- All 36 `spur-net` tests passed.
- Workspace Clippy passed without warnings.
- Full workspace test suite passed.
- LXD validation confirmed:
  - layers were cached under `~/.spur/images/.layers`;
  - a second import reused the layer without another registry request;
  - `SPUR_IMAGE_CACHE` remained effective;
  - cache creation and write failures produced warnings without failing the import.

## Submission Checklist

### DIFF
--- a/crates/spur-net/src/oci.rs
+++ b/crates/spur-net/src/oci.rs
@@ -24,7 +24,7 @@ use anyhow::{bail, Context};
 use base64::{engine::general_purpose::STANDARD, Engine as _};
 use reqwest::header::{ACCEPT, AUTHORIZATION};
 use serde::Deserialize;
-use tracing::{debug, info};
+use tracing::{debug, info, warn, Level};
 
 /// A parsed container image reference.
 #[derive(Debug, Clone)]
@@ -157,7 +157,11 @@ pub async fn pull_image(image: &str, output_dir: &Path) -> anyhow::Result<PathBu
     let rootfs_dir = tmp_dir.join("rootfs");
     std::fs::create_dir_all(&rootfs_dir)?;
 
-    let result = pull_and_extract(&image_ref, &rootfs_dir).await;
+    let cache_dir = layer_cache_dir(
+        output_dir,
+        std::env::var_os("SPUR_IMAGE_CACHE").map(PathBuf::from),
+    );
+    let result = pull_and_extract(&image_ref, &rootfs_dir, &cache_dir).await;
     if let Err(e) = &result {
         let _ = std::fs::remove_dir_all(&tmp_dir);
         return Err(anyhow::anyhow!("{}", e));
@@ -211,7 +215,11 @@ pub async fn pull_image(image: &str, output_dir: &Path) -> anyhow::Result<PathBu
 }
 
 /// Download manifest and layers, extract to rootfs directory.
-async fn pull_and_extract(image_ref: &ImageRef, rootfs_dir: &Path) -> anyhow::Result<()> {
+async fn pull_and_extract(
+    image_ref: &ImageRef,
+    rootfs_dir: &Path,
+    cache_dir: &Path,
+) -> anyhow::Result<()> {
     let client = reqwest::Client::builder().user_agent("spur/0.1").build()?;
 
     // Get auth token
@@ -279,12 +287,25 @@ async fn pull_and_extract(image_ref: &ImageRef, rootfs_dir: &Path) -> anyhow::Re
 
     info!(layers = manifest.layers.len(), "downloading layers");
 
-    // Layer cache directory
-    let cache_dir = PathBuf::from(
-        std::env::var("SPUR_IMAGE_CACHE")
-            .unwrap_or_else(|_| "/var/spool/spur/images/.layers".into()),
-    );
-    let _ = std::fs::create_dir_all(&cache_dir);
+    let cache_enabled = match std::fs::create_dir_all(cache_dir) {
+        Ok(()) => true,
+        Err(error) => {
+            if tracing::enabled!(Level::WARN) {
+                warn!(
+                    path = %cache_dir.display(),
+                    %error,
+                    "failed to create image layer cache; continuing without cache"
+                );
+            } else {
+                eprintln!(
+                    "Warning: failed to create image layer cache at {}; continuing without cache: {}",
+                    cache_dir.display(),
+                    error
+                );
+            }
+            false
+        }
+    };
 
     // Download layers in parallel, then extract sequentially (order matters)
     let mut layer_data: Vec<(usize, bytes::Bytes)> = Vec::new();
@@ -294,11 +315,11 @@ async fn pull_and_extract(image_ref: &ImageRef, rootfs_dir: &Path) -> anyhow::Re
     for (i, layer) in manifest.layers.iter().enumerate() {
         let digest = layer.digest.clone();
         let size = layer.size;
-        let cache_path = cache_dir.join(digest.replace(':', "_"));
+        let cache_path = cache_enabled.then(|| cache_dir.join(digest.replace(':', "_")));
 
         // Check layer cache
-        if cache_path.exists() {
-            if let Ok(cached) = std::fs::read(&cache_path) {
+        if let Some(cache_path) = &cache_path {
+            if let Ok(cached) = std::fs::read(cache_path) {
                 info!(
                     layer = i + 1,
                     total = manifest.layers.len(),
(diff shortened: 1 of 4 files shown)


yansun1996:
This PR is on top of PR #500 , wait for the merge of the previous PR then proceed with this one.

yansun1996:
Hi @maybeharshit , we've merged your previous PR #500 , please rebase this PR on top of latest upstream main branch.

crates/spur-net/src/oci.rs
```diff
@@ -155,7 +155,11 @@ pub async fn pull_image(image: &str, output_dir: &Path) -> anyhow::Result<PathBu
     let rootfs_dir = tmp_dir.join("rootfs");
     std::fs::create_dir_all(&rootfs_dir)?;
 
-    let result = pull_and_extract(&image_ref, &rootfs_dir).await;
+    let cache_dir = layer_cache_dir(
+        output_dir,
+        std::env::var_os("SPUR_IMAGE_CACHE").map(PathBuf::from),
```
yansun1996: `var_os(...).map(PathBuf::from)` treats `SPUR_IMAGE_CACHE=""` as an explicit empty override, so `layer_cache_dir` returns `""`, `create_dir_all("")` fails, and the cache is disabled — which reintroduces the "re-download every layer" symptom this PR fixes. The sibling env resolver in `spur-cli/src/image.rs` guards this (`if !dir.is_empty()`); would it be worth mirroring that here?

```rust
std::env::var_os("SPUR_IMAGE_CACHE")
    .filter(|v| !v.is_empty())
    .map(PathBuf::from),
```

crates/spur-net/src/oci.rs
```diff
@@ -277,12 +285,25 @@ async fn pull_and_extract(image_ref: &ImageRef, rootfs_dir: &Path) -> anyhow::Re
...
-    // Layer cache directory
-    let cache_dir = PathBuf::from(
-        std::env::var("SPUR_IMAGE_CACHE")
-            .unwrap_or_else(|_| "/var/spool/spur/images/.layers".into()),
-    );
-    let _ = std::fs::create_dir_all(&cache_dir);
+    let cache_enabled = match std::fs::create_dir_all(cache_dir) {
+        Ok(()) => true,
+        Err(error) => {
+            if tracing::enabled!(Level::WARN) {
```
yansun1996: This `tracing::enabled!(WARN)` → `warn!` else `eprintln!` block is duplicated verbatim at the layer-write site below — worth pulling into a small shared helper.

One subtlety with the gate: `tracing::enabled!(Level::WARN)` is also false when a subscriber *is* installed but WARN is filtered out. In that case `warn!` is suppressed and the `eprintln!` branch never runs, so the warning is lost in exactly the situation the fallback was meant to cover. Always calling `warn!` (and letting the CLI install a subscriber) may be simpler and avoids that gap.

crates/spur-net/src/oci.rs
```diff
@@ -845,4 +885,25 @@ mod tests {
             "nvcr.io+nvidia+pytorch+24.01"
         );
     }
+
+    #[test]
+    fn layer_cache_defaults_below_output_directory() {
```
yansun1996, line 909: These two tests exercise only the pure `layer_cache_dir` helper, so they'd still pass if the core of the fix (the env read in `pull_image`, the `create_dir_all` failure handling, and the `Option` gating on the layer read/write) were reverted. Could the actual degradation path get coverage? Extracting a small synchronous cache read/write seam would let you assert the disable-on-unwritable and round-trip behaviors with `tempfile`, no network needed. At minimum, an empty-`SPUR_IMAGE_CACHE` case here would lock in the guard suggested above.

yansun1996 (commented):
Thanks for this — deriving the layer cache from the resolved `output_dir` is the right fix for #345, and degrading gracefully instead of silently swallowing cache failures is a nice touch. A few suggestions before merge, mostly an edge case and test coverage — details inline.

yansun1996 (changes requested):

crates/spur-net/src/oci.rs
maybeharshit: Good catch — fixed in 1d48fa3. Rather than filtering at the call site, the guard now lives inside `layer_cache_dir` itself, so every caller (including the tests) goes through the same rule:

```rust
match override_dir.filter(|dir| !dir.is_empty()) {
    Some(dir) => PathBuf::from(dir),
    None => output_dir.join(".layers"),
}
```

An empty `SPUR_IMAGE_CACHE` is now treated as unset and falls back to `<output_dir>/.layers`, matching the `SPUR_IMAGE_DIR` handling in `spur-cli/src/image.rs`.

crates/spur-net/src/oci.rs
maybeharshit: Agreed on both points — the `tracing::enabled!` gate is gone entirely as of 1d48fa3.

Cache handling now sits behind a small `LayerCache` type that calls `warn!` unconditionally at both the create and write sites, so there is no duplicated block and no way for a subscriber with WARN filtered out to swallow the message without the `eprintln!` fallback firing.

To make sure those warnings actually reach users, `spur-cli` now installs a stderr subscriber in `main`, defaulting to `warn` and honouring `RUST_LOG` for more verbosity. Previously the CLI had no subscriber at all, which is what the `eprintln!` fallback was compensating for.

crates/spur-net/src/oci.rs
maybeharshit: You were right that the original tests would have survived a revert of the actual fix. 1d48fa3 extracts the synchronous read/write seam you suggested — a `LayerCache` type with `open`/`read_layer`/`write_layer` — so the degradation paths are now covered directly with `tempfile`, no network needed:

- `layer_cache_round_trips_layers` — write then read back through the cache.
- `layer_cache_disabled_when_directory_cannot_be_created` — `open` on an unwritable parent degrades to a disabled cache, and reads/writes become no-ops instead of errors.
- `layer_cache_write_failure_leaves_pull_usable` — a failing write does not propagate, so the pull continues.
- `layer_cache_ignores_empty_environment_override` — locks in the empty-value guard from the thread above.

All 40 `spur-net` tests pass, along with workspace clippy and the full test suite.

maybeharshit:
@yansun1996 thanks for the review — all three points are addressed in 1d48fa3, with details in the inline threads. Summary:

- **Empty `SPUR_IMAGE_CACHE`**: the guard now lives inside `layer_cache_dir` (`override_dir.filter(|dir| !dir.is_empty())`), so an empty value is treated as unset and falls back to `<output_dir>/.layers`.
- **Duplicated warning block**: removed along with the `tracing::enabled!` gate. Cache handling moved behind a `LayerCache` type that calls `warn!` unconditionally, and `spur-cli` now installs a stderr subscriber (default `warn`, `RUST_LOG`-aware) so the warnings actually reach users — which was the real gap the `eprintln!` fallback was papering over.
- **Test coverage**: the create/read/write seam is now unit testable, so the degradation paths are exercised directly with `tempfile` rather than only the pure helper — round-trip, disable-on-unwritable-directory, write-failure, and the empty-override case.

Validated with `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` (clean) and the full `cargo test --locked` suite. Could you take another look when you get a chance?

yansun1996 (dismissed):
LGTM

shiv-tyagi (approved):
LGTM

PR #508, opened 2026-07-27
fix(spurctld): distinguish missing vs unavailable QOS in validation errors

### LINKED ISSUE #419: sbatch QoS/account errors don't hint at accounting status
When accounting is not connected (PostgreSQL unreachable), `sbatch --qos=X` returns:

```
code: 'Client specified an invalid argument', message: "QOS 'amd-burst-qos' does not exist"
```

This is misleading because the QoS does exist in the database — it's just not loaded because the accounting service failed to start.

**Expected behavior:**

When the controller has no accounting pool (i.e. the service isn't registered), QoS/account validation errors should include a hint:

```
QOS 'amd-burst-qos' does not exist (hint: accounting may not be enabled on this controller — check controller logs)
```

Or better: the controller should distinguish between "QoS genuinely doesn't exist" (accounting is connected, QoS not found) vs "cannot validate QoS" (accounting not available).

ssenthilnathan3:
Fixes #419 

This PR adds an `is_loaded()` flag to `QosCache`, following the same pattern as `AssociationCache`, and appends a hint to the error when the cache hasn't been populated yet. I believe this makes the failure mode clearer for users without changing existing behavior.

### DIFF
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -1523,7 +1523,12 @@ impl ClusterManager {
                     anyhow::bail!("cannot clear a job's QOS");
                 }
                 if self.qos_cache.get(&q).is_none() {
-                    anyhow::bail!("QOS '{q}' does not exist");
+                    let hint = if self.qos_cache.is_loaded() {
+                        ""
+                    } else {
+                        " (hint: accounting may not be enabled on this controller — check controller logs)"
+                    };
+                    anyhow::bail!("QOS '{q}' does not exist{hint}");
                 }
                 // Treat account="" as unset so we authorize against the
                 // job's existing account, rather than erroring on a blank
@@ -4741,7 +4746,14 @@ fn apply_default_qos(
 
     if let Some(name) = spec.qos.as_deref().filter(|n| !n.is_empty()) {
         if qos_cache.get(name).is_none() {
-            return Err(SubmitError::invalid(format!("QOS '{name}' does not exist")));
+            let hint = if qos_cache.is_loaded() {
+                String::new()
+            } else {
+                " (hint: accounting may not be enabled on this controller — check controller logs)".into()
+            };
+            return Err(SubmitError::invalid(format!(
+                "QOS '{name}' does not exist{hint}"
+            )));
         }
         if !qos_permitted(&allowed_qos, default_qos_for_auth.as_deref(), name) {
             return Err(SubmitError::invalid(format!(
@@ -4774,8 +4786,13 @@ fn apply_default_qos(
     let fallback = accounting.default_qos.trim();
     if !fallback.is_empty() {
         if qos_cache.get(fallback).is_none() {
+            let hint = if qos_cache.is_loaded() {
+                String::new()
+            } else {
+                " (hint: accounting may not be enabled on this controller — check controller logs)".into()
+            };
             return Err(SubmitError::invalid(format!(
-                "configured default QOS '{fallback}' does not exist"
+                "configured default QOS '{fallback}' does not exist{hint}"
             )));
         }
         if qos_permitted(&allowed_qos, default_qos_for_auth.as_deref(), fallback) {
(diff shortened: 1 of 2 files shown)


crates/spurctld/src/cluster.rs
```diff
@@ -10254,6 +10273,21 @@ mod tests {
...
+        let dir = TempDir::new().unwrap();
+        let cm = test_cluster(&dir).await;
+        let id = submit_and_wait(&cm, basic_spec("q"));
+        assert_eq!(cm.get_job(id).unwrap().spec.qos, None);
+
+        let err = cm
+            .update_job(id, None, None, None, None, None, Some("any-qos".into()))
+            .unwrap_err();
+        assert!(err.to_string().contains("QOS 'any-qos' does not exist"));
+        assert!(err.to_string().contains("accounting may not be enabled"));
```
yansun1996, line 10289: Would it be worth adding the negative assertion for the loaded-cache case on this path too? `update_job_qos_validates_against_cache` rejects a missing QOS with a loaded cache but only checks `.contains("QOS 'ghost' does not exist")`, which would still pass if a hint were accidentally appended. A one-line `assert!(!err.to_string().contains("accounting may not be enabled"))` there would lock down the no-hint behavior for the update-job path, matching the exact-match coverage the `apply_default_qos` sites already have.

crates/spurctld/src/cluster.rs
```diff
@@ -1523,7 +1523,12 @@ impl ClusterManager {
                     anyhow::bail!("cannot clear a job's QOS");
                 }
                 if self.qos_cache.get(&q).is_none() {
-                    anyhow::bail!("QOS '{q}' does not exist");
+                    let hint = if self.qos_cache.is_loaded() {
+                        ""
+                    } else {
+                        " (hint: accounting may not be enabled on this controller — check controller logs)"
```
yansun1996: Minor: this hint string (and the `is_loaded()` branch around it) is repeated at all three sites. A small `QosCache` helper returning the hint (or `""`) would keep the wording in one place. Optional — and if you do, using an ASCII `--` instead of the em-dash `—` would be a touch safer for a runtime message surfaced through the CLI on odd terminals.

crates/spurctld/src/limits_cache.rs
```diff
@@ -19,27 +20,36 @@ use spur_core::accounting::{Qos, QosLimits, QosPreemptMode, TresRecord};
 
 pub struct QosCache {
     qos: RwLock<HashMap<String, Qos>>,
+    loaded: AtomicBool,
```
yansun1996: One design-consistency thought: `AssociationCache` keeps its `loaded` flag *inside* the `RwLock<Snapshot>` so data and flag can never be read out of sync. Here it's a separate `AtomicBool`. For this use it's fine — it only gates a hint and the `Release` store happens after the map write — but folding it under the existing `RwLock` would drop the atomic and match the sibling exactly. Non-blocking.

yansun1996 (commented):
Thanks for this — the approach mirrors `AssociationCache` cleanly and the fix is nicely targeted. A few optional thoughts and one test suggestion below; nothing blocking from my side.

- The three hinted sites are exactly the cache-backed existence checks, and the DB-backed `add_user` paths in `accounting/grpc.rs` are correctly left un-hinted — nice consistency.
- One small test gap worth closing: the loaded-cache path for `scontrol update job qos=` doesn't currently assert the hint is *absent* (see inline note).
- A couple of optional polish items (hint-string duplication, em-dash in a runtime string) inline.

yansun1996 (changes requested):

yansun1996 (approved):
Thanks for the quick turnaround — the follow-up commit addresses everything cleanly. The hint is now a single `const`, the `loaded` flag lives inside the `RwLock<Snapshot>` so it can never be read out of sync with the map, and the no-hint behavior is locked down on the update-job path too. Nicely consistent with `AssociationCache`. LGTM.

PR #509, opened 2026-07-27
fix(spur-core): default JobSpec.pty so pre-0.6 Raft logs replay

yansun1996:
## What this fixes

Upgrading a controller from v0.5.1 to v0.6.0 crashes on startup with a Raft log deserialization error (*missing field `pty`*).

## Root cause

`spurctld` persists Raft log entries as JSON (`serde_json`, `raft.rs`). `WalOperation::JobSubmit` embeds `Box<JobSpec>`. The PTY-session work (#482) added `pub pty: bool` to `JobSpec` **without** `#[serde(default)]`. On upgrade, the controller replays `JobSubmit` entries (and snapshots) whose serialized `JobSpec` predates the field, so serde fails with *missing field `pty`*. Because strict WAL recovery is the startup default, spurctld refuses to start rather than degrading — a hard crash loop until the operator intervenes.

`pty` is the only field added to any WAL-embedded type since v0.5.1 that lacks a serde default; every sibling (`srun_job`, `run_attempt`, reservation `owner`, node `k0s_*`, …) already has one.

## Approach

- Annotate `JobSpec.pty` with `#[serde(default)]`, consistent with how every other post-v0.5.1 field was handled.
- Add a **frozen-fixture regression test** (`wal.rs`): a real v0.5.1 `JobSubmit` JSON payload — captured before the `gpus*`, `pty`, and `srun_job` fields existed — that current code must still deserialize. The existing `wal.rs` serde tests all round-trip through *today's* schema, so a missing `#[serde(default)]` is invisible to them (the field is present on both sides); that is exactly how this shipped. The frozen literal is intentionally immovable — any future non-defaulted `JobSpec` field breaks it, pointing the author at the fix before release rather than at a controller crash on upgrade.

## Testing

- `cargo test -p spur-core --locked` — 375 passed. Both new tests fail without the `pty` annotation and pass with it.
- `cargo clippy -p spur-core --all-targets --locked` — clean.

### DIFF
--- a/crates/spur-core/src/job.rs
+++ b/crates/spur-core/src/job.rs
@@ -1098,8 +1098,7 @@ mod tests {
         assert_eq!(effective_memory_mb(&spec, 1), 0);
     }
 
-    // A JobSpec serialized before the `pty` field existed (v0.5.1 Raft log
-    // entries) must still deserialize, or spurctld crashes on upgrade replay.
+    // Pre-`pty` JobSpec must still deserialize, or spurctld crashes on replay.
     #[test]
     fn job_spec_deserializes_without_pty_field() {
         let mut value = serde_json::to_value(JobSpec::default()).unwrap();
--- a/crates/spur-core/src/job.rs
+++ b/crates/spur-core/src/job.rs
@@ -486,6 +486,7 @@ pub struct JobSpec {
     pub open_mode: Option<String>,
 
     // Interactive PTY
+    #[serde(default)]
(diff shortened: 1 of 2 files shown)


yansun1996:
This is a blocking issue for upgrading to v0.6.0, the backward compatibility has been broken, this PR is fixing the issue and try to make sure that the pre v0.6.0 logs can be replayed on the v0.6.0 controller.

PR #516, opened 2026-07-27
fix(spur-cli): target the -w node when attaching an interactive PTY

yansun1996:
## What this fixes

`srun --jobid=N --overlap --pty -w <node>` ignored `-w` entirely: attaching an interactive shell to a multi-node job always landed on the same node no matter which one you asked for.

## Root cause

Two independent drops of the requested node:

1. The CLI's `--jobid` attach branch called `run_interactive_pty` with only the controller/job/command — `args.nodelist` (`-w`) was never forwarded.
2. `CreateJobStepRequest` had no node field, and the controller's `create_job_step` always resolved the step's agent address from `allocated_nodes.first()`.

So every attach resolved to `allocated_nodes[0]`.

## Approach

- Add an optional `node` field to `CreateJobStepRequest` (additive, field 8).
- On the attach path the CLI forwards the `-w` node. Since an interactive PTY attaches to exactly one node, a comma list is reduced to its first entry client-side.
- The controller validates the requested node is one the job actually holds (`select_step_node`) and resolves the agent address from it. A node outside the allocation is rejected with `InvalidArgument`; an empty request keeps the previous first-node default, so non-attach callers are unchanged.

## Known limitations

- Hostlist bracket syntax (e.g. `node[001-002]`) is not expanded on the attach path; pass a single node name. An unmatched name is rejected with a clear error.
- The non-interactive in-allocation step path (`dispatch_step`) still defaults to the first node; its placement is driven separately and it does not consume the returned address, so it is left unchanged here.

## Testing

- Unit tests for node selection (empty/named/not-allocated/empty-list/comma-joined) and the client-side `-w` reduction.
- `cargo clippy` + `cargo test` on the touched crates — clean.
- Validated end-to-end on an isolated 2-node deployment: attaching with `-w nodeA` vs `-w nodeB` opens the shell on the corresponding host; no `-w` falls back to the first allocated node; a non-allocated `-w` is rejected; a comma list reduces to its first node.

### DIFF
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -912,7 +912,7 @@ fn first_node(nodelist: &str) -> String {
     nodelist
         .split(',')
         .next()
-        .unwrap_or(nodelist)
+        .unwrap_or_default()
         .trim()
         .to_string()
 }
@@ -922,7 +922,7 @@ async fn try_stream_output(
     nodelist: &str,
     job_id: u32,
 ) -> bool {
-    let first_node = nodelist.split(',').next().unwrap_or(nodelist).trim();
+    let first_node = nodelist.split(',').next().unwrap_or_default().trim();
     if first_node.is_empty() {
         return false;
     }
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -212,7 +212,9 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
         if !args.overlap {
             anyhow::bail!("--jobid requires --overlap");
         }
-        let exit_code = run_interactive_pty(&args.controller, job_id, args.command.clone()).await?;
+        let node = args.nodelist.clone().unwrap_or_default();
+        let exit_code =
+            run_interactive_pty(&args.controller, job_id, args.command.clone(), node).await?;
         std::process::exit(exit_code);
     }
 
@@ -673,6 +675,7 @@ async fn dispatch_step(
             overlap: false,
             pty: false,
             winsize: None,
+            node: String::new(),
         })
         .await
         .context("failed to create job step")?
@@ -801,7 +804,13 @@ async fn run_standalone_srun(args: &SrunArgs, hooks: &HooksConfig, work_dir: &st
     if args.pty {
         ctrl_c_handle.abort();
         eprintln!("srun: opening interactive session on {}", nodelist);
-        let result = run_interactive_pty(&args.controller, job_id, args.command.clone()).await;
+        let result = run_interactive_pty(
+            &args.controller,
+            job_id,
+            args.command.clone(),
+            String::new(),
+        )
+        .await;
         let _ = client
             .cancel_job(CancelJobRequest {
                 job_id,
@@ -1104,7 +1113,12 @@ fn warn_unsupported_cpu_bind(environment: &HashMap<String, String>) {
 ///
 /// Retries transient failures (job not yet visible on agent after controller
 /// reports it as Running) up to a few seconds before giving up.
-async fn run_interactive_pty(controller: &str, job_id: u32, command: Vec<String>) -> Result<i32> {
+async fn run_interactive_pty(
+    controller: &str,
+    job_id: u32,
+    command: Vec<String>,
+    node: String,
+) -> Result<i32> {
     let channel = spur_client::connect_channel(controller)
         .await
         .context("cannot connect to controller")?;
@@ -1132,6 +1146,7 @@ async fn run_interactive_pty(controller: &str, job_id: u32, command: Vec<String>
                     overlap: true,
                     pty: true,
                     winsize: Some(winsize),
+                    node: node.clone(),
                 })
                 .await
             {
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -212,7 +212,7 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
         if !args.overlap {
             anyhow::bail!("--jobid requires --overlap");
         }
-        let node = args.nodelist.clone().unwrap_or_default();
+        let node = first_node(args.nodelist.as_deref().unwrap_or_default());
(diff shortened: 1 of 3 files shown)


proto/slurm.proto
```diff
@@ -1178,6 +1178,7 @@ message CreateJobStepRequest {
   bool overlap = 5;           // Share resources with existing step (--overlap)
   bool pty = 6;               // Allocate a PTY for the step
   WindowSize winsize = 7;     // Initial terminal dimensions
+  string node = 8;            // Target node (-w); empty = first allocated node
```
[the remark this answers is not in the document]
yansun1996: Good catch — updated the comment so `node_addr` reflects the address of the selected node (the `-w` target, or the first allocated node when unset), rather than always the first allocated node.

crates/spurctld/src/server.rs
```diff
@@ -1827,6 +1829,23 @@ pub async fn serve(
     Ok(())
 }
 
+/// Resolve the node an interactive step attaches to. Empty request = the first
+/// allocated node (legacy default); a named node must be one the job actually
+/// holds, otherwise the step would attach outside the allocation.
+fn select_step_node<'a>(allocated: &'a [String], requested: &str) -> Result<&'a str, String> {
```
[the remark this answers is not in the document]
yansun1996: Agreed — reworded the doc comment to describe it as resolving the target node for any job step, since it runs for every step request, not just interactive PTY attaches.

crates/spurctld/src/server.rs
```diff
@@ -1247,15 +1250,11 @@ impl SlurmController for ControllerService {
...
-            .map(|n| {
-                let host = n.address.as_deref().unwrap_or(&n.name);
-                format!("{}:{}", host, n.port)
-            })
-            .unwrap_or_default();
+        let node = self.cluster.get_node(target_node).ok_or_else(|| {
+            Status::unavailable(format!("node {target_node} is not currently registered"))
+        })?;
+        let host = node.address.as_deref().unwrap_or(&node.name);
+        let node_addr = format!("{}:{}", host, node.port);
```
shiv-tyagi: This `get_node` resolution runs after `create_step` (above). When the target node is allocated but not currently registered, the controller returns `Status::unavailable`, and `is_retryable_status` (srun.rs:1221) treats `Unavailable` as retryable. So `run_interactive_pty` retries up to 5 times, and each attempt creates a new step before failing, leaking steps.

Previously this path returned an empty `node_addr` and the CLI bailed non-retryably, so only one step was ever created. Resolve the address before creating the step so a registration failure short-circuits cleanly:

```rust
let node = self.cluster.get_node(target_node).ok_or_else(|| {
    Status::unavailable(format!("node {target_node} is not currently registered"))
})?;
let host = node.address.as_deref().unwrap_or(&node.name);
let node_addr = format!("{}:{}", host, node.port);

// ... then build `step` and call create_step ...
```

crates/spur-cli/src/srun.rs
```diff
@@ -895,6 +904,19 @@ fn build_command_script(command: &[String]) -> Result<String> {
...
+/// then defaults to the first allocated node). Hostlist brackets are not
+/// expanded here — the controller rejects an unmatched name with a clear error.
+fn first_node(nodelist: &str) -> String {
+    nodelist
+        .split(',')
+        .next()
+        .unwrap_or(nodelist)
+        .trim()
+        .to_string()
+}
```
shiv-tyagi, line 918: Nit: `str::split(',')` always yields at least one element (even for `""`), so `.next()` is always `Some` and the `.unwrap_or(nodelist)` fallback is dead code. It reads as if an empty input needs a fallback when it doesn't.

```suggestion
fn first_node(nodelist: &str) -> String {
    nodelist
        .split(',')
        .next()
        .unwrap_or_default()
        .trim()
        .to_string()
}
```

crates/spurctld/src/server.rs
```diff
@@ -1222,6 +1222,9 @@ impl SlurmController for ControllerService {
             )));
         }
 
+        let target_node =
+            select_step_node(&job.allocated_nodes, &req.node).map_err(Status::invalid_argument)?;
```
shiv-tyagi, line 1226: The step is created (below) with `nodes: job.allocated_nodes.clone()` even when `-w` targets a single node, so the recorded step spans all allocated nodes while it actually attaches to `target_node`. This matches prior behavior, so not a regression.

Q: Should the step record only `target_node` so step listings reflect where the PTY actually runs?

shiv-tyagi (commented):
posted comments. ptal.

crates/spurctld/src/server.rs
yansun1996: Fixed — the target node's address is now resolved before the step is created, so an unregistered node returns `Unavailable` before any step exists and the client's retries can't each leak a step. Added a `create_job_step` integration test that drives the retry path against an unregistered target and asserts the step count is unchanged.

crates/spur-cli/src/srun.rs
yansun1996: Good catch — changed both to `.unwrap_or_default()` (here and the identical spot in `try_stream_output`). The old fallback was unreachable since `split(',')` always yields at least one element.

crates/spurctld/src/server.rs
yansun1996: Good question. `step.nodes` isn't currently surfaced anywhere (`JobStepInfo` has no nodes field, and step dispatch resolves nodes from the job's allocation rather than the step), so recording only `target_node` wouldn't change any observable behavior today. Since it's pre-existing and orthogonal to this attach fix, I'd prefer to leave it and address it as a follow-up alongside exposing per-step nodes in step listings — happy to scope it here instead if you'd prefer.

yansun1996:
On the low-confidence note about `create_job_step` returning an empty `node_addr` for an unknown node: the current code already returns a `Status::unavailable` for that case rather than an empty address, and the step-ordering fix in this update keeps that error ahead of step creation. So the actionable-error concern is handled.

shiv-tyagi (approved):
LGTM

PR #517, opened 2026-07-27
test(spur-core): freeze per-variant WAL payloads to guard replay back-compat

### LINKED ISSUE #515: test(spurctld): Raft/WAL types can silently lose backward-compat, crashing spurctld on upgrade replay, need a test coverage
## Problem

spurctld persists the Raft log and snapshots as JSON and replays them on startup under strict recovery. If a field is added to any type reachable from `WalOperation` (or the snapshot state) without `#[serde(default)]` (or making it `Option`), old persisted entries fail to deserialize and the controller **crashes on upgrade**.

This already happened once: `JobSpec.pty` was added without a default, so pre-0.6 Raft logs failed to replay. That specific case is fixed, and a single frozen-payload test now guards `JobSubmit`/`JobSpec`.

## Why the current guard is insufficient

The fix added one frozen v0.5.1 `JobSubmit` fixture. That only exercises the types reachable from that one variant (`JobSpec`). The rest of the persisted surface is unguarded:

- Other `WalOperation` variants: `JobStart`, `JobStepCreate` (`JobStep`), `NodeRegister`/`NodeUpdate` (`ResourceSet`), `TokenCreate` (`AdmissionToken`), `ReservationCreate` (`Reservation`), `NodeStateChange` (`NodeState`), the k0s variants, etc.
- The snapshot surface (`PersistedSnapshot` / `restore_from_snapshot`), a separate persisted blob from log entries.

A new field on any of those, or a brand-new WAL variant, can reintroduce the same upgrade crash with no test catching it. Round-trip serde tests do not catch it — they use today's schema on both sides.

## Expected

Adding a field to a persisted type, or a new WAL variant, without preserving backward-compatible deserialization should fail a test locally / in CI — not crash a real controller on upgrade.

## Reproduce

1. Add a non-defaulted field to any type reachable from a `WalOperation` variant other than `JobSubmit` (e.g. a field on `ResourceSet`).
2. Serialize an entry of that variant under the old schema; attempt to deserialize under the new one.
3. Deserialization fails; under strict recovery this aborts spurctld startup.

yansun1996:
## What

Adds a backward-compatibility test harness for the Raft WAL log-entry surface in `crates/spur-core/src/wal.rs`.

`spurctld` persists every `WalOperation` as JSON and replays it on startup under strict recovery. A field added to any type reachable from `WalOperation` without `#[serde(default)]` (or `Option`) makes old entries fail to deserialize and crashes the controller on upgrade. A single-variant frozen fixture was recently added for the `JobSpec.pty` incident; this generalizes that guard across the whole log-entry enum.

## Approach

- `FIXTURES`: one frozen JSON blob per `WalOperation` variant, captured from a real serialization. If a new field lands without a serde default, the matching blob stops deserializing and the test fails with an actionable message.
- `variant_tag`: an exhaustive match with **no `_` arm**, so adding a new `WalOperation` variant breaks compilation until an arm — and a matching fixture — is added. `every_variant_has_a_fixture` backs this at runtime.
- Fixture collections carry at least one element (a `GpuResource`, an `AllocatedDevice`) so serde descends into the nested element structs rather than skipping them behind an empty `Vec`/`Map`.

The guard was verified by temporarily adding a non-defaulted field to a nested type and confirming the frozen fixture fails to deserialize with the intended message.

## Scope / known limitations

- **Directional guard.** Catches an *added non-defaulted field* (the crash class). It does not catch a renamed/removed field or variant — serde ignores unknown fields (no `deny_unknown_fields`), and renames/removals are inherently unsafe for old-log replay regardless.
- **Snapshot surface not covered.** The controller also restores a separate `ClusterSnapshot` blob (full `Job`/`Node` structs) under the same strict path. That is a distinct surface needing its own fixture test — tracked as follow-up.

## Testing

`cargo test -p spur-core`, `cargo clippy -p spur-core --all-targets`, and `cargo fmt --check` all pass. The change is test-only — no runtime behavior changes.

### DIFF
--- a/crates/spur-core/src/wal.rs
+++ b/crates/spur-core/src/wal.rs
@@ -340,21 +340,8 @@ mod reservation_wal_tests {
     }
 }
 
-// Backward-compatibility guard for the WAL log-entry surface. spurctld persists
-// every `WalOperation` as JSON and replays it on startup under strict recovery, so
-// a field added to any embedded type without `#[serde(default)]` (or `Option`)
-// makes old entries fail to deserialize and crashes the controller on upgrade.
-//
-// Each `FIXTURES` entry is a frozen JSON blob for one variant; collections hold at
-// least one element so serde descends into the element structs (an empty Vec/Map
-// would leave them unguarded). They must NOT be regenerated: a failure here means a
-// new field needs `#[serde(default)]`, not a fixture edit. `variant_tag` is an
-// exhaustive match (no `_` arm), so a new `WalOperation` variant breaks compilation
-// until an arm — and a matching frozen fixture — is added.
-//
-// Scope/limits: this guards only the added-non-defaulted-field case (serde ignores
-// unknown fields, so renames/removals are not caught here). The separate snapshot
-// blob (`ClusterSnapshot`, full `Job`/`Node`) is a distinct surface, not covered.
+// Frozen old WAL payloads: a new field without `#[serde(default)]` breaks replay and
+// crashes spurctld on upgrade. Do not regenerate — fix the field, not the fixture.
 #[cfg(test)]
 mod backcompat_fixtures {
     use super::*;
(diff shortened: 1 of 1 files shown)


yansun1996:
close, we will update the AGENTS.md to help mark the PR that could break the compatibility between release versions, instead of maintaining hard coded test case

PR #519, opened 2026-07-27
fix(spur-cli): honor format= in sacctmgr show account

### LINKED ISSUE #487: sacctmgr show account does not honor format= selection
## What happens

`sacctmgr show account` prints a fixed set of columns (Account, Descr, Org, Parent, Share, GrpTRES). Unlike `sacctmgr show qos` and `sacctmgr show user`, it ignores a `format=` argument, so a user cannot select or reorder the displayed fields.

## Example

```
$ sacctmgr show account format=Account,GrpTRES
```

Expected: a two-column table with just `Account` and `GrpTRES`.
Actual: the full fixed six-column table is printed; `format=` has no effect.

## Notes

`show qos` and `show user` already route through the shared format engine (`resolve_format`) to honor `format=`; the account branch does not. Surfaced while exposing account-level GrpTRES in the CLI.

### LINKED ISSUE #536: sacctmgr show account ignores where name= filter
## What happens

`sacctmgr show account where name=<acct>` lists **all** accounts, ignoring the `name=` filter. Only `format=` is honored on the account path.

## What's expected

Consistent with `sacctmgr show qos`, which honors `name=` and returns only the matching entry (e.g. `sacctmgr show qos where name=normal` lists just `normal`).

## Reproduce

```
sacctmgr -i add account name=physics
sacctmgr -i add account name=chem
sacctmgr show account where name=physics   # lists physics AND chem
```

## Notes

The account path issues `ListAccountsRequest {}` and applies no name filter client-side, whereas the qos path filters by name. `ListAccountsRequest` currently has no name field, so a full fix likely wants server-side filtering rather than mirroring the qos client-side approach.

yansun1996:
## What

`sacctmgr show account` printed a fixed six-column table and ignored a `format=` argument, so users could not select or reorder the displayed fields. `show qos` already routes through the shared format engine (`resolve_format`) to honor `format=`; the account branch did not. Fixes #487.

## Approach

Route the `account` branch of `show` through `format_engine::resolve_format` + `print_header` + `format_row`, mirroring the existing `qos` branch. Adds the account field/header/spec helpers and default/all format strings:

- `format=Account,GrpTRES` selects and narrows columns.
- `format=GrpTRES,Account` reorders them.
- `format=all` expands to every column (adds `MaxJobs`).
- An unrecognized `format=` value errors with the list of valid fields.

Field-name aliases follow the same convention as `qos` (case-insensitive, common Slurm spellings). `MaxJobs` renders blank when unset (0), consistent with the qos zero-handling.

## Behavior change

The empty-account-list placeholder row (`(no accounts configured)`) is removed in favor of header-only output. This matches real `sacctmgr` and the `qos` path; there is no implicit default account to synthesize (unlike the qos `normal` default).

## Testing

- Unit tests for field selection, reordering, default/all format strings, alias case-insensitivity, unknown-field rejection, and the `resolve_format` `all`/error paths.
- An e2e test (`TestSacctmgrShowAccount`) covering default columns and `format=` selection, alongside the existing qos coverage.
- `cargo clippy --workspace --exclude spur-ffi --all-targets` clean, `cargo fmt --check` clean, full `spur-cli` suite passes.
- Exercised end-to-end against an isolated single-node deployment: default output, column selection/reordering, `format=all`, the `list` alias, and the unknown-field error path all behave as expected.

### DIFF
--- a/crates/spur-cli/src/sacctmgr.rs
+++ b/crates/spur-cli/src/sacctmgr.rs
@@ -707,6 +707,15 @@ async fn show(entity: &str, params: &[String], addr: &str) -> Result<()> {
 
     match entity.to_lowercase().as_str() {
         "account" | "accounts" => {
+            let fields = format_engine::resolve_format(
+                p.get("format").map(String::as_str),
+                ACCOUNT_DEFAULT_FORMAT,
+                ACCOUNT_ALL_FORMAT,
+                &account_field_spec,
+                &account_header,
+                "Account, Descr, Org, Parent, Share, GrpTRES, MaxJobs",
+            )?;
+
             let mut client = connect(addr).await?;
             let resp = client
                 .list_accounts(ListAccountsRequest {})
@@ -715,34 +724,13 @@ async fn show(entity: &str, params: &[String], addr: &str) -> Result<()> {
 
             let accounts = resp.into_inner().accounts;
 
-            println!(
-                "{:<20} {:<30} {:<15} {:<10} {:<10} {:<10}",
-                "Account", "Descr", "Org", "Parent", "Share", "GrpTRES"
-            );
-            println!("{}", "-".repeat(100));
+            format_engine::print_header(&fields);
 
-            if accounts.is_empty() {
+            for a in &accounts {
                 println!(
-                    "{:<20} {:<30} {:<15} {:<10} {:<10} {:<10}",
-                    "(no accounts configured)", "", "", "", "", ""
+                    "{}",
+                    format_engine::format_row(&fields, &|spec| resolve_account_field(a, spec))
                 );
-            } else {
-                for a in &accounts {
-                    let parent = if a.parent_account.is_empty() {
-                        ""
-                    } else {
-                        &a.parent_account
-                    };
-                    println!(
-                        "{:<20} {:<30} {:<15} {:<10} {:<10} {:<10}",
-                        a.name,
-                        a.description,
-                        a.organization,
-                        parent,
-                        a.fairshare_weight as u32,
-                        a.grp_tres,
-                    );
-                }
             }
             Ok(())
         }
@@ -858,6 +846,55 @@ async fn show(entity: &str, params: &[String], addr: &str) -> Result<()> {
 fn filter_qos_by_name(qos_list: &mut Vec<QosInfo>, filter: &str) {
     let names: Vec<&str> = filter.split(',').map(str::trim).collect();
     qos_list.retain(|q| names.iter().any(|n| n.eq_ignore_ascii_case(&q.name)));
+}
+
+const ACCOUNT_DEFAULT_FORMAT: &str = "%-20N %-30D %-15O %-10P %-10S %-10G";
+
+const ACCOUNT_ALL_FORMAT: &str = "%-20N %-30D %-15O %-10P %-10S %-10G %-10J";
+
+fn account_header(spec: char) -> &'static str {
+    match spec {
+        'N' => "Account",
+        'D' => "Descr",
+        'O' => "Org",
+        'P' => "Parent",
+        'S' => "Share",
+        'G' => "GrpTRES",
+        'J' => "MaxJobs",
+        _ => "?",
+    }
+}
+
+fn account_field_spec(name: &str) -> Option<char> {
+    match name.to_lowercase().as_str() {
+        "account" | "name" => Some('N'),
+        "description" | "descr" => Some('D'),
+        "organization" | "org" => Some('O'),
+        "parent" | "parentaccount" => Some('P'),
+        "share" | "fairshare" | "fairshareweight" => Some('S'),
+        "grptres" => Some('G'),
+        "maxjobs" | "maxrunningjobs" => Some('J'),
+        _ => None,
+    }
+}
+
+fn resolve_account_field(a: &AccountInfo, spec: char) -> String {
+    match spec {
+        'N' => a.name.clone(),
+        'D' => a.description.clone(),
+        'O' => a.organization.clone(),
+        'P' => a.parent_account.clone(),
+        'S' => (a.fairshare_weight as u32).to_string(),
+        'G' => a.grp_tres.clone(),
+        'J' => {
+            if a.max_running_jobs == 0 {
+                String::new()
+            } else {
+                a.max_running_jobs.to_string()
+            }
+        }
+        _ => String::new(),
+    }
 }
 
 const QOS_DEFAULT_FORMAT: &str = "%-15N %-8p %-10P %-12U %-10J %-10S %-10W %-10w %-20T %-20V %-20G";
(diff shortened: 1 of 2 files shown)


tests/native_host/e2e/test_accounting.py
```diff
@@ -132,6 +132,34 @@ def test_show_qos_renders_all_limit_columns(self, accounting_cluster):
...
+        c = accounting_cluster
+        c.sacctmgr(["add", "account", "name=physics", "description=Physics",
+                     "organization=sciences", "grptres=cpu=16"])
+        time.sleep(15)
+        out = c.sacctmgr(["show", "account"])
+        header = next(line for line in out.splitlines() if line.strip())
+        for column in ("Account", "Descr", "Org", "Parent", "Share", "GrpTRES"):
+            assert column in header, f"missing column {column!r}: {header!r}"
+        assert "physics" in out
+
```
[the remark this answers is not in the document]
yansun1996: Good idea — added `test_empty_account_list_is_header_only`, which runs `show account` before any account exists and asserts header-only output: the header columns are present, `(no accounts configured)` is absent, and there are no data rows. That locks down both regressions you mentioned.

tests/native_host/e2e/test_accounting.py
```diff
@@ -132,6 +132,45 @@ def test_show_qos_renders_all_limit_columns(self, accounting_cluster):
...
+    def test_empty_account_list_is_header_only(self, accounting_cluster):
+        """No accounts prints the header only, not a placeholder or blank row."""
+        c = accounting_cluster
+        out = c.sacctmgr(["show", "account"])
+        rows = [line for line in out.splitlines() if line.strip()]
+        assert rows, f"expected a header line: {out!r}"
+        for column in ("Account", "Descr", "Org"):
+            assert column in rows[0], f"missing column {column!r}: {rows[0]!r}"
+        assert "no accounts configured" not in out
+        assert rows[1:] == [], f"expected no data rows: {rows[1:]!r}"
```
shiv-tyagi: This test can never pass. `format_engine::print_header` prints the header plus a dashed separator line, so `rows` is always `[header, "-------  -----  ..."]` and `rows[1:]` is never empty.

Verified locally against the account default format:

```
LINE0[Account              Descr                          Org             Parent     Share      GrpTRES]
LINE1[-------              -----                          ---             ------     -----      -------]
```

```suggestion
        rows = [line for line in out.splitlines() if line.strip()]
        assert len(rows) >= 2, f"expected header and separator: {out!r}"
        header, separator, *data_rows = rows
        for column in ("Account", "Descr", "Org"):
            assert column in header, f"missing column {column!r}: {header!r}"
        assert "no accounts configured" not in out
        assert data_rows == [], f"expected no data rows: {data_rows!r}"
```

tests/native_host/e2e/test_accounting.py
```diff
@@ -132,6 +132,45 @@ def test_show_qos_renders_all_limit_columns(self, accounting_cluster):
...
 
+class TestSacctmgrShowAccount:
+    """Verify sacctmgr show account honors format= field selection, the same
+    way show qos does."""
+
+    def test_default_output_shows_legacy_columns(self, accounting_cluster):
+        c = accounting_cluster
+        c.sacctmgr(["add", "account", "name=physics", "description=Physics",
+                     "organization=sciences", "grptres=cpu=16"])
+        time.sleep(15)
```
shiv-tyagi: `ListAccounts` reads Postgres directly in the request path (`crates/spurctld/src/accounting/grpc.rs`), so there is nothing to wait on after `sacctmgr add account`. This sleep and the one on L154 add 30s of wall clock for no reason.

The sleeps in the qos tests target the controller's QOS cache refresh loop. `show account` never touches it.

crates/spur-cli/src/sacctmgr.rs
```diff
@@ -1457,4 +1494,149 @@ mod tests {
...
+    ) -> anyhow::Result<Vec<format_engine::FormatToken>> {
+        format_engine::resolve_format(
+            fmt,
+            ACCOUNT_DEFAULT_FORMAT,
+            ACCOUNT_ALL_FORMAT,
+            &account_field_spec,
+            &account_header,
+            "Account, Descr, Org, Parent, Share, GrpTRES, MaxJobs",
+        )
+    }
```
shiv-tyagi: This helper re-implements the call site in `show()`, including the `valid_hint` literal. Change the default format or the hint in `show()` and every test here still passes.

Extract the wiring into production code and call it from both:

```rust
fn account_format_fields(format_param: Option<&str>) -> Result<Vec<format_engine::FormatToken>> {
    format_engine::resolve_format(
        format_param,
        ACCOUNT_DEFAULT_FORMAT,
        ACCOUNT_ALL_FORMAT,
        &account_field_spec,
        &account_header,
        "Account, Descr, Org, Parent, Share, GrpTRES, MaxJobs",
    )
}
```

`show()` becomes `let fields = account_format_fields(p.get("format").map(String::as_str))?;` and these tests drop the local helper.

crates/spur-cli/src/sacctmgr.rs
```diff
@@ -860,6 +848,55 @@ fn filter_qos_by_name(qos_list: &mut Vec<QosInfo>, filter: &str) {
...
+        'P' => a.parent_account.clone(),
+        'S' => (a.fairshare_weight as u32).to_string(),
+        'G' => a.grp_tres.clone(),
+        'J' => {
+            if a.max_running_jobs == 0 {
+                String::new()
+            } else {
+                a.max_running_jobs.to_string()
+            }
+        }
```
shiv-tyagi: The blank-if-zero pattern is now in five places in this file (four in `resolve_qos_field`, one here). Extract it:

```rust
fn blank_if_zero(v: u32) -> String {
    if v == 0 {
        String::new()
    } else {
        v.to_string()
    }
}
```

```suggestion
        'J' => blank_if_zero(a.max_running_jobs),
```

crates/spur-cli/src/sacctmgr.rs
```diff
@@ -860,6 +848,55 @@ fn filter_qos_by_name(qos_list: &mut Vec<QosInfo>, filter: &str) {
...
+        _ => String::new(),
```
shiv-tyagi: Nit: this falls back to `String::new()` while `resolve_qos_field` and `account_header` both fall back to `"?"`. A spec added to `account_header` but missed here renders as a silently blank column instead of a visible `?`.

```suggestion
        _ => "?".into(),
```

crates/spur-cli/src/sacctmgr.rs
```diff
@@ -707,6 +707,15 @@ async fn show(entity: &str, params: &[String], addr: &str) -> Result<()> {
...
     match entity.to_lowercase().as_str() {
         "account" | "accounts" => {
+            let fields = format_engine::resolve_format(
+                p.get("format").map(String::as_str),
+                ACCOUNT_DEFAULT_FORMAT,
+                ACCOUNT_ALL_FORMAT,
+                &account_field_spec,
+                &account_header,
+                "Account, Descr, Org, Parent, Share, GrpTRES, MaxJobs",
+            )?;
```
shiv-tyagi: Q: `show qos` honors `name=` as a filter, but `show account` still ignores every param except `format=`, so `sacctmgr show account where name=physics` lists all accounts. Deliberately out of scope here?

shiv-tyagi (changes requested):
posted comments PTAL

tests/native_host/e2e/test_accounting.py
yansun1996: Fixed — the assertion now unpacks `header, separator, *data_rows` and checks `data_rows == []`, accounting for the separator line `print_header` emits. This was the native-host e2e failure; verified the empty case now prints header + separator only.

tests/native_host/e2e/test_accounting.py
yansun1996: Removed both sleeps — right, `ListAccounts` reads Postgres directly so there's nothing to wait on after `add account`. The QOS-test sleeps stay since those wait on the controller's cache-refresh loop.

crates/spur-cli/src/sacctmgr.rs
yansun1996: Extracted `account_format_fields` into production code and pointed both `show()` and the tests at it, so the default/all formats and the valid-field hint live in one place.

crates/spur-cli/src/sacctmgr.rs
yansun1996: Extracted `blank_if_zero(u32)` and applied it to all five sites (four in `resolve_qos_field`, one here).

crates/spur-cli/src/sacctmgr.rs
yansun1996: Changed the fallback to `"?"` to match `resolve_qos_field` and `account_header`, and added a test so a header spec missing a resolver renders a visible `?` rather than a silent blank.

crates/spur-cli/src/sacctmgr.rs
yansun1996: Good catch — `show account` does ignore `where name=` today (it fetches all and only applies `format=`). It's a distinct feature from this format= change (likely wants server-side filtering, since `ListAccountsRequest` has no name field yet), so I've filed it as a follow-up: ROCm/spur#536.

crates/spur-cli/src/sacctmgr.rs
```diff
@@ -860,6 +841,73 @@ fn filter_qos_by_name(qos_list: &mut Vec<QosInfo>, filter: &str) {
...
+        "maxjobs" | "maxrunningjobs" => Some('J'),
+        _ => None,
+    }
+}
+
+/// Resolve `show account` output columns from an optional `format=` param. Shared by the handler
+/// and its tests so the default/all formats and the valid-field hint stay in one place.
+fn account_format_fields(
+    format_param: Option<&str>,
+) -> anyhow::Result<Vec<format_engine::FormatToken>> {
```
shiv-tyagi, line 878: Nit: `Result` is already imported at the top of this file and every other signature here uses the bare alias.

```suggestion
) -> Result<Vec<format_engine::FormatToken>> {
```

shiv-tyagi (approved):
All six comments addressed, verified against d427032. `cargo test -p spur-cli` passes (245, up one for the new fallback test) and `cargo clippy -p spur-cli --all-targets` is clean.

- The empty-list assertion now unpacks `header, _separator, *data_rows`, which was the real fix for the native-host e2e failure.
- `account_format_fields` in production code removes the duplicated format/hint wiring, and the tests now exercise the same path as `show()`.
- `blank_if_zero` applied to all five sites, not just the one I flagged.
- The `"?"` fallback is pinned by a test, so the silent-blank case can't regress.
- Dropping the two sleeps takes 30s off the e2e suite, and keeping the qos ones is correct since those wait on the controller's cache-refresh loop.
- `where name=` split out to ROCm/spur#536 is the right call for this PR.

One non-blocking nit inline.

PR #522, opened 2026-07-28
fix(spur-cli): default sbatch --ntasks to one task per node

Nikhil-Nunna:
## What

`sbatch` gave `--ntasks` a fixed clap default of `1`, independent of `--nodes`. A multi-node job submitted without an explicit `-n` was therefore recorded with a single task for the whole allocation, so `SLURM_NTASKS` and the derived `SLURM_TASKS_PER_NODE` both came out as `1` instead of one task per node. Tools that read those variables to work out per-node task counts (MPI launchers in particular) then lay ranks out incorrectly.

## Approach

`SbatchArgs::ntasks` becomes `Option<u32>` with the static default dropped, and the effective value is resolved when the job spec is built, falling back to the requested node count. A clap `default_value` cannot reference another argument, so the fallback has to happen after parsing rather than in the derive attribute.

An explicit `-n`, an `SBATCH_*` env value, or a `#SBATCH --ntasks` directive all still take precedence: the CLI > env > directive merge in `resolve_sbatch_args` is unchanged, since optional fields already flow through it the same way `--ntasks-per-node` does.

This also fixes the per-node CPU estimate in backfill, which derives `num_tasks / num_nodes` and previously integer-divided to zero CPUs per node for a multi-node job left at the default.

Deliberately out of scope: `--ntasks-per-node=K` with no `-n` still resolves to `num_tasks = nodes` rather than `nodes * K`. That is a separate default with wider reach (it feeds memory-per-CPU and accounting totals), so it is left for a follow-up rather than folded in here.

## Testing

`cargo test -p spur-cli` passes (236 tests), including three new ones covering the node-count default, an explicit `-n` overriding it, and a `#SBATCH --ntasks` directive still being honoured. Also run: the full `cargo test --workspace --exclude spur-ffi`, `cargo fmt --all --check`, and `cargo clippy -p spur-cli --all-targets -- -D warnings`, all clean.

### DIFF
--- a/crates/spur-cli/src/sbatch.rs
+++ b/crates/spur-cli/src/sbatch.rs
@@ -1057,6 +1057,33 @@ echo "hello world"
         assert_eq!(effective_ntasks(args.ntasks, args.nodes), 8);
     }
 
+    #[tokio::test]
+    #[serial(env_injection)]
+    async fn test_submit_path_builds_spec_with_defaulted_ntasks() {
+        // Guarded because a stray SBATCH_* var in the runner's environment
+        // could fail the parse long before the spec is built.
+        let _env = SbatchEnvGuard::new();
+        let err = main_with_args(vec![
+            "sbatch".into(),
+            "--controller".into(),
+            "http://127.0.0.1:1".into(),
+            "--export".into(),
+            "NONE".into(),
+            "-N".into(),
+            "4".into(),
+            "--wrap".into(),
+            "echo hi".into(),
+        ])
+        .await
+        .unwrap_err();
+        // Failing at the connect means the whole JobSpec was built first,
+        // including num_tasks defaulting to the node count.
+        assert!(
+            err.to_string().contains("failed to connect to spurctld"),
+            "expected a connect failure, got: {err}"
+        );
+    }
+
     #[test]
     fn test_vec_args_accumulate_from_both_sources() {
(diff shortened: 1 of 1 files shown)


yansun1996 (approved):
Nice fix — `-N` now scales the task default correctly, and it also quietly fixes the backfill per-node CPU estimate (the old `num_tasks/num_nodes` integer-divided to 0 CPUs/node for a defaulted multi-node job). Precedence (CLI > env > directive) is preserved and the `effective_ntasks` helper reads cleanly. Approving.

Three small things, none blocking:

1. `--ntasks-per-node=K` with no `-n` still resolves `num_tasks` to `nodes` rather than `nodes * K`. The job does run the right number of tasks, but `SLURM_NTASKS`/accounting report `nodes` — worth a test documenting the current value so it's visible, even though the full fix is deferred.
2. `test_submit_path_builds_spec_with_defaulted_ntasks` only asserts the connect failure; it never checks `num_tasks`, so it would pass even if the default regressed. Consider strengthening it (assert the built spec's `num_tasks`) or dropping it, since the `effective_ntasks` unit tests already cover the logic.
3. Follow-up only: `srun`/`salloc` still default `--ntasks` to a fixed `1`, so they now diverge from `sbatch` for `-N`. Might be worth a linked issue.

PR #524, opened 2026-07-28
fix(spur-cli): sort squeue output and honor -S/--sort

### LINKED ISSUE #525: squeue does not sort output and ignores -S/--sort
## What happened

`squeue` prints jobs in an apparently arbitrary order, and the `-S`/`--sort` option has no effect.

Submitting six jobs (IDs 70–75) and running `squeue` yields output ordered like:

```
70 75 74 71 72 73
```

The order is not by job ID, nor by any other visible column. Adding `-S i` (or any other sort field) does not change the ordering. Attempting a descending sort fails to parse entirely:

```
$ squeue -S -i
error: unexpected argument '-i' found
```

## What was expected

Slurm's `squeue`:

- Sorts by a default field order when `-S` is omitted (documented default is `P,t,-p` — partition, then state, then descending priority).
- Honors `-S <fields>` with comma-separated fields and optional `+`/`-` direction prefixes, e.g. `-S i`, `-S -i`, `-S P,i`.

Spur currently does neither: the default output is unsorted, and `-S` is accepted but has no effect.

## How to reproduce

1. Submit several jobs so more than one appears in the queue.
2. Run `squeue` — observe the rows are not in job-ID order (or any consistent order).
3. Run `squeue -S i` — observe the ordering is unchanged.
4. Run `squeue -S -i` — observe it errors instead of sorting descending.

yansun1996:
Closes #525

## What this fixes

`squeue` had three sorting gaps versus Slurm:

1. **No ordering.** Jobs were printed in the controller's internal `HashMap` iteration order — effectively unsorted, and unstable across restarts.
2. **`-S`/`--sort` was dead.** The flag parsed into a field that nothing ever read, so `-S i` had no effect.
3. **Descending sort wouldn't parse.** `squeue -S -i` failed with `unexpected argument '-i'` because the arg rejected hyphen-prefixed values.

## Approach

Sort **client-side** in `squeue.rs`, matching Slurm's architecture (Slurm's `squeue` sorts locally; every field needed is already on `JobInfo`). This keeps the controller a pure data source and avoids coupling display ordering to the core.

- Parse `-S`/`--sort` as comma-separated fields, each optionally prefixed `+` (ascending, default) or `-` (descending), reusing the existing format-spec letters.
- Apply Slurm's documented default sort `P,t,-p` (partition asc, job-state asc, priority desc) when `-S` is absent, with a trailing job-id tiebreak so equal-priority jobs order deterministically.
- Add `allow_hyphen_values` to the `-S` arg so `-S -i` parses.
- Comparator sorts numerically for numeric fields (job id, priority, nodes, cpus), by epoch seconds for timestamps, by Slurm's base state order for `t`/`T` (so SUSPENDED sorts right after RUNNING, ahead of terminal states), and lexically otherwise.
- Unrecognized sort fields are rejected with an `Invalid sort specification` error.

## Known limitations / deviations

- **Unknown sort field handling is stricter than Slurm.** Slurm silently ignores a genuinely-unknown sort letter (no-op) yet errors on some *format-only* fields; the behavior is inconsistent. This change instead accepts every field Spur can meaningfully sort and returns a clear error for anything unrecognized.
- **`-S` consumes the following token** (a consequence of allowing hyphen-prefixed values). Use the attached form for descending sorts — `-S -i` or `--sort=-i` — not `-S` split from an unrelated following flag.
- Non-sortable/heterogeneous string fields (e.g. the reason/nodelist column) sort lexically on their rendered text.

## Testing

- Unit tests cover sort-spec parsing (direction prefixes, empty/malformed/unknown rejection), ascending/descending job-id, the default `P,t,-p` ordering, priority-desc with job-id tiebreak, state ordering (SUSPENDED placement), time-limit (UNLIMITED sorts last), and timestamp sort.
- `cargo clippy` clean; full `spur-cli` suite passes.
- Validated end-to-end on an isolated multi-node deployment: confirmed the default view is partition-grouped and deterministic, `-S i` / `-S -i` / multi-key `-S P,i` order correctly, unknown fields error, and outcomes match Slurm for the default sort and the `-S` variants.

### DIFF
--- a/crates/spur-cli/src/squeue.rs
+++ b/crates/spur-cli/src/squeue.rs
@@ -47,8 +47,8 @@ pub struct SqueueArgs {
     #[arg(short = 'h', long)]
     pub noheader: bool,
 
-    /// Sort by field
-    #[arg(short = 'S', long)]
+    /// Sort by field(s), comma-separated, each optionally prefixed with + (asc) or - (desc)
+    #[arg(short = 'S', long, allow_hyphen_values = true)]
     pub sort: Option<String>,
 
     /// Controller address
@@ -113,7 +113,15 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
         .await
         .context("failed to get jobs")?;
 
-    let jobs = response.into_inner().jobs;
+    let mut jobs = response.into_inner().jobs;
+
+    // Slurm sorts client-side; its default job sort is "P,t,-p". The trailing
+    // jobid tiebreak is ours, so equal-priority jobs order deterministically.
+    let sort_keys = match args.sort.as_deref() {
+        Some(s) => parse_sort_arg(s)?,
+        None => default_sort_keys(),
+    };
+    sort_jobs(&mut jobs, &sort_keys);
 
     // Print header
     if !args.noheader {
@@ -171,6 +179,132 @@ fn resolve_job_field(job: &spur_proto::proto::JobInfo, spec: char) -> String {
         }
         _ => "?".into(),
     }
+}
+
+/// One field in a sort spec, with direction.
+#[derive(Debug, Clone, Copy, PartialEq)]
+struct SortKey {
+    spec: char,
+    descending: bool,
+}
+
+/// Slurm's documented default job sort: partition asc, state asc, priority desc.
+/// A trailing jobid asc keeps equal-priority jobs deterministically ordered.
+fn default_sort_keys() -> Vec<SortKey> {
+    vec![
+        SortKey {
+            spec: 'P',
+            descending: false,
+        },
+        SortKey {
+            spec: 't',
+            descending: false,
+        },
+        SortKey {
+            spec: 'p',
+            descending: true,
+        },
+        SortKey {
+            spec: 'i',
+            descending: false,
+        },
+    ]
+}
+
+/// Parse `-S` / `--sort`: comma-separated fields, each optionally prefixed with
+/// `+` (asc, default) or `-` (desc). Reuses the format-spec letters.
+fn parse_sort_arg(s: &str) -> Result<Vec<SortKey>> {
+    let mut keys = Vec::new();
+    for token in s.split(',').map(str::trim).filter(|t| !t.is_empty()) {
+        let (descending, rest) = match token.strip_prefix('-') {
+            Some(r) => (true, r),
+            None => (false, token.strip_prefix('+').unwrap_or(token)),
+        };
+        let mut chars = rest.chars();
+        let spec = chars
+            .next()
+            .filter(|_| chars.next().is_none())
+            .ok_or_else(|| anyhow::anyhow!("Invalid sort specification: {token}"))?;
+        if format_engine::squeue_header(spec) == "?" {
+            anyhow::bail!("Invalid sort specification: {token}");
+        }
+        keys.push(SortKey { spec, descending });
+    }
+    if keys.is_empty() {
+        anyhow::bail!("Invalid sort specification: (empty)");
+    }
+    Ok(keys)
+}
+
+fn sort_jobs(jobs: &mut [spur_proto::proto::JobInfo], keys: &[SortKey]) {
+    jobs.sort_by(|a, b| {
+        for key in keys {
+            let ord = compare_field(a, b, key.spec);
+            let ord = if key.descending { ord.reverse() } else { ord };
+            if ord != std::cmp::Ordering::Equal {
+                return ord;
+            }
+        }
+        std::cmp::Ordering::Equal
+    });
+}
+
+/// Compare two jobs on a single field. Numeric fields compare numerically;
+/// timestamps by epoch seconds; everything else lexically on the rendered value.
+fn compare_field(
+    a: &spur_proto::proto::JobInfo,
+    b: &spur_proto::proto::JobInfo,
+    spec: char,
+) -> std::cmp::Ordering {
+    match spec {
+        'i' | 'A' => a.job_id.cmp(&b.job_id),
+        'p' => a.priority.cmp(&b.priority),
+        'D' => a.num_nodes.cmp(&b.num_nodes),
+        'C' => a.cpus_per_task.cmp(&b.cpus_per_task),
+        't' | 'T' => state_sort_rank(a.state).cmp(&state_sort_rank(b.state)),
+        'M' => run_time_secs(a).cmp(&run_time_secs(b)),
+        'l' => time_limit_secs(a).cmp(&time_limit_secs(b)),
+        'S' => ts_secs(a.start_time.as_ref()).cmp(&ts_secs(b.start_time.as_ref())),
+        'V' => ts_secs(a.submit_time.as_ref()).cmp(&ts_secs(b.submit_time.as_ref())),
+        'e' => ts_secs(a.end_time.as_ref()).cmp(&ts_secs(b.end_time.as_ref())),
+        _ => resolve_job_field(a, spec).cmp(&resolve_job_field(b, spec)),
+    }
+}
+
+/// Rank a proto state by Slurm's base job-state order (SUSPENDED sits right
+/// after RUNNING, before the terminal states) so `-S t` matches Slurm.
+fn state_sort_rank(state: i32) -> u8 {
+    use spur_proto::proto::JobState as S;
+    match S::try_from(state) {
+        Ok(S::JobPending) => 0,
+        Ok(S::JobRunning) => 1,
+        Ok(S::JobSuspended) => 2,
+        Ok(S::JobCompleting) => 3,
+        Ok(S::JobCompleted) => 4,
+        Ok(S::JobCancelled) => 5,
+        Ok(S::JobFailed) => 6,
+        Ok(S::JobTimeout) => 7,
+        Ok(S::JobNodeFail) => 8,
+        Ok(S::JobPreempted) => 9,
+        Ok(S::JobDeadline) => 10,
+        Ok(S::JobOutOfMemory) => 11,
+        Err(_) => u8::MAX,
+    }
+}
+
+fn run_time_secs(job: &spur_proto::proto::JobInfo) -> i64 {
+    job.run_time.as_ref().map(|d| d.seconds).unwrap_or(0)
+}
+
+fn time_limit_secs(job: &spur_proto::proto::JobInfo) -> i64 {
+    job.time_limit
+        .as_ref()
+        .map(|d| d.seconds)
+        .unwrap_or(i64::MAX)
+}
+
+fn ts_secs(ts: Option<&prost_types::Timestamp>) -> i64 {
+    ts.map(|t| t.seconds).unwrap_or(0)
 }
 
 fn state_code(state: i32) -> String {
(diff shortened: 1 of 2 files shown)


crates/spur-cli/src/squeue.rs
```diff
@@ -173,6 +181,132 @@ fn resolve_job_field(job: &spur_proto::proto::JobInfo, spec: char) -> String {
...
+            None => (false, token.strip_prefix('+').unwrap_or(token)),
+        };
+        let mut chars = rest.chars();
+        let spec = chars
+            .next()
+            .filter(|_| chars.next().is_none())
+            .ok_or_else(|| anyhow::anyhow!("Invalid sort specification: {token}"))?;
+        if format_engine::squeue_header(spec) == "?" {
+            anyhow::bail!("Invalid sort specification: {token}");
+        }
```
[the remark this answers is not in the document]
yansun1996: Good catch — `L` was the one accepted spec with no comparator, so `-S L` silently compared everything equal. Added a `TIME_LEFT` comparator (time_limit − run_time, with unlimited sorting last) so it now orders meaningfully, plus a test for the ordering.

crates/spur-cli/src/squeue.rs
```diff
@@ -173,6 +181,132 @@ fn resolve_job_field(job: &spur_proto::proto::JobInfo, spec: char) -> String {
...
+        'D' => a.num_nodes.cmp(&b.num_nodes),
+        'C' => a.cpus_per_task.cmp(&b.cpus_per_task),
+        't' | 'T' => state_sort_rank(a.state).cmp(&state_sort_rank(b.state)),
+        'M' => run_time_secs(a).cmp(&run_time_secs(b)),
+        'l' => time_limit_secs(a).cmp(&time_limit_secs(b)),
+        'S' => ts_secs(a.start_time.as_ref()).cmp(&ts_secs(b.start_time.as_ref())),
+        'V' => ts_secs(a.submit_time.as_ref()).cmp(&ts_secs(b.submit_time.as_ref())),
+        'e' => ts_secs(a.end_time.as_ref()).cmp(&ts_secs(b.end_time.as_ref())),
+        _ => resolve_job_field(a, spec).cmp(&resolve_job_field(b, spec)),
+    }
```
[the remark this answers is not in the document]
yansun1996: Agreed. Added direct comparison arms for the common string specs (partition/user/name/account/qos) that borrow the proto fields instead of cloning through `resolve_job_field`, keeping the fallback only for computed/display-only specs. Added a partition-sort test to guard the change.

crates/spur-cli/src/squeue.rs
```diff
@@ -113,7 +113,15 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
...
-    let jobs = response.into_inner().jobs;
+    let mut jobs = response.into_inner().jobs;
+
+    // Slurm sorts client-side; its default job sort is "P,t,-p". The trailing
+    // jobid tiebreak is ours, so equal-priority jobs order deterministically.
+    let sort_keys = match args.sort.as_deref() {
+        Some(s) => parse_sort_arg(s)?,
+        None => default_sort_keys(),
+    };
+    sort_jobs(&mut jobs, &sort_keys);
```
shiv-tyagi, line 134: Sort parsing happens after the `get_jobs` round trip, so an invalid spec never surfaces its real error. With no controller reachable, `squeue -S x` reports `failed to connect to spurctld` instead of `Invalid sort specification: x`. Even against a live controller, a typo costs a full fetch before failing.

Hoist the parse next to the state parsing at L82, which already validates before any network I/O:

```rust
let sort_keys = match args.sort.as_deref() {
    Some(s) => parse_sort_arg(s)?,
    None => default_sort_keys(),
};
```

and leave only the application here:

```rust
let mut jobs = response.into_inner().jobs;
sort_jobs(&mut jobs, &sort_keys);
```

crates/spur-cli/src/squeue.rs
```diff
@@ -173,6 +181,147 @@ fn resolve_job_field(job: &spur_proto::proto::JobInfo, spec: char) -> String {
...
+        Ok(S::JobCancelled) => 5,
+        Ok(S::JobFailed) => 6,
+        Ok(S::JobTimeout) => 7,
+        Ok(S::JobNodeFail) => 8,
+        Ok(S::JobPreempted) => 9,
+        Ok(S::JobDeadline) => 10,
+        Ok(S::JobOutOfMemory) => 11,
+        Err(_) => u8::MAX,
+    }
+}
```
shiv-tyagi: This is a fourth copy of the state list, and it matches on the proto enum directly even though `state_code` and `state_name` just below both route through `spur_core::job::JobState`. Sort order is a property of the state, so it belongs next to `code()` and `display()` in `spur-core/src/job.rs`:

```rust
/// Sort rank for `squeue -S t`, following Slurm's base job-state order.
pub fn sort_rank(&self) -> u8 {
    match self {
        Self::Pending => 0,
        Self::Running => 1,
        Self::Suspended => 2,
        Self::Completing => 3,
        // ...
    }
}
```

Then this collapses to:

```rust
fn state_sort_rank(state: i32) -> u8 {
    spur_core::job::JobState::from_proto_i32(state)
        .map(|s| s.sort_rank())
        .unwrap_or(u8::MAX)
}
```

crates/spur-cli/src/squeue.rs
```diff
@@ -173,6 +181,147 @@ fn resolve_job_field(job: &spur_proto::proto::JobInfo, spec: char) -> String {
...
 
+/// One field in a sort spec, with direction.
+#[derive(Debug, Clone, Copy, PartialEq)]
+struct SortKey {
+    spec: char,
+    descending: bool,
+}
+
+/// Slurm's documented default job sort: partition asc, state asc, priority desc.
+/// A trailing jobid asc keeps equal-priority jobs deterministically ordered.
```
shiv-tyagi, line 202: Nit: the call-site comment at L118-119 restates this doc comment verbatim (Slurm default `P,t,-p`, jobid tiebreak is ours). Keep this one, drop the call-site copy.

crates/spur-cli/src/squeue.rs
```diff
@@ -173,6 +181,147 @@ fn resolve_job_field(job: &spur_proto::proto::JobInfo, spec: char) -> String {
...
+    for token in s.split(',').map(str::trim).filter(|t| !t.is_empty()) {
+        let (descending, rest) = match token.strip_prefix('-') {
+            Some(r) => (true, r),
+            None => (false, token.strip_prefix('+').unwrap_or(token)),
+        };
+        let mut chars = rest.chars();
+        let spec = chars
+            .next()
+            .filter(|_| chars.next().is_none())
+            .ok_or_else(|| anyhow::anyhow!("Invalid sort specification: {token}"))?;
```
shiv-tyagi: Nit: mutating `chars` inside the `filter` predicate hides the "exactly one char" intent behind a side effect.

```suggestion
        let mut chars = rest.chars();
        let spec = match (chars.next(), chars.next()) {
            (Some(c), None) => c,
            _ => anyhow::bail!("Invalid sort specification: {token}"),
        };
```

crates/spur-cli/src/squeue.rs
```diff
@@ -47,8 +47,8 @@ pub struct SqueueArgs {
     #[arg(short = 'h', long)]
     pub noheader: bool,
 
-    /// Sort by field
-    #[arg(short = 'S', long)]
+    /// Sort by field(s), comma-separated, each optionally prefixed with + (asc) or - (desc)
+    #[arg(short = 'S', long, allow_hyphen_values = true)]
     pub sort: Option<String>,
```
shiv-tyagi, line 62: `allow_hyphen_values` is the fix for the `squeue -S -i` parse failure, but nothing tests it. The parser tests all call `parse_sort_arg` directly, so they would still pass if this attribute were dropped.

One prerequisite: `SqueueArgs::try_parse_from` panics on debug builds because `-h` is claimed by both `noheader` and the auto-generated `--help`. With `#[command(disable_help_flag = true)]` on the struct:

```rust
#[test]
fn parses_descending_sort_flag() {
    let args = SqueueArgs::try_parse_from(["squeue", "-S", "-i"]).unwrap();
    assert_eq!(args.sort.as_deref(), Some("-i"));
}
```

crates/spur-cli/src/squeue.rs
```diff
@@ -173,6 +181,147 @@ fn resolve_job_field(job: &spur_proto::proto::JobInfo, spec: char) -> String {
...
+        .as_ref()
+        .map(|d| d.seconds)
+        .unwrap_or(i64::MAX)
+}
+
+fn ts_secs(ts: Option<&prost_types::Timestamp>) -> i64 {
+    ts.map(|t| t.seconds).unwrap_or(0)
+}
+
+// Unlimited time limit sorts last, matching `-S l`.
```
shiv-tyagi: Nit: this is the `L` (TIME_LEFT) helper, not `l` (TIME_LIMIT). Also `///` to match the doc comments on the neighbouring helpers.

```suggestion
/// Unlimited time limit sorts last, matching `-S L`.
```

shiv-tyagi (commented):
Client-side sort matching Slurm's architecture is the right call, and so is the default `P,t,-p` with a jobid tiebreak.

Verified locally: `cargo test -p spur-cli` passes (248 tests), `cargo clippy -p spur-cli --all-targets` is clean, and `-S -i` / `--sort=-i` both parse against the built binary.

One thing worth fixing before merge: the sort spec is validated after the gRPC fetch, so `squeue -S <typo>` reports a connection error instead of the invalid-spec error. The rest is a refactor suggestion, a missing test, and three nits.

crates/spur-cli/src/squeue.rs
yansun1996: Hoisted the sort parse up next to the state parse, before the connect — so `squeue -S <bad>` now reports `Invalid sort specification` immediately instead of a connect failure, and a typo no longer costs a fetch. Only `sort_jobs` remains after the response.

crates/spur-cli/src/squeue.rs
yansun1996: Moved the ranking into `JobState::sort_rank()` in spur-core (next to `code()`/`display()`) and collapsed `state_sort_rank` to route through it. Kept it an explicit match since the enum's declaration order differs from Slurm's sort order, and added a test guarding that.

crates/spur-cli/src/squeue.rs
yansun1996: Dropped the duplicated call-site comment; kept the one on `default_sort_keys`.

crates/spur-cli/src/squeue.rs
yansun1996: Switched to matching `(chars.next(), chars.next())` so the "exactly one char" intent is explicit rather than a filter side effect.

crates/spur-cli/src/squeue.rs
yansun1996: Added `disable_help_flag = true` (squeue's `-h` is `--noheader` by Slurm convention) and a `try_parse_from(["squeue","-S","-i"])` test that exercises `allow_hyphen_values` through the real parser.

crates/spur-cli/src/squeue.rs
yansun1996: Fixed — corrected to `-S L` (TIME_LEFT) and made it a `///` doc comment to match the neighbours.

crates/spur-cli/src/squeue.rs
```diff
@@ -9,7 +9,12 @@ use crate::format_engine;
 
 /// View information about jobs in the scheduling queue.
 #[derive(Parser, Debug)]
-#[command(name = "squeue", about = "View the job queue")]
+// -h is squeue's --noheader (Slurm convention), so disable clap's auto -h/--help.
+#[command(
+    name = "squeue",
+    about = "View the job queue",
+    disable_help_flag = true
+)]
```
shiv-tyagi: `disable_help_flag` drops `--help` too, not just `-h`. `squeue --help` now fails:

```
$ squeue --help
error: unexpected argument '--help' found
```

Slurm binds `-h` to `--noheader` but keeps `--help` working, and every other command in this CLI still prints help. Re-add it as long-only:

```rust
/// Print help
#[arg(long, action = clap::ArgAction::Help)]
pub help: Option<bool>,
```

I verified this locally: `--help` prints the full help, `-h` still means `--noheader`, `-S -i` still parses, and tests plus clippy stay clean.

Worth narrowing the comment above too, since with this only the short `-h` is reclaimed and `--help` stays.

shiv-tyagi (changes requested):
All six points from the last round are addressed. Verified against `f286190`:

- Sort parse hoisted above the network call: `squeue -S x` now prints `Invalid sort specification: x` and `-S ""` prints `Invalid sort specification: (empty)`, neither touching the controller.
- `state_sort_rank` delegates to `JobState::sort_rank()` in spur-core, and the uniqueness-across-`ALL` assertion in the new test is a nice touch.
- Duplicate comment gone, tuple match applied, `-S L` comment corrected.
- The clap test now runs `try_parse_from` under debug assertions without panicking, so the `-h` fix is real.

`cargo test -p spur-cli` 249 passing, `-p spur-core` 389 passing, `cargo clippy` clean on both.

One regression to fix before merge: `disable_help_flag` also removed `--help`, so `squeue --help` no longer works. Details inline, with a fix I verified locally.

Separately, and outside this PR: `sinfo`, `sprio`, and `sshare` carry the identical `-h` collision and still panic on every invocation in a debug build (`Command sinfo: Short option names must be unique for each argument, but '-h' is in use by both 'noheader' and 'help'`). Release builds compile the assert out, which is why CI never caught it. Not yours to fix here, but this PR does leave `squeue` as the only one of the four that works. Happy to file an issue if you'd rather keep it separate.

crates/spur-cli/src/squeue.rs
yansun1996: Good catch — re-added `--help` as a long-only arg (`ArgAction::Help`), so `-h` stays `--noheader` while `--help` prints help again, matching the other commands. Narrowed the comment to say only `-h` is reclaimed. Added two tests: `--help` returns `DisplayHelp` and `-h` still sets `noheader`. Verified `--help`, `-h`, `-S -i`, and `-S <bad>` all behave on the built binary.

On the siblings: `sinfo` already got this fix on `main` (just not on this branch yet), but `sprio` and `sshare` still carry the collision. Filed a follow-up: #539. Thanks for flagging.

yansun1996:
> All six points from the last round are addressed. Verified against `f286190`:
> 
> * Sort parse hoisted above the network call: `squeue -S x` now prints `Invalid sort specification: x` and `-S ""` prints `Invalid sort specification: (empty)`, neither touching the controller.
> * `state_sort_rank` delegates to `JobState::sort_rank()` in spur-core, and the uniqueness-across-`ALL` assertion in the new test is a nice touch.
> * Duplicate comment gone, tuple match applied, `-S L` comment corrected.
> * The clap test now runs `try_parse_from` under debug assertions without panicking, so the `-h` fix is real.
> 
> `cargo test -p spur-cli` 249 passing, `-p spur-core` 389 passing, `cargo clippy` clean on both.
> 
> One regression to fix before merge: `disable_help_flag` also removed `--help`, so `squeue --help` no longer works. Details inline, with a fix I verified locally.
> 
> Separately, and outside this PR: `sinfo`, `sprio`, and `sshare` carry the identical `-h` collision and still panic on every invocation in a debug build (`Command sinfo: Short option names must be unique for each argument, but '-h' is in use by both 'noheader' and 'help'`). Release builds compile the assert out, which is why CI never caught it. Not yours to fix here, but this PR does leave `squeue` as the only one of the four that works. Happy to file an issue if you'd rather keep it separate.

Thanks for the suggestion, I've fixed the `disable_help_flag` and then filed separate issues to track the `sprio` and `sshare` side of issues mentioned above.

shiv-tyagi (approved):
`--help` is restored and the two added tests pin down both halves of the `-h` / `--help` split. All points addressed.

PR #526, opened 2026-07-28
fix(spur-cli): honor SPUR_* env-var twins in sbatch

### LINKED ISSUE #448: sbatch ignores SPUR_* env-var twins that srun/salloc honor
## What happens

`sbatch` resolves environment-variable defaults through clap's single `#[arg(env = "SBATCH_*")]` attribute — one env var per flag. As a result, the native `SPUR_*` twins are silently ignored by `sbatch`.

For example, `SPUR_PARTITION` sets the default partition for `srun` and `salloc` but has no effect on `sbatch`.

## What was expected

`SPUR_*` env vars should behave consistently across the submission CLIs. If `SPUR_PARTITION` is honored by `srun`/`salloc`, a user reasonably expects it to work for `sbatch` too.

## How to reproduce

```bash
export SPUR_PARTITION=gpu

srun ...     # picks up partition=gpu
salloc ...   # picks up partition=gpu
sbatch ...   # partition NOT set from SPUR_PARTITION (silently ignored)
```

Any fix needs to preserve sbatch's existing three-tier precedence (CLI > env > #SBATCH directive) and its directive-merge logic

maybeharshit:
Fixes #448.

`SPUR_PARTITION` sets the default partition for `srun` and `salloc` but had no effect on `sbatch`. The cause is that `sbatch` declared each env default with clap's `#[arg(env = "SBATCH_...")]` attribute, which accepts exactly one variable name per flag. `srun`/`salloc` outgrew that in #421, which moved them to a shared resolver that walks an ordered list of candidate names so each flag honors both the Slurm-style name and Spur's native `SPUR_*` twin. `sbatch` never moved, so its twins were silently ignored.

## Approach

`sbatch` now resolves env defaults through the same `env_defaults` helpers, with the `SPUR_*` twin listed ahead of the `SBATCH_*` name, matching the ordering `srun`/`salloc` already use. Twin names reuse the ones those commands accept (`SPUR_MEM_PER_NODE`, `SPUR_TIMELIMIT`) rather than introducing parallel spellings.

`sbatch` has a third precedence tier the others do not: CLI > env > `#SBATCH` directive. That previously worked only because clap tags env-sourced values as `ValueSource::EnvVariable`, which the directive merge treated as already-set. With env resolved outside clap, `resolve_sbatch_args` applies env *after* the CLI-over-directive merge and only to flags the CLI did not set, so each field lands on exactly one tier in that order. `--controller` keeps its clap `env` attribute, since `SPUR_CONTROLLER_ADDR` is already a native name and this matches how `srun`/`salloc` declare the same flag.

## Design choices

`SPUR_JOB_NAME` and `SPUR_NODELIST` are deliberately **not** read. Both are injected into every running job's environment by `spurd`, `salloc`, and the K8s agent, so honoring them would make an `sbatch` submitted from inside a job silently inherit the enclosing job's name and node placement. Real Slurm's `sbatch` inherits neither. Those two flags keep their `SBATCH_*` name only, and a test pins the behavior. Every other previously env-backed flag gains its twin.

`--nodelist` and `--nodefile` are two spellings of one choice, so an env nodelist has to clear a directive nodefile to take effect, and must stand down when either spelling came from the command line.

This also replaces `sbatch.rs`'s local `SbatchEnvGuard` test guard with the shared `EnvGuard`, now extended to the `SBATCH_` prefix. The local guard derived the variables to clear from the clap `env=` attributes, so removing those attributes would have left it clearing nothing and made the env tests read the runner's environment.

## Follow-up

Unrelated but noticed while mapping the variables: `--mem` reads `SBATCH_MEM`, whereas Slurm's name is `SBATCH_MEM_PER_NODE`, so a migrating user with the real name exported gets silence. Left out of scope here; worth its own issue.

## Testing

Nine tests added, covering the twins, `SPUR_*` taking precedence over `SBATCH_*`, env beating a directive but losing to the CLI, non-inheritance of the two allocation-output variables, and the nodelist/nodefile interaction.

- `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` — clean
- `cargo fmt --all -- --check` — clean
- `cargo test --locked` — passes, except six pre-existing `spurd` failures in my sandbox (cgroup creation and `openpty` denied), reproduced identically on pristine `main` with the change stashed.

### DIFF
--- a/crates/spur-cli/src/env_defaults.rs
+++ b/crates/spur-cli/src/env_defaults.rs
@@ -1,7 +1,7 @@
 // Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-//! Environment-variable defaults for `srun` and `salloc` flags.
+//! Environment-variable defaults for `srun`, `salloc`, and `sbatch` flags.
 //!
 //! clap's `#[arg(env = ...)]` accepts a single variable name, but Slurm
 //! defines several aliases per flag (e.g. `--nodes` reads `SLURM_NNODES` and
@@ -136,10 +136,10 @@ where
 
 /// Test-only guard that isolates env-var tests from the runner's environment.
 ///
-/// Clears every `SPUR_/SLURM_/SALLOC_/SRUN_`-prefixed variable on construction
-/// and on drop, so a CI runner that injects `SLURM_*` vars cannot perturb the
-/// resolvers and a panicking test cannot leak state into the next one. Tests
-/// using it must be `#[serial]`.
+/// Clears every `SPUR_/SLURM_/SALLOC_/SRUN_/SBATCH_`-prefixed variable on
+/// construction and on drop, so a CI runner that injects `SLURM_*` vars cannot
+/// perturb the resolvers and a panicking test cannot leak state into the next
+/// one. Tests using it must be `#[serial]`.
 #[cfg(test)]
 pub(crate) struct EnvGuard;
 
@@ -162,6 +162,7 @@ impl EnvGuard {
                     || k.starts_with("SLURM_")
                     || k.starts_with("SALLOC_")
                     || k.starts_with("SRUN_")
+                    || k.starts_with("SBATCH_")
             })
             .collect();
         for k in stale {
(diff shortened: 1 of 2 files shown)


yansun1996 (approved):
LGTM. This nicely finishes the migration that #421 started for srun/salloc — moving sbatch onto the shared env_defaults resolver so the SPUR_* twins finally apply. Traced the precedence end to end (CLI > env > #SBATCH directive) and it holds for every field, the twin spellings line up with srun/salloc, and the nodelist/nodefile handling looks right. Nice call deliberately not reading SPUR_JOB_NAME / SPUR_NODELIST, and pinning it with a test. Thanks for the thorough test coverage.

PR #527, opened 2026-07-28
fix(accounting): preserve unset fields on sacctmgr modify

biluriuday:
## Summary

`sacctmgr modify <entity> set <field>=<val>` reset every field the command
didn't mention back to its default instead of leaving it untouched — so
changing one attribute silently wiped the rest ([SPUR-96]). This makes
`modify` a true partial patch and hardens the accounting write path around it.

## Approach

Partial-patch semantics, end to end:

- **CLI** (`sacctmgr`) sends only the fields the user restated, via proto3
  field presence; `add` still sets everything.
- **Proto** marks settable fields `optional` — unset means "leave unchanged",
  set means "write" (per-field clear-vs-store semantics documented inline).
- **Controller** maps that presence onto a dynamic `INSERT ... ON CONFLICT DO
  UPDATE` that touches only the provided columns and preserves the rest with
  `COALESCE`/`CASE`.

## Also in this PR

- **One default account per user**: enforced with a partial unique index plus a
  dedup + column-drop backfill in `migrate()`, serialized by a
  transaction-scoped advisory lock. `add_user` demotes the user's other rows
  before the upsert. Drops the redundant, write-only `associations.is_default`
  so `users.default_account` is the single source of truth.
- **QOS allow-list validation** collapsed from N+1 lookups to one batch query.
- **SQL-injection hardening**: table and conflict targets are closed enums, so
  dynamically built SQL can only ever splice known literals.
- **Fail-loud parsing**: bad numeric fields error instead of silently
  defaulting; fairshare rejects fractional/out-of-range values instead of
  truncating; errors name the exact alias the user typed.
- **Dead code**: removed the unused `spur-core::Association` type.
- **CI**: a new `test-db` job runs the (normally `#[ignore]`d) accounting DB
  tests against an ephemeral, pinned, secret-free Postgres service.

## Design notes

- Kept `users.default_account` denormalized (dropped the duplicate flag rather
  than normalizing into its own table) — removes drift with a minimal change.
- The migration rewrites data and rebuilds the index in a single transaction
  under a fixed advisory lock, so HA controllers can't apply it concurrently.

## Testing

- `cargo clippy` (no warnings), `cargo fmt --check`, and `cargo test`.
- Accounting DB integration tests against a real Postgres: partial-patch
  preservation, default-account demotion, the migration/dedup upgrade path,
  concurrent `migrate()` serialization, and batch QOS validation.
- e2e accounting tests assert `modify` leaves unrelated fields intact.

[SPUR-96]: https://amd.atlassian.net/browse/SPUR-96

### DIFF
--- a/crates/spur-cli/src/sacctmgr.rs
+++ b/crates/spur-cli/src/sacctmgr.rs
@@ -198,6 +198,34 @@ fn parse_wall_limit(key: &str, val: &str) -> Result<u32> {
     parse_wall_time(val).ok_or_else(|| anyhow::anyhow!("invalid value for {key}=: '{val}'"))
 }
 
+/// Parse a floating-point field (e.g. `fairshare=`/`usagefactor=`), failing
+/// loudly instead of silently defaulting so a typo never quietly changes a
+/// weight on a partial `modify`.
+fn parse_f64(key: &str, val: &str) -> Result<f64> {
+    val.parse()
+        .map_err(|_| anyhow::anyhow!("invalid value for {key}=: '{val}'"))
+}
+
+/// Parse a signed integer field (e.g. `priority=`), failing loudly rather than
+/// silently defaulting, for the same reason as `parse_f64`.
+fn parse_i32(key: &str, val: &str) -> Result<i32> {
+    val.parse()
+        .map_err(|_| anyhow::anyhow!("invalid value for {key}=: '{val}'"))
+}
+
+/// Look up a field by its accepted aliases in priority order, returning the key
+/// the caller actually wrote alongside its value. Parsers use the returned key
+/// so an invalid-value error names the exact parameter the user passed (e.g.
+/// `maxrunningjobs`) rather than a canonical alias (`maxjobs`).
+fn find_alias<'a>(
+    p: &'a std::collections::HashMap<String, String>,
+    keys: &[&'a str],
+) -> Option<(&'a str, &'a str)> {
+    keys.iter()
+        .copied()
+        .find_map(|k| p.get(k).map(|v| (k, v.as_str())))
+}
+
 /// Fields shared by `add user` and `modify user` (both upsert via the same
 /// `AddUserRequest`).
 #[derive(Debug, PartialEq)]
@@ -214,6 +242,26 @@ struct UserUpsertFields {
     max_wall_minutes: u32,
 }
 
+/// Resolve the association account for `add`/`modify user`. `account=` picks the
+/// association to write and `defaultaccount=` marks the user's default; both map
+/// to one `(account, is_default)` pair, so two different accounts are rejected —
+/// honoring it would modify one association while silently clearing the default.
+fn resolve_user_account(p: &std::collections::HashMap<String, String>) -> Result<String> {
+    let account = match (p.get("account"), p.get("defaultaccount")) {
+        (Some(a), Some(d)) if a != d => bail!(
+            "account={a} and defaultaccount={d} name different accounts; \
+             set the default with defaultaccount= alone or a matching account="
+        ),
+        (Some(a), _) => a,
+        (None, Some(d)) => d,
+        (None, None) => bail!("account= required"),
+    };
+    if account.is_empty() {
+        bail!("account= must not be empty");
+    }
+    Ok(account.clone())
+}
+
 /// Parse the key=value params for `add user`/`modify user` into the shared
 /// upsert shape. Numeric/TRES aliases mirror `add account`'s
 /// `maxrunningjobs`/`maxjobs` and `add qos`'s `maxwall` for consistency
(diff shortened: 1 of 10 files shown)


yansun1996 (commented):
Nice fix — the partial-patch semantics read cleanly end to end, and the tri-state (unset preserves / empty-0 clears / value sets) is consistent across the CLI, gRPC, and DB layers. The migration dedup + partial unique index + advisory-lock serialization look correct, and the injection hardening via the closed `UpsertTable` enum is a good call.

Just a few small comments inline — all optional, nothing blocking.

crates/spurctld/src/accounting/db.rs
```diff
@@ -1266,6 +1549,412 @@ mod job_history_tests {
...
+        .await?;
+        assert_eq!(
+            default_account(pool.clone(), account.clone(), user.clone())
+                .await
+                .as_deref(),
+            Some(account.as_str())
+        );
+
+        // A partial update that doesn't restate is_default must preserve
+        // default_account (the reported bug cleared it via the ON CONFLICT gap).
```
yansun1996: Small nit: this comment (and the one a few lines below, "...the change the old UPDATE dropped on the floor") narrates the bug/fix history rather than the invariant. Per the repo's comment convention, would it be worth rewording to state what the assertion guarantees — e.g. "an unrestated is_default must not touch default_account" — and leaving the bug context to the commit message?

crates/spurctld/src/accounting/db.rs
```diff
@@ -555,94 +703,148 @@ pub struct AccountRecord {
...
-/// Add a user-account association. Like `upsert_account`/`upsert_qos`, this
-/// is a full resend on modify, not a partial patch: an empty `default_qos`,
-/// or a `None`/zero limit, clears any existing value rather than preserving
-/// it. Numeric limits use `None` (not 0) for "no limit", matching how
-/// `list_associations`/`AssociationCache` read an unset limit back out.
-#[allow(clippy::too_many_arguments)]
-pub async fn add_user(
-    pool: &PgPool,
+/// Update the partition-less association's provided columns, inserting the row
+/// if none exists. Serialized per (user, account) by the caller's advisory lock
```
yansun1996: One concern: "Serialized per (user, account) by the caller's advisory lock" holds against concurrent `add_user` calls (which take the per-user lock), but `remove_user` takes no advisory lock, so the UPDATE-then-INSERT here isn't actually serialized against a concurrent removal. The real-world risk looks low (admin path), but would it be worth softening the comment so the guarantee isn't overstated?

crates/spur-cli/src/sacctmgr.rs
```diff
@@ -349,18 +387,19 @@ async fn add(entity: &str, params: &[String], addr: &str) -> Result<()> {
...
-                        .unwrap_or(true),
-                    default_qos: fields.default_qos.clone(),
-                    allowed_qos: fields.allowed_qos.clone(),
-                    max_running_jobs: fields.max_running_jobs,
-                    max_submit_jobs: fields.max_submit_jobs,
-                    max_tres_per_job: fields.max_tres_per_job.clone(),
-                    grp_tres: fields.grp_tres.clone(),
-                    max_wall_minutes: fields.max_wall_minutes,
+                    admin_level: Some(fields.admin.clone()),
+                    is_default: Some(
```
yansun1996: Not for this PR, but flagging for awareness: on the `add` path, when `defaultaccount=` is omitted `is_default` defaults to `Some(true)`, which now triggers the new demotion logic — so `add user name=alice account=B` on a user who already has a default would silently demote it. Slurm's `add user` doesn't demote on re-add. Might be worth a follow-up to send `None` (preserve) when `defaultaccount=` is absent. Could you open a new ticket to track this so it's handled separately from this PR?

crates/spurctld/src/accounting/db.rs
biluriuday: fixed the comments across the changeset

crates/spur-cli/src/sacctmgr.rs
biluriuday: handled the issue in same PR

crates/spurctld/src/accounting/db.rs
biluriuday: updated the comment

yansun1996 (approved):
Thanks for the thorough follow-up — all the earlier review comments look addressed: the conflicting `account`/`defaultaccount` case is now rejected before it can clear a default, `add user` no longer demotes an existing default when `defaultaccount=` is omitted, and the comments now state the invariants rather than the fix history. The partial-patch upsert, the one-default-per-user migration (transaction-scoped advisory lock, idempotent), and the dynamic-SQL hardening all read cleanly.

One small thing surfaced while reviewing: `remove_user`'s two DELETEs aren't wrapped in a transaction, so they aren't atomic. That's already being handled in #451, so no action needed here — just flagging that both PRs rewrite `remove_user`, so whichever lands second will need a quick reconcile.

LGTM.

PR #531, opened 2026-07-28
fix(spur-cli): scale default sbatch --ntasks by --ntasks-per-node

Nikhil-Nunna:
## What
sbatch --ntasks-per-node=K without -n resolved num_tasks to the node count instead of nodes * K, so SLURM_NTASKS, the PMIx world size derived from num_tasks, and accounting all reported fewer tasks than the job actually ran. Scheduling/CPU allocation were unaffected since those paths already prefer tasks_per_node. This completes the gap deliberately left out of scope in #522.

## Approach
effective_ntasks now folds --ntasks-per-node into the default, returning nodes * K instead of just nodes. Precedence is untouched: merge_resolved already collapses CLI > env > #SBATCH directive into args.ntasks before this function runs, so an explicit -n/--ntasks still wins outright via early return. A zero --ntasks-per-node is treated as unset, matching the "0 means unset" convention already used in spurd/agent_server.rs and scheduler_loop.rs.

## Behavior change
SLURM_NTASKS, the PMIx world size, and accounting now correctly reflect nodes * K for the --ntasks-per-node-without--n case. No change to actual task placement/launch behavior — the scheduler already derived that from tasks_per_node, not num_tasks.

## Testing
Four new unit tests: the nodes * K default, explicit -n overriding it (CLI), an #SBATCH --ntasks=N directive overriding it, and --ntasks-per-node supplied via directive. The three pre-existing effective_ntasks tests from #522 pass unchanged. cargo test -p spur-cli --locked (242 passed), cargo fmt --all --check, and cargo clippy -p spur-cli --all-targets --locked -- -D warnings all clean.

### DIFF
--- a/crates/spur-cli/src/sbatch.rs
+++ b/crates/spur-cli/src/sbatch.rs
@@ -704,9 +704,18 @@ fn candidate_script_path(cli_args: &[String]) -> Option<String> {
 }
 
 /// Slurm allocates one task per node unless `-n` is given, so the default
-/// scales with `-N` instead of being a fixed 1.
-fn effective_ntasks(ntasks: Option<u32>, nodes: u32) -> u32 {
-    ntasks.unwrap_or(nodes)
+/// scales with `-N` instead of being a fixed 1. `--ntasks-per-node=K` raises
+/// that default to `nodes * K`, keeping `SLURM_NTASKS` and accounting in step
+/// with the number of tasks actually launched.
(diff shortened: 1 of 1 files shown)


yansun1996 (approved):
Nice targeted fix — verified `-n` still wins outright and `num_tasks`/`tasks_per_node` stay mutually consistent for the PMIx universe size and accounting.

PR #532, opened 2026-07-28
fix(spur-cli): default srun and salloc --ntasks to one task per node

Nikhil-Nunna:
## What
srun and salloc gave --ntasks a fixed default of 1 regardless of --nodes, so a multi-node job submitted without -n reported a single task for the whole allocation. #522 fixed this for sbatch, leaving the three commands inconsistent for -N — the follow-up @yansun1996 flagged in review there. Confirmed salloc is a real, independently-implemented subcommand here (not an alias of srun), so it needed its own fix.

## Approach
Both now scale the default with -N by reusing sbatch's effective_ntasks helper (made pub(crate)) rather than duplicating it, matching the existing convention where salloc/srun already call other sbatch helpers like parse_gpu_flag. srun's --ntasks also resolves from SLURM_NTASKS/SLURM_NPROCS env vars, so an apply_num_opt variant of the numeric env helper was added for Option<u32> targets, sharing its parse/error logic with the existing apply_num.

## Behavior change
Precedence matches sbatch: explicit -n wins outright, and for srun an inherited SLURM_NTASKS/SLURM_NPROCS still takes priority over the node-count default, so a step inside an existing allocation stays sized to that allocation rather than jumping to the node count.

## Testing
Two new tests per command (node-count default, explicit -n override), plus env-precedence and invalid-env coverage for srun. cargo test -p spur-cli --locked (243 passed), cargo fmt --all --check, and cargo clippy -p spur-cli --all-targets --locked -- -D warnings all clean.

### DIFF
--- a/crates/spur-cli/src/env_defaults.rs
+++ b/crates/spur-cli/src/env_defaults.rs
@@ -112,9 +112,26 @@ pub fn apply_flag(matches: &ArgMatches, id: &str, names: &[&str], target: &mut b
     }
 }
 
-/// Apply a numeric env default. The error names the specific variable that
-/// supplied the bad value so a misconfigured `SLURM_NNODES=abc` is easy to
-/// diagnose.
+/// Parse the first set variable in `names`. The error names the specific
+/// variable that supplied the bad value so a misconfigured `SLURM_NNODES=abc`
+/// is easy to diagnose.
+fn parse_first_num<T>(names: &[&str]) -> Result<Option<T>>
+where
+    T: FromStr,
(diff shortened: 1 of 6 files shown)


yansun1996 (approved):
LGTM — rebased cleanly on top of the merged #531, and the srun/salloc call sites correctly pass `None` for the new `ntasks_per_node` param. Precedence (`-n` > env > per-node default) checks out and the mock-controller dispatch tests exercise the real paths. Let's wait for CI to go green before merging.

PR #533, opened 2026-07-28
fix(spur-cli): apply sinfo --states filter to node query

Nikhil-Nunna:
## What
sinfo declared -t/--states but never used the parsed value — it sent a hardcoded empty state list in GetNodesRequest, so the filter was silently discarded and sinfo -t idle returned nodes in every state, not just the requested one. spur nodes and spur info alias to sinfo and were affected identically.

## Approach
build_get_nodes_request now parses --states and forwards it, mirroring squeue's existing pattern: parse into typed states before connecting (so an invalid -t fails without a network round-trip), reject unknown tokens rather than silently widening the filter, and treat a whole-string all as no filter. Added NodeState::from_short_or_name in spur-core, directly parallel to the existing JobState::from_code_or_name, accepting both short sinfo suffixes (idle, alloc, mix) and full names (allocated, mixed), case-insensitively.

## Behavior change
sinfo -t <state> (and spur nodes/spur info) now actually filters as documented. Omitting -t still shows all states, so the default output is unchanged. The server-side filter was already correct and is untouched.

## Testing
Seven new CLI tests (including one asserting sinfo -t idle produces states == [NodeIdle]) plus two new core tests for the string/enum mapping. cargo test -p spur-cli --locked (244 passed), cargo fmt --all --check, and cargo clippy -p spur-cli -p spur-core --all-targets --locked -- -D warnings all clean.

### DIFF
--- a/crates/spur-cli/src/sinfo.rs
+++ b/crates/spur-cli/src/sinfo.rs
@@ -11,7 +11,11 @@ use crate::format_engine;
 
 /// View information about nodes and partitions.
 #[derive(Parser, Debug)]
-#[command(name = "sinfo", about = "View cluster information")]
+#[command(
+    name = "sinfo",
+    about = "View cluster information",
+    disable_help_flag = true
+)]
 pub struct SinfoArgs {
     /// Show only this partition
     #[arg(short = 'p', long)]
@@ -48,6 +52,10 @@ pub struct SinfoArgs {
         default_value = "http://localhost:6817"
     )]
     pub controller: String,
+
+    /// Print help
+    #[arg(long, action = clap::ArgAction::Help)]
+    pub help: Option<bool>,
 }
 
 pub async fn main() -> Result<()> {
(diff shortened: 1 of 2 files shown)


crates/spur-cli/src/sinfo.rs
```diff
@@ -310,6 +363,202 @@ mod tests {
...
+                    }))
+                }
+
+                async fn get_nodes(
+                    &self,
+                    request: Request<pb::GetNodesRequest>,
+                ) -> Result<Response<pb::GetNodesResponse>, Status> {
+                    *self.nodes_req.lock().unwrap() = Some(request.into_inner());
+                    Ok(Response::new(pb::GetNodesResponse {
+                        nodes: vec![make_node("n1", NodeState::NodeIdle, "batch")],
```
yansun1996, line 471: The stub returns this same idle node regardless of the request, so the end-to-end test proves the request *carries* `[NodeIdle]` but not that a non-matching node is actually filtered out. Would it be worth having the stub return a mix (e.g. idle + down + alloc) and asserting `-t down` renders only the down node? That would exercise the real server-side filter path too, not just the request plumbing.

yansun1996 (approved):
Nice fix — `-t/--states` now actually reaches `GetNodesRequest`, parsed before the connect so an invalid state fails fast, with unknown tokens rejected and `all` treated as no filter. Handling the `-h`/`--help` collision at the same time is a good catch. LGTM.

One small suggestion on test coverage left inline — not blocking.

PR #534, opened 2026-07-28
feat(spur-cli): accept --ntasks-per-node in srun

Nikhil-Nunna:
## What
srun rejected --ntasks-per-node outright (`error: unexpected argument '--ntasks-per-node' found`), so any launcher that passes it unconditionally — e.g. Open MPI's PRTE plm:slurm, which always includes it when spawning remote daemons — failed at argument parsing before ever running anything. It's a standard, documented Slurm srun flag, and sbatch already supports it here.

## Approach
Declared ntasks_per_node on SrunArgs using the same clap attributes sbatch already uses, and forwarded it to the job spec as tasks_per_node, which the scheduler and agent already consume for per-node task distribution and SLURM_TASKS_PER_NODE. resolve_srun_env also derives the task total as nodes × ntasks_per_node when a count isn't requested directly, so the flag is genuinely wired in rather than just accepted-and-ignored.

## Behavior change
srun --ntasks-per-node=K -N N now launches nodes * K tasks instead of erroring or silently falling back to the -n 1 default. An explicit --ntasks, from the CLI or inherited from an allocation, still takes precedence. Known limitation: step mode (srun inside an existing allocation) carries the derived total task count but not the per-node value itself, since CreateJobStepRequest has no tasks_per_node field — extending that is a proto change beyond this fix's scope.

## Testing
Five new unit tests: both flag spellings, the job-spec field, the derived task count, and --ntasks precedence over it. cargo test -p spur-cli --locked (242 passed), cargo fmt --all --check, and cargo clippy -p spur-cli --all-targets --locked -- -D warnings all clean. Also confirmed spur srun --help now lists --ntasks-per-node.

### DIFF
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -1976,6 +1976,13 @@ mod tests {
             crate::sbatch::effective_ntasks(multi.ntasks, multi.ntasks_per_node, multi.nodes),
             16
         );
+
+        // Zero per-node is treated as unset, so the count falls back to nodes.
+        let zero = resolve_from(&["srun", "-N", "4", "--ntasks-per-node=0", "hostname"]);
+        assert_eq!(
+            crate::sbatch::effective_ntasks(zero.ntasks, zero.ntasks_per_node, zero.nodes),
+            4
+        );
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


yansun1996:
we have merged your PR #532 , please rebase this PR to the latest upstream main branch

yansun1996 (approved):
LGTM, merge after CI tests pass

PR #538, opened 2026-07-29
fix(dispatch): back off between launch retries and drain unusable nodes

### LINKED ISSUE #520: spurd: agent launch rejection black-holes jobs — failed node is neither drained nor avoided, and requeue hot-loops to JobHoldMaxRequeue
## What happened

On a production v0.6.0 deployment, a single node with a node-local filesystem problem black-holed jobs. `spurd` on `crsuse2-m2m-226` rejected every launch with `agent rejected job: create job spool dir`, and the controller drove the job through a tight requeue loop until it landed in `JobHoldMaxRequeue` — roughly 4 requeue attempts per second, exhausting `max_batch_requeue` (default 5) in ~1.5 seconds.

Controller log (trimmed):

```
ERROR spurctld::scheduler_loop: dispatch to agent failed job_id=631 node=crsuse2-m2m-226 error=agent rejected job: create job spool dir
ERROR spurctld::scheduler_loop: all dispatches failed — requeueing job to Pending job_id=631 failures=1
INFO  spurctld::cluster: job requeued after dispatch failure job_id=631 from=RUNNING
... (repeats ~4x/sec) ...
INFO  spurctld::cluster: job held at max requeue limit job_id=631
```

```
$ spur show job 631
JobId=631 JobName=gha-runner JobState=CANCELLED Reason=JobHoldMaxRequeue
```

The underlying trigger was environmental: `create_job_spool_dir` (`crates/spurd/src/executor.rs:1077`) failed to `mkdir` under **both** candidate roots (`/var/spool/spur` and `/tmp/spur`) — consistent with a full or read-only root filesystem on that node. But the environmental trigger is not the point of this issue; the point is how Spur amplified one sick node into a job-losing storm.

## Three distinct defects

**1. The real error is swallowed.** The agent returns the failure via `anyhow`'s `Display` (`e.to_string()` at `crates/spurd/src/agent_server.rs:1110`), which prints only the outermost context (`create job spool dir`). The underlying OS errno captured in `last_err` at `executor.rs:1098` (`ENOSPC`? `EROFS`? `EACCES`?) never reaches the controller log, so the node problem is undiagnosable from the controller side.

**2. The failed node is neither drained nor avoided.** The scheduler's node selection (`crates/spur-sched/src/backfill.rs:327-357`) sorts on resource load only — earliest start, most free GPUs, weight, least allocated CPUs. There is no term for "this node just rejected a launch." Because the spool `mkdir` fails *before* any resources are consumed, the broken node still looks maximally idle and sorts to the **top** of the candidate list on the very next cycle — so it is re-selected immediately. Nothing drains it or lowers its desirability. spurd only drains on *prolog* failure (`agent_server.rs:1113`); a plain launch rejection penalizes neither the node nor future scheduling.

**3. Requeue has no backoff and no hold.** `requeue_job` (`crates/spurctld/src/cluster.rs:1149`) flips the job Failed→Pending unconditionally and instantly, making it eligible for immediate re-dispatch on the next scheduler cycle. There is no delay and no hold, so the loop runs at scheduler speed and burns the entire `max_batch_requeue` budget in well under two seconds. The user's window to notice or intervene is effectively zero, and the job ends CANCELLED rather than parked for inspection.

## Expected behavior (Slurm reference)

Verified against Slurm slurmctld source (`SchedMD/slurm`). When slurmd reports a batch-launch/setup failure, `_slurm_rpc_complete_batch_script` (`src/slurmctld/proc_req.c`) does, in order:

1. **Drains the offending node** — `drain_nodes(node, "batch job complete failure", ...)`. A generic slurmd setup failure is treated as fatal; the default branch is literally "All others drain the node." Only a small allowlist (`SLURM_COMMUNICATIONS_SEND_ERROR`, `ESLURM_USER_ID_MISSING`, `ESLURMD_INVALID_ACCT_FREQ`, `ESPANK_JOB_FAILURE`) is non-fatal.
2. **Requeues onto a different node** — because the bad node is now DRAIN, it is out of the candidate set when the job returns to Pending. Slurm doesn't need smarter scoring; draining removes the node structurally.
3. **Holds the job by default** — on `JOB_LAUNCH_FAILED`, `job_requeue_internal` (`src/slurmctld/job_mgr.c:18027`) sets `priority = 0` with reason `"launch failed requeued held"`. It does not silently re-queue at full priority; `nohold_on_prolog_fail` (default off) is required to opt out.
4. **Bumps counters** — `slurmctld_diag_stats.jobs_failed++` plus a per-job `batch_flag++`, holding with `WAIT_MAX_REQUEUE` once `MaxBatchRequeue` (default 5, same as Spur) is exceeded.

Net effect in Slurm: one sick node drains itself, the job moves elsewhere (or is held for an admin), and there is never a busy-loop. Spur v0.6.0 replicates only the final max-requeue cap and none of the mechanisms that make it resilient.

## Impact

A single node with a transient or persistent local-FS fault can silently consume and cancel a stream of jobs, with no actionable diagnostic in the controller log and no protection for other jobs (the node keeps advertising itself as idle and keeps getting picked). This is a fault-tolerance and Slurm-compatibility gap, not just a cosmetic logging miss.

## Reproduction

1. Make the spool roots un-writable on one compute node (e.g. fill the fs so both `/var/spool/spur` and `/tmp/spur` fail `mkdir`, or mount read-only).
2. Submit a batch job that the scheduler places on that node.
3. Observe: controller logs `create job spool dir` with no errno; the job hot-requeues ~4x/sec onto the same node; within ~1.5s it reaches `JobHoldMaxRequeue`; the node is never drained and remains eligible for subsequent jobs.

## Notes / scope

- Not a regression of #489 (GPU reservation stranded mid-launch) or #498 (completion/eviction desync). Both are in v0.6.0 and both concern GPU-allocation bookkeeping; their rejection signature is `controller-allocated GPUs unavailable`, distinct from `create job spool dir`, which fails before any GPU setup.
- The Raft `Unreachable node target=3` warnings seen alongside this in the same deployment are a separate controller-peer connectivity problem, unrelated to the spool failure.
- Any drain-on-launch-failure behavior should mirror Slurm's conservatism: a node that cannot create a batch/spool dir is genuinely broken and recovery (clearing the fault, resuming the node) is reasonably an operator action rather than silent auto-retry.

shiv-tyagi:
Fixes #520.

A node whose spool filesystem was read-only or full rejected every launch. The controller re-dispatched to that same node about four times a second, exhausted `max_batch_requeue` in roughly 1.5 seconds, and cancelled the job with no errno for the operator to act on.

## Approach

`LaunchError`'s `Display` now uses `{e:#}`, so the errno reaches the controller instead of being truncated to `create job spool dir`. An I/O failure on the node-owned spool root is classified as a node fault and drains the node, which removes it from the candidate set structurally rather than needing scheduler scoring changes. Requeued jobs carry a `begin_time` hold that doubles per attempt, capped by the new `controller.max_launch_backoff_secs` (default 300).

Pairing a drain with an automatic retry changes what a drain costs: a failure that recurs everywhere drains a node per retry. Slurm bounds this by never letting a drain and an automatic retry coexist unless the failure is provably the node's fault. This PR follows the same rule.

**Node faults retry.** `classify_spool_error` gates on the root-owned spool tree, where every path is built by spurd from the job id and no user input can steer the errno, so a retry elsewhere converges. Structured as an exclusion list (any errno except `EDQUOT`) mirroring Slurm's "all others drain the node" default in `_slurm_rpc_complete_batch_script`. The world-writable `/tmp` fallback and the user-controlled `work_dir` stay excluded: either would let one user drain the cluster node by node.

**Prolog failures hold.** A prolog receives the job's own uid, gid and work_dir, so an unprovisioned account or an unreachable NFS mount fails identically on every node. The agent no longer self-drains or reports a fake completion; it returns a classified rejection (`LaunchFailureKind::PROLOG`) and the controller drains the node and parks the job at priority 0 with Slurm's `state_desc` of `launch failed requeued held`. `scontrol release` resumes it. `controller.hold_on_prolog_fail = false` restores the retry, which is Slurm's `SchedulerParameters=nohold_on_prolog_fail`, still backstopped by `max_batch_requeue`. Interactive jobs are cancelled rather than held, matching Slurm and avoiding a stranded `srun`.

The drain also uses the controller rather than a completion report from the agent: a job that never ran has no exit code, and reporting one races the requeue and can finalize a still-retryable job to Failed. Reaching `drain_node` automatically from the agent opened a deadlock window, so it now takes jobs before nodes to match `apply_operation`.

## Design notes

Slurm keeps two fields for a job's wait reason: `state_reason` (an enum) and `state_desc` (free text that both `squeue` and `scontrol` print in preference to the code). Spur had only the enum, so `Job` gains `pending_reason_desc`. Writes to either field go through `set_pending_reason` / `set_pending_reason_desc`, which move them together, so a description can never outlive the reason it explains.

One deliberate divergence. Slurm applies no backoff to its node-fault requeue, because its drain is synchronous under the job write lock and the broken node is already out of the candidate set before the requeue happens. Spur's node-fault drain is an agent-initiated RPC racing the controller's requeue, so the backoff is what covers that window. Moving that drain controller-side later, as this PR already does for prolog, would restore Slurm's ordering and demote the backoff to a safety net.

One march survives, and Slurm has it too: a prolog script broken cluster-wide drains one node per job until every node is out. The per-job hold bounds each job to a single drain, but nothing bounds the aggregate. A global circuit breaker is the fix and is filed as a follow-up.

## Rolling upgrades

`#[serde(default)]` covers old WAL replayed by a new binary, but a pre-538 follower applying a 538 leader's `JobStateChange` silently drops the unknown `begin_time` field, so on that replica the job is immediately eligible; if it wins leadership mid-upgrade the backoff vanishes. Upgrade all controllers before relying on the backoff. `pending_reason_desc` has the same caveat with only a cosmetic consequence: an old follower reports `JobHeldUser` instead of the description. The hold itself still applies, since the reason and the zeroed priority both survive.

The proto field is additive and degrades cleanly in both directions. An old agent sends `failure_kind = UNSPECIFIED` and its prolog failures keep today's requeue behavior; an old controller ignores the field entirely.

## Behavior change

A batch job whose prolog fails used to end `Failed` and leave the queue. It now parks Pending at priority 0 and stays there until `scontrol release`. That is Slurm's behavior and the reason the drain is bounded, but anything scripted around "prolog failure means the job is gone" needs updating. Interactive jobs are unaffected: they are cancelled, as before.

## Testing

Clippy clean, 2330 tests pass. Verified on a 4 node bare-metal cluster: the node drains with the errno in its reason and the job requeues and completes elsewhere.

### DIFF
--- a/crates/spurd/src/agent_server.rs
+++ b/crates/spurd/src/agent_server.rs
@@ -798,13 +798,9 @@ fn warn_mpi_mpirun_skipped_affinity(job_id: u32, source: &HashMap<String, String
     }
 }
 
-/// Drain this node without reporting a job completion.
-///
-/// A launch failure is not a completion: the job never ran, and the controller's
-/// dispatch path already owns its fate (requeue when every dispatch failed,
-/// eviction to NodeFail on a partial failure). Reporting an exit code here would
-/// race that path and could finalize a still-retryable job to Failed, which no
-/// requeue path recovers.
+/// Drain this node without reporting a job completion. The controller's dispatch
+/// path already owns the job's fate, so reporting an exit code here would race it
+/// and could finalize a still-retryable job to Failed, which no requeue recovers.
 async fn request_node_drain(controller_addr: &str, node_name: &str, reason: &str, job_id: u32) {
     let result = retry_controller_rpc(move |attempt| async move {
         let channel = spur_client::connect_channel(controller_addr)
--- a/crates/spurd/src/agent_server.rs
+++ b/crates/spurd/src/agent_server.rs
@@ -1062,31 +1062,17 @@ impl SlurmAgent for AgentService {
                 memory_mb: spec.memory_per_node_mb,
             };
             if let Err(e) = spur_core::hooks::run_hook(prolog, &ctx).await {
-                let err_msg = format!("prolog failed: {e}");
+                // No completion report and no self-drain: the controller owns
+                // both decisions here, because only it can pair the drain with
+                // the hold that stops the job walking the cluster.
+                let err_msg = format!("prolog failed: {e:#}");
                 error!(job_id, error = %err_msg, "prolog hook failed before launch");
-                let controller = self.reporter.controller_addr.clone();
-                let node_name = self.reporter.hostname.clone();
-                let drain_reason = err_msg.clone();
-                tokio::spawn(async move {
-                    let drain = DrainRequest {
-                        reason: drain_reason.clone(),
-                    };
-                    report_completion(
-                        &controller,
-                        job_id,
-                        -1,
-                        0,
-                        run_attempt,
-                        &node_name,
-                        Some(&drain),
-                    )
-                    .await;
-                });
                 return Ok(Response::new(LaunchJobResponse {
                     success: false,
                     error: err_msg,
                     stdout_path: String::new(),
                     stderr_path: String::new(),
+                    failure_kind: LaunchFailureKind::LaunchFailureProlog as i32,
                 }));
             }
         }
@@ -1225,6 +1211,7 @@ impl SlurmAgent for AgentService {
                         error: "reservation reclaimed during launch".into(),
                         stdout_path: String::new(),
                         stderr_path: String::new(),
+                        failure_kind: LaunchFailureKind::LaunchFailureUnspecified as i32,
                     }));
                 }
 
@@ -1274,11 +1261,18 @@ impl SlurmAgent for AgentService {
                     error: String::new(),
                     stdout_path,
                     stderr_path,
+                    failure_kind: LaunchFailureKind::LaunchFailureUnspecified as i32,
                 }))
             }
             Err(e) => {
                 // reservation_guard releases the allocation and PMI on this return.
                 let drain_reason = e.drain_reason();
+                let failure_kind = match e {
+                    executor::LaunchError::PrologFailed(_) => {
+                        LaunchFailureKind::LaunchFailureProlog
+                    }
+                    _ => LaunchFailureKind::LaunchFailureUnspecified,
+                };
                 let err_msg = e.to_string();
                 error!(job_id, error = %err_msg, "failed to launch job");
 
@@ -1295,6 +1289,7 @@ impl SlurmAgent for AgentService {
                     error: err_msg,
                     stdout_path: String::new(),
                     stderr_path: String::new(),
+                    failure_kind: failure_kind as i32,
                 }))
             }
         }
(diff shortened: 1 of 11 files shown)


crates/spurd/src/agent_server.rs
```diff
@@ -670,6 +670,76 @@ fn warn_mpi_mpirun_skipped_affinity(job_id: u32, source: &HashMap<String, String
...
+                    name: node_name.to_string(),
+                    reason: reason.to_string(),
+                };
+                match client.drain_node(req).await {
+                    Ok(resp) => {
+                        warn!(
+                            job_id,
+                            node = %node_name,
+                            state = %resp.into_inner().actual_state,
+                            reason,
```
yansun1996: Nit: this `reason,` shorthand formats the `&str` with `Debug`, so it logs quoted — inconsistent with the `reason = %reason` and `error = %e` used elsewhere in this function. `reason = %reason` keeps the logs uniform.

crates/spurd/src/agent_server.rs
```diff
@@ -670,6 +670,76 @@ fn warn_mpi_mpirun_skipped_affinity(job_id: u32, source: &HashMap<String, String
...
 }
 
+/// Drain this node without reporting a job completion.
+///
+/// A launch failure is not a completion: the job never ran, and the controller's
+/// dispatch path already owns its fate (requeue when every dispatch failed,
+/// eviction to NodeFail on a partial failure). Reporting an exit code here would
+/// race that path and could finalize a still-retryable job to Failed, which no
+/// requeue path recovers.
+async fn request_node_drain(controller_addr: &str, node_name: &str, reason: &str, job_id: u32) {
```
yansun1996, line 804: This retry path — 3 attempts, the 1s sleep, the give-up after the loop, and the `controller_rpc_retryable` branch that returns early on a non-retryable code — doesn't appear to be covered by a test. Would it be worth a focused test with a mock controller `DrainNode` endpoint, so a regression here (e.g. sleeping/looping on a non-retryable code, or not giving up) can't ship silently? The rename of `controller_rpc_retryable` itself is covered, so it's really the loop control flow that's untested.

crates/spurd/src/executor.rs
```diff
@@ -1092,10 +1172,35 @@ fn create_job_spool_dir(job_id: JobId, uid: u32, gid: u32) -> anyhow::Result<Pat
...
     }
-    bail!("failed to create job spool dir: {last_err:?}")
+    Err(spool_dir_error(failures))
+}
+
+/// Build the error for a spool dir that could not be created under any candidate
+/// root.
+///
+/// Reports the owned root's failure when there is one, in preference to the temp
+/// fallback's: that is the root an operator configured and the only one whose
```
yansun1996: This doc block (and a fair number of others in the PR — the largest are here in `executor.rs`, plus `cluster.rs` and `agent_server.rs`) runs well past a couple of lines. The repo leans toward short comments that capture only the non-obvious *why*. The rationale here is genuinely worth keeping — so rather than dropping it, could it be tightened to a line or two, with the fuller explanation moving to the commit message / PR description? Same suggestion applies to the other long blocks (e.g. `drain_reason` and `launch_backoff_secs`).

yansun1996 (commented):
Thanks for this — thorough fix with genuinely strong test coverage (frozen old-payload deserialization tests per new persisted field, the backoff schedule verified as arithmetic, both total and partial prolog-failure paths exercised through real gRPC dispatch). The proto change is additive and the WAL/snapshot fields default cleanly, so rolling upgrades look safe. A few small, non-blocking notes inline: a logging-consistency nit, a suggested test for the drain retry loop, and a note on trimming the longer comment blocks. Nice work overall.

yansun1996 (changes requested):

crates/spurd/src/agent_server.rs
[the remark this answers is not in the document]
shiv-tyagi: Fixed in d6ded2b. Now `reason = %reason`, matching `node` and `state` on the same line.

shiv-tyagi: Fixed in d6ded2b. Now `reason = %reason`.

crates/spurd/src/agent_server.rs
shiv-tyagi: Covered in d6ded2b, though not with a mock controller. `SlurmController` has 39 RPCs, so a mock costs ~37 `unimplemented!()` stubs and still only exercises this one caller.

Instead I pulled the policy out. `report_completion` had a byte-identical copy of this loop, both hardcoding the 3 attempts and the 1s sleep, so `retry_controller_rpc` now owns the retry decision and both callers supply only their RPC and log fields. The four behaviors you listed are tested against it directly under `#[tokio::test(start_paused = true)]`, asserting attempt counts and elapsed virtual time:

```rust
async fn non_retryable_error_gives_up_without_retrying() {
    let start = tokio::time::Instant::now();
    let (result, calls) = run_script(vec![
        Err(Status::invalid_argument("unknown node")),
        Ok(()),
        Ok(()),
    ])
    .await;

    assert!(matches!(result, Err(ControllerRpcError::Rpc(_))));
    assert_eq!(calls, 1);
    assert_eq!(start.elapsed(), Duration::ZERO);
}
```

The trailing `Ok`s are deliberate: a regression that retries a non-retryable code fails the assertion instead of running off the end of the script. Verified by removing the `!e.retryable()` guard, which fails that test and no other.

crates/spurd/src/executor.rs
shiv-tyagi: Compressed in 450fbff across `executor.rs` and `cluster.rs`, cutting 45 comment lines to 26.

What went: restatements the code already carries. `drain_reason` no longer explains that only `NodeFault` qualifies (the `match` says so), `is_node_fault_io_error` no longer lists the errnos it covers, and the backoff constant no longer walks through the cap interaction.

What stayed, in shorter form:

```rust
/// The `io::Error` must stay a source rather than be formatted into the message:
/// [`is_node_fault_io_error`] detects the fault by walking the chain, so a
/// flattened errno would silently downgrade a node fault to a job failure.
```

That one is a constraint on how the error is built, not a description of it, and flattening it fails quietly. `write_job_scratch_keeps_the_errno_downcastable` catches it, but the comment is what tells you why the test exists.

I left the `[controller]` key docs in `config.rs` alone since those are the reference for the TOML surface.

shiv-tyagi: Following up: I'd scoped the earlier pass to `executor.rs` and `cluster.rs` and missed the `agent_server.rs` block you named. Now trimmed too, on `request_node_drain`:

```rust
/// Drain this node without reporting a job completion. The controller's dispatch
/// path already owns the job's fate, so reporting an exit code here would race it
/// and could finalize a still-retryable job to Failed, which no requeue recovers.
```

The enumeration of what the dispatch path does went; the race and its irrecoverable outcome stayed, since that is the reason this is a separate RPC rather than a flag on the completion report.

Branch is also rebased onto current main. Note the rebase renamed the commits my earlier replies cite: they are now 28269e0 (retry loop) and 1d2b9bb (comments).

yansun1996 (approved):

PR #540, opened 2026-07-29
fix(spur-cli): accept --qos and SALLOC_QOS in salloc

yansun1996:
## What

`spur alloc` / `salloc` could not select a QoS for an interactive allocation: it exposed no `-q/--qos` flag and had no `SALLOC_QOS` env fallback, even though the sibling submission commands both support QoS (`srun`: `-q/--qos`; `sbatch`: `-q/--qos` + `SBATCH_QOS`). Attempting `salloc --qos ...` failed with `unexpected argument '--qos'`.

Fixes SPUR-110.

## Approach

- Add the `-q/--qos` flag to `SallocArgs`, matching the `srun` definition.
- Resolve it from the environment in `resolve_salloc_env` using the same `SPUR_*` + `SALLOC_*` twin ordering as salloc's other fields (`SPUR_QOS`, `SALLOC_QOS`), with CLI overriding env. This mirrors real Slurm, where salloc reads the command-prefixed `SALLOC_QOS`.
- Thread the resolved value into the submitted `JobSpec.qos`.

The inline `JobSpec` construction in `main_with_args` is extracted into a small `build_salloc_job_spec` helper (mirroring `build_srun_job_spec`) so the spec wiring is independently unit-testable rather than only reachable through the network path.

### Note on env alias scope

salloc intentionally reads only `SPUR_QOS`/`SALLOC_QOS`, not `SLURM_QOS`. This matches salloc's existing convention for every other field (partition, account, etc. use `SPUR_*`+`SALLOC_*` only) and real Slurm: salloc reads its own command-prefixed `SALLOC_QOS`, while the generic `SLURM_QOS` / `*_JOB_QOS` twins that `srun` honors are the vars an *inherited* allocation exports — relevant to srun-inside-an-allocation, not to salloc which creates the allocation.

## Tests

- Parse: `-q` and `--qos` short/long.
- Env: `SALLOC_QOS` default, CLI override of env, and `SPUR_QOS` alias precedence.
- Spec: `build_salloc_job_spec` sets `qos`, and defaults it empty when unset.

Validated with `cargo clippy` (clean), `cargo fmt --check`, and the targeted `spur-cli` test suite. Also verified end-to-end on an isolated single-node deployment: both `--qos <name>` and `SALLOC_QOS=<name>` round-trip into the granted allocation's recorded QoS (surfaced to the interactive shell as `SPUR_JOB_QOS`/`SLURM_JOB_QOS`), and the no-QoS baseline stays empty.

### DIFF
--- a/crates/spur-cli/src/salloc.rs
+++ b/crates/spur-cli/src/salloc.rs
@@ -277,7 +277,6 @@ fn build_salloc_job_spec(args: &SallocArgs, nodelist: Option<String>) -> Result<
         .transpose()?
         .unwrap_or(0);
 
-    // sleep infinity holds the allocation until the interactive shell exits.
     Ok(JobSpec {
         name: args
             .job_name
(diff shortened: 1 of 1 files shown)


shiv-tyagi (approved):
LGTM

PR #544, opened 2026-07-30
fix(spurd): retry RunCommand job lookup until dispatch registers

Nikhil-Nunna:
## What

A job step dispatched via `srun` (or any other `RunCommand` caller) targeting a node can fail with `job <id> not running on this node`, even though `squeue`/`scontrol` already report the job as `RUNNING` -- sometimes for several hundred milliseconds after `RUNNING` first becomes visible.

## Approach

`spurctld` transitions a job to `Running` (the state `squeue`/`scontrol` read) synchronously in the scheduler loop, then dispatches `LaunchJob` to each assigned node concurrently and out of band via a spawned task. Each node's `spurd` agent only inserts the job into its local `running` map after its own launch pipeline (resource allocation, GPU device-injection plan, process spawn) finishes. A step targeting that node can therefore arrive before its own dispatch has completed, even though the controller already reports the job `Running` -- there is no signal that bridges "controller says Running" and "this specific node has registered the job locally."

`run_command`'s job lookup previously did a single, non-retried `HashMap::get` and failed immediately on a miss. This extracts that lookup into `wait_for_running_job`, which polls the same map every 100ms for up to ~3s before giving up, on the basis that the job is expected to appear within one launch pipeline's worth of time, not never.

## Behavior change

None for the non-racy path -- a job already in `running` resolves immediately, same as before. A job not yet in `running` now gets a short retry window instead of failing on the first miss; genuinely non-existent jobs still fail, just after the retry budget (~3s) is exhausted rather than instantly.

## Testing

Reproduced live: a 2-node allocation whose script immediately fires `srun` at its peer node while a separate session polls `squeue` at 10Hz. Idle-cluster runs didn't trigger it (dispatch skew was tens of ms); 10 concurrent 2-node allocations (to load the scheduler/agent RPC paths) reproduced it reliably, with the peer node's own script not starting until ~670ms after squeue first reported `RUNNING`, well after the failing node had already been retrying against it.

Added two tests using `tokio::test(start_paused = true)`: one confirms the not-found path still fires correctly once the retry budget is exhausted; one exercises the actual race directly -- spawns the call, lets it hit the first failing lookup and enter its retry sleep, then inserts the job and advances paused time, asserting the retry picks it up on the next poll.

All existing `spurd` tests pass (153 total), `cargo fmt --all --check` clean, `cargo clippy -p spurd --all-targets --locked -- -D warnings` clean.

Note: this is a surgical fix (retry-on-miss) rather than a deeper architectural change (blocking the `Running` transition until nodes confirm dispatch, mirroring the existing `register_allocation_on_nodes` pattern already used for standalone `srun` allocations) -- that alternative would change job-admission latency/scalability tradeoffs and seemed like a maintainer decision rather than something to bake in unilaterally.

### DIFF
--- a/crates/spurd/src/agent_server.rs
+++ b/crates/spurd/src/agent_server.rs
@@ -366,6 +366,12 @@ struct DrainRequest {
 /// past this bound is reclaimed.
 const LAUNCHING_TTL: std::time::Duration = std::time::Duration::from_secs(600);
 
+/// Poll interval and attempt count for `wait_for_running_job`'s retry of a
+/// job that hasn't reached `running` yet. ~3s total, well above the launch
+/// latency skew observed between nodes of the same job under scheduler load.
+const JOB_LOOKUP_RETRY_INTERVAL: std::time::Duration = std::time::Duration::from_millis(100);
+const JOB_LOOKUP_RETRY_ATTEMPTS: u32 = 30;
+
 /// Reclaim allocations whose job is no longer tracked and is not mid-launch,
 /// using the running set as ground truth. Callers hold the `running` lock
 /// across building `running` and this call so the live set is a consistent
@@ -1481,27 +1487,8 @@ impl SlurmAgent for AgentService {
         };
         let step_id = req.step_id;
 
-        let (gpu_devices, partition, cpus, memory_mb, nodelist, job_mpi) = {
-            let jobs = self.running.lock().await;
-            let tracked = jobs.get(&job_id).ok_or_else(|| {
-                Status::not_found(format!("job {} not running on this node", job_id))
-            })?;
-            let nodelist = if tracked.nodelist.is_empty() {
-                hostname::get()
-                    .map(|h| h.to_string_lossy().to_string())
-                    .unwrap_or_else(|_| "localhost".into())
-            } else {
-                tracked.nodelist.clone()
-            };
-            (
-                tracked.gpu_devices.clone(),
-                tracked.partition.clone(),
-                tracked.cpus,
-                tracked.memory_mb,
-                nodelist,
-                tracked.mpi.clone(),
-            )
-        };
+        let (gpu_devices, partition, cpus, memory_mb, nodelist, job_mpi) =
+            self.wait_for_running_job(job_id).await?;
 
         let agent_hostname = self.reporter.hostname.clone();
         let node_names: Vec<&str> = nodelist.split(',').filter(|s| !s.is_empty()).collect();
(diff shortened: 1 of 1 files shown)


crates/spurd/src/agent_server.rs
```diff
@@ -2255,6 +2242,54 @@ impl AgentService {
...
+    /// dispatch happens concurrently, out of band, and each agent only adds
+    /// the job to `running` after its own local launch pipeline (resource
+    /// allocation, GPU device injection, process/container spawn) finishes.
+    /// A step targeting this node (`srun` from within the batch script, or
+    /// any other RunCommand caller) can therefore arrive here before this
+    /// node's own dispatch has completed, even though the controller already
+    /// reports the job Running. Poll briefly rather than failing on the
+    /// first miss, since the job is expected to appear within one launch
+    /// pipeline's worth of time, not never.
+    async fn wait_for_running_job(
```
yansun1996, line 2259: The same early-arrival window applies to `stream_job_output` (`:1731`) and `job_entry`/`exec_in_job` (`:2294`) — both do `running.get(&job_id)` and return the identical `not running on this node` error. Would it make sense to hoist this wait into the shared lookup so those paths are covered too, rather than retrying only here? Otherwise the same symptom can resurface via attach/exec.

crates/spurd/src/agent_server.rs
```diff
@@ -2921,6 +2959,85 @@ mod tests {
...
+                job_id,
+                ..Default::default()
+            });
+            svc_call.run_command(req).await
+        });
+
+        // Let the first (failing) lookup run and start its retry sleep before
+        // the job appears, so this exercises the retry path, not a lucky win
+        // on the first attempt.
+        tokio::task::yield_now().await;
```
yansun1996, line 2998: A single `yield_now()` may not be enough to guarantee the spawned task has reached its first `sleep` before the insert — `run_command` does arg validation and a lock acquire first. If the insert wins that race, attempt 0 hits and the retry path isn't exercised, yet the test still passes. Advancing one interval before inserting would make the first miss deterministic.

crates/spurd/src/agent_server.rs
```diff
@@ -2255,6 +2242,54 @@ impl AgentService {
...
+                    tracked.gpu_devices.clone(),
+                    tracked.partition.clone(),
+                    tracked.cpus,
+                    tracked.memory_mb,
+                    nodelist,
+                    tracked.mpi.clone(),
+                ));
+            }
+            drop(jobs);
+            if attempt + 1 < JOB_LOOKUP_RETRY_ATTEMPTS {
```
yansun1996, line 2283: Since the sleep is skipped on the final attempt, a job appearing exactly on the last poll is the edge worth pinning — inserting right before the final attempt would guard this off-by-one.

crates/spurd/src/agent_server.rs
```diff
@@ -366,6 +366,12 @@ struct DrainRequest {
 /// past this bound is reclaimed.
 const LAUNCHING_TTL: std::time::Duration = std::time::Duration::from_secs(600);
 
+/// Poll interval and attempt count for `wait_for_running_job`'s retry of a
+/// job that hasn't reached `running` yet. ~3s total, well above the launch
```
yansun1996, line 370: Small thing: this comes out to ~2.9s (29 sleeps between 30 attempts) rather than ~3s.

yansun1996 (commented):
Thanks for the clear write-up and the live repro — easy to follow, and the `start_paused` tests keep it deterministic.

Retry-at-the-node looks like the right call; deferring the controller-side alternative is fair since it changes admission latency clusterwide.

**Main question: scope.** The same early-arrival window isn't unique to `run_command` — `stream_job_output` (`srun --attach`) and `job_entry` (`exec_in_job`) do the same `running.get(&job_id)` and return the identical `not running on this node` error. This fixes one of three doors; the symptom could resurface via attach/exec. Would it be worth lifting the wait into the shared lookup so all three are covered? (The most complete fix is controller-side — extending the `register_allocation_on_nodes`-style "wait for all nodes before Running" that the standalone-srun path already uses to the batch/managed dispatch path — which closes the race for every consumer at once. Agreed that's a larger, admission-latency-affecting change and fine to treat as a follow-up rather than something to bake in here.)

Nikhil-Nunna:
Closing PR in favor of #550

PR #549, opened 2026-07-30
fix(spur-cli): accept multi-node hostlists in scontrol update NodeName

shiv-tyagi:
## What this fixes

`scontrol update NodeName=` only worked for a single node. The value was forwarded to the controller as one opaque string, so any multi-node argument failed with `node <entire list> not found`. Draining a set of nodes required one invocation per node.

## Approach

Node arguments resolve through the existing `spur_core::hostlist::expand`, so this is wiring rather than new parsing. `scontrol update NodeName=` now accepts comma-separated lists (`node1,node2`), ranges (`node[01-05]`), mixed forms (`node[01-05],login01`), ranges with commas (`node[1,3,5-7]`), and `ALL`. `ALL` resolves by querying registered nodes, matching Slurm. The same expansion applies to `spur node drain`, `remove`, and `label`, without `ALL`, since those are native commands rather than a Slurm-compatible surface.

RPCs stay single-node and the CLI issues one call per node. No proto changes.

## Behavior change

An unrecognized state previously warned on stderr and coerced to `IDLE`, so `State=IDEL` against a drained node silently resumed it and exited 0. Multi-node support turns that from a one-node mistake into a fleet-wide one, so unknown states are now rejected. State is parsed before any RPC is issued, so a typo leaves every node untouched instead of updating a prefix and failing partway. Errors echo what the caller typed rather than the lowercased form.

I did not mark the title `!`, on the grounds that the old path silently discarded invalid input rather than honoring a documented contract. Say so if you read it differently.

## Also included

One gRPC connection is reused for the whole update. `update_node` previously reconnected per call, so draining 12 nodes opened 13 connections.

## Testing

10 new unit tests (6 hostlist, 4 state parsing). Full suite passes, clippy clean, rebased on `main`.

Manual on a 4-node bare-metal cluster: 92 checks, no failures. Covered every accepted form, `ALL` casing, `DRAIN`/`DOWN`/`RESUME` casings, nonexistent node, reversed range, and reason persistence. Fail-fast verified in both directions: a typo across 4 nodes leaves all 4 `idle`, and with 2 nodes already drained a typo leaves both `drain` instead of resuming them. Scheduling checked end to end so drain affects placement and not just reported state: a job stayed `PENDING` while all nodes were drained and completed after resume, `-N3` spread across 3 nodes, and with 3 nodes drained a job landed on the single idle node.

Hostlist expansion is client-side, so a large range becomes that many sequential RPCs. Fine at this scale, worth revisiting if operators drive thousands of nodes at once.

### DIFF
--- a/crates/spur-cli/src/node.rs
+++ b/crates/spur-cli/src/node.rs
@@ -169,15 +169,15 @@ async fn cmd_drain(controller: &str, node_pattern: String, reason: Option<String
                 } else {
                     println!("Node {node} set to drain");
                 }
+                if let Some(r) = &reason {
+                    println!("  reason: {r}");
+                }
             }
             Err(e) => {
                 eprintln!("error: {node}: {e}");
                 failed.push(node.clone());
             }
         }
-    }
-    if let Some(r) = &reason {
-        println!("  reason: {r}");
     }
     if !failed.is_empty() {
         bail!(
@@ -224,15 +224,15 @@ async fn cmd_remove(
                 } else {
                     println!("Node {node} removed from cluster");
                 }
+                if let Some(r) = &reason {
+                    println!("  reason: {r}");
+                }
             }
             Err(e) => {
                 eprintln!("error: {node}: {e}");
                 failed.push(node.clone());
             }
         }
-    }
-    if let Some(r) = &reason {
-        println!("  reason: {r}");
     }
     if !failed.is_empty() {
         bail!(
--- a/crates/spur-cli/src/node.rs
+++ b/crates/spur-cli/src/node.rs
@@ -103,12 +103,13 @@ fn parse_label_args(label_args: &[String]) -> Result<(HashMap<String, String>, V
 }
 
 async fn cmd_label(controller: &str, node_pattern: String, label_args: Vec<String>) -> Result<()> {
-    let mut client = spur_proto::controller_client(spur_client::connect_channel(controller).await?);
     let (set_labels, remove_labels) = parse_label_args(&label_args)?;
     let nodes = expand_node_pattern(&node_pattern)?;
+    let mut client = spur_proto::controller_client(spur_client::connect_channel(controller).await?);
 
+    let mut failed: Vec<String> = Vec::new();
     for node in &nodes {
-        client
+        match client
             .update_node(UpdateNodeRequest {
                 name: node.to_string(),
                 state: None,
@@ -116,44 +117,75 @@ async fn cmd_label(controller: &str, node_pattern: String, label_args: Vec<Strin
                 labels: set_labels.clone(),
                 remove_labels: remove_labels.clone(),
             })
-            .await?;
-
-        for (k, v) in &set_labels {
-            println!("  {node}: {k}={v}");
-        }
-        for k in &remove_labels {
-            println!("  {node}: {k} removed");
+            .await
+        {
+            Ok(_) => {
+                for (k, v) in &set_labels {
+                    println!("  {node}: {k}={v}");
+                }
+                for k in &remove_labels {
+                    println!("  {node}: {k} removed");
+                }
+            }
+            Err(e) => {
+                eprintln!("error: {node}: {e}");
+                failed.push(node.clone());
+            }
         }
     }
 
+    if !failed.is_empty() {
+        bail!(
+            "failed on {} of {} node(s): {}",
+            failed.len(),
+            nodes.len(),
+            failed.join(", ")
+        );
+    }
     Ok(())
 }
 
 async fn cmd_drain(controller: &str, node_pattern: String, reason: Option<String>) -> Result<()> {
-    let mut client = spur_proto::controller_client(spur_client::connect_channel(controller).await?);
     let nodes = expand_node_pattern(&node_pattern)?;
+    let mut client = spur_proto::controller_client(spur_client::connect_channel(controller).await?);
 
+    let mut failed: Vec<String> = Vec::new();
     for node in &nodes {
-        let resp = client
+        match client
             .drain_node(spur_proto::proto::DrainNodeRequest {
                 name: node.to_string(),
                 reason: reason.clone().unwrap_or_default(),
             })
-            .await?
-            .into_inner();
-
-        if resp.running_jobs > 0 {
-            println!(
-                "Node {node} set to draining ({} running job{} will finish first)",
-                resp.running_jobs,
-                if resp.running_jobs == 1 { "" } else { "s" }
-            );
-        } else {
-            println!("Node {node} set to drain");
+            .await
+        {
+            Ok(resp) => {
+                let resp = resp.into_inner();
+                if resp.running_jobs > 0 {
+                    println!(
+                        "Node {node} set to draining ({} running job{} will finish first)",
+                        resp.running_jobs,
+                        if resp.running_jobs == 1 { "" } else { "s" }
+                    );
+                } else {
+                    println!("Node {node} set to drain");
+                }
+            }
+            Err(e) => {
+                eprintln!("error: {node}: {e}");
+                failed.push(node.clone());
+            }
         }
     }
     if let Some(r) = &reason {
         println!("  reason: {r}");
+    }
+    if !failed.is_empty() {
+        bail!(
+            "failed on {} of {} node(s): {}",
+            failed.len(),
+            nodes.len(),
+            failed.join(", ")
+        );
     }
     Ok(())
 }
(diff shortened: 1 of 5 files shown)


crates/spur-cli/src/node.rs
```diff
@@ -102,90 +102,105 @@ fn parse_label_args(label_args: &[String]) -> Result<(HashMap<String, String>, V
...
+        }
     }
-    if let Some(r) = reason {
+    if let Some(r) = &reason {
         println!("  reason: {r}");
     }
     Ok(())
 }
 
+fn expand_node_pattern(pattern: &str) -> Result<Vec<String>> {
```
yansun1996, line 248: `hostlist::expand` returns empty-string elements for empty segments (`"n1,"` → `["n1", ""]`, `"n1,,n2"` → `["n1", "", "n2"]`, `""` → `[""]`), and this list is looped straight into per-node RPCs (here and in `resolve_node_names` on the scontrol path). A stray comma therefore issues an `UpdateNode`/`DrainNode`/`DeregisterNode` with `name == ""`, possibly after earlier nodes were already mutated. Could we drop empty segments and error on an all-empty result? Doing it inside `expand` itself would fix every call site at once (and matches the `.filter(|s| !s.is_empty())` already used in `create_reservation`).

crates/spur-core/src/hostlist.rs
```diff
@@ -259,4 +259,62 @@ mod tests {
...
+        assert!(expand("node[1-3").is_err());
+    }
+
+    #[test]
+    fn test_reversed_range() {
+        assert!(expand("node[5-3]").is_err());
+    }
+
+    #[test]
+    fn test_empty_string() {
```
yansun1996, line 308: This test pins `expand("")` to `vec![""]`, which locks in the empty-name behavior noted above. If empty input becomes an error (or empties are filtered), this should flip to asserting `is_err()` / an empty vec.

crates/spur-cli/src/node.rs
```diff
@@ -102,90 +102,105 @@ fn parse_label_args(label_args: &[String]) -> Result<(HashMap<String, String>, V
     Ok((set_labels, remove_labels))
 }
 
-async fn cmd_label(controller: &str, node: String, label_args: Vec<String>) -> Result<()> {
+async fn cmd_label(controller: &str, node_pattern: String, label_args: Vec<String>) -> Result<()> {
     let mut client = spur_proto::controller_client(spur_client::connect_channel(controller).await?);
```
yansun1996: `cmd_label`/`cmd_drain`/`cmd_remove` connect to the controller before calling `expand_node_pattern`, so a malformed pattern (`node[5-3]`) pays the full connection cost/latency before the local parse error surfaces. `scontrol` validates state before connecting — could the pattern expansion move above `connect_channel` here for the same fail-fast behavior?

crates/spur-cli/src/node.rs
```diff
@@ -102,90 +102,105 @@ fn parse_label_args(label_args: &[String]) -> Result<(HashMap<String, String>, V
...
+                resp.running_jobs,
+                if resp.running_jobs == 1 { "" } else { "s" }
+            );
+        } else {
+            println!("Node {node} set to drain");
+        }
     }
-    if let Some(r) = reason {
+    if let Some(r) = &reason {
         println!("  reason: {r}");
```
yansun1996: The `reason:` line prints once after the loop, so for a multi-node drain/remove it reads as if it applied only to the last node. `cmd_label` interleaves its output per node — worth matching that, or printing the reason once as a batch header before the loop.

crates/spur-cli/src/scontrol.rs
```diff
@@ -632,8 +633,21 @@ async fn parse_and_update(controller: &str, params: &[String]) -> Result<()> {
     }
 
     // Node update takes priority if NodeName is specified
-    if let Some(name) = node_name {
-        return update_node(controller, &name, node_state.as_deref(), node_reason).await;
+    if let Some(node_pattern) = node_name {
+        // Reject an invalid state before touching any node, so a typo cannot
+        // leave part of the list updated.
```
yansun1996: These two comments (and the `// Echo what the caller typed…` further down) narrate the fix's rationale; per `AGENTS.md` that context is better in the commit/PR than the source. The parse-before-connect ordering and passing the un-lowercased state are self-evident, so they could be dropped.

yansun1996 (changes requested):
Nice change — the up-front state parsing, single reused connection, and reject-instead-of-coerce for unknown states are all solid improvements, and the state-parse tests are thorough.

One correctness issue worth addressing before merge: `hostlist::expand` keeps empty segments, so a trailing/leading/doubled comma (`NodeName=n1,`, `n1,,n2`, or an empty `NodeName=`) expands to include an empty string, which then gets sent as a real single-node RPC — potentially after earlier nodes in the list have already been updated. The existing `create_reservation` path already guards against this by filtering empties; the new node paths don't. Filtering empties (ideally once, inside `expand`) would close it and keep behavior consistent.

A few smaller notes inline. Two things that might be worth a follow-up rather than this PR: the per-node loop is fail-fast (if node *k*'s RPC errors, nodes `1..k` are already changed and the rest are skipped, with no summary of what applied) — reasonable, but worth documenting the contract; and there's currently no test exercising the expansion→loop wiring itself (the new tests cover `expand` and `parse_node_state` in isolation), so the partial-application behavior is uncovered.

crates/spur-cli/src/node.rs
shiv-tyagi: Fixed. `expand()` now filters empties with `results.retain(|s| !s.is_empty())` before returning, so every caller benefits. Updated `test_empty_string` to expect an empty vec and added tests for trailing/doubled commas.

crates/spur-core/src/hostlist.rs
shiv-tyagi: Done. Flipped to `assert!(hosts.is_empty())`.

crates/spur-cli/src/node.rs
shiv-tyagi: Fixed. `expand_node_pattern()` (and `parse_label_args()` for `cmd_label`) now runs before `connect_channel` in all three functions.

crates/spur-cli/src/node.rs
shiv-tyagi: Moved the reason line into each per-node success block so it prints immediately after each node's status. Also converted all loops to best-effort (try every node, collect failures, report at end).

crates/spur-cli/src/scontrol.rs
shiv-tyagi: Removed all three.

crates/spur-cli/src/scontrol.rs
```diff
@@ -662,35 +676,61 @@ async fn parse_and_update(controller: &str, params: &[String]) -> Result<()> {
...
+            .into_iter()
+            .map(|n| n.name)
+            .collect();
+        if names.is_empty() {
+            bail!("no nodes registered in the cluster");
+        }
+        return Ok(names);
+    }
+    spur_core::hostlist::expand(pattern).context("invalid node name pattern")
+}
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed at source in `hostlist::expand()` by filtering empties before returning.

crates/spur-cli/src/node.rs
```diff
@@ -102,90 +102,105 @@ fn parse_label_args(label_args: &[String]) -> Result<(HashMap<String, String>, V
     Ok((set_labels, remove_labels))
 }
 
-async fn cmd_label(controller: &str, node: String, label_args: Vec<String>) -> Result<()> {
+async fn cmd_label(controller: &str, node_pattern: String, label_args: Vec<String>) -> Result<()> {
     let mut client = spur_proto::controller_client(spur_client::connect_channel(controller).await?);
     let (set_labels, remove_labels) = parse_label_args(&label_args)?;
+    let nodes = expand_node_pattern(&node_pattern)?;
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed. Validation now runs before `connect_channel` in all three functions.

crates/spur-cli/src/node.rs
```diff
@@ -102,90 +102,105 @@ fn parse_label_args(label_args: &[String]) -> Result<(HashMap<String, String>, V
...
-    if let Some(r) = reason {
+    if let Some(r) = &reason {
         println!("  reason: {r}");
     }
     Ok(())
 }
 
+fn expand_node_pattern(pattern: &str) -> Result<Vec<String>> {
+    spur_core::hostlist::expand(pattern).context("invalid node name pattern")
+}
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed at source in `hostlist::expand()`.

yansun1996 (approved):
Re-reviewed against the latest commit — all my earlier comments are addressed:

- Empty-segment handling fixed at source in `hostlist::expand` (the `retain(|s| !s.is_empty())`), so every call site benefits and `test_empty_string` now asserts an empty vec.
- All three node commands (and the scontrol path) validate/expand before connecting, so a malformed pattern fails fast without paying the connection cost.
- Per-node `reason:` now prints inside each node's success block.
- The narration comments are gone.

Fail-fast (state parsed before any RPC) and best-effort partial-failure reporting both look correct and are well covered by the new unit tests.

A couple of non-blocking follow-ups worth a note for later, not this PR: `ALL` is only wired into the scontrol path (`spur node drain ALL` would treat "ALL" as a literal name), and the `ALL` resolution path itself isn't unit-tested yet. Neither needs to hold up merge.

LGTM — thanks for the thorough testing writeup.

yansun1996:
Filed the `ALL` keyword follow-up noted in the review as ROCm/spur#566 — the native `spur node drain/remove/label` commands don't accept `ALL` the way `scontrol update NodeName=ALL` now does. Tracked separately so it doesn't hold up this PR.

PR #550, opened 2026-07-30
fix(spurctld): confirm batch dispatch on every node before Running

Nikhil-Nunna:
## What

An earlier attempt at this (#544, closed in favor of this PR) widened the retry window for a step (`srun`) that races a node's own dispatch, but the underlying gap is structural — `spurctld` marks a job `Running` (visible to `squeue`/`scontrol`) synchronously as soon as it assigns nodes, then dispatches `LaunchJob` to each node concurrently and out of band via a spawned task. A step targeting a node can therefore still be issued before that node's own launch has actually completed, if the gap ever exceeds a retry budget -- a retry is a mitigation, not a fix for the root timing gap.

## Approach

Batch dispatch (`sbatch`, and the `srun`-as-batch-script fallback) now confirms `LaunchJob` on every assigned node -- awaited, not spawned -- *before* `start_job` is allowed to flip the job to `Running`. This mirrors the pattern `register_allocation_on_nodes` already uses for standalone `srun`/`salloc` allocations, just against the heavier `LaunchJob` RPC (which spawns the actual process) rather than the lighter `RegisterJobAllocation`.

Because the job is now confirmed before it ever becomes visibly `Running`, a dispatch failure is handled as a `Pending`-job admission failure rather than the old `Running -> Failed/NodeFail -> Held` detour:
- A node whose prolog rejected the job is drained, unchanged from before.
- If `hold_on_prolog_fail` applies, the job is held via a new `hold_job_for_launch_failure` (a `Pending`-compatible equivalent of the old post-`Running` hold path, since there's no `Running` state to transition out of here). Interactive jobs are cancelled instead, matching the existing `PrologSlurmctld`-failure behavior.
- Otherwise the job is simply requeued to `Pending` for the next scheduler tick.

One accepted trade-off: a non-prolog dispatch failure pre-`Running` no longer goes through `requeue_after_launch_failure`'s exponential backoff/requeue-count bookkeeping -- it falls through to the next tick immediately, the same simplification `register_allocation_on_nodes`'s own failure arms already make for the `srun` path.

Because every assigned node's `spurd` agent has confirmed and registered the job locally *before* it's ever visible as `Running` anywhere, this closes the race at its root for every consumer that looks the job up post-`Running` -- not just the `srun`-step case #544 targeted, but also `srun --attach` and `exec_in_job`, which had the identical unguarded-lookup pattern.

## Behavior change

Job admission for multi-node batch jobs now waits for every node's `LaunchJob` to resolve before the job appears `Running`. Measured with a synthetic multi-node test (mocked per-node launch delay): added latency tracks the slowest single node, not the sum across nodes -- ~155ms at 2 nodes, ~178ms at 64 nodes in that harness. A partial dispatch failure now aborts admission cleanly (hold/requeue while still `Pending`) instead of a job reaching `Running` with only some nodes actually launched.

## Testing

`cargo fmt --all --check`, `cargo clippy --all-targets --locked -- -D warnings`, and `cargo test --locked` all pass, including a synthetic latency test and a rewritten dispatch-failure test suite covering the new pre-`Running` hold/requeue semantics.

### DIFF
--- a/crates/spurctld/src/scheduler_loop.rs
+++ b/crates/spurctld/src/scheduler_loop.rs
@@ -258,7 +258,74 @@ pub async fn run(cluster: Arc<ClusterManager>, raft: Arc<RaftHandle>) {
                 }
             }
 
-            // Transition job to Running
+            // Build peer_nodes list with addresses for cross-node communication
+            // and the effective per-node task count. Both are needed by the
+            // LaunchJob dispatch below, which — unlike the old fire-and-forget
+            // spawn — now runs (and is awaited) *before* the job is allowed to
+            // become visibly Running.
+            let peer_addrs: Vec<String> = all_nodes
+                .iter()
+                .filter_map(|name| {
+                    cluster
+                        .get_node(name)
+                        .and_then(|n| n.address.as_ref().map(|a| format!("{}:{}", a, n.port)))
+                })
+                .collect();
+
+            let tasks_per_node = if let Some(tpn) = spec.tasks_per_node {
+                tpn
+            } else {
+                (spec.num_tasks / spec.num_nodes.max(1)).max(1)
+            };
+
+            // Batch dispatch (plain sbatch jobs, and the srun-as-batch-script
+            // fallback) must have every assigned node confirm its LaunchJob
+            // *before* start_job flips the job to Running — otherwise squeue
+            // reports Running while a node may still be mid-launch (the race
+            // PR #544 papers over with a retry window on the agent side). The
+            // pure interactive srun_step_dispatch case has nothing left to
+            // dispatch here: register_allocation_on_nodes above already
+            // confirmed the allocation, and the actual step launch is a later
+            // RunCommand, not this LaunchJob RPC.
+            let dispatch_spec = if !spec.srun_job {
+                Some(spec.clone())
+            } else if !srun_step_dispatch {
+                let mut batch_spec = spec.clone();
+                batch_spec.srun_job = false;
+                Some(batch_spec)
+            } else {
+                None
+            };
+
+            if let Some(dspec) = dispatch_spec {
+                // The run epoch start_job_impl is about to persist for this
+                // dispatch. Safe to read ahead of that call: this iteration is
+                // the only place that can advance a Pending job's run_attempt,
+                // and nothing here yields back to another iteration for the
+                // same job in between.
+                let prospective_run_attempt = job.run_attempt.saturating_add(1);
+
+                match confirm_dispatch_on_nodes(
+                    cluster.clone(),
+                    job_id,
+                    dispatch_nodes.clone(),
+                    dspec,
+                    peer_addrs.clone(),
+                    per_node_allocs.clone(),
+                    allocated_nodelist.clone(),
+                    tasks_per_node,
+                    prospective_run_attempt,
+                )
+                .await
+                {
+                    DispatchConfirmOutcome::Aborted => continue,
+                    DispatchConfirmOutcome::Confirmed => {}
+                }
+            }
+
+            // Transition job to Running. Reached only once every assigned node
+            // has confirmed (LaunchJob for batch dispatch above, or
+            // RegisterJobAllocation for the pure interactive case above that).
             let start_result = if srun_step_dispatch {
                 cluster.start_job_impl(
                     job_id,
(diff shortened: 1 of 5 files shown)


crates/spurctld/src/scheduler_loop.rs
```diff
@@ -1204,102 +1295,64 @@ async fn dispatch_job_to_nodes(
...
-        // keyed only by job_id, a cancel still in flight would otherwise be able
-        // to terminate a freshly re-dispatched attempt on the same node. Ordering
-        // the awaited cancel ahead of the requeue closes that race — the launched
-        // processes are stopped before the job can be relaunched anywhere.
-        cancel_job_on_nodes(&cluster, job_id, &succeeded_nodes, 9).await;
-        if let Err(e) = cluster.evict_job(job_id) {
-            error!(job_id, error = %e, "failed to evict job after partial dispatch failure");
-        }
+    } else if let Err(e) = cluster.requeue_job(job_id) {
+        error!(job_id, error = %e, "failed to requeue after dispatch confirmation failure");
```
yansun1996: This `requeue_job` on a still-Pending job returns early without applying any backoff, requeue-count increment, or begin-time hold (the backoff logic only runs once a job has reached Running). Since the node here isn't drained (that only happens for prolog failures), a node that's momentarily unreachable or has no resolved address will cause this whole confirm-then-abort cycle to repeat every scheduler tick until it recovers — a tight retry loop plus a steady stream of error/warn logs. Would it be worth applying a Pending-safe backoff (e.g. a begin-time hold + requeue-count bump so `max_batch_requeue` still bounds it), or draining a node that repeatedly fails to accept a dispatch even without a prolog classification?

crates/spurctld/src/scheduler_loop.rs
```diff
@@ -3011,15 +2985,24 @@ mod tests {
...
+            // reachable from Pending. See confirm_dispatch_on_nodes's doc
+            // comment for the trade-off. The node still drains either way.
             let job = cm.get_job(job_id).unwrap();
-            assert_eq!(job.pending_reason, PendingReason::JobLaunchFailure);
-            assert_ne!(job.priority, 0);
-            assert!(job.spec.begin_time.is_some_and(|b| b > Utc::now()));
+            assert_eq!(job.state, JobState::Pending);
+            assert_eq!(
+                job.pending_reason, reason_before_dispatch,
+                "no hold/backoff is applied pre-Running when hold_on_prolog_fail is off"
```
yansun1996: This assertion effectively locks in the no-backoff behavior from the abort path above (it asserts the pending reason is unchanged after a failed dispatch). If the non-prolog retry path grows a backoff/hold, this test would need to move with it — worth updating together so it protects the intended behavior rather than the current one.

crates/spurctld/src/scheduler_loop.rs
```diff
@@ -3111,6 +3102,542 @@ mod tests {
...
+            assert!(eight_nodes >= PER_NODE_LAUNCH_DELAY);
+            assert!(sixty_four_nodes >= PER_NODE_LAUNCH_DELAY);
+
+            // ...and, crucially, stays close to it as the node count grows by
+            // 32x (2 -> 64) rather than scaling with the sum across nodes
+            // (which would be ~32x slower, i.e. ~4.8s here). The multiplier
+            // below leaves headroom for scheduler/OS jitter on a loaded box;
+            // it is not evidence of a real per-node cost.
+            assert!(
+                sixty_four_nodes < PER_NODE_LAUNCH_DELAY * 5,
```
yansun1996: This is a wall-clock upper-bound assertion (64-node confirmation under 5x the per-node delay). On a loaded CI box that can get flaky, and it adds real sleep time to the suite. The lower-bound checks (`>= PER_NODE_LAUNCH_DELAY`) already demonstrate the concurrency shape; would it be worth dropping the absolute upper bound, or gating this behind an ignored perf test?

crates/spurd/src/agent_server.rs
```diff
@@ -1652,6 +1652,11 @@ impl SlurmAgent for AgentService {
         };
         let step_id = req.step_id;
 
+        // No retry on a miss here: the controller only lets a step reach this
```
yansun1996: The reasoning here is sound for the launch race. One case it doesn't cover: if this agent restarts while a job is still Running, its in-memory running map starts empty while the controller still reports the job Running, so a lookup here can miss with no retry. Might be worth narrowing the comment to the launch race specifically, and noting the agent-restart case as out of scope (or handled elsewhere).

crates/spurd/src/agent_server.rs
```diff
@@ -1910,6 +1915,19 @@ impl SlurmAgent for AgentService {
         let req = request.into_inner();
         let job_id = req.job_id;
 
+        // No retry on a miss here, for the same reason as run_command: a job
```
yansun1996: Same thought as the run_command lookup above: the "no retry needed" guarantee holds for the launch race but not for an agent restart mid-job (the in-memory map is empty until jobs re-register), and this path is reached directly by attach/sattach with no controller-side proxy. A one-line narrowing of the claim would avoid it reading as fully race-free.

yansun1996 (commented):
Thanks for this — the core approach reads well, and confirming every node's launch before the job goes Running closes the race cleanly. I tried it out on a small 2-node setup and the happy path and prolog-failure paths behaved exactly as described. I did notice one thing I think is worth addressing before merge, plus a few smaller suggestions inline. Nothing blocking beyond the first point.

Main concern: for a non-prolog dispatch failure (e.g. a node whose agent is briefly unreachable, or a node with no resolved address), admission is aborted and the job is left Pending with no backoff, so it re-dispatches on essentially every scheduler tick until the node recovers. Prolog failures are safe here because they drain the node, but the non-prolog path isn't bounded. Details inline.

Smaller items inline: a test that appears to pin the no-backoff behavior above, a wall-clock timing assertion that may be flaky in CI, and two comments about retry-on-miss that I think slightly over-state the guarantee. Really nice work overall.

yansun1996 (dismissed):

yansun1996 (approved):

PR #551, opened 2026-07-30
fix(auth): enforce job ownership on exec, attach, and output streaming

shiv-tyagi:
## Summary

Users could `spur exec` or `sattach` into jobs they do not own, executing commands or attaching interactive shells as the job owner. This closes two privilege-escalation vulnerabilities:

- **SPUR-108**: `spur exec <jobid> <cmd>` allowed any user to run commands inside another user's job.
- **SPUR-109**: `srun --jobid=N --overlap --pty` / `sattach` allowed any user to attach an interactive PTY to another user's job.

## Approach

Dual-layer ownership enforcement at both controller and agent:

- **Controller** gates `ExecInJob` and `CreateJobStep` RPCs before dispatching to the agent.
- **Agent** gates `ExecInJob`, `InteractiveSession`, and `StreamJobOutput` RPCs directly (defense-in-depth for direct-to-agent connections like `sattach`).
- Shared `spur_core::auth::check_job_owner()` implements the allow rule: empty requester (daemon/internal), root, or matching owner passes; all else denied with `PermissionDenied`.
- CLI sends the local username in each relevant proto message (`ExecInJobRequest`, `CreateJobStepRequest`, `StreamJobOutputRequest`, `InitSession`, `RegisterJobAllocationRequest`).

## Design choices

- **Empty-owner jobs are root-only**: Jobs submitted via REST without a `user` field, or allocations registered by older controllers during rolling upgrades, have no recorded owner. These jobs run as root (the executor never drops privileges for uid 0), so granting access to any caller would be a privilege escalation. Only root and daemon (empty user) can exec/attach/stream/cancel them. During a rolling upgrade this means the real submitter temporarily needs root to interact with their job; this is acceptable since the alternative is handing root shells to any local user.

- **Self-reported username**: The `user` field is trust-on-first-use from the CLI, matching what `scancel` already did. It raises the bar from "no check at all" to "must lie deliberately", and is not a substitute for authenticated identity. Real enforcement requires the mTLS/JWT layer. On lookup failure the CLI sends a value that cannot be a UNIX username, so a `whoami` failure denies rather than colliding with a real account.

- **`RunStep`/`RunCommand` remain ungated**: These carry no owner and are not gated here. `srun` dispatches `RunStep` directly with a caller-supplied `job_id`, so this is a real gap, but a pre-existing one that this PR neither introduces nor widens. Closing it properly means giving `JobStep` an owner and threading identity through both handlers, which belongs with the ownership framework rather than bolted onto this fix. Tracked separately.

## Test plan

- [x] Unit tests in `spur-core` covering owner, root, daemon (empty user), empty owner denial, and denial cases
- [x] Controller integration tests for `exec_in_job` and `create_job_step` authorization, including empty-owner denial
- [x] Agent integration tests for `check_job_access` on the attach and stream paths, including empty-owner denial
- [x] Full 38-point authorization matrix on the bare-metal testbed, with `spurd` running as root so the privilege drop is real: owner allowed, cross-user denied, root allowed, `scancel` ownership, empty-owner escalation blocked, edge cases (non-existent job, cancelled job), standalone `srun` allocation propagation
- [x] Pre-fix binaries confirmed to reproduce both tickets, so the post-fix denials are not vacuous
- [x] `cargo test --locked` all pass

### DIFF
--- a/crates/spur-cli/src/interactive.rs
+++ b/crates/spur-cli/src/interactive.rs
@@ -19,10 +19,10 @@ pub async fn connect_agent(addr: &str) -> Result<SlurmAgentClient<tonic::transpo
 }
 
 /// Local username sent with exec/attach requests so the server can enforce job
-/// ownership. Falls back to a non-matching placeholder so lookup failure denies
-/// access rather than granting it.
+/// ownership. The fallback is deliberately not a legal UNIX username, so a
+/// lookup failure cannot collide with a real account and grant access.
 pub fn current_user() -> String {
-    whoami::username().unwrap_or_else(|_| "unknown".into())
+    whoami::username().unwrap_or_else(|_| "<lookup-failed>".into())
 }
 
 pub fn get_terminal_size() -> spur_proto::proto::WindowSize {
--- a/crates/spur-cli/src/interactive.rs
+++ b/crates/spur-cli/src/interactive.rs
@@ -18,6 +18,13 @@ pub async fn connect_agent(addr: &str) -> Result<SlurmAgentClient<tonic::transpo
         .max_encoding_message_size(spur_proto::MAX_GRPC_REQUEST_SIZE))
 }
 
+/// Local username sent with exec/attach requests so the server can enforce job
+/// ownership. Falls back to a non-matching placeholder so lookup failure denies
+/// access rather than granting it.
+pub fn current_user() -> String {
+    whoami::username().unwrap_or_else(|_| "unknown".into())
+}
+
 pub fn get_terminal_size() -> spur_proto::proto::WindowSize {
     let (cols, rows) = crossterm::terminal::size().unwrap_or((80, 24));
     spur_proto::proto::WindowSize {
@@ -54,6 +61,7 @@ pub async fn open_interactive_session(
             winsize: Some(winsize),
             argv,
             env: HashMap::new(),
+            user: current_user(),
         })),
     };
 
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -726,6 +726,7 @@ async fn dispatch_step(
             pty: false,
             winsize: None,
             node: String::new(),
+            user: crate::interactive::current_user(),
         })
         .await
         .context("failed to create job step")?
@@ -1007,6 +1008,7 @@ async fn try_stream_output(
         .stream_job_output(StreamJobOutputRequest {
             job_id,
             stream: "stdout".into(),
+            user: crate::interactive::current_user(),
         })
         .await
     {
@@ -1219,6 +1221,7 @@ async fn run_interactive_pty(
                     pty: true,
                     winsize: Some(winsize),
                     node: node.clone(),
+                    user: crate::interactive::current_user(),
                 })
                 .await
             {
--- a/crates/spur-core/src/auth.rs
+++ b/crates/spur-core/src/auth.rs
@@ -88,13 +88,12 @@ impl Identity {
 
 /// Check that `user` is allowed to perform `action` on a job owned by `owner`.
 ///
-/// Allows the owner, root, an empty `user` (daemon calls and clients predating
-/// the identity field), and an empty `owner` (an unattributed job matches no
-/// caller, so denying would lock out its real owner rather than an intruder).
-/// A non-empty placeholder owner is not covered by that last rule and so stays
-/// restricted to root.
+/// Allows the owner, root, or an empty `user` (daemon calls and clients
+/// predating the identity field). Jobs with an empty owner are restricted to
+/// root only, since they run as root and granting access to any caller would
+/// be a privilege escalation.
 pub fn check_job_owner(user: &str, owner: &str, action: &str) -> Result<(), AuthError> {
-    if user.is_empty() || owner.is_empty() || user == "root" || user == owner {
+    if user.is_empty() || user == "root" || user == owner {
         return Ok(());
     }
     Err(AuthError::NotJobOwner {
@@ -249,14 +248,16 @@ mod tests {
         );
     }
 
-    /// A job with no recorded submitter must stay reachable by its real owner.
-    /// Reachable via REST submission without `user`, and via an allocation
-    /// registered by a controller predating the identity field.
+    /// Jobs with an empty owner run as root, so only root and daemon (empty
+    /// user) are allowed. A regular user must be denied.
     #[test]
-    fn test_check_job_owner_allows_any_user_when_owner_unrecorded() {
-        assert!(check_job_owner("alice", "", "exec").is_ok());
-        assert!(check_job_owner("bob", "", "attach to").is_ok());
+    fn test_check_job_owner_empty_owner_restricts_to_root() {
+        assert!(check_job_owner("root", "", "exec").is_ok());
         assert!(check_job_owner("", "", "exec").is_ok());
+        assert!(
+            check_job_owner("alice", "", "exec").is_err(),
+            "empty-owner jobs run as root; granting access is a privilege escalation"
+        );
     }
 
     /// A non-empty placeholder owner matches no caller, so it restricts the job
--- a/crates/spur-core/src/auth.rs
+++ b/crates/spur-core/src/auth.rs
@@ -92,8 +92,15 @@ impl Identity {
 /// before the identity field existed send nothing; root is an admin override.
 /// Callers that reach this from a self-reported client field get only as much
 /// assurance as that field carries.
+///
+/// An empty `owner` is also accepted: a job whose submitter was never recorded
+/// (REST submission without `user`, or an allocation registered by a controller
+/// predating the field) is unattributable, so it cannot be matched against any
+/// caller. Denying it would lock the real owner out of their own job instead of
+/// stopping an intruder. Note this reasoning does not extend to a non-empty
+/// placeholder owner, which matches nobody and so restricts the job to root.
 pub fn check_job_owner(user: &str, owner: &str, action: &str) -> Result<(), AuthError> {
-    if user.is_empty() || user == "root" || user == owner {
+    if user.is_empty() || owner.is_empty() || user == "root" || user == owner {
         return Ok(());
     }
     Err(AuthError::NotJobOwner {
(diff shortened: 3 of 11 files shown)


crates/spur-core/src/auth.rs
```diff
@@ -86,6 +86,30 @@ impl Identity {
...
+/// Callers that reach this from a self-reported client field get only as much
+/// assurance as that field carries.
+///
+/// An empty `owner` is also accepted: a job whose submitter was never recorded
+/// (REST submission without `user`, or an allocation registered by a controller
+/// predating the field) is unattributable, so it cannot be matched against any
+/// caller. Denying it would lock the real owner out of their own job instead of
+/// stopping an intruder. Note this reasoning does not extend to a non-empty
+/// placeholder owner, which matches nobody and so restricts the job to root.
+pub fn check_job_owner(user: &str, owner: &str, action: &str) -> Result<(), AuthError> {
```
yansun1996, line 95: The empty-`owner` bypass is a reasonable back-compat choice, but two consequences may be worth confirming: (a) it's a behavior change for `cluster.rs` — the old local check rejected a non-owner on an empty-owner job, and `cancel`/`complete`/`suspend`/`resume` all route through this delegate now, so any user can act on an unattributed-owner job where they previously couldn't; and (b) in a current deployment, a REST submit that omits `user` records an empty owner, so those jobs become exec/attach/stream-able by anyone. Both may be intended given JWT is coming — but they'd be worth calling out in the PR description, and maybe a test pinning the widened `cancel` behavior so it's a decision rather than a side effect of the refactor.

crates/spurctld/src/cluster.rs
```diff
@@ -650,15 +649,7 @@ impl ClusterManager {
     /// Check that `user` is allowed to perform `action` on a job owned by `owner`.
     /// Empty user (internal/daemon calls) and root are always allowed.
```
yansun1996: Since this now delegates to the shared `check_job_owner`, the doc comment ("Empty user … and root are always allowed") no longer mentions the empty-*owner* bypass it inherits. Worth updating to match.

crates/spur-cli/src/interactive.rs
```diff
@@ -18,6 +18,13 @@ pub async fn connect_agent(addr: &str) -> Result<SlurmAgentClient<tonic::transpo
         .max_encoding_message_size(spur_proto::MAX_GRPC_REQUEST_SIZE))
 }
 
+/// Local username sent with exec/attach requests so the server can enforce job
+/// ownership. Falls back to a non-matching placeholder so lookup failure denies
+/// access rather than granting it.
+pub fn current_user() -> String {
+    whoami::username().unwrap_or_else(|_| "unknown".into())
```
yansun1996: The `"unknown"` fallback on `whoami` failure is a legal UNIX username, so in the (rare) lookup-failure case it could match a job actually owned by `unknown`. A sentinel that can't be a username (or hard-failing on lookup failure) would deny more reliably. Low likelihood, cheap to harden.

crates/spur-core/src/auth.rs
```diff
@@ -86,6 +86,30 @@ impl Identity {
     }
 }
 
+/// Check that `user` is allowed to perform `action` on a job owned by `owner`.
```
yansun1996, line 89: This doc block runs ~13 lines. The empty-user/empty-owner rationale is genuinely useful *why*, but per the repo's comment conventions it could collapse to a line or two with the detail moving to the commit/PR body.

yansun1996 (changes requested):
This is a solid, well-tested set of gates — the dual-layer controller + agent checks, the shared `check_job_owner`, and the owner/root/unattributed test matrix all look good.

**Main item worth resolving before merge — the `RunStep` → `RunCommand` path looks like it reopens the same hole.** `srun` dispatches `RunStep` directly (`spur-cli/src/srun.rs:748`) with a caller-supplied `job_id`, `command`, and `uid`/`gid`, and neither the controller `run_step` (`spurctld/src/server.rs`, ~line 1580) nor the agent `run_command` handler checks ownership — they only verify the job exists, has allocated nodes, and that the step id exists (and `JobStep` carries no owner). So a client can target another user's `job_id` with `step_id` 0, skip the now-gated `CreateJobStep`/`ExecInJob`, and run an arbitrary command inside the victim's allocation — and because `run_command` privdrops to the request's `uid`/`gid`, potentially as an arbitrary uid. `RunStepRequest`/`RunCommandRequest` are the only running-job RPCs that didn't get a `user` field; threading one in (new proto tags) and gating both handlers the same way as `exec_in_job` would close it. (Flagging here rather than inline since these handlers are outside the diff.)

A couple of smaller consistency points around the empty-owner fail-open (it also widens `scancel`/suspend/resume/complete, since those delegate to the same shared fn), and a note on the self-reported `user` field — which the PR already scopes honestly as the pre-identity-layer posture. Details inline. Nothing blocking beyond the first item.

On tests: if the `RunStep`/`RunCommand` gate is added, a cross-user DENY test on that path would be the highest-value addition. Two smaller gaps in the current set — the `interactive_session` handler's gate placement is only exercised via the `check_job_access` helper (not the handler), and there's no test asserting the CLI actually populates a non-empty `user` in the outgoing request, so a regression to an empty `user` would pass silently.

crates/spur-cli/src/interactive.rs
```diff
@@ -18,6 +18,13 @@ pub async fn connect_agent(addr: &str) -> Result<SlurmAgentClient<tonic::transpo
...
+}
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed. The fallback is now `"<lookup-failed>"` (contains angle brackets, not a legal UNIX username) so lookup failure reliably denies rather than accidentally matching a real account.

crates/spurctld/src/cluster.rs
```diff
@@ -650,15 +649,7 @@ impl ClusterManager {
...
-        if user.is_empty() || user == "root" || user == owner {
-            return Ok(());
-        }
-        Err(AuthError::NotJobOwner {
-            user: user.into(),
-            owner: owner.into(),
-            action: action.into(),
-        }
-        .into())
+        spur_core::auth::check_job_owner(user, owner, action).map_err(Into::into)
```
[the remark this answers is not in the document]
shiv-tyagi: Updated. The doc now reads: "Delegates to `spur_core::auth::check_job_owner`; see there for the bypass rules."

Also, the empty-owner bypass no longer exists (removed in 0e28883). Empty-owner jobs are now restricted to root only.

crates/spurd/src/agent_server.rs
```diff
@@ -2426,6 +2440,20 @@ impl AgentService {
         });
     }
 
+    /// Reject `user` unless they own job `job_id`.
+    ///
+    /// Enforced here as well as on the controller because `sattach` and the
+    /// output stream dial the agent's port directly.
```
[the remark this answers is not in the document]
shiv-tyagi: Reworded to: "Gate `user` for `action` on job `job_id`, per `spur_core::auth::check_job_owner`." This references the shared function directly instead of restating (potentially incorrectly) the rules it enforces.

crates/spur-core/src/auth.rs
shiv-tyagi: Both concerns resolved:

(a) The empty-owner bypass is removed entirely (0e28883). `check_job_owner` no longer has `owner.is_empty()` in the allow path. Empty-owner jobs (REST submissions without `user`) are now restricted to root only, since they run as root and granting access to any caller would be privilege escalation.

(b) Because the bypass is gone, the widening of `cancel`/`suspend`/`resume` on unattributed jobs no longer applies. Non-root users are denied on empty-owner jobs across all gated paths.

The PR description has been updated to document this decision and the rolling-upgrade tradeoff.

crates/spurctld/src/cluster.rs
shiv-tyagi: Done. Doc now says: "Delegates to `spur_core::auth::check_job_owner`; see there for the bypass rules." The empty-owner bypass it previously inherited has also been removed (0e28883), so this is now accurate.

crates/spur-cli/src/interactive.rs
shiv-tyagi: Changed to `"<lookup-failed>"`. The angle brackets make it an illegal UNIX username, so it cannot collide with any real account.

crates/spur-core/src/auth.rs
shiv-tyagi: Trimmed to 4 lines. The rationale for the empty-owner restriction and the rolling-upgrade tradeoff now lives in the commit message for 0e28883.

shiv-tyagi:
Addressed all inline comments. A few items from the review body:

**RunStep/RunCommand ownership gap**

Agreed this is a real gap. The `RunStep` controller handler and `RunCommand` agent handler accept a caller-supplied `uid`/`gid` and do not verify the caller owns the job. An attacker who knows a valid `job_id` + `step_id` (both sequential integers) can craft a raw gRPC `RunStepRequest` and get code execution inside the victim's allocation.

This is preexisting: these handlers never had ownership checks, and they are outside the diff for this PR. The proper fix is adding a `user` field to `RunStepRequest`/`RunCommandRequest` and gating both handlers the same way we gate `ExecInJob`. Will file a follow-up issue.

**Test gaps (interactive_session handler, CLI populating non-empty user)**

Fair points. The `interactive_session` handler's gate is exercised via the `check_job_access` helper test and the end-to-end bare-metal smoke tests (which confirmed denial and success on real sattach calls), but not via a unit test that calls the handler directly. Similarly, there's no unit test asserting the CLI sends a non-empty `user`. Both are worth adding but are additive coverage, not correctness issues in the current code. Will address in a follow-up.

**Empty-owner bypass**

Removed entirely in 0e28883. See inline replies for details.

yansun1996:
Thanks for the quick turnaround — inline items all addressed, and empty-owner now denies non-root at both layers with unit + handler coverage. Two small things left:

**1. PR description still describes the old fail-open behavior.** The "Fail-open on empty owner" bullet still says empty-owner jobs *"stay accessible rather than becoming root-only"* — the opposite of the shipped rule. This squash-merges into the commit message, so worth updating.

**2. Test follow-ups.** Agreed on the CLI-populates-`user` and `interactive_session` handler-gate tests — the empty-user bypass means a dropped `user` fails *open*, so the CLI test earns its keep.

One more on the `srun.rs` owner fallback, left inline. Nothing blocking once [1] is sorted.

crates/spur-cli/src/srun.rs
```diff
@@ -726,6 +726,7 @@ async fn dispatch_step(
             pty: false,
             winsize: None,
             node: String::new(),
+            user: crate::interactive::current_user(),
```
yansun1996, line 729: Minor owner-fallback inconsistency: `build_srun_job_spec` records the job owner via `whoami::username().unwrap_or_else(|_| "unknown")`, whereas `current_user()` (used here and on the exec/attach paths) falls back to `"<lookup-failed>"`. On a `whoami` failure the job is recorded as owned by `"unknown"` but a later `exec` / `sattach` / `srun --pty` step requests as `"<lookup-failed>"`, so the real owner is denied their own job. Might be worth routing owner-recording through the same `current_user()` sentinel. Pre-existing and low-likelihood, so fine as a follow-up.

shiv-tyagi: Good catch on the inconsistency. This is preexisting and outside this PR's scope (the owner-recording paths in sbatch/srun/salloc all predate this change). The scenario itself requires `whoami::username()` to fail, which is extremely unlikely in practice (needs a uid with no passwd entry). Will address as a follow-up.

yansun1996 (approved):
lgtm pls file a ticket so that we won't lose the record of that finding

shiv-tyagi:
Sure. Thanks @yansun1996.

shiv-tyagi:
Filed #567

PR #552, opened 2026-07-30
fix(spurd): run sbatch batch script once per node, not once per task

### LINKED ISSUE #535: sbatch executes the batch script once per task instead of once per allocation
## Problem

`sbatch` executes the entire batch script — every line, including setup and cleanup steps, not just the intended workload — once per task implied by `--ntasks-per-node` / `--ntasks`, concurrently, on every allocated node. No `srun` call is required to trigger this; it happens automatically based on the task count alone.

This comes from spurd's batch launch path, which replaces the user's script with a generated wrapper whenever the per-node task count exceeds 1:

- `crates/spurd/src/agent_server.rs:697-702` derives `tasks_per_node` from the job spec, falling back to `num_tasks / num_nodes`.
- `crates/spurd/src/agent_server.rs:889-917` writes the user's script to `.spur_user_<job_id>.sh` and substitutes `spur_core::task_launch::build_multi_task_wrapper(...)` in its place.
- `crates/spur-core/src/task_launch.rs:310-341` is that wrapper: a `for LOCAL_RANK in $(seq 0 $((_TASKS_ON_NODE - 1)))` loop that backgrounds one `bash "<user script>"` per rank and then `wait`s.
- `crates/spurctld/src/scheduler_loop.rs:1113-1156` dispatches the launch to every allocated node with `task_offset = node_idx * tasks_per_node`, so the replication also multiplies across nodes — a 2-node, 8-tasks-per-node job runs the whole script 16 times.

Standalone `srun` does not have this problem — on native-host nodes it goes through a separate, correct per-task dispatch path (`crates/spurctld/src/scheduler_loop.rs:212-259`). Only the raw batch-script body is affected.

## Expected

Under Slurm semantics, `sbatch` runs the batch script exactly once, on the first allocated node, regardless of `--ntasks` / `--ntasks-per-node`. Task multiplicity is advertised to the script purely through environment variables (`SLURM_NTASKS`, `SLURM_NTASKS_PER_NODE`, `SLURM_NNODES`, ...); fanning out to multiple concurrent processes is something the user's script does explicitly, typically via an `srun` call inside it. A script with sequential setup → payload → cleanup steps and no `srun` call should run each step exactly once, not N times concurrently.

## Reproduce

```bash
#!/bin/bash
#SBATCH -N 1
#SBATCH --ntasks-per-node=4
#SBATCH -t 00:02:00

echo "=== SETUP pid=$$ rank=$SLURM_PROCID ==="
mkdir "myonlydir_${SLURM_JOB_ID}" 2>&1
echo "=== PAYLOAD pid=$$ rank=$SLURM_PROCID ==="
echo "=== CLEANUP pid=$$ rank=$SLURM_PROCID ==="

### PR DESCRIPTION
**Pull request by Nikhil-Nunna**
## What

Closes #535. When a job requests multiple tasks per node (`--ntasks-per-node=K`), `AgentService::launch_job` unconditionally wrapped the batch script in a multi-copy launcher whenever `tasks_per_node > 1`, forking K concurrent copies of the *entire* script per node. Real Slurm runs a batch script exactly once per allocation; the script itself is responsible for any internal fan-out via `srun`. Any script with more than a single trivial command (setup, file writes, cleanup) silently races itself K times under the old behavior.

## Approach

Added a `task_fanout` field to `LaunchJobRequest` so the controller can tell the agent whether a dispatch is a genuine multi-task fan-out or a plain batch script:

- `task_fanout=true`: set only for standalone `srun` requests that fall back to the batch dispatch path (e.g. Kubernetes-inclusive allocations, where the dispatched "command" is literally what should repeat N times) -- unchanged multi-copy behavior.
- Everything else -- a plain `sbatch` script, regardless of `--mpi` value -- now runs exactly once per node. `SPUR_TASKS_PER_NODE`/`LOCAL_WORLD_SIZE`/`NPROC_PER_NODE` are still set so the script can fan out itself via `srun` if it needs to.

This is scoped to intra-node duplication only. Per-*node* dispatch (once per node, with `NODE_RANK`/`MASTER_ADDR`/`WORLD_SIZE` set, for multi-node distributed scripts) is untouched -- internal `srun` calls issued from within a running batch script require the job to already be tracked on every allocated node, so collapsing dispatch to a single head node would break multi-node internal `srun` fan-out. That's a separate, more invasive change and out of scope here.

**Scope note (per review):** an earlier version of this PR also kept the old multi-copy behavior for `--mpi=pmix`/`--mpi=pmi1` batch jobs, to preserve existing multi-rank PMIx launch behavior. That carve-out has been dropped at @sgopinath1's request, since it re-introduces the same "replicate the whole script per task" model this PR removes for plain `sbatch`, and would collide with upcoming multi-node PMIx work. Multi-rank direct PMIx batch launch is intentionally left to a follow-up MPI PR.

## Behavior change

A plain `sbatch` script requesting multiple tasks per node now executes once, not once per task -- this is an intentional breaking-behavior fix, not something gated behind a flag. The only legitimate prior use of the multi-copy path (the `srun`-fallback dispatch case) is explicitly preserved.

This now also includes `--mpi=pmix`/`--mpi=pmi1` batch jobs: `sbatch --mpi=pmix -n4 batch.sh` no longer forks the script 4x per node. Users who want multi-rank PMIx launched directly from a batch script should instead do the fan-out explicitly inside the script, e.g.:

```bash
#!/bin/bash
#SBATCH --mpi=pmix
srun --mpi=pmix -n4 ./hello_mpi
```

(or equivalent step dispatch). Direct multi-rank PMIx batch launch (without an internal `srun`) can land in the follow-up multi-node MPI PR.

## Testing

Regression tests in `agent_server.rs`, modeled on the reproducer from #535 (a counter plus a colliding `mkdir` to detect concurrent re-execution): one confirms a plain script now runs exactly once regardless of `--ntasks-per-node` (verified this fails on the old code, passes after the fix), one confirms `task_fanout` dispatch still replicates correctly, and one confirms a documented `--mpi=pmix` batch job is no longer a special case and also runs once. Also corrected a native-host e2e test (`test_devices.py`) that had encoded the old buggy behavior as its expected outcome, and added `scheduler_loop.rs` tests asserting `task_fanout` is set correctly for both the plain-batch and srun-fallback dispatch paths.

`cargo fmt --all --check`, `cargo clippy --all-targets --locked -- -D warnings`, and `cargo test --locked --workspace` all pass. Patch coverage on the changed Rust files is 100%.

### DIFF
--- a/crates/spurd/src/agent_server.rs
+++ b/crates/spurd/src/agent_server.rs
@@ -1091,12 +1091,19 @@ impl SlurmAgent for AgentService {
             (script, crate::container::RootfsMode::Extracted)
         };
 
-        // Multi-task per-node: wrap the user script so it forks N processes,
-        // each with a distinct LOCAL_RANK. The wrapper backgrounds N copies and
-        // waits for all to finish, so TrackedJob only tracks a single PID (the
-        // wrapper shell). GPU devices are partitioned across tasks via
-        // ROCR_VISIBLE_DEVICES / CUDA_VISIBLE_DEVICES overrides in each fork.
-        let launch_script = if tasks_per_node > 1 {
+        // Under Slurm semantics a batch script runs exactly once per node,
+        // regardless of --ntasks-per-node or --mpi: task multiplicity is only
+        // advertised via environment variables (SPUR_TASKS_PER_NODE,
+        // LOCAL_WORLD_SIZE, NPROC_PER_NODE, ...) and any fan-out — including
+        // multi-rank MPI — is the script's own responsibility, typically via
+        // an internal `srun` call. The one exception is `task_fanout`: this
+        // dispatch is itself the fan-out mechanism for a standalone `srun`
+        // request routed through the batch dispatch path (e.g. an allocation
+        // spanning Kubernetes-backed nodes, which bypass the native-host
+        // per-task step path), where the dispatched "script" is the literal
+        // command `srun` was asked to run `tasks_per_node` times on this node
+        // — real srun semantics, not a batch script.
+        let launch_script = if tasks_per_node > 1 && req.task_fanout {
             // Write the user script to disk first so the wrapper can reference it
             let user_script_path = format!("{}/.spur_user_{}.sh", work_dir, job_id);
             std::fs::write(&user_script_path, &launch_script)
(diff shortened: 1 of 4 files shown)


crates/spurd/src/agent_server.rs
```diff
@@ -1089,12 +1089,26 @@ impl SlurmAgent for AgentService {
...
+        // - `req.task_fanout`: this dispatch is itself the fan-out mechanism
+        //   for a standalone `srun` request routed through the batch dispatch
+        //   path (e.g. an allocation spanning Kubernetes-backed nodes, which
+        //   bypass the native-host per-task step path). Here the "script" is
+        //   the literal command `srun` was asked to run N times.
+        // - single-node `--mpi=pmix`/`--mpi=pmi1` batch jobs: documented
+        //   convenience (`sbatch --mpi=pmix -n4 batch.sh`) that launches the
+        //   user's MPI binary as N ranks directly, without requiring an
+        //   internal `srun` call.
+        let is_multi_rank_mpi_job = spec.mpi == MPI_PMIX || spec.mpi == "pmi1";
```
yansun1996: The `|| spec.mpi == "pmi1"` clause (and the `pmi1` mentions in the nearby comment and the proto doc) advertise `--mpi=pmi1` as supported, but the CLI `--mpi` parser accepts only `none`/`pmix` and rejects everything else, so a CLI-submitted job can never have `spec.mpi == "pmi1"` — that value is only reachable via the Kubernetes CRD path (which copies the field unvalidated). Could we either drop the `pmi1` clause and its doc mentions to match what the CLI actually accepts, or keep it deliberately and note that it's the K8s-only path? As-is the three sources (proto doc, parser, this check) disagree on which MPI values exist. Minor related nit: `"pmi1"` is a bare string here while `MPI_PMIX` is a const — a `MPI_PMI1` const would keep them consistent.

crates/spurd/src/agent_server.rs
```diff
@@ -1089,12 +1089,26 @@ impl SlurmAgent for AgentService {
...
-        // waits for all to finish, so TrackedJob only tracks a single PID (the
-        // wrapper shell). GPU devices are partitioned across tasks via
-        // ROCR_VISIBLE_DEVICES / CUDA_VISIBLE_DEVICES overrides in each fork.
-        let launch_script = if tasks_per_node > 1 {
+        // Under Slurm semantics a batch script runs exactly once per node,
+        // regardless of --ntasks-per-node: task multiplicity is only advertised
+        // via environment variables (SPUR_TASKS_PER_NODE, LOCAL_WORLD_SIZE,
+        // NPROC_PER_NODE, ...) and any further fan-out is the script's own
+        // responsibility, typically via `srun`. Replicating the whole script
+        // per task here (see #535) silently reran setup/cleanup steps N times
```
yansun1996: Small convention note: the contributor guidelines ask that issue/PR numbers not appear in code comments (they belong in the commit message / PR description). There are a few `(see #535)` / issue-URL references in the new comments here and in scheduler_loop.rs and the e2e test — worth dropping them to keep the source clean.

crates/spurd/src/agent_server.rs
yansun1996: Coverage suggestion (not blocking): the `is_multi_rank_mpi_job` carve-out is the one behavior this change deliberately preserves, but I don't see a direct test for it — a small unit test asserting that `mpi=pmix` with `tasks_per_node>1` and `task_fanout=false` still fans out would guard against a future edit here silently regressing pmix batch jobs. A multi-node once-per-node integration test would also round out the single-node unit coverage nicely.

yansun1996 (commented):
Thanks for fixing this — running a batch script once per node rather than once per task is the right behavior, and the env-var advertisement (SPUR_TASKS_PER_NODE etc.) so a script can still fan out via srun is a nice touch. I tried a multi-node multi-task job and it ran exactly once per node as intended. A couple of small things worth tidying before merge, plus one coverage suggestion — all inline, none blocking.

Main item: the MPI carve-out references `pmi1`, but the CLI `--mpi` parser only accepts `none`/`pmix`, so that branch and the surrounding docs advertise a mode users can't actually select via the CLI (it's only reachable through the Kubernetes path). Worth either dropping the `pmi1` reference or making it intentional. Details inline, along with a note on a few issue-number references in comments and a small coverage gap.

Appreciate the thorough tests and the corrected e2e expectation.

sgopinath1:
@Nikhil-Nunna, @yansun1996 Please allow me some time to review this PR.
https://github.com/ROCm/spur/pull/473 fixed the same issue with srun.
https://github.com/ROCm/spur/pull/528 supported single node MPI.
I am about to open a PR to support multi-node MPI.
Let me review this PR and make sure we are having a consistent design.

yansun1996:
> @Nikhil-Nunna, @yansun1996 Please allow me some time to review this PR. #473 fixed the same issue with srun. #528 supported single node MPI. I am about to open a PR to support multi-node MPI. Let me review this PR and make sure we are having a consistent design.

sure, @sgopinath1 , from the high level this PR is trying to make the `sbatch` and `srun` behaves similar to Slurm and reduce the  redundant  script run, if there are other PRs already covering this, we can decide which one to go. More information is in this ticket https://github.com/ROCm/spur/issues/553

sgopinath1:
@Nikhil-Nunna Thanks for fixing #535 — running the batch script once per node (instead of once per task) is the right Slurm semantics, and task_fanout is a clean way to preserve the K8s srun batch-fallback path.

Could we keep this PR scoped to the sbatch bug and drop the PMIx carve-out? The #535 fix only needs to stop unconditionally wrapping plain batch scripts when tasks_per_node > 1. The mpi == MPI_PMIX exception is a separate concern and will collide with upcoming multi-node PMIx work (PreparePmix / per-rank direct launch).

**Why drop the PMIx logic here**
- Under Slurm, a batch script runs once; multi-rank MPI is the script’s job via srun, not the launcher re-running the whole script N times.
- The PMIx exception re-introduces the same “replicate the entire script per task” model we’re fixing for plain sbatch.
- Multi-rank sbatch --mpi=pmix belongs in a dedicated MPI PR, not bundled into this compatibility fix.
- Expected behavior change (acceptable for this PR)

**Expected behavior change (acceptable for this PR)**

`sbatch --mpi=pmix -n4 batch.sh would no longer fork the script 4× on each node. Users should do:
`
```
#!/bin/bash
#SBATCH --mpi=pmix
srun --mpi=pmix -n4 ./hello_mpi

```
(or equivalent step dispatch). Direct multi-rank PMIx batch launch can land in the MPI follow-up.

Nikhil-Nunna:
> @Nikhil-Nunna Thanks for fixing #535 — running the batch script once per node (instead of once per task) is the right Slurm semantics, and task_fanout is a clean way to preserve the K8s srun batch-fallback path.
> 
> Could we keep this PR scoped to the sbatch bug and drop the PMIx carve-out? The #535 fix only needs to stop unconditionally wrapping plain batch scripts when tasks_per_node > 1. The mpi == MPI_PMIX exception is a separate concern and will collide with upcoming multi-node PMIx work (PreparePmix / per-rank direct launch).
> 
> **Why drop the PMIx logic here**
> 
> * Under Slurm, a batch script runs once; multi-rank MPI is the script’s job via srun, not the launcher re-running the whole script N times.
> * The PMIx exception re-introduces the same “replicate the entire script per task” model we’re fixing for plain sbatch.
> * Multi-rank sbatch --mpi=pmix belongs in a dedicated MPI PR, not bundled into this compatibility fix.
> * Expected behavior change (acceptable for this PR)
> 
> **Expected behavior change (acceptable for this PR)**
> 
> `sbatch --mpi=pmix -n4 batch.sh would no longer fork the script 4× on each node. Users should do: `
> 
> ```
> #!/bin/bash
> #SBATCH --mpi=pmix
> srun --mpi=pmix -n4 ./hello_mpi
> ```
> 
> (or equivalent step dispatch). Direct multi-rank PMIx batch launch can land in the MPI follow-up.

PMIx carve-out has been dropped.

yansun1996 (approved):
LGTM. The once-per-node semantics are the right call, and the `task_fanout` flag cleanly separates a genuine batch script from the srun-as-batch fallback.

I walked through the sbatch and srun cases and the behavior matches: a plain `sbatch --ntasks-per-node=N` script executes exactly once per node (and once per node for a multi-node allocation), while `srun -nN` still replicates to N tasks — so the fix lands without regressing the srun path. Dropping the pmix/pmi1 carve-out reads correctly for the stated scope, with direct multi-rank PMIx batch left to the follow-up.

sgopinath1 (approved):
LGTM

PR #554, opened 2026-07-31
fix(spurctld): report wall-time expiry as TIMEOUT

### LINKED ISSUE #546: time-limit expiry reports FAILED instead of TIMEOUT unless the job ignores SIGTERM
## What happened

A job killed by its wall-time limit lands in `FAILED` with `ExitCode=0:15`. Whether it ever reaches `TIMEOUT` depends on how the job reacts to SIGTERM.

Two jobs, same `--time=00:00:20`, same cluster, same 4-node bare-metal testbed.

Job that exits promptly on SIGTERM:

```
$ cat plain.sh
#!/bin/bash
sleep 300

$ scontrol show job 25
JobState=FAILED Reason=RaisedSignal:15(Terminated)
ExitCode=0:15 DerivedExitCode=0:0
```

Job that ignores SIGTERM and survives the grace period:

```
$ cat trap.sh
#!/bin/bash
trap "" TERM
sleep 300

$ scontrol show job 26
JobState=TIMEOUT Reason=None
ExitCode=-1:0 DerivedExitCode=0:0
```

The well-behaved job is the one that gets misreported.

## What was expected

Slurm records both as `TIMEOUT`. From the sacct job state codes:

- `TO TIMEOUT` job terminated upon reaching its time limit
- `F FAILED` job terminated with non-zero exit code or other failure condition

Exceeding the wall-time limit is a `TIMEOUT` regardless of which signal ended the process. A user cannot currently distinguish "ran out of time" from "crashed" without inspecting the exit signal, and the common case reports the wrong one.

## How to reproduce

1. Submit `sleep 300` with `--time=00:00:20`.
2. Wait for the limit to pass plus the grace period.
3. `scontrol show job <id>` reports `JobState=FAILED`, `ExitCode=0:15`.
4. Repeat with `trap "" TERM` at the top of the script. That job reports `JobState=TIMEOUT`.

## Notes

`enforce_time_limits` in `crates/spurctld/src/scheduler_loop.rs` is a two-phase watchdog: SIGTERM once the deadline passes, then `JobState::Timeout` plus SIGKILL after a 30 second grace period. A job that exits on the SIGTERM reports completion through the normal agent path first, and that path derives the state from the signal before the kill phase runs. Only a job that outlives the grace period is ever labelled `Timeout`.

This also reaches requeue behaviour: `--requeue` handling and the `max_batch_requeue` hold both key off `Timeout`/`NodeFail`, so jobs that die cleanly on SIGTERM skip the timeout requeue path as well.

Reproduced on 4-node bare metal at 1d2b9bb. Also present on upstream/main at 7a0b2c9.

maybeharshit:
## Motivation

A job killed by its wall-time limit landed in `FAILED` with `ExitCode=0:15`, and whether it ever reached `TIMEOUT` depended on how it reacted to SIGTERM — the well-behaved job was the one misreported.

The time-limit watchdog tracked which jobs it had signalled in two in-memory maps (`warned_jobs`, `warn_times`) owned by the scheduler task, so the SIGTERM decision was invisible to the completion path. A job that exits promptly on that SIGTERM is finalized from its agent's completion report, where `Job::derived_completion` sees only "signaled" and returns `Failed`. Only a job that ignored SIGTERM survived to the kill phase, which sets `Timeout` explicitly.

Fixes #546.

## Technical Details

The expiry is now recorded on the job itself as `time_limit_signaled_at`, written through a new `JobTimeLimitSignaled` WAL entry before the SIGTERM goes out.

Replicating it is not optional: completions are finalized inside the Raft apply, so a non-replicated flag could not be consulted there without replicas disagreeing on the outcome. Deriving the grace period from the same field then falls out for free — it removes both in-memory maps and makes the window survive a leadership change, which previously restarted it from scratch.

Two smaller pieces follow from that:

- `Job::completion_verdict` decides the final state and reason together, so they cannot drift. `Completing -> Timeout` was missing from the state machine, since every completion routes through `Completing` (Slurm's `JOB_TIMEOUT | JOB_COMPLETING`).
- `PendingReason::TimeLimit` mirrors Slurm's `FAIL_TIMEOUT`, replacing the `RaisedSignal:15(Terminated)` that would otherwise sit incongruously next to `JobState=TIMEOUT`.

Design choices worth flagging:

[1] The agent's real exit status survives the verdict rather than being normalized, so a job that ended on SIGTERM stays distinguishable from one the watchdog had to SIGKILL. The forced kill-phase path keeps its synthetic `-1`, because it is the backstop that guarantees finalization when no agent report is coming.

[2] An OOM kill still outranks the time limit. It is direct kernel evidence of a distinct failure the user has to act on, whereas the timeout is the controller's own reason for signalling.

[3] Correct attribution routes these jobs into the existing `Timeout` requeue path, which `--requeue` jobs dying on SIGTERM previously skipped. The issue names this as affected behaviour. `time_limit_signaled_at` is cleared in `clear_run_state_for_requeue`, so a requeued run starts with a fresh budget instead of being marked timed out the moment it ends.

## Test Plan

- Unit tests in `spur-core` for the verdict itself: SIGTERM death, a trapped-SIGTERM clean exit, an unrelated signal death, OOM precedence, and the `Completing -> Timeout` transition.
- Unit tests in `spurctld` driving the real WAL apply: completion after a signal reports `Timeout`/`TimeLimit`, a signal arriving after the run ended is a no-op, the forced kill path carries the reason, and requeue clears the marker.
- New `tests/native_host/e2e/test_time_limit.py` covering the issue's two scripts plus a control job that fails inside its limit.
- Clippy and the full workspace suite.
- Live two-node LXD cluster (Ubuntu 24.04), `--time=00:00:20`.

## Test Result

Clippy clean; full workspace suite passes.

On the LXD cluster:

| job | script | result |
| --- | --- | --- |
| `tl-plain` | `sleep 300` | `JobState=TIMEOUT Reason=TimeLimit ExitCode=143:0` |
| `tl-trap` | `trap "" TERM; sleep 300` | `JobState=TIMEOUT Reason=TimeLimit ExitCode=-1:0` |
| `tl-fastfail` | `exit 3` | `JobState=FAILED Reason=NonZeroExitCode ExitCode=3:0` |

The controller log shows `tl-plain` reaching `TIMEOUT` with no `grace period expired` line, confirming it was finalized from the agent's completion report — the exact path that used to yield `FAILED`. `tl-trap` was force-killed 30s after its SIGTERM, as before.

`ExitCode` for `tl-plain` is `143:0` rather than the issue's `0:15` because this environment's bash exits `128+15` instead of dying from the signal. That is a property of the shell, not of the fix, so the e2e test only rules out the synthetic `-1` rather than asserting an exact status.

A `--requeue` job at the same limit was requeued repeatedly with `from=TIMEOUT`, each run logging a full `elapsed_secs=27` against `limit_secs=20`, which confirms both the new requeue routing and that the marker is cleared between runs.

## Submission Checklist

### DIFF
--- a/crates/spur-core/src/wal.rs
+++ b/crates/spur-core/src/wal.rs
@@ -68,6 +68,15 @@ pub enum WalOperation {
         exit_code: i32,
         signal: i32,
     },
+    /// The time-limit watchdog signalled a running job for exhausting its wall
+    /// clock. Durable so the grace period survives a leadership change and so
+    /// every replica finalizes the run as `Timeout` rather than reading the
+    /// terminating signal as an ordinary failure. `at` is stamped on the leader
+    /// so replicas share one instant instead of consulting their own clocks.
+    JobTimeLimitSignaled {
+        job_id: JobId,
+        at: chrono::DateTime<chrono::Utc>,
+    },
     /// An srun job step finished. Records the step's exit code durably so the
     /// job's DerivedExitCode (running max over steps) survives restart/replay.
     JobStepComplete {
--- a/crates/spur-core/src/wal.rs
+++ b/crates/spur-core/src/wal.rs
@@ -721,6 +721,24 @@ mod suspend_wal_tests {
             }
         }
     }
+
+    #[test]
+    fn job_time_limit_signaled_op_round_trips() {
+        let at = chrono::Utc::now();
+        let op = WalOperation::JobTimeLimitSignaled { job_id: 13, at };
+        let json = serde_json::to_string(&op).unwrap();
+        let back: WalOperation = serde_json::from_str(&json).unwrap();
+        match back {
+            WalOperation::JobTimeLimitSignaled {
+                job_id,
+                at: at_back,
+            } => {
+                assert_eq!(job_id, 13);
+                assert_eq!(at_back, at);
+            }
+            _ => panic!("wrong variant"),
+        }
+    }
 }
 
 #[cfg(test)]
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -991,6 +991,16 @@ impl ClusterManager {
         Ok(())
     }
 
+    /// Record that a running job has exhausted its time limit, before the
+    /// caller sends SIGTERM. Durably marking the run first is what lets the
+    /// completion path report `Timeout` instead of reading the terminating
+    /// signal as an ordinary failure — a job that exits promptly on SIGTERM
+    /// reports back long before the grace period is up.
+    pub fn signal_time_limit(&self, job_id: JobId, at: DateTime<Utc>) -> anyhow::Result<()> {
+        self.propose(WalOperation::JobTimeLimitSignaled { job_id, at })?;
+        Ok(())
+    }
+
     /// Preempt a running job per its partition's PreemptMode. Does the
     /// controller-side state change; the caller dispatches the signal named by
     /// the returned `PreemptOutcome`. `Off` is rejected.
@@ -3187,6 +3197,7 @@ impl ClusterManager {
         job.allocated_nodes.clear();
         job.allocated_resources = None;
         job.per_node_alloc.clear();
+        job.time_limit_signaled_at = None;
         job.set_pending_reason(PendingReason::None);
         // Stale after requeue (points at nodes the job left); next dispatch resets it.
         job.actual_stdout_path = None;
@@ -3589,11 +3600,8 @@ impl ClusterManager {
                         let (derived_state, final_exit, raw_signal) =
                             Job::derived_completion(&job.node_completions, &primary);
                         let final_signal = raw_signal & !spur_core::job::OOM_SIGNAL_FLAG;
-                        let final_state = if oom {
-                            JobState::OutOfMemory
-                        } else {
-                            derived_state
-                        };
+                        let (final_state, final_reason) =
+                            job.completion_verdict(derived_state, final_exit, final_signal, oom);
                         match job.transition(final_state) {
                             Ok(()) => {
                                 job.exit_code = Some(final_exit);
@@ -3602,15 +3610,7 @@ impl ClusterManager {
                                 // steps, accumulated live by JobStepComplete; a
                                 // job with no srun steps keeps 0 (Slurm parity),
                                 // not the batch exit. Left as-is here.
-                                job.set_pending_reason(if oom {
-                                    PendingReason::OutOfMemory
-                                } else if final_signal != 0 {
-                                    PendingReason::RaisedSignal
-                                } else if final_exit != 0 {
-                                    PendingReason::NonZeroExitCode
-                                } else {
-                                    PendingReason::None
-                                });
+                                job.set_pending_reason(final_reason);
                                 job.end_time = Some(timestamp);
                                 job.node_completions.clear();
                                 Some((final_state, final_exit))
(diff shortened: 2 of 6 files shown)


yansun1996:

crates/spurctld/src/cluster.rs
```diff
@@ -991,6 +991,16 @@ impl ClusterManager {
...
 
+    /// Record that a running job has exhausted its time limit, before the
+    /// caller sends SIGTERM. Durably marking the run first is what lets the
+    /// completion path report `Timeout` instead of reading the terminating
+    /// signal as an ordinary failure — a job that exits promptly on SIGTERM
+    /// reports back long before the grace period is up.
+    pub fn signal_time_limit(&self, job_id: JobId, at: DateTime<Utc>) -> anyhow::Result<()> {
+        self.propose(WalOperation::JobTimeLimitSignaled { job_id, at })?;
+        Ok(())
+    }
```
shiv-tyagi, line 1016: `complete_job` doesn't consult `time_limit_signaled_at`, so two callers bypass the new verdict:

- `force_finish_completing_job` (completing-timeout path): a multi-node job where some nodes trap SIGTERM enters Completing, the watchdog skips it, and `enforce_completing_timeout` eventually force-finishes it as `Failed` instead of `Timeout`. The requeue path never fires.
- `finish_srun_job`: derives state from `exit_code` alone (`0 → Completed`, else `Failed`), never checking the marker.

Both paths call `complete_job`, so a single check there covers all current and future callers:

```suggestion
    pub fn signal_time_limit(&self, job_id: JobId, at: DateTime<Utc>) -> anyhow::Result<()> {
        self.propose(WalOperation::JobTimeLimitSignaled { job_id, at })?;
        Ok(())
    }
```

(The suggestion block is a no-op placeholder since GitHub doesn't support suggestions on unchanged lines. The actual fix is in `complete_job` above, inside the existing validation block:)

```rust
pub fn complete_job(&self, job_id: JobId, exit_code: i32, state: JobState) -> anyhow::Result<()> {
    let state = {
        let jobs = self.jobs.read();
        let job = jobs
            .get(&job_id)
            .ok_or_else(|| anyhow::anyhow!("job {} not found", job_id))?;
        if job.state.is_terminal() {
            anyhow::bail!("invalid transition from {:?} to {:?}", job.state, state);
        }
        if job.time_limit_signaled_at.is_some()
            && matches!(state, JobState::Failed | JobState::Completed)
        {
            JobState::Timeout
        } else {
            state
        }
    };

    let resp = self.propose(WalOperation::JobComplete {
        job_id,
        exit_code,
        state,
    })?;
    self.run_all_finalized_side_effects(&resp);
    debug!(job_id, exit_code, "job completed");
    Ok(())
}
```

The override is narrow: only `Failed` and `Completed` promote to `Timeout`. `Cancelled` (user cancel, preemption) and `Timeout` (watchdog force-kill, already correct) pass through unchanged. The WAL entry carries the corrected state, so every replica applies `Timeout` directly and the existing `PendingReason::TimeLimit` branch in the `JobComplete` apply fires.

crates/spur-core/src/wal.rs
```diff
@@ -68,6 +68,15 @@ pub enum WalOperation {
...
     },
+    /// The time-limit watchdog signalled a running job for exhausting its wall
+    /// clock. Durable so the grace period survives a leadership change and so
+    /// every replica finalizes the run as `Timeout` rather than reading the
+    /// terminating signal as an ordinary failure. `at` is stamped on the leader
+    /// so replicas share one instant instead of consulting their own clocks.
+    JobTimeLimitSignaled {
+        job_id: JobId,
+        at: chrono::DateTime<chrono::Utc>,
+    },
```
shiv-tyagi, line 79: Nit: every other WAL variant has a round-trip serialization test in this module (`preempt_requeue_op_round_trips`, `suspend_resume_ops_round_trip`, `job_evict_op_round_trips`, etc.). `JobTimeLimitSignaled` carries a `DateTime<Utc>` that must survive the round-trip, so a test here would guard against future serde regressions.

shiv-tyagi (commented):
Posted comments. PTAL.

crates/spur-core/src/wal.rs
maybeharshit: Addressed in 5eab022 — moved the promotion into complete_job and added the WAL round-trip test

shiv-tyagi (approved):
LGTM.

shiv-tyagi:
@yansun1996 Please review and merge.

yansun1996 (approved):

PR #557, opened 2026-07-31
feat(spur-net): normalize node comm addresses for agent reachability

sgopinath1:
## Summary

Multi-node jobs and agent RPCs need a routable comm address (NodeAddr), not just the cluster node name. Agents on hosts where the short hostname resolves to loopback (common on cloud bare-metal) were registering unusable addresses and breaking inter-node TCP.

This PR separates node identity from reachability:

- **NodeName** — cluster name (`Node.name`, `-w`, partitions)
- **NodeHostname** — OS hostname at registration (`Node.hostname`, WAL-backed)
- **NodeAddr** — routable IP/FQDN for agent gRPC and inter-node TCP (`Node.address`)

Registration now normalizes comm addresses, prefers the gRPC peer IP when the advertised address is loopback or link-local, and updates comm metadata on re-register even when resources are unchanged. Dispatch and agent RPC paths require a registered comm address; they no longer fall back to the node name.

## Approach

- New `spur-net::comm_addr` module: normalize, validate, and detect unusable addresses (loopback, unspecified, link-local)
- `spurd`: keep `--address`, add `--comm-address` alias and `SPUR_COMM_ADDRESS`; normalize explicit addresses at startup
- `spurctld`: `resolve_registration_comm_addr()` at registration; Skip-path metadata updates; scheduler/gRPC dispatch via `comm_addr()`
- `NodeConfig.address` fills only when the agent has not registered an address yet (non-breaking for existing configs)
- Optional `[network] reject_loopback_comm_addr` to refuse non-routable registrations at the controller

Proto and wire format unchanged. WAL gains optional `hostname` on `NodeRegister`/`NodeUpdate` with `#[serde(default)]` for replay compatibility.

## Test plan

- [x] `cargo test --locked`
- [x] Unit tests: comm address normalization, registration peer-IP fallback, re-register comm update, NodeConfig non-override, dispatch abort without comm address

### DIFF
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -556,7 +556,8 @@ pub struct NodeConfig {
     pub gres: Vec<String>,
     #[serde(default)]
     pub features: Vec<String>,
-    /// Override address (if different from hostname).
+    /// Default comm address when the agent has not registered one yet. Does not
+    /// override an agent-registered address.
     pub address: Option<String>,
     /// Scheduling weight. Higher weight = preferred for scheduling.
     #[serde(default = "default_one")]
@@ -581,6 +582,9 @@ pub struct NetworkConfig {
     /// Agent gRPC listen port (default: 6818).
     #[serde(default = "default_agent_port")]
     pub agent_port: u16,
+    /// Reject agent registrations whose comm address resolves to loopback.
+    #[serde(default)]
+    pub reject_loopback_comm_addr: bool,
 }
 
 fn default_wg_cidr() -> String {
@@ -604,6 +608,7 @@ impl Default for NetworkConfig {
             wg_interface: "spur0".into(),
             wg_port: 51820,
             agent_port: 6818,
+            reject_loopback_comm_addr: false,
         }
     }
 }
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -582,7 +582,8 @@ pub struct NetworkConfig {
     /// Agent gRPC listen port (default: 6818).
     #[serde(default = "default_agent_port")]
     pub agent_port: u16,
-    /// Reject agent registrations whose comm address resolves to loopback.
+    /// Reject agent registrations whose comm address is not routable (loopback,
+    /// unspecified, or link-local).
     #[serde(default)]
     pub reject_loopback_comm_addr: bool,
 }
--- a/crates/spur-core/src/node.rs
+++ b/crates/spur-core/src/node.rs
@@ -355,20 +355,6 @@ impl Node {
         self.address.as_deref()
     }
 
-    /// Host used to reach this agent; prefers comm address over node name.
-    pub fn agent_host(&self) -> &str {
-        self.comm_addr().unwrap_or(&self.name)
-    }
-
-    /// OS hostname reported at registration; falls back to node name.
-    pub fn node_hostname(&self) -> &str {
-        if self.hostname.is_empty() {
-            &self.name
-        } else {
-            &self.hostname
-        }
-    }
-
     /// Update state based on allocation level.
     pub fn update_state_from_alloc(&mut self) {
         if self.state.is_admin_hold() {
@@ -627,23 +613,5 @@ mod tests {
     fn node_state_from_short_or_name_rejects_unknown() {
         assert_eq!(NodeState::from_short_or_name("bogus"), None);
         assert_eq!(NodeState::from_short_or_name(""), None);
-    }
-
-    #[test]
-    fn agent_host_prefers_comm_addr() {
-        let mut node = Node::new("node1".into(), ResourceSet::default());
-        assert_eq!(node.agent_host(), "node1");
-        node.address = Some("10.0.0.5".into());
-        assert_eq!(node.agent_host(), "10.0.0.5");
-    }
-
-    #[test]
-    fn node_hostname_falls_back_to_name() {
-        let node = Node::new("node1".into(), ResourceSet::default());
-        assert_eq!(node.node_hostname(), "node1");
-
-        let mut node = Node::new("node1".into(), ResourceSet::default());
-        node.hostname = "host1.example.com".into();
-        assert_eq!(node.node_hostname(), "host1.example.com");
     }
 }
--- a/crates/spur-core/src/node.rs
+++ b/crates/spur-core/src/node.rs
@@ -226,7 +226,11 @@ pub enum NodeSource {
 /// A compute node in the cluster.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Node {
+    /// Cluster node name (NodeName): used by `-w`, partitions, and display.
     pub name: String,
+    /// OS hostname reported at agent registration (NodeHostname).
+    #[serde(default)]
+    pub hostname: String,
     pub state: NodeState,
     pub state_reason: Option<String>,
     /// When true, the current state was set by an operator (admin API, drain,
@@ -261,7 +265,7 @@ pub struct Node {
     pub agent_start_time: Option<DateTime<Utc>>,
     pub last_heartbeat: Option<DateTime<Utc>>,
 
-    /// Agent address for gRPC communication.
+    /// Routable comm address for agent gRPC and inter-node TCP (NodeAddr).
     pub address: Option<String>,
     /// Agent gRPC listen port.
     pub port: u16,
@@ -297,6 +301,7 @@ impl Node {
     pub fn new(name: String, resources: ResourceSet) -> Self {
         Self {
             name,
+            hostname: String::new(),
             state: NodeState::Unknown,
             state_reason: None,
             admin_locked: false,
@@ -343,6 +348,25 @@ impl Node {
     /// Whether this node can accept new work.
     pub fn is_schedulable(&self) -> bool {
         self.state.is_available()
+    }
+
+    /// Routable comm address (NodeAddr), when registered.
+    pub fn comm_addr(&self) -> Option<&str> {
+        self.address.as_deref()
+    }
+
+    /// Host used to reach this agent; prefers comm address over node name.
+    pub fn agent_host(&self) -> &str {
+        self.comm_addr().unwrap_or(&self.name)
+    }
+
+    /// OS hostname reported at registration; falls back to node name.
+    pub fn node_hostname(&self) -> &str {
+        if self.hostname.is_empty() {
+            &self.name
+        } else {
+            &self.hostname
+        }
     }
 
     /// Update state based on allocation level.
@@ -603,5 +627,23 @@ mod tests {
     fn node_state_from_short_or_name_rejects_unknown() {
         assert_eq!(NodeState::from_short_or_name("bogus"), None);
         assert_eq!(NodeState::from_short_or_name(""), None);
+    }
+
+    #[test]
+    fn agent_host_prefers_comm_addr() {
+        let mut node = Node::new("node1".into(), ResourceSet::default());
+        assert_eq!(node.agent_host(), "node1");
+        node.address = Some("10.0.0.5".into());
+        assert_eq!(node.agent_host(), "10.0.0.5");
+    }
+
+    #[test]
+    fn node_hostname_falls_back_to_name() {
+        let node = Node::new("node1".into(), ResourceSet::default());
+        assert_eq!(node.node_hostname(), "node1");
+
+        let mut node = Node::new("node1".into(), ResourceSet::default());
+        node.hostname = "host1.example.com".into();
+        assert_eq!(node.node_hostname(), "host1.example.com");
     }
 }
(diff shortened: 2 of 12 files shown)


crates/spurctld/src/server.rs
```diff
@@ -27,6 +27,82 @@ use crate::sched_stats::SchedStatsCollector;
...
+        .ok_or_else(|| Status::unavailable(format!("node {node_name} has no comm address")))?;
+    Ok(format!("http://{host}:{}", node.port))
+}
+
+fn node_comm_socket(node: &spur_core::node::Node, node_name: &str) -> Result<String, Status> {
+    let host = node
+        .comm_addr()
+        .ok_or_else(|| Status::unavailable(format!("node {node_name} has no comm address")))?;
+    Ok(format!("{host}:{}", node.port))
+}
```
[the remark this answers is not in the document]
sgopinath1: Fixed. node_comm_http_url / node_comm_socket now use spur_net::format_comm_http_url and format_comm_socket, which bracket IPv6 literals.

crates/spurctld/src/scheduler_loop.rs
```diff
@@ -27,6 +27,20 @@ use crate::raft::RaftHandle;
...
+}
+
+fn node_comm_http_url(node: &Node) -> Option<String> {
+    let host = node.comm_addr()?;
+    Some(format!("http://{host}:{}", node.port))
+}
+
+fn node_comm_endpoint(node: &Node) -> Option<(String, u16)> {
+    Some((node.comm_addr()?.to_string(), node.port))
+}
```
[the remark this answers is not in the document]
sgopinath1: Same fix. Scheduler paths use the shared formatters; removed raw http://host:port construction.

crates/spurd/src/main.rs
```diff
@@ -179,14 +179,49 @@ async fn main() -> anyhow::Result<()> {
...
+        .or_else(|| std::env::var("SPUR_COMM_ADDRESS").ok());
+    let node_address = if let Some(ref addr) = explicit_addr {
+        match spur_net::normalize_comm_address(addr) {
+            Ok(normalized) => {
+                if spur_net::comm_addr_is_unusable(&normalized) {
+                    warn!(
+                        comm_addr = %normalized,
+                        input = %addr,
+                        "explicit comm address resolves to loopback; inter-node jobs may fail"
+                    );
```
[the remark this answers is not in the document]
sgopinath1: Updated: warn message now says "not routable" to match comm_addr_is_unusable (loopback, unspecified, link-local).

crates/spur-net/src/detect.rs
```diff
@@ -70,21 +70,18 @@ fn get_wg_address(interface: &str) -> Option<String> {
...
+        Ok(addr) => {
+            if crate::comm_addr::comm_addr_is_unusable(&addr) {
                 warn!(
                     hostname,
-                    "hostname resolved to no addresses, using 127.0.0.1"
+                    comm_addr = %addr,
+                    "hostname resolved to loopback; set --address/--comm-address to a routable IP"
                 );
-                "127.0.0.1".to_string()
             }
```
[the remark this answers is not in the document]
sgopinath1: Same wording change.

crates/spur-core/src/config.rs
```diff
@@ -581,6 +582,9 @@ pub struct NetworkConfig {
     /// Agent gRPC listen port (default: 6818).
     #[serde(default = "default_agent_port")]
     pub agent_port: u16,
+    /// Reject agent registrations whose comm address resolves to loopback.
+    #[serde(default)]
+    pub reject_loopback_comm_addr: bool,
```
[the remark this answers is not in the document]
sgopinath1: Comment updated to describe rejection of loopback, unspecified, and link-local addresses.

crates/spur-net/src/comm_addr.rs
```diff
@@ -0,0 +1,147 @@
...
+    let normalized = normalize_comm_address(input)?;
+    if reject_loopback {
+        if let Ok(ip) = normalized.parse::<IpAddr>() {
+            if is_unusable_comm_ip(ip) {
+                return Err(CommAddressError::Unusable(input.to_string(), normalized));
+            }
+        }
+    }
+    Ok(normalized)
+}
```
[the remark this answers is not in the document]
sgopinath1: Doc comment aligned with actual behavior.

crates/spur-net/src/comm_addr.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    if let Ok(ip) = input.parse::<IpAddr>() {
+        return Ok(ip.to_string());
+    }
+
+    let addrs: Vec<_> = format!("{input}:0")
+        .to_socket_addrs()
+        .map_err(|e| CommAddressError::Unresolvable(input.to_string(), e.to_string()))?
+        .map(|sa| sa.ip())
+        .collect();
```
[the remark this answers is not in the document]
sgopinath1: Documented blocking DNS on normalize_comm_address, and moved resolution to tokio::task::spawn_blocking in register_agent and spurd startup.

crates/spurctld/src/server.rs
```diff
@@ -27,6 +27,82 @@ use crate::sched_stats::SchedStatsCollector;
...
+            comm_addr = %addr,
+            candidate = %candidate,
+            "node registered with loopback or link-local comm address"
+        );
+        return Ok(addr);
+    }
+
+    Err(Status::invalid_argument(format!(
+        "invalid comm address: {}",
+        last_err.unwrap()
```
yansun1996: `last_err.unwrap()` is unreachable given the loop structure today, but it's an `unwrap()` in library code (which AGENTS.md asks us to avoid) and is fragile against future edits to the candidate loop. Would it be worth an explicit fallback, e.g. `match last_err { Some(e) => format!("invalid comm address: {e}"), None => "no comm address candidate resolved".into() }`?

crates/spurctld/src/server.rs
```diff
@@ -27,6 +27,82 @@ use crate::sched_stats::SchedStatsCollector;
...
+    if candidates.is_empty() {
+        return Err(Status::invalid_argument(
+            "no comm address available for registration",
+        ));
+    }
+
+    let mut last_err = None;
+    let mut unusable_result = None;
+    for candidate in candidates {
+        match spur_net::validate_comm_address(candidate, reject_loopback) {
```
yansun1996, line 57: Minor: `comm_addr_is_unusable(&addr)` re-normalizes an address that `validate_comm_address` already normalized. It's cheap (IP re-parse, no second DNS), but `addr.parse::<IpAddr>().is_ok_and(spur_net::is_unusable_comm_ip)` would avoid the round-trip.

crates/spur-core/src/node.rs
```diff
@@ -345,6 +350,25 @@ impl Node {
         self.state.is_available()
     }
 
+    /// Routable comm address (NodeAddr), when registered.
+    pub fn comm_addr(&self) -> Option<&str> {
+        self.address.as_deref()
+    }
+
+    /// Host used to reach this agent; prefers comm address over node name.
+    pub fn agent_host(&self) -> &str {
```
yansun1996: `agent_host()` (and `node_hostname()` just below) don't appear to have production callers — the dispatch paths all go through `comm_addr()` + the URL helpers. `agent_host()` also reintroduces the `unwrap_or(&self.name)` name-fallback that the rest of this PR deliberately removes, so if it's wired up later it'd quietly undo the fail-closed behavior. Might be cleanest to drop both until there's a caller?

crates/spur-net/src/comm_addr.rs
```diff
@@ -0,0 +1,180 @@
...
+    #[error("comm address is empty")]
+    Empty,
+    #[error("comm address {0:?} is not resolvable: {1}")]
+    Unresolvable(String, String),
+    #[error("comm address {0} is not routable ({1})")]
+    Unusable(String, String),
+}
+
+/// True when `addr` is loopback or an unspecified address.
+pub fn is_loopback_ip(addr: IpAddr) -> bool {
```
yansun1996: `is_loopback_ip` doesn't seem to have any callers (the code uses `is_unusable_comm_ip` everywhere). Worth removing it and trimming it from the `pub use` in `lib.rs`? Same for `comm_host_for_socket` / `is_unusable_comm_ip` being re-exported when they're only used inside this module.

crates/spurctld/src/cluster.rs
```diff
@@ -4154,6 +4196,11 @@ impl ClusterManager {
             if node_config_matches(nc, &node.name, &node.labels) {
                 node.features = nc.features.clone();
                 node.weight = nc.weight;
+                if node.address.is_none() {
```
yansun1996, line 4206: This config-default-address branch (adopt `nc.address` when the node has none) is new behavior but I don't see a test for it — `node_config_address_does_not_override_registered_comm_addr` covers the opposite direction. Could we add one that registers with an empty address + a `NodeConfig.address` and asserts the node inherits it?

crates/spurctld/src/scheduler_loop.rs
```diff
@@ -325,9 +335,14 @@ async fn process_assignment(
...
         .iter()
         .filter_map(|name| {
-            cluster
-                .get_node(name)
-                .and_then(|n| n.address.as_ref().map(|a| format!("{}:{}", a, n.port)))
+            let n = cluster.get_node(name)?;
+            if n.comm_addr().is_none() {
+                warn!(
+                    node = %name,
+                    "no comm address for peer list; node omitted from multi-node peers"
```
yansun1996: Small consistency note: this peer-list path warns and drops the node, while the sibling dispatch/confirm/allocation paths fail the job closed when a comm address is missing. It's harmless today because `confirm_dispatch_on_nodes` runs over the same node set and aborts first, but aligning them (or consolidating onto a shared `Node::comm_socket()` helper) would remove the latent divergence.

yansun1996 (commented):
Nice cleanup separating node identity from reachability — the peer-IP fallback and the shared IPv6-safe formatters read well, and the earlier IPv6-bracketing / warning-wording / blocking-DNS points look fully addressed. A few smaller items below (a stray `unwrap`, some dead helpers, and back-compat/test coverage for the new persisted field); nothing blocking correctness that I could find.

Two coverage suggestions with no anchor line:
- The peer-IP fallback in `resolve_registration_comm_addr` is well unit-tested, but not through the real `register_agent` gRPC path — one test registering an agent that advertises loopback from a routable peer would close that.
- IPv6 bracketing is covered at the formatter level; a dispatch-layer test driving an IPv6 comm address through `node_comm_http_url` would prove the end-to-end URL.

crates/spurctld/src/server.rs
sgopinath1: Good catch — replaced with an explicit match last_err { Some(e) => …, None => … } so there’s no unwrap() and the error path stays safe if the candidate loop changes later.

crates/spurctld/src/server.rs
sgopinath1: Agreed. Added normalized_comm_addr_is_unusable for already-normalized IPs and switched the registration path to use that instead of calling comm_addr_is_unusable again.

crates/spur-core/src/node.rs
sgopinath1: Removed both — no callers, and agent_host() would’ve brought back the name fallback this PR is trying to avoid.

crates/spur-net/src/comm_addr.rs
sgopinath1: Removed is_loopback_ip and trimmed the public re-exports to what callers actually need.

crates/spurctld/src/cluster.rs
sgopinath1: Added node_config_address_fills_when_agent_address_empty: register with empty agent address + configured NodeConfig.address, assert the node inherits it.

crates/spurctld/src/scheduler_loop.rs
sgopinath1: Aligned with the other dispatch paths — peer-list construction now aborts when any node lacks a comm address instead of warning and dropping it.

sgopinath1:
> Nice cleanup separating node identity from reachability — the peer-IP fallback and the shared IPv6-safe formatters read well, and the earlier IPv6-bracketing / warning-wording / blocking-DNS points look fully addressed. A few smaller items below (a stray `unwrap`, some dead helpers, and back-compat/test coverage for the new persisted field); nothing blocking correctness that I could find.
> 
> Two coverage suggestions with no anchor line:
> 
> * The peer-IP fallback in `resolve_registration_comm_addr` is well unit-tested, but not through the real `register_agent` gRPC path — one test registering an agent that advertises loopback from a routable peer would close that.
> * IPv6 bracketing is covered at the formatter level; a dispatch-layer test driving an IPv6 comm address through `node_comm_http_url` would prove the end-to-end URL.

Thanks for reviewing @yansun1996.

The unit tests cover resolve_registration_comm_addr directly; I haven’t added a full gRPC registration test yet. Happy to add one in a follow-up if you think it’s worth the harness cost.

Added node_comm_http_url_brackets_ipv6 in server.rs to assert the dispatch URL helper brackets IPv6 literals end-to-end at that layer.

yansun1996 (dismissed):
LGTM, please rebase once before merge, there are commits merged while this PR is pending

PR #561, opened 2026-08-03
feat(mpi): multi-node PMIx prepare/join with TCP modex exchange

sgopinath1:
## Summary

- Add multi-node `--mpi=pmix` support via a two-phase controller dispatch: parallel `PreparePmix` on all allocated agents (start embedded PMIx server, bind TCP modex listener, verify peer reachability), then `LaunchJob` with `pmix_prepared=true` to join the prepared namespace and exec ranks.
- Implement cross-node modex exchange in `spur_mpi_pmix.so` (`modex_exchange.c`, `fence_nb` callback) so Open MPI ranks can complete `MPI_Init` across nodes.
- Fix PMIx init stability when single- and multi-node jobs run back-to-back on the same agent (always register `fence_nb` at `PMIx_server_init`; do not finalize/reinit OpenPMIx between job types).
- On partial batch dispatch failure after prepare, release PMIx on **all** agents that participated in prepare, not only nodes whose launch failed.
- Extend native-host deployment docs, CI/release artifact wiring for the plugin, and native-host MPI e2e tests for multi-node `srun` / batch steps.

## Approach

- **Plan:** controller builds a per-node `PmixLaunchPlan` (ranks, peer FQDNs, modex timeouts, namespace `spur.{job_id}`).
- **Prepare:** `pmix_dispatch::prepare_pmix_on_nodes` runs in parallel; any failure rolls back with `ReleasePmix` on agents that succeeded.
- **Launch:** agents call `join_prepared_pmix`, per-rank `setup_fork` env injection, direct exec (Slurm parity — no `mpirun`).
- **Modex:** each agent listens on `16819 + (job_id % 8000)`; fence exchanges blobs over ephemeral TCP connections.
- **Cleanup:** job completion, cancel, prepare rollback, and partial dispatch abort all tear down PMIx/modex listeners.

Multi-node PMIx is rejected on K8s virtual agents. Requires routable agent comm addresses on every node in the allocation (depends on #557).

## Test plan

- [x] `cargo test --locked`
- [x] Unit test: `partial_multinode_pmix_dispatch_releases_on_all_agents`
- [x] Native-host MPI e2e: single-node, multi-node (2-node), batch + in-allocation `srun` steps
- [x ] Manual: single-node PMIx on node A, then multi-node job including node A (regression for fence_nb / `PMIX_SERVER_URI`)
- [x ] Manual: Crusoe / bare-metal cluster smoke (`hello_mpi` single + multi-node)


crates/spur-mpi-pmix/c/modex_exchange.c
```diff
@@ -0,0 +1,603 @@
...
+}
+
+static void *accept_loop(void *arg) {
+    spur_modex_session_t *session = (spur_modex_session_t *)arg;
+    while (session->accept_running) {
+        int client = accept(session->listen_fd, NULL, NULL);
+        if (client < 0) {
+            if (!session->accept_running) {
+                break;
+            }
```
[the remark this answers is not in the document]
sgopinath1: Fixed: accept_running is now _Atomic bool with atomic_load/atomic_store in the accept loop and destroy path.

docs/deployment/native-host.rst
```diff
@@ -181,73 +185,216 @@ The default can be changed in ``spur.conf``:
...
+
+   scontrol ping
+   sinfo                                    # all agents idle/ready
+   srun --mpi=list                          # expect: none, pmix
+   srun --mpi=pmix -n4 /path/to/hello_mpi   # single-node smoke test
+   srun --mpi=pmix -N2 -n4 /path/to/hello_mpi   # multi-node smoke test
+
+For multi-node ``srun``, the command path and any binaries or scripts it
+``exec``\ s must exist at the **same path on every participating agent** (for
+example ``/tmp/hello_mpi`` on each node, not only on the submission host).
```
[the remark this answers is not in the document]
sgopinath1: Fixed: added a note that multi-node -n must be divisible by -N, with valid/invalid examples.

install.sh
```diff
@@ -164,6 +164,13 @@ EXTRACTED=$(find "${TMPDIR}" -maxdepth 1 -type d -name 'spur-*' | head -1)
 cp -f "${EXTRACTED}"/bin/* "${INSTALL_DIR}/"
 chmod +x "${INSTALL_DIR}/spur" "${INSTALL_DIR}/spurctld" "${INSTALL_DIR}/spurd"
 
+PLUGIN_DIR="${INSTALL_DIR%/bin}/lib/spur"
+if [ -f "${EXTRACTED}/lib/spur/spur_mpi_pmix.so" ]; then
+    mkdir -p "${PLUGIN_DIR}"
+    cp -f "${EXTRACTED}/lib/spur/spur_mpi_pmix.so" "${PLUGIN_DIR}/"
+    log "Installed MPI plugin to ${PLUGIN_DIR}/spur_mpi_pmix.so"
+fi
```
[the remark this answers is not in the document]
sgopinath1: Fixed.

crates/spurd/src/mpi_plugin.rs
```diff
@@ -361,6 +428,139 @@ impl MpiPluginHost {
...
+            .contains_key(&job_id);
+        if active {
+            return self.release_pmix_server(job_id);
+        }
+        if was_prepared.is_none() {
+            return Ok(());
+        }
+        let namespace = PmixLaunchPlan::namespace_for_job(job_id);
+        self.call_server_stop(job_id, &namespace)
+    }
```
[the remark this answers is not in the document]
sgopinath1: Fixed: release_prepared_pmix only acts when the job is still in the prepared map; if already launched, it returns without calling release_pmix_server. PMIx teardown on running jobs stays on the job-monitor path. Added unit test release_prepared_does_not_stop_active_launched_namespace.

yansun1996 (commented):
Thanks for this — the two-phase prepare/join structure is clean and the persisted-type handling (WAL defaults, `NodeSource`, config, append-only proto) is all correct. A few concurrency/lifecycle items I think are worth resolving before merge, plus some test-coverage gaps around the new rollback/cancel paths. Details inline. Nothing blocking on the compatibility front.

A couple of design notes that don't fit on a diff line:

- **Modex listener trust model** — the modex listener binds `INADDR_ANY` and authenticates peers only by `job_id` + a static magic. Worth a one-line note in the deployment docs that multi-node PMIx assumes a trusted/isolated node network (no per-job secret on the wire today), so operators know the assumption.
- **Port collision** — `16819 + job_id % 8000` means two concurrent multi-node jobs whose ids are congruent mod 8000 on a shared node collide on the listener port; with `SO_REUSEADDR` the second bind can succeed and cross-wire. Fine to defer, but worth a follow-up to allocate the port and pass it through the plan.
- **Test gaps:** the prepare-failure→rollback path and cancel-between-prepare-and-launch aren't covered; and `modex_port_for_job` asserts the Rust value but nothing asserts it equals the C constant in `modex_exchange.h` (silent drift would break all cross-node connects).

crates/spur-mpi-pmix/c/pmix_server.c
```diff
@@ -117,6 +169,88 @@ static int ensure_server_init(char *errbuf, size_t errlen) {
...
+        cbfunc(PMIX_ERR_NOT_SUPPORTED, NULL, 0, cbdata, NULL, NULL);
+#endif
+        return PMIX_SUCCESS;
+    }
+    spur_modex_session_t *modex = session->modex;
+    pthread_mutex_unlock(&g_session_lock);
+
+    char *merged = NULL;
+    size_t merged_len = 0;
+    int modex_rc = spur_modex_fence_collect(modex, data, ndata, &merged, &merged_len);
```
yansun1996: `spur_fence_nb` reads `modex = session->modex` under the lock, unlocks at :217, then blocks in `spur_modex_fence_collect` for up to the fence timeout (120s default). If `spur_mpi_pmix_server_stop` runs concurrently (the normal job completion/cancel path), it takes the same lock and calls `spur_modex_session_destroy(session->modex)` — freeing the struct the fence caller is still using, and the caller then reads `session->node_index` lock-free at :248. `spur_mpi_pmix_verify_peers` (:844-847) has the same shape. Could we refcount/own the modex handle across the unlock, or have `server_stop` mark-and-defer the destroy until the in-flight caller drops its reference?

crates/spur-core/src/job.rs
```diff
@@ -786,7 +790,23 @@ impl Job {
             srun_step_dispatch: false,
             actual_stdout_path: None,
             actual_stderr_path: None,
+            launch_failure_detail: None,
+        }
+    }
+
+    /// Slurm-style state reason, including PMIx bootstrap detail when present.
+    pub fn state_reason(&self) -> String {
+        if self.state == JobState::Pending {
```
yansun1996: `state_reason()` returns `LaunchFailure (…)` for *any* `Pending` job that still holds a `launch_failure_detail`, regardless of `pending_reason`. But `clear_run_state_for_requeue` (cluster.rs:3308) resets `pending_reason` without clearing `launch_failure_detail` — only `JobStart` clears it. So a job that hit a launch failure, got requeued, and is now Pending for an ordinary reason (e.g. waiting on resources) keeps reporting the stale `LaunchFailure` in `squeue`/`scontrol`. Could we clear `launch_failure_detail` in `clear_run_state_for_requeue`, or gate the first branch on `pending_reason == JobLaunchFailure` instead of bare `Pending`?

crates/spurd/src/mpi_plugin.rs
```diff
@@ -361,6 +428,139 @@ impl MpiPluginHost {
...
+        }
+        if existing_attempt.is_some() {
+            self.release_prepared_pmix(plan.job_id)?;
+        }
+
+        let mut plan = plan.clone();
+        self.apply_modex_timeouts(&mut plan);
+        mpi::validate_pmix_plan(&plan)?;
+        self.load_plugin()?;
+        self.call_server_start(&plan)?;
```
yansun1996: The ordering here is `call_server_start` (:461, starts the C server + accept thread) → insert into `active_namespaces` (:480) → insert into `prepared` (:491). If a cancel fires between server-start and the `prepared` insert, `release_prepared_pmix` checks the `prepared` map first and returns early — so the already-started server and accept thread are never reaped. Could the `prepared` entry be inserted before/with the side effect, or `release` fall back to stopping any `active_namespaces` entry with `refs == 0`?

crates/spurctld/src/pmix_dispatch.rs
```diff
@@ -0,0 +1,146 @@
...
+        }
+    }
+
+    if errors.is_empty() {
+        return Ok(());
+    }
+
+    let detail = errors.join("; ");
+    error!(job_id, error = %detail, "PMIx prepare failed — rolling back prepared agents");
+    release_pmix_on_agents(&prepared_agents, job_id).await;
```
yansun1996: Rollback releases only `prepared_agents` — the agents that returned `Ok`. If an agent commits its namespace but the `PreparePmix` RPC then times out or the connection drops (:52), that agent holds an active prepared namespace the controller never learns about, so it's skipped here and leaks until job completion. The batch launch-failure path already releases *all* participants; could rollback do the same (release every node it dispatched prepare to, not just the acked ones)?

crates/spur-mpi-pmix/c/pmix_server.c
```diff
@@ -117,6 +169,88 @@ static int ensure_server_init(char *errbuf, size_t errlen) {
...
+    (void)ninfo;
+    if (cbfunc == NULL || procs == NULL) {
+        return PMIX_ERR_BAD_PARAM;
+    }
+
+    pthread_mutex_lock(&g_session_lock);
+    spur_pmix_session_t *session = find_session_by_proc(&procs[0]);
+    if (session == NULL) {
+        pthread_mutex_unlock(&g_session_lock);
+#if PMIX_VERSION_MAJOR >= 6
```
yansun1996: Several `#if PMIX_VERSION_MAJOR >= 6 / #else` blocks (here and ~:209, :233, :241) call `cbfunc(...)` with byte-identical arguments in both arms. These look like dead conditionals — could they collapse to a single unconditional call?

yansun1996 (changes requested):
Requesting changes for the four must-fix items in the inline comments above (fence/verify use-after-free, stale launch-failure reason on requeue, prepare cancel-window leak, rollback leaking non-acked agents). The rest are non-blocking.

crates/spur-mpi-pmix/c/pmix_server.c
sgopinath1: Fixed: we now hold g_session_lock for the full spur_modex_fence_collect / spur_modex_verify_peers call so server_stop cannot destroy modex mid-flight. Comment documents the trade-off (blocks other sessions on that agent for up to the fence timeout); per-modex refcounts can follow if that becomes hot.

crates/spur-core/src/job.rs
sgopinath1: Fixed: detail is only appended when pending_reason == JobLaunchFailure, and leaving that reason via set_pending_reason clears it. state_reason_display() now delegates to state_reason(). Also fixed repeated backoff wiping detail by not routing requeue reset through set_pending_reason(None).

crates/spurd/src/mpi_plugin.rs
sgopinath1: Fixed: register active_namespaces before call_server_start. release_prepared_pmix now tears down an unref'd namespace even when the job is not yet in prepared. Unit test: release_prepared_stops_unrefd_namespace_before_prepared_insert.

crates/spurctld/src/pmix_dispatch.rs
sgopinath1: Fixed: rollback calls release_pmix_on_agents on every target in the prepare set, not just RPC successes. Matches the partial batch-dispatch cleanup path.

crates/spur-mpi-pmix/c/pmix_server.c
sgopinath1: Fixed in spur_fence_nb: collapsed the identical #if PMIX_VERSION_MAJOR >= 6 / #else pairs to single calls. Left version branches where the callback signature actually differs (e.g. spur_client_connected).

sgopinath1:
> Thanks for this — the two-phase prepare/join structure is clean and the persisted-type handling (WAL defaults, `NodeSource`, config, append-only proto) is all correct. A few concurrency/lifecycle items I think are worth resolving before merge, plus some test-coverage gaps around the new rollback/cancel paths. Details inline. Nothing blocking on the compatibility front.
> 
> A couple of design notes that don't fit on a diff line:
> 
> * **Modex listener trust model** — the modex listener binds `INADDR_ANY` and authenticates peers only by `job_id` + a static magic. Worth a one-line note in the deployment docs that multi-node PMIx assumes a trusted/isolated node network (no per-job secret on the wire today), so operators know the assumption.
> * **Port collision** — `16819 + job_id % 8000` means two concurrent multi-node jobs whose ids are congruent mod 8000 on a shared node collide on the listener port; with `SO_REUSEADDR` the second bind can succeed and cross-wire. Fine to defer, but worth a follow-up to allocate the port and pass it through the plan.
> * **Test gaps:** the prepare-failure→rollback path and cancel-between-prepare-and-launch aren't covered; and `modex_port_for_job` asserts the Rust value but nothing asserts it equals the C constant in `modex_exchange.h` (silent drift would break all cross-node connects).

Thanks for the thorough review, @yansun1996 

Must-fix (all in c39dd05): fence/verify UAF (hold g_session_lock through modex), stale launch-failure reason (gate on JobLaunchFailure + preserve detail across backoff), prepare cancel window (active_namespaces before server_start, release unref'd ns), prepare rollback (release every target in the prepare set, not just acked RPCs). Replied inline on each thread.

Manual smoke: single- and multi-node hello_mpi --mpi=pmix on bare-metal (3-node) and Crusoe (8-agent) after c39dd05.

Design notes:
- Trust model: agree; will add a one-liner to native-host.rst that modex TCP assumes a trusted/isolated agent network (job_id + magic only, no per-job secret today).
- Port collision: acknowledged; happy to track a follow-up to allocate/pass the listener port in the plan instead of job_id % 8000.
- Test gaps: prepare-failure rollback and cancel-during-prepare mocks are still on my list; Rust/C port constant parity test is a good catch and I'll add it (or a build-time assert) in a follow-up unless you want it in this PR.

crates/spurd/src/agent_server.rs
```diff
@@ -1812,13 +1857,13 @@ impl SlurmAgent for AgentService {
 
             if num_tasks > 1 {
                 senv.set("SPUR_TASK_OFFSET", req.task_offset);
-            } else {
-                SpurEnv::apply_task_rank(&mut senv, req.task_offset, 0, 1, step_mpi);
+            } else if !req.label {
+                SpurEnv::apply_task_rank(&mut senv, req.task_offset, 0, 1);
             }
```
shiv-tyagi, line 1994: `apply_task_rank` is now skipped for labeled single-task steps. This block is only entered when `num_tasks > 1 || req.label`, so reaching this `else` means `num_tasks == 1`, which implies `req.label == true`. The `!req.label` guard is therefore always false here, so `srun -l <cmd>` with one task no longer gets `SLURM_PROCID`/`SLURM_LOCALID` (nothing else on this path sets them). Pre-PR this `else` was unconditional.

```suggestion
            if num_tasks > 1 {
                senv.set("SPUR_TASK_OFFSET", req.task_offset);
            } else {
                SpurEnv::apply_task_rank(&mut senv, req.task_offset, 0, 1);
            }
```

crates/spur-mpi-pmix/c/modex_exchange.c
```diff
@@ -0,0 +1,604 @@
...
+        return SPUR_MODEX_ERR_BLOB;
+    }
+    int push_rc = push_local_blob(session, local_data, local_len);
+    if (push_rc != SPUR_MODEX_OK) {
+        return push_rc;
+    }
+    int wait_rc = wait_for_remotes(session);
+    if (wait_rc != SPUR_MODEX_OK) {
+        return wait_rc;
+    }
```
shiv-tyagi, line 684: Q: `remote[].present` is never reset between fences, so `spur_modex_fence_collect` is effectively one-shot. The session is created once per job and reused by `fence_nb` on every fence, so a second collective fence (the `MPI_Finalize` barrier, or any app that fences more than once) finds `all_remotes_present()` already true, skips `wait_for_remotes`, and `merge_blobs` returns stale round-1 data. Clearing each `remote[i].present`/`data` at the start of the collect would fix it.

Related: `merge_blobs` reads `remote[]` without holding `session->lock` while the accept thread is still live, so a late or duplicate frame could `free()` a buffer mid-copy. Taking the lock around the merge closes both.

crates/spur-mpi-pmix/include/spur_mpi_plugin.h
```diff
@@ -31,19 +32,35 @@ typedef struct spur_mpi_launch_plan {
...
 int spur_mpi_pmix_server_start(const spur_mpi_launch_plan_t *plan, char *errbuf, size_t errlen);
 int spur_mpi_pmix_server_stop(const char *namespace_, char *errbuf, size_t errlen);
+int spur_mpi_pmix_verify_peers(const spur_mpi_launch_plan_t *plan, char *errbuf, size_t errlen);
 int spur_mpi_pmix_env(
     const spur_mpi_launch_plan_t *plan,
     uint32_t rank,
     const char *key,
     char *val,
     size_t vallen
 );
```
shiv-tyagi: `spur_mpi_pmix_env` is dead after the v3 switch to `setup_fork_env`. The Rust host now loads `spur_mpi_pmix_setup_fork_env`/`_free` and no longer looks up `spur_mpi_pmix_env`; it has no callers in the header, `pmix_server.c`, `stub_server.c`, or tests, and it duplicates the URI4/URI3 aliasing that `mpi_plugin.rs` already does. Drop the declaration plus both definitions.

crates/spur-mpi-pmix/c/pmix_server.c
```diff
@@ -44,6 +51,32 @@ static int local_procs_match(const spur_pmix_session_t *session, const spur_mpi_
 static spur_pmix_session_t g_sessions[SPUR_PMIX_MAX_SESSIONS];
 static pthread_mutex_t g_session_lock = PTHREAD_MUTEX_INITIALIZER;
 static int g_server_initialized = 0;
+static uint32_t g_multinode_fence_sessions = 0;
```
shiv-tyagi: `g_multinode_fence_sessions` is write-only. `update_multinode_fence_hook` increments and decrements it but nothing ever reads the value (the only read clamps its own decrement). Since `fence_nb` is now always registered, this counter drives nothing. Remove it and the two `update_multinode_fence_hook` call sites.

crates/spurctld/src/server.rs
```diff
@@ -2113,6 +2203,11 @@ impl SlurmController for ControllerService {
             ));
         }
 
+        if needs_pmix_prepare {
+            let addrs: Vec<String> = dispatches.iter().map(|d| d.agent_addr.clone()).collect();
+            pmix_dispatch::release_pmix_on_agents(&addrs, job_id).await;
+        }
```
shiv-tyagi, line 2332: Prepared PMIx servers leak if this handler's future is dropped between prepare and here. `prepare_pmix_on_nodes` succeeds earlier, but the only release is this call after the full `RunCommand` set resolves, with several awaits in between. If the srun client disconnects mid-step, tonic drops the handler and this release never runs, leaving a prepared namespace (`refs == 0`) on every participating agent until the job is later cancelled or completed. `confirm_dispatch_on_nodes` has the same shape on controller shutdown or leadership loss between prepare and its release. A drop guard that releases on all prepared agents unless disarmed on the success path would close it.

shiv-tyagi (commented):
Posted comments. PTAL.

crates/spurd/src/agent_server.rs
sgopinath1: Good catch — regression. Removed the !req.label guard; single-task labeled steps get apply_task_rank again.

crates/spur-mpi-pmix/c/modex_exchange.c
sgopinath1: Fixed both: reset remote[] at the start of each collect, and hold session->lock through merge_blobs.

crates/spur-mpi-pmix/include/spur_mpi_plugin.h
sgopinath1: Agreed — dropped the declaration and both C/stub definitions. Host only uses setup_fork_env now.

crates/spur-mpi-pmix/c/pmix_server.c
sgopinath1: Removed the counter and both update_multinode_fence_hook call sites.

crates/spurctld/src/server.rs
sgopinath1: Added PmixPreparedReleaseGuard on the srun step path (and batch dispatch); disarms on the normal release path, spawns rollback on drop. Batch confirm_dispatch_on_nodes gets the same guard.

sgopinath1:
> Posted comments. PTAL.

Thanks for reviewing @shiv-tyagi. I have addressed your comments.

crates/spur-mpi-pmix/c/pmix_server.c
```diff
@@ -369,6 +1004,18 @@ int spur_mpi_pmix_server_stop(const char *namespace_, char *errbuf, size_t errle
...
     if (session != NULL) {
+        if (session->local_fence.active) {
+            finish_local_fence(session, PMIX_ERR_JOB_TERMINATED, NULL, 0);
+        } else {
+            clear_local_fence(&session->local_fence, session->num_local_procs);
+        }
+        if (session->modex != NULL) {
+            spur_modex_session_t *modex = session->modex;
+            session->modex = NULL;
+            spur_modex_session_retain(modex);
```
yansun1996, line 1015: `session->modex` owns the initial reference created by `spur_modex_session_create`. After `session->modex = NULL`, the new `retain` and following `release` cancel each other out, leaving the original owner reference unreachable. `spur_modex_session_abort` does not close the listener, so every stopped multi-node job leaks its listener socket and accept thread. Remove the extra retain: abort, then release the detached session-owned reference. An in-flight fence already owns its temporary reference. Please add a start/stop cleanup regression test.

yansun1996 (changes requested):
One new blocking lifecycle issue remains; details inline.

crates/spur-mpi-pmix/c/pmix_server.c
yansun1996: suggested diff
```c
                if (session->modex != NULL) {
                    spur_modex_session_t *modex = session->modex;
                    session->modex = NULL;
-> remove           spur_modex_session_retain(modex);
                    spur_modex_session_abort(modex);
                    spur_modex_session_release(modex);
                }
```

yansun1996 (approved):
LGTM, @sgopinath1 @shiv-tyagi  I restarted a failed k8s failure, most likely an infra failure, once it passes we can merge this PR.

PR #562, opened 2026-08-03
refactor(spurd)!: drop --comm-address alias and SPUR_COMM_ADDRESS

sgopinath1:
## Summary

#557 added `--comm-address` and `SPUR_COMM_ADDRESS` as Spur-only aliases for the existing `--address` / `SPUR_NODE_ADDRESS` pair. Slurm has no `slurmd` counterpart: `NodeAddr` is set in `slurm.conf`, not via a daemon CLI flag. This PR removes the extra surface and documents `--address` only.

Internal NodeAddr handling from #557 (normalization, peer-IP fallback, controller-side `comm_addr()`) is unchanged.

## Breaking change

Anyone who adopted `--comm-address` or `SPUR_COMM_ADDRESS` from the #557 docs should switch to `--address` or `SPUR_NODE_ADDRESS`.

## Test plan

- [x] `cargo test --locked`

### DIFF
--- a/crates/spur-net/src/detect.rs
+++ b/crates/spur-net/src/detect.rs
@@ -78,7 +78,7 @@ fn resolve_hostname(hostname: &str) -> String {
                 warn!(
                     hostname,
                     comm_addr = %addr,
-                    "hostname resolved to a non-routable address; set --address/--comm-address to a routable IP"
+                    "hostname resolved to a non-routable address; set --address to a routable IP"
                 );
             }
             addr
--- a/crates/spur-net/src/detect.rs
+++ b/crates/spur-net/src/detect.rs
@@ -78,7 +78,7 @@ fn resolve_hostname(hostname: &str) -> String {
                 warn!(
                     hostname,
                     comm_addr = %addr,
-                    "hostname resolved to a non-routable address; set --address to a routable IP"
+                    "hostname resolved to a non-routable address; set --address to a routable IP or FQDN"
                 );
             }
             addr
--- a/docs/deployment/native-host.rst
+++ b/docs/deployment/native-host.rst
@@ -132,7 +132,8 @@ Start the agent:
        --address 10.44.0.2 \
        --listen [::]:6818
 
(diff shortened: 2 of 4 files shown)


crates/spur-net/src/detect.rs
```diff
@@ -78,7 +78,7 @@ fn resolve_hostname(hostname: &str) -> String {
                 warn!(
                     hostname,
                     comm_addr = %addr,
-                    "hostname resolved to a non-routable address; set --address/--comm-address to a routable IP"
+                    "hostname resolved to a non-routable address; set --address to a routable IP"
```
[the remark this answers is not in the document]
sgopinath1: Fixed.

docs/deployment/native-host.rst
```diff
@@ -132,8 +132,7 @@ Start the agent:
        --address 10.44.0.2 \
        --listen [::]:6818
 
-``--address`` and ``--comm-address`` are equivalent. You can also set
-``SPUR_NODE_ADDRESS`` or ``SPUR_COMM_ADDRESS``. Pass a routable IP or FQDN,
+Or set the ``SPUR_NODE_ADDRESS`` environment variable. Pass a routable IP or FQDN,
 not the short hostname alone when ``/etc/hosts`` maps it to loopback.
```
shiv-tyagi: Nit: the "Or" dangles a bit now that the preceding line no longer mentions `--comm-address`. Consider:

```suggestion
``--address`` sets the advertised comm address. Alternatively, set the
``SPUR_NODE_ADDRESS`` environment variable. Pass a routable IP or FQDN,
not the short hostname alone when ``/etc/hosts`` maps it to loopback.
```

shiv-tyagi (commented):
Posted comments. PTAL.

docs/deployment/native-host.rst
sgopinath1: Fixed.

yansun1996 (approved):
LGTM

PR #569, opened 2026-08-05
fix(spur-cli): reject unresolved usernames

### LINKED ISSUE #567: fix(spur-cli): unify whoami fallback sentinel across all CLI commands
Multiple CLI commands record the job owner or send the requesting username via `whoami::username()`, but use inconsistent fallback values on failure:

- `interactive.rs` (`current_user()`) falls back to `"<lookup-failed>"` (illegal UNIX username, cannot collide)
- `srun.rs`, `sbatch.rs`, `salloc.rs`, `scancel.rs`, `scontrol.rs`, `scrontab.rs` fall back to `"unknown"` (legal username, could collide with a real account)

If `whoami` fails (uid with no `/etc/passwd` entry, e.g. in some container setups), a job gets recorded with owner `"unknown"` but later exec/attach sends `"<lookup-failed>"`. The mismatch denies the legitimate owner access to their own job.

**Proposed fix:** Make `current_user()` return `Result<String>` and propagate the error to all CLI entry points. If we cannot determine who the caller is, refuse to proceed rather than recording a sentinel that may mismatch later. This is the cleanest solution since any sentinel (even a consistent one) has edge cases when two users both experience lookup failures.

**Workaround:** Ensure all compute nodes have proper `/etc/passwd` entries for all users (standard practice).

**Likelihood:** Extremely low. `whoami::username()` reads `/etc/passwd` (or equivalent) and only fails if the running uid has no entry.

Discovered during review of #551.

01xjw:
## Summary

- Fail closed when the current username cannot be resolved.
- Resolve the owner once per command entry point and reuse it for submission, step creation, PTY, output streaming, and cancellation.
- Rebase onto latest `main` while preserving response-warning handling.

Closes #567.

## Review follow-up

- Fixed the duplicate username lookup throughout the `srun` flow.
- Kept the `k8s.rs` fallback out of scope as agreed; tracked separately in #603.

## Validation

- `cargo test --workspace --exclude spurd --locked`
- Unmapped-UID runtime checks fail closed and leave the crontab file intact.

## Disclosure

This change was prepared with assistance from the radeon-issue automation and independently checked by a separately configured validation model. Maintainer review is still required.

### DIFF
--- a/crates/spur-cli/src/exec.rs
+++ b/crates/spur-cli/src/exec.rs
@@ -46,7 +46,7 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
         .exec_in_job(ExecInJobRequest {
             job_id: args.job_id,
             command: args.command.clone(),
-            user: crate::interactive::current_user(),
+            user: crate::interactive::current_user()?,
         })
         .await
         .context("exec failed")?;
--- a/crates/spur-cli/src/interactive.rs
+++ b/crates/spur-cli/src/interactive.rs
@@ -18,11 +18,13 @@ pub async fn connect_agent(addr: &str) -> Result<SlurmAgentClient<tonic::transpo
         .max_encoding_message_size(spur_proto::MAX_GRPC_REQUEST_SIZE))
 }
 
-/// Local username sent with exec/attach requests so the server can enforce job
-/// ownership. The fallback is deliberately not a legal UNIX username, so a
-/// lookup failure cannot collide with a real account and grant access.
-pub fn current_user() -> String {
-    whoami::username().unwrap_or_else(|_| "<lookup-failed>".into())
+/// Local username sent with authenticated job requests.
+///
+/// Refuse to continue when the operating system cannot resolve the caller:
+/// recording a sentinel can either collide with a real account or disagree
+/// with a later exec/attach request.
+pub fn current_user() -> Result<String> {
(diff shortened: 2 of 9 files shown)


yansun1996 (commented):
LGTM — clean fail-closed fix.

yansun1996 (approved):
Approving. One optional follow-up worth tracking separately: `k8s.rs` `effective_user()` (line 112) still uses the old `"unknown"` sentinel and feeds `caller` into the k0s/kubeconfig RPCs — the last one left in `spur-cli`. Not a bypass (`is_k0s_admin` rejects `"unknown"`), so fine outside this PR.

yansun1996:
Hi @01xjw could you please rebase + resolve the conflicts ? also please take a look at the above Copilot posted comments

yansun1996 (changes requested):
could you please rebase + resolve the conflicts ? also please take a look at the above Copilot posted comments

01xjw:
Addressed all requested items:

- rebased onto latest `main` and preserved the new warning handling;
- reused one resolved owner across `srun` step, PTY, output, and cancellation requests;
- tracked the out-of-scope `k8s.rs` fallback in #603, as agreed.

Formatting, strict Clippy, workspace tests, and unmapped-UID checks all pass.

yansun1996 (approved):

PR #578, opened 2026-08-06
fix(spur-net): support gzip, zstd, and uncompressed OCI layers

### LINKED ISSUE #347: fix(spur-net): OCI puller assumes all layers are gzip — zstd and uncompressed layers fail
## Problem

In `pull_and_extract()` (`spur-net/src/oci.rs`), the extraction condition `media_type.contains("gzip") || digest.starts_with("sha256:")` is always true (all digests are sha256), routing every layer through `GzDecoder`. For zstd-compressed layers (`application/vnd.oci.image.layer.v1.tar+zstd`) or uncompressed layers (`…layer.v1.tar`), `GzDecoder` fails with an invalid gzip header error.

The `docker save` path in `image.rs` is more robust — it checks magic bytes (`0x1f, 0x8b`) and falls back to plain tar. The OCI path has no such detection.

Neither path supports zstd decompression.

## Expected Behavior

- Detect layer compression from `mediaType` and/or magic bytes
- Support gzip, zstd, and uncompressed tar layers
- Remove the meaningless `digest.starts_with("sha256:")` fallback

## Acceptance Criteria

- [ ] Layer decompression selected by `mediaType` with magic-byte fallback
- [ ] Uncompressed tar layers (`…layer.v1.tar`) extracted correctly
- [ ] Zstd layers (`…tar+zstd`) extracted correctly (requires `zstd` crate)
- [ ] `sha256:` digest check removed from extraction gate
- [ ] Consistent detection logic between OCI pull and `docker save` import paths

joshkmartinez:
## What

OCI pulls treated every layer as gzip-compressed because the `sha256:` digest check made the extraction gate unconditional. Zstd and uncompressed layers therefore failed, while Docker/Podman-save imports also lacked zstd support.

Closes #347.

## Approach

- Add one streaming layer decoder in `spur-net` and use it for registry pulls and Docker/Podman-save imports.
- Prefer declared gzip, zstd, or plain-tar media types; use gzip/zstd magic bytes and then plain tar when the media type is absent or unknown.
- Support concatenated gzip/zstd frames and zstd skippable frames.
- Fail clearly on malformed layers and clean temporary Docker/Podman import data on errors.

## Behavior change

Docker/Podman-save imports now abort if any layer fails to decode or unpack. Previously they printed a warning and continued, which could produce an incomplete image. The registry path continues to tolerate individual entry-unpack errors for rootless or special-file cases.

## Testing

- `cargo test -p spur-net --locked` (55 passed, including ordinary OCI whiteout extraction)
- `cargo test -p spur-cli --locked` (342 unit tests and the broken-pipe integration test passed)
- `cargo clippy -p spur-net -p spur-cli --all-targets --locked -- -D warnings`
- Local end-to-end smoke through the real `spur image import` CLI using a loopback OCI registry and a deterministic `docker save` fixture with gzip, zstd, uncompressed, and unknown-media zstd layers
- Real `mksquashfs` packaging and `unsquashfs` verification of exact rootfs contents and layer overwrite order for both import paths

Workspace-wide tests and Clippy were also attempted; on this macOS host they stop in the pre-existing `spur-mpi-pmix` build script because it passes GNU `ld` flags unsupported by Apple's linker. TLS/auth registries, multi-arch manifests, a live Docker/Podman daemon, and Linux `spurd` mounting remain CI/platform boundaries.

### DIFF
--- a/crates/spur-cli/src/image.rs
+++ b/crates/spur-cli/src/image.rs
@@ -132,12 +132,10 @@ async fn cmd_import_dockerd(image: &str) -> Result<()> {
     let rootfs_str = rootfs.to_string_lossy();
     let output_path_str = output_path.to_string_lossy();
 
-    // Extract the docker save tar, then extract each layer
-    extract_docker_save_tar(&output.stdout, &rootfs_str)?;
-
-    // Pack into squashfs
-    pack_squashfs(&rootfs_str, &output_path_str)?;
+    let import_result = extract_docker_save_tar(&output.stdout, &rootfs_str)
+        .and_then(|()| pack_squashfs(&rootfs_str, &output_path_str));
     let _ = std::fs::remove_dir_all(&tmp_dir);
+    import_result?;
 
     let size = std::fs::metadata(&output_path)
         .map(|m| m.len())
@@ -184,9 +182,10 @@ async fn cmd_import_podman(image: &str) -> Result<()> {
     let rootfs_str = rootfs.to_string_lossy();
     let output_path_str = output_path.to_string_lossy();
 
-    extract_docker_save_tar(&output.stdout, &rootfs_str)?;
-    pack_squashfs(&rootfs_str, &output_path_str)?;
+    let import_result = extract_docker_save_tar(&output.stdout, &rootfs_str)
+        .and_then(|()| pack_squashfs(&rootfs_str, &output_path_str));
     let _ = std::fs::remove_dir_all(&tmp_dir);
+    import_result?;
 
     let size = std::fs::metadata(&output_path)
         .map(|m| m.len())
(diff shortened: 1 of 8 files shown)


crates/spur-cli/src/image.rs
```diff
@@ -202,61 +201,49 @@ async fn cmd_import_podman(image: &str) -> Result<()> {
...
+            let mut archive = tar::Archive::new(reader);
             archive.set_overwrite(true);
-            archive.unpack(dest)
-        };
-
-        if let Err(e) = result {
-            eprintln!("Warning: layer {} extraction had errors: {}", layer_file, e);
+            archive
+                .unpack(dest)
+                .with_context(|| format!("failed to extract layer {}", layer_file))?;
```
yansun1996, line 239: This path now aborts the whole import if any layer fails to unpack, whereas the previous code logged a warning per layer and continued. The registry path (`extract_tar` in `oci.rs`) still tolerates per-entry unpack errors via `unpack_in` + a `debug!` skip, so the two import paths now differ in how forgiving they are — e.g. a rootless `docker save` import that hits an unpackable special file would now fail outright. Is the stricter behavior intended here? If so, no change needed — just worth a note in the description.

crates/spur-net/src/oci.rs
```diff
@@ -611,42 +611,38 @@ async fn resolve_manifest_list(
...
+        let path = entry.path()?.to_path_buf();
+        let filename = path.file_name().and_then(|f| f.to_str()).unwrap_or("");
+        if filename.starts_with(".wh.") {
+            // Whiteout: delete the corresponding file
+            let target = if filename == ".wh..wh..opq" {
+                // Opaque whiteout: directory should be empty
+                // (skip for now — complex to handle)
+                continue;
+            } else {
+                let real_name = filename.strip_prefix(".wh.").unwrap();
```
yansun1996: Minor: `AGENTS.md` asks for no `unwrap()` in library code. It's provably safe here (guarded by the `starts_with(".wh.")` check above), but since this line was already touched by the refactor, would `strip_prefix(".wh.").unwrap_or(filename)` be worth using to keep it clean?

crates/spur-net/src/image_layer.rs
```diff
@@ -0,0 +1,172 @@
...
+    Uncompressed,
+    Zstd,
+}
+
+fn detect_compression(media_type: Option<&str>, data: &[u8]) -> LayerCompression {
+    match media_type.filter(|value| !value.is_empty()) {
+        Some(value) if value.ends_with("+gzip") || value.ends_with(".gzip") => {
+            LayerCompression::Gzip
+        }
+        Some(value) if value.ends_with("+zstd") || value.ends_with(".zstd") => {
```
yansun1996, line 20: Small suggestion: the zstd arm matches `+zstd` and `.zstd` but not `.zst`. Standard OCI/Docker media types use `+zstd`, so this is only defensive, and a `.zst`-declared layer with real zstd bytes still decodes via the magic-byte fallback below — so feel free to skip. If you'd like the declared-type fast path to cover it too, adding `|| value.ends_with(".zst")` would do it.

yansun1996 (commented):
Nice cleanup — one decoder for all three compression cases, reused across both paths, with solid tests. A few small notes; nothing blocking.

Worth a line in the PR description: both paths now hard-fail on a malformed layer where the old code warned and continued. More correct, but images that imported "with warnings" will now fail.

Possible follow-ups (pre-existing, not this PR): the docker/podman-save path has no `.wh.*` whiteout handling unlike the registry path, and decompressed layers have no size cap (zip-bomb risk, inherited from the old gzip path).

One test gap: the `.wh.*` whiteout branch in `extract_tar` was refactored here but isn't exercised.

crates/spur-cli/src/image.rs
joshkmartinez: Yes, the stricter behavior is intentional. Continuing after a layer fails can produce a silently incomplete image. Ack'ed it in the PR desc as well.

crates/spur-net/src/oci.rs
joshkmartinez: Good catch. I replaced the guarded unwrap() with unwrap_or(filename) and added a regression test.

crates/spur-net/src/image_layer.rs
joshkmartinez: I think we should leave this unchanged b/c OCI media types use +zstd, while .zst is a filename extension. Nonstandard layers containing valid zstd data are already supported through magic-byte fallback, so recognizing .zst as authoritative could misclassify a custom media type.

crates/spur-cli/src/image.rs
```diff
@@ -202,61 +201,49 @@ async fn cmd_import_podman(image: &str) -> Result<()> {
...
+            .first()
+            .and_then(|manifest| manifest.get("Layers"))
+            .and_then(|layers| layers.as_array())
+            .ok_or_else(|| anyhow::anyhow!("no layers in manifest.json"))?;
+
+        for layer_path in layers {
+            let layer_file = layer_path
+                .as_str()
+                .ok_or_else(|| anyhow::anyhow!("invalid layer path"))?;
+            let data = std::fs::read(tmp.join(layer_file))
```
yansun1996: `layer_file` comes straight from the untrusted `manifest.json` inside the docker-save tarball. `PathBuf::join` replaces the base entirely when the joined component is absolute, so a `Layers` entry like `/etc/shadow` would make this read an arbitrary file on the host instead of something scoped under `tmp`. The old `format!("{}/{}", tmp, layer_file)` didn't have this escape. Worth rejecting absolute/`..` components in `layer_file` before joining, or asserting the resulting path still starts with `tmp`.

yansun1996 (commented):
Thanks for addressing the earlier round of feedback — the whiteout fix plus its regression test look solid, and the description update covers the hard-fail-behavior question. One thing worth fixing before merge, left as an inline note below.

yansun1996 (changes requested):

crates/spur-cli/src/image.rs
joshkmartinez: good catch. I now reject absolute and parent-relative manifest paths, resolve both the manifest and each layer against the staging directory, and reject symlink escapes

yansun1996 (approved):

PR #579, opened 2026-08-06
test(e2e): expand native-host and k8s coverage for untested subsystems

biluriuday:
## Summary

This PR closes a long-standing e2e gap — 65 of the last 120 feat/fix commits shipped without touching `tests/`, and whole subsystems had never had any e2e coverage — then fixes the product bugs that the new coverage exposed, and finally restructures the suite so it runs in parallel. Collected tests go from **218 → 585**.

The commits are ordered so each builds on the last: harness first, then the tests, then the fixes those tests surfaced, then the CI split.

## 1. Harness + coverage expansion (`52bf35e`)

The harness had to grow before the subsystems could be tested:

- **`SpurCluster`** gains `salloc`, `sattach`, `spur exec`, read-only CLI wrappers, `cli_with_env` (submit-side env vars), curl-over-SSH HTTP helpers, plugstack + C-fixture compilation, and per-node controller control so a 3-controller Raft cluster can be driven.
- **k8s fixture** gains `assert_pod_spec`, spec builders, operator restart/log access, and a quota variant.

New coverage for previously untested surfaces:

- **native-host**: scheduling/placement, topology, task layout, SPANK, REST API, metrics, resource limits (cgroups/OOM), FFI (compiled C driver), PTY/attach, licenses, job ownership, images, hetjobs, dispatch resilience, read-side CLI formatting/filtering, submit-time env defaults, burst buffer, WireGuard mesh, and multi-controller Raft HA.
- **k8s**: SpurJob CRD spec→PodSpec propagation, failure modes (OOMKilled, image-pull, restart/orphan cleanup), quota projection, operator health, and node watcher.

Three k8s subsystems turned out to be **unreachable** from the suite rather than merely untested, so the manifests were fixed too: the operator never passed `--enable-quota`, the ClusterRole lacked the quota controller's resources plus `escalate`/`bind`, nothing provided the Postgres that `SlurmAccounting` needs, and `spurctld` exposed neither the REST nor the metrics port.

## 2. Product bugs surfaced by the new coverage

Real defects the suite caught against a live cluster:

- **Burst buffer directives silently dropped** (`6a73ddb`) — `spurd` read `SPUR_BURST_BUFFER` from its own process env instead of the job env, so every `stage_in:`/`stage_out:` was a no-op (a job with `stage_in:exit 7` completed with exit 0). Now read from `cfg.environment`.
- **`scancel` filters always errored** (`6a73ddb`) — `--partition`/`--account` were declared and passed to `get_jobs`, but the "did the caller select anything?" guard only checked job IDs/`--user`/`--name`, so both flags failed with *"no job IDs or filters specified"*. Replaced with a `has_selection()` predicate (unit-tested).
- **`--mem` overruns paged out instead of OOM-killing** (`760ebb5`) — `spurd` set `memory.max` but never capped swap, so a job that outgrew its limit never reached `OUT_OF_MEMORY`. Swap is now capped via a `[cgroup]` section mirroring Slurm's `ConstrainSwapSpace`/`AllowedSwapSpace`, defaulting **off** so deployed clusters keep today's behavior.
- **Multi-node nodelist mis-parsed** (`760ebb5`) — `sattach`/`srun` took the first comma-separated field of the nodelist, yielding `node[1` once the controller compresses an allocation. Both now expand the hostlist first.
- **k8s operator wedged on permanent rejections** (`760ebb5`) — it retried *every* failed submit, so a SpurJob the controller permanently rejected (unknown partition, denied account) sat without status forever. Permanent rejections are now marked `Failed`; only transient codes requeue.
- **Operator backoff was flat, not exponential** (`f7b7a63`) — `error_policy` requeued after a flat 60s despite a comment claiming exponential backoff; kube-rs uses that `Action` verbatim, so the first transient blip cost a full minute of submission latency. Now doubles 1s→60s cap, per-SpurJob, reset on a clean pass (cleanup reconciles take the same path). This is what broke `test_state_survives_leader_failover` in CI.

## 3. Test/harness robustness (`bfdd00b`, `f7b7a63`)

Several failures were the tests' fault, not the product's — they assumed placement or enforcement the environment didn't guarantee:

- Assertions now read the job's own `NodeList` (extracted into shared `job_node_names`/`job_node_indices` helpers) instead of guessing the first N nodes.
- The SPANK failing-hook test searches every agent's log rather than node 0.
- The OOM tests now **preflight** the same cgroup-delegation sequence `spurd` uses and skip (naming the node) on hosts where the memory controller isn't delegated — instead of failing where `--mem` can't be enforced.
- Tests skip rather than fail where the environment lacks `wireguard-tools` or inter-node hostname resolution.

Verified on the 3-node bare-metal cluster: resource limits + topology + placement **39/39** with no skips (so the preflight doesn't mask genuine OOM cases there), and the SPANK fix flips that test from 5/5 failing to 5/5 passing.

**New `e2e-fixtures` CI job** (`f7b7a63`): runs `pytest --setup-plan` over `tests/`, resolving every fixture without starting a cluster. It catches things like a fixture missing its `@pytest.fixture` decorator (which had errored 19 `test_metrics.py` tests at setup) in **under a second** instead of an hour into the E2E run. E2E gates on this workflow's conclusion.

## 4. Parallelizing e2e (`f885805`)

native-host ran as one ~60 min job and k8s as one ~40 min job. Both are now partitioned by pytest markers so CI runs them in parallel:

- **native-host** → six suite matrix jobs (`scheduling`, `policy`, `api`, `runtime`, `fabric`, `ha`), each on its own ephemeral 4-VM cluster. `RUN_DIR`, the VM prefix, and the results artifact are scoped per suite so the jobs never collide on the shared run id.
- **k8s** → five suites (`core`, `spec`, `quota`, `ha`, `nodes`) run concurrently on one bootstrapped cluster, each in its own namespace via `tests/k8s/e2e/run_suites.sh`. Safe because the destructive node tests stay skipped, so no suite mutates shared cluster-scoped `Node`s.
- Every e2e file carries **exactly one** `suite_*` / `suite_k8s_*` marker. A collection guard in each `conftest` plus a parity check in `e2e-fixtures` fail fast if a file has zero or two markers, so a new test can't silently escape a suite. The marker rule and the node-mutation guardrail are documented in `AGENTS.md`.

## Addressing the review feedback

- **CI target failures** — the native-host and k8s failures called out earlier were the placement/enforcement and operator-backoff issues above; they're resolved across `6a73ddb`, `760ebb5`, `bfdd00b`, and `f7b7a63`.
- **e2e time benchmarking** — one caveat on measuring this: E2E runs via `workflow_run`, which GitHub always evaluates from the **default branch**, so the parallel split only takes effect *after* this PR merges. The before number is visible now (native-host ~60 min single job, k8s ~37 min single job); the after (per-suite parallel) numbers will land on the first `main` run and I'll post them here. Target is ≤15 min per suite; if any suite overshoots we rebalance markers, which is a one-line change per file.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,6 +115,30 @@ jobs:
       - name: Test
         run: cargo test --locked
 
+  e2e-fixtures:
+    name: E2E fixture resolution
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.12"
+
+      - name: Install E2E test dependencies
+        run: pip install -r tests/requirements.txt
+
+      # --setup-plan resolves every fixture without running one, so a missing or
+      # misnamed fixture fails here rather than an hour into E2E. The env vars
+      # only satisfy pytest_ignore_collect; nothing connects anywhere.
+      - name: Check every E2E fixture resolves
+        run: pytest --setup-plan -q
+        working-directory: tests
+        env:
+          SPUR_TEST_NODES: 192.0.2.1
+          SPUR_TEST_SSH_USER: nobody
+          KUBECONFIG: /dev/null
+
   test-db:
     name: DB tests (accounting)
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,6 +139,32 @@ jobs:
           SPUR_TEST_SSH_USER: nobody
           KUBECONFIG: /dev/null
 
+      - name: Native-host suite markers are complete and disjoint
+        working-directory: tests
+        env:
+          SPUR_TEST_NODES: 192.0.2.1
+          SPUR_TEST_SSH_USER: nobody
+          KUBECONFIG: /dev/null
+        run: |
+          all=$(pytest native_host/e2e --collect-only -q 2>/dev/null | grep -c '::' || true)
+          sel=$(pytest native_host/e2e --collect-only -q \
+            -m "suite_scheduling or suite_policy or suite_api or suite_runtime or suite_fabric or suite_ha" \
+            2>/dev/null | grep -c '::' || true)
+          echo "collected=$all selected-by-suites=$sel"
+          test "$all" -eq "$sel"
+
+      - name: K8s suite markers are complete and disjoint
+        working-directory: tests
+        env:
+          KUBECONFIG: /dev/null
+        run: |
+          all=$(pytest k8s/e2e --collect-only -q 2>/dev/null | grep -c '::' || true)
+          sel=$(pytest k8s/e2e --collect-only -q \
+            -m "suite_k8s_core or suite_k8s_spec or suite_k8s_quota or suite_k8s_ha or suite_k8s_nodes" \
+            2>/dev/null | grep -c '::' || true)
+          echo "collected=$all selected-by-suites=$sel"
+          test "$all" -eq "$sel"
+
   test-db:
     name: DB tests (accounting)
     runs-on: ubuntu-latest
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -68,14 +68,33 @@ jobs:
       matrix:
         include:
           - cluster_type: native-host
-            STATUS_CONTEXT: "E2E / native-host"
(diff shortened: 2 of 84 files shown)


yansun1996 (changes requested):
1. there are many CI test targets failure PTAL
2. could you please also benchmarking the e2e total time change after this PR ?

PR #581, opened 2026-08-06
fix(spurctld): filter reservations by name in scontrol show

shiv-tyagi:
## What

`scontrol show reservation <name>` listed every reservation instead of just the named one (SPUR-119).

## Why

The CLI dropped the `name` argument and `ListReservations` had no filter field, so the handler always returned all reservations. This diverged from `scontrol show node`/`show partition`, which filter server-side.

## Approach

- `proto/slurm.proto`: add `string name = 1;` to `ListReservationsRequest` (additive, wire-safe: new tag, no renames/retypes).
- `spurctld` `list_reservations`: filter by name on the controller (`name.is_empty() || r.name == name`), mirroring `get_nodes`/`get_partitions`. Leader-forward path preserved.
- `spur-cli` scontrol: forward the `name` argument; when a name is given but nothing matches, error with `Reservation <name> not found` (exit 1), matching Slurm.

## Tests

- `spurctld` unit test `list_reservations_filters_by_name` exercises the real handler: one match by name, all when empty, none for an unknown name.
- Extended `tests/native_host/e2e/test_reservations.py` to create two reservations and assert the filtered `show` returns only the requested one.

## Validation

- `cargo test --locked`: pass.
- Bare-metal testbed (4 nodes): `show reservation rocm_patch` returns only `rocm_patch`, no-name returns all, unknown name exits 1.

### DIFF
--- a/crates/spur-cli/src/scontrol.rs
+++ b/crates/spur-cli/src/scontrol.rs
@@ -735,11 +735,21 @@ async fn show(controller: &str, entity: &str, name: Option<&str>) -> Result<()> 
         }
         "reservation" | "reservations" => {
             let resp = client
-                .list_reservations(spur_proto::proto::ListReservationsRequest {})
+                .list_reservations(spur_proto::proto::ListReservationsRequest {
+                    name: name.unwrap_or("").into(),
+                })
                 .await
                 .context("failed to list reservations")?;
 
-            for res in resp.into_inner().reservations {
+            let reservations = resp.into_inner().reservations;
+            if reservations.is_empty() {
+                if let Some(name) = name {
+                    bail!("Reservation {} not found", name);
+                }
+                return Ok(());
+            }
+
+            for res in reservations {
                 println!("ReservationName={}", res.name);
                 println!("   StartTime={}", res.start_time);
                 println!("   EndTime={}", res.end_time);
--- a/crates/spur-cli/src/scontrol.rs
+++ b/crates/spur-cli/src/scontrol.rs
@@ -734,6 +734,7 @@ async fn show(controller: &str, entity: &str, name: Option<&str>) -> Result<()> 
             }
         }
         "reservation" | "reservations" => {
+            let name = normalize_show_name(name);
             let resp = client
                 .list_reservations(spur_proto::proto::ListReservationsRequest {
                     name: name.unwrap_or("").into(),
@@ -744,7 +745,7 @@ async fn show(controller: &str, entity: &str, name: Option<&str>) -> Result<()> 
             let reservations = resp.into_inner().reservations;
             if reservations.is_empty() {
                 if let Some(name) = name {
-                    bail!("Reservation {} not found", name);
+                    bail!("Reservation {name} not found");
                 }
                 return Ok(());
             }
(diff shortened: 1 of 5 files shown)


yansun1996 (commented):
Nice, focused change — the server-side filter is clean, the read-forwarding refactor correctly matches the `get_nodes` shape, and the unit test genuinely exercises the filter (it goes red if the filter is removed). One behavioral edge worth tightening before merge, plus a couple of optional follow-ups.

**One suggestion (edge case):** the server trims the name (`req.name.trim()`) but the CLI sends and tests it untrimmed. So `scontrol show reservation "  "` sends `"  "`, the server trims to empty and returns *all* reservations, and the CLI — seeing a non-empty list — prints them all instead of reporting not-found. Trimming in the CLI once (and basing both the request field and the empty/not-found decision on the trimmed value) would keep the two sides in agreement and also make the "not found" message echo the cleaned name. Details inline.

**Optional / follow-ups (not blocking):**
- The reservation branch now errors on an unknown name, while `show node`/`show partition`/`show job` still print nothing and exit 0 on no match. Erroring is the more Slurm-faithful behavior, so this arm is arguably the right one — but it's worth a note that the divergence is intentional, and maybe a follow-up to align the siblings.
- Consider a small test for the not-found path (e.g. an e2e asserting `scontrol show reservation <nonexistent>` exits non-zero), and one server case pinning the whitespace-name behavior once the trim seam above is settled — the current tests cover the positive filter well but not these.

crates/spur-cli/src/scontrol.rs
```diff
@@ -735,11 +735,21 @@ async fn show(controller: &str, entity: &str, name: Option<&str>) -> Result<()>
         }
         "reservation" | "reservations" => {
             let resp = client
-                .list_reservations(spur_proto::proto::ListReservationsRequest {})
+                .list_reservations(spur_proto::proto::ListReservationsRequest {
+                    name: name.unwrap_or("").into(),
```
yansun1996, line 740: The name is passed through untrimmed here, but the server trims it (`req.name.trim()`), so the two sides can disagree: `scontrol show reservation "  "` sends `"  "`, the server trims to empty and returns every reservation, and since the returned list is non-empty this prints all of them rather than reporting not-found. Trimming once in the CLI and reusing that value for both the request field and the not-found check would keep them consistent (and make the `bail!` message print the cleaned name). Something like:

```rust
let name = name.map(str::trim).filter(|s| !s.is_empty());
// ListReservationsRequest { name: name.unwrap_or("").into() }
// if reservations.is_empty() { if let Some(name) = name { bail!("Reservation {name} not found"); } return Ok(()); }
```

yansun1996 (dismissed):
LGTM, please re-trigger the CI workflow

shiv-tyagi:
Trimmed the name once in the CLI (`normalize_show_name`) and reused it for both the request field and the not-found check, so `scontrol show reservation "  "` now behaves the same on both sides. Added a CLI test for the helper, a server case pinning the whitespace-name behavior, and an e2e asserting an unknown name exits non-zero. Left the unknown-name error as-is (intentional divergence from node/partition/job). CI re-triggered.

crates/spur-cli/src/scontrol.rs
shiv-tyagi: Fixed.

yansun1996 (approved):

PR #583, opened 2026-08-06
fix(spurctld): evict aged terminal jobs to bound controller memory

yansun1996:
## What this fixes

Terminal jobs were retained in the controller's in-memory job map — and in every Raft snapshot — indefinitely. Controller state therefore grew without bound with the total number of jobs ever submitted. On a busy long-running cluster this drives the leader's memory up until it OOMs, and it inflates snapshots past the replication message limit, which in turn stalls Raft log compaction and makes state growth worse.

## Approach

A leader-only maintenance pass (alongside the existing reservation-purge passes in the scheduler loop) evicts finished jobs whose `end_time` is older than a configurable retention window. It does so through a new replicated WAL operation carrying the leader-resolved job ids, so every replica drops the same jobs from its in-memory map and from all subsequent snapshots. Because the snapshot is built from the in-memory map, bounding the map bounds the snapshot.

- New config knob `controller.terminal_job_retention_secs` (default 3600s), validated at load against a one-year upper bound so an out-of-range value cannot overflow the eviction cutoff and panic the controller.
- The eviction set is computed once on the leader and shipped in the WAL op; apply re-checks `is_finalized()` per id so a job requeued between propose and apply is spared. Every replica evicts an identical set — no state divergence from clock skew.
- `is_finalized()` jobs are evicted (terminal, plus a `Preempted` job stranded with an `end_time`); all active states are spared. A job's steps are dropped alongside it.

Accounting history lives in PostgreSQL and is unaffected — a finished job remains visible to `scontrol show job` for the retention window, then ages out of controller memory only (`sacct` history is permanent).

### Correctness guards

- **Dependencies:** a terminal job still referenced by a non-finalized job's dependency is spared, so eviction can't cancel an `afterok`/`afternotok`/`aftercorr` dependent (or early-release an `after`/`afterany` one) by making its target resolve to "unknown".
- **Job-id reuse:** `next_job_id` is persisted in the snapshot and restored as `max(rebuilt-from-survivors, persisted)`, so evicting the high-id tail can't lower the counter and reissue a used id after a snapshot+restart.
- **Accounting reconcile:** effective retention is floored to the accounting reconcile interval, so a job always survives at least one reconcile pass before eviction and can't be stranded as `RUNNING` in `sacct` (this also makes `retention = 0` safe).

## Trade-offs / notes

- The retention default (1 hour) is more generous than Slurm's `MinJobAge` (5 min) to keep post-run inspection convenient; operators can lower it to tighten the memory bound (floored to the reconcile interval).
- A Postgres outage longer than the retention window can still evict a job before its accounting row is durably correct — an operator trade-off, documented on the config field.
- Array-parent views are synthesized from surviving per-task jobs; once all tasks age out, the parent id returns nothing (same as any evicted job).

## Upgrade ordering

`EvictTerminalJobs` is a new WAL variant. Once a leader commits one, an older controller replaying the log can't deserialize it — so upgrade all controllers before this feature emits the op, and do not roll a controller back across this version once an `EvictTerminalJobs` entry has committed.

## Tests

- Apply-path test: an aged terminal job (and a stranded `Preempted` one) is evicted while a recent terminal job, a running job, and a pending job are spared; the evicted job's steps are removed while a live job's steps are kept; a named-but-live id is a no-op.
- Replica-determinism test: divergent per-replica `end_time`s do not change the evicted set (apply reads state, not `end_time`).
- Dependency-sparing test: a target referenced by a begin-held pending `afterok` child is spared until the child finalizes.
- Retention-floor test: `retention = 0` does not evict a just-completed job on the next tick; a job past the floored window is evicted.
- Snapshot-restore test: `next_job_id` does not regress after the high-id tail is evicted.
- Config validation rejects an out-of-range retention value.

## How it was tested

Validated end-to-end on an isolated single-node deployment: a completed job survived past its configured retention (proving the reconcile-interval floor), an unreferenced completed job was evicted once it aged out, and a completed job referenced by a begin-held pending dependent was spared while the dependent stayed `PENDING` (not cancelled). An out-of-range retention value was rejected cleanly at config load rather than panicking at runtime.

### DIFF
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -320,10 +320,19 @@ pub struct ControllerConfig {
     /// the inverse `SchedulerParameters=nohold_on_prolog_fail`.
     #[serde(default = "default_hold_on_prolog_fail")]
     pub hold_on_prolog_fail: bool,
+
+    /// Seconds a terminal job stays in controller memory/snapshots before eviction
+    /// (default 3600); Postgres accounting history is unaffected.
+    #[serde(default = "default_terminal_job_retention_secs")]
+    pub terminal_job_retention_secs: u64,
 }
 
 fn default_max_batch_requeue() -> u32 {
     5
+}
+
+fn default_terminal_job_retention_secs() -> u64 {
+    3600
 }
 
 fn default_hold_on_prolog_fail() -> bool {
@@ -374,6 +383,7 @@ impl Default for ControllerConfig {
             max_batch_requeue: default_max_batch_requeue(),
             max_launch_backoff_secs: default_max_launch_backoff_secs(),
             hold_on_prolog_fail: default_hold_on_prolog_fail(),
+            terminal_job_retention_secs: default_terminal_job_retention_secs(),
         }
     }
 }
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -348,6 +348,10 @@ fn default_max_launch_backoff_secs() -> u64 {
 /// out of chrono's representable range.
 pub const MAX_LAUNCH_BACKOFF_SECS: u64 = 86_400;
 
+/// Upper bound for `terminal_job_retention_secs` (one year). Keeps the eviction
+/// cutoff (`now - retention`) inside chrono's representable range.
+pub const MAX_TERMINAL_JOB_RETENTION_SECS: u64 = 366 * 24 * 60 * 60;
+
 fn default_listen_addr() -> String {
     "[::]:6817".into()
 }
@@ -1156,6 +1160,16 @@ impl SlurmConfig {
                 ),
             });
         }
+        // Bounded so the eviction cutoff `now - retention` stays representable.
+        if self.controller.terminal_job_retention_secs > MAX_TERMINAL_JOB_RETENTION_SECS {
+            return Err(ConfigError::InvalidValue {
+                field: "controller.terminal_job_retention_secs".into(),
+                value: format!(
+                    "{} (must be at most {})",
+                    self.controller.terminal_job_retention_secs, MAX_TERMINAL_JOB_RETENTION_SECS
+                ),
+            });
+        }
         if self.cluster.enabled {
             if self.cluster.distro != "k0s" {
                 return Err(ConfigError::InvalidValue {
@@ -1930,6 +1944,43 @@ max_launch_backoff_secs = 0
         assert!(
             err.to_string().contains("max_launch_backoff_secs"),
             "unexpected error: {err}"
+        );
+    }
+
+    #[test]
+    fn controller_config_rejects_out_of_range_terminal_job_retention_secs() {
+        // Past this bound the eviction cutoff overflows chrono and panics the
+        // controller, so it must be refused at load rather than at first use.
+        let toml = format!(
+            r#"
+cluster_name = "test"
+
+[controller]
+terminal_job_retention_secs = {}
+"#,
+            MAX_TERMINAL_JOB_RETENTION_SECS + 1
+        );
+        let err = SlurmConfig::load_from_str(&toml).unwrap_err();
+        assert!(
+            err.to_string().contains("terminal_job_retention_secs"),
+            "unexpected error: {err}"
+        );
+
+        let ok = format!(
+            r#"
+cluster_name = "test"
+
+[controller]
+terminal_job_retention_secs = {MAX_TERMINAL_JOB_RETENTION_SECS}
+"#
+        );
+        assert_eq!(
+            SlurmConfig::load_from_str(&ok)
+                .unwrap()
+                .controller
+                .terminal_job_retention_secs,
+            MAX_TERMINAL_JOB_RETENTION_SECS,
+            "the bound itself must be accepted"
         );
     }
 
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -321,8 +321,10 @@ pub struct ControllerConfig {
     #[serde(default = "default_hold_on_prolog_fail")]
     pub hold_on_prolog_fail: bool,
 
-    /// Seconds a terminal job stays in controller memory/snapshots before eviction
-    /// (default 3600); Postgres accounting history is unaffected.
+    /// Seconds a terminal job stays in controller memory before eviction (default
+    /// 3600). `sacct` history is unaffected but `scontrol show job` stops finding
+    /// it after the window; runs once per `scheduler.interval_secs` and is floored
+    /// to the accounting reconcile interval so a job's DB row can be repaired first.
     #[serde(default = "default_terminal_job_retention_secs")]
     pub terminal_job_retention_secs: u64,
 }
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -348,8 +348,8 @@ fn default_max_launch_backoff_secs() -> u64 {
 /// out of chrono's representable range.
 pub const MAX_LAUNCH_BACKOFF_SECS: u64 = 86_400;
 
-/// Upper bound for `terminal_job_retention_secs` (one year). Keeps the eviction
-/// cutoff (`now - retention`) inside chrono's representable range.
+/// Upper bound for `terminal_job_retention_secs` (one year). Operational
+/// guardrail: retaining terminal jobs longer defeats the memory bound.
 pub const MAX_TERMINAL_JOB_RETENTION_SECS: u64 = 366 * 24 * 60 * 60;
 
 fn default_listen_addr() -> String {
@@ -1160,7 +1160,7 @@ impl SlurmConfig {
                 ),
             });
         }
-        // Bounded so the eviction cutoff `now - retention` stays representable.
+        // Guardrail: retention beyond the cap defeats the memory bound.
         if self.controller.terminal_job_retention_secs > MAX_TERMINAL_JOB_RETENTION_SECS {
             return Err(ConfigError::InvalidValue {
                 field: "controller.terminal_job_retention_secs".into(),
@@ -1949,8 +1949,8 @@ max_launch_backoff_secs = 0
 
     #[test]
     fn controller_config_rejects_out_of_range_terminal_job_retention_secs() {
-        // Past this bound the eviction cutoff overflows chrono and panics the
-        // controller, so it must be refused at load rather than at first use.
+        // Beyond the guardrail, load must reject up front rather than silently
+        // retaining terminal jobs long enough to defeat the memory bound.
         let toml = format!(
             r#"
 cluster_name = "test"
--- a/crates/spur-core/src/wal.rs
+++ b/crates/spur-core/src/wal.rs
@@ -262,6 +262,12 @@ pub enum WalOperation {
     NodeK0sClear {
         name: String,
     },
+
+    /// Evict terminal jobs whose end_time is older than `before` from the
+    /// in-memory map and snapshots, bounding controller memory.
+    EvictTerminalJobs {
+        before: chrono::DateTime<chrono::Utc>,
+    },
 }
 
 impl WalOperation {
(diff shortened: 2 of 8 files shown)


crates/spurctld/src/cluster.rs
```diff
@@ -4845,6 +4863,21 @@ impl ClusterManager {
                 }
                 k0s.reset_requested = *reset_requested;
             }
+            WalOperation::EvictTerminalJobs { before } => {
+                let evicted: HashSet<JobId> = jobs
+                    .iter()
+                    .filter(|(_, j)| {
+                        j.state.is_terminal() && j.end_time.is_some_and(|t| t < *before)
+                    })
+                    .map(|(&id, _)| id)
```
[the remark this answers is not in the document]
yansun1996: Good catch — this was a real divergence risk. `end_time` is stamped per-replica at apply time, so comparing it against `before` in the apply path meant replicas could evict different sets from the same entry. Changed `EvictTerminalJobs` to carry the resolved `job_ids` (the leader computes the aged-out set once); apply is now a pure id-membership removal, so every replica evicts identically. Added a two-replica test with skewed `end_time`s to lock it in.

crates/spur-core/src/config.rs
```diff
@@ -1923,6 +1947,43 @@ max_launch_backoff_secs = 0
         );
     }
 
+    #[test]
+    fn controller_config_rejects_out_of_range_terminal_job_retention_secs() {
+        // Past this bound the eviction cutoff overflows chrono and panics the
+        // controller, so it must be refused at load rather than at first use.
```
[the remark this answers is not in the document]
yansun1996: You are right, it is not a chrono representability limit — a one-year-old cutoff is nowhere near overflowing. Reworded to describe what it actually is: an operational guardrail (retaining terminal jobs past this defeats the memory bound). No behavior change.

crates/spurctld/src/cluster.rs
```diff
@@ -3303,6 +3303,28 @@ impl ClusterManager {
...
+    pub fn evict_expired_terminal_jobs(&self) {
+        let retention = self.config().controller.terminal_job_retention_secs;
+        let before = Utc::now() - chrono::Duration::seconds(retention as i64);
+        // is_finalized is load-bearing, not redundant: end_time survives a
+        // requeue, so a re-dispatched job's stale end_time alone would reap it.
+        let job_ids: Vec<JobId> = self
+            .jobs
+            .read()
+            .iter()
+            .filter(|(_, j)| j.state.is_finalized() && j.end_time.is_some_and(|t| t < before))
```
shiv-tyagi: Evicting a terminal job that a pending job still depends on will cancel the dependent. `check_dependencies` resolves a target via `resolve_target_state`, which returns `None` once the job leaves the map; for `afterok`/`afternotok`/`aftercorr` that maps to `DependencyResult::Failed`, and `cancel_unsatisfiable_dependency_jobs` turns `Failed` into a real `JobComplete{Cancelled}`.

Concrete trigger: `sbatch --begin=now+2h --dependency=afterok:100`. A begin-time hold is not an `is_scheduling_hold()`, so the child is not shielded. Job 100 completes, evicts at the 1h default, and the next tick cancels the child ~1h before its begin time. `afterany`/`after` flip the other way (missing target treated as satisfied/releasable), so an evicted target can also release a dependent early or skip an `after:+delay` window.

This path did not exist before, since terminal jobs lived in the map forever. Spare any job still referenced by a non-terminal job's `spec.dependency` (by id, array-parent id, or name/user), or resolve dependencies without requiring the target record to survive.

crates/spurctld/src/cluster.rs
```diff
@@ -4845,6 +4867,21 @@ impl ClusterManager {
...
+            WalOperation::EvictTerminalJobs { job_ids } => {
+                // Re-check finalized: spare an id requeued between propose and
+                // apply. Deterministic — every replica applies in the same order.
+                let evicted: HashSet<JobId> = job_ids
+                    .iter()
+                    .filter(|id| jobs.get(id).is_some_and(|j| j.state.is_finalized()))
+                    .copied()
+                    .collect();
+                if !evicted.is_empty() {
+                    jobs.retain(|id, _| !evicted.contains(id));
```
shiv-tyagi, line 4900: This is the first code that ever removes entries from the jobs map, which silently breaks a `next_job_id` invariant and can lead to job ID reuse.

`next_job_id` is an in-memory `AtomicU32` and is not part of `ClusterSnapshot`. `restore_from_snapshot` reconstructs it as `next_id = first_job_id` then `max(job.job_id + 1)` over surviving jobs only. Before this PR the highest-ever job id was always present (terminal jobs stayed in the map), so reconstruction was always safe. Once eviction removes the high-id tail and a snapshot compacts away those `JobSubmit` log entries, a restarted controller or a freshly promoted follower resets `next_job_id` down and reissues already-used IDs. Postgres `jobs` is keyed by `job_id`, so a reused id UPSERTs over the prior job's `sacct` row, and a dangling `afterok:<id>` binds to the wrong job.

This compiles and passes tests but corrupts a running cluster after a snapshot+restart. Persist the high-water mark: add `next_job_id` to `ClusterSnapshot` (with `#[serde(default)]`) and restore as `max(reconstructed, persisted)`, or carry it in the evict WAL op. Worth a snapshot-restore test asserting no id regression after eviction.

crates/spurctld/src/cluster.rs
```diff
@@ -3303,6 +3303,28 @@ impl ClusterManager {
...
+            .jobs
+            .read()
+            .iter()
+            .filter(|(_, j)| j.state.is_finalized() && j.end_time.is_some_and(|t| t < before))
+            .map(|(&id, _)| id)
+            .collect();
+        if job_ids.is_empty() {
+            return;
+        }
+        if let Err(e) = self.propose(WalOperation::EvictTerminalJobs { job_ids }) {
```
shiv-tyagi, line 3344: Eviction can defeat the accounting reconciliation safety net. `accounting/reconcile.rs::run_once` repairs missing/stale Postgres rows (from a `notify_job_end` that exhausted its retries during a DB outage) by scanning the in-memory job map, which is its only data source. Reconcile runs every 120s; default retention is 3600s, so a transient outage is covered. But an outage longer than retention, or `terminal_job_retention_secs = 0` (supported and tested), evicts the job before its row is durably correct and permanently strands it as `RUNNING`/no-`end_time` in `sacct`.

Floor the effective retention above the reconcile interval, reject or round up `retention = 0`, or skip jobs not yet confirmed durable in accounting.

crates/spur-core/src/wal.rs
```diff
@@ -262,6 +262,12 @@ pub enum WalOperation {
     NodeK0sClear {
         name: String,
     },
+
+    /// Evict the named terminal jobs to bound controller memory. The leader
+    /// resolves the aged-out set so every replica evicts identically.
+    EvictTerminalJobs {
```
shiv-tyagi, line 268: Q: worth calling out the upgrade ordering in the PR/runbook. `WalOperation` is an externally-tagged serde enum, so once a leader commits an `EvictTerminalJobs` entry, an older controller replaying that log (rollback, or an old follower during a rolling upgrade) hits an unknown variant and fails to deserialize. This is inherent to any new WAL variant and consistent with the append-only rule, but it means there is no clean rollback past the first emitted entry; all replicas must be upgraded before the feature can emit the op.

crates/spur-core/src/config.rs
```diff
@@ -320,12 +320,21 @@ pub struct ControllerConfig {
     /// the inverse `SchedulerParameters=nohold_on_prolog_fail`.
     #[serde(default = "default_hold_on_prolog_fail")]
     pub hold_on_prolog_fail: bool,
+
+    /// Seconds a terminal job stays in controller memory/snapshots before eviction
+    /// (default 3600); Postgres accounting history is unaffected.
+    #[serde(default = "default_terminal_job_retention_secs")]
+    pub terminal_job_retention_secs: u64,
```
shiv-tyagi, line 329: Nit: document the user-visible behavior here. After the retention window a finished job disappears from `scontrol show job <id>` (history stays in `sacct`), and `terminal_job_retention_secs = 0` can evict a completion on the very next scheduler tick, racing anything polling the controller (rather than `sacct`) to detect completion. Also note eviction only runs as often as `scheduler_interval_secs`, so the effective window is retention rounded up to a tick.

shiv-tyagi (commented):
Reviewed the eviction path across happy and failure flows. The core mechanism (leader-gated, leader-resolved id set, apply-time `is_finalized()` re-check) is sound, but three correctness issues stand out, plus two smaller notes. Inline comments have detail.

Blockers:
1. Evicting a terminal job still referenced by a pending job's dependency cancels the dependent (`afterok`/`afternotok`/`aftercorr` resolve missing target to Failed).
2. Job ID reuse: `next_job_id` is rebuilt from the job map on snapshot restore; this is the first code to remove jobs from the map, so after a snapshot+restart ids get reissued and overwrite `sacct` rows.
3. Eviction can outrun accounting reconciliation (which scans the map), stranding rows during a Postgres outage longer than retention or at `retention = 0`.

Verified safe: lock ordering, node resource accounting, late agent RPCs, array-parent display, metrics, REST/FFI by-id paths, and the retention bounds math. The Preempted widening does not race a live requeue (preempt-requeue is atomic to Pending).

shiv-tyagi:
An E2E is also failing, please check if it is related or a flaky test.

crates/spurctld/src/cluster.rs
yansun1996: Good catch — confirmed the regression. The producer now builds a spare-set from every non-finalized job's parsed dependencies and excludes those target ids (and their array-parent ids) from eviction, so a terminal target still referenced by a pending `afterok`/`afternotok`/`aftercorr`/`after`/`afterany` dependent is kept until the dependent itself finalizes. Added a `Dependency::target_job_id()` helper, a unit test for the `afterok` begin-hold case, and verified live: an evict-eligible target referenced by a begin-held pending child is spared and the child stays Pending.

crates/spurctld/src/cluster.rs
yansun1996: Confirmed — this was a real id-reuse path across snapshot+restart. Persisted the high-water mark as a `#[serde(default)] next_job_id` on `ClusterSnapshot`; restore now takes `max(rebuilt-from-survivors, persisted)` so evicting the high-id tail can't lower it. Added a snapshot→restore test asserting no regression after the tail is evicted.

crates/spurctld/src/cluster.rs
yansun1996: Agreed. Added a shared `RECONCILE_INTERVAL_SECS` constant and floored the effective retention to it in the producer, so a job always survives at least one reconcile pass before it can be evicted — this closes the `retention = 0` hole too. `main.rs` now uses the same constant for the reconcile loop so they can't drift. The longer-than-retention outage remains an operator trade-off, now documented on the config field.

crates/spur-core/src/wal.rs
yansun1996: Right — this is inherent to adding any WAL variant. No code change; I've noted the upgrade ordering in the PR description (upgrade all controllers before the feature emits the op; no rollback across this version once an `EvictTerminalJobs` entry has committed).

crates/spur-core/src/config.rs
yansun1996: Done — extended the `terminal_job_retention_secs` doc-comment to note that `sacct` history is retained but `scontrol show job` stops finding the job after the window, that eviction runs once per `scheduler.interval_secs`, and that the window is floored to the reconcile interval.

yansun1996:
The earlier `E2E / native-host` failure was an unrelated infra flake, not a regression in this PR: the run errored in test setup with `docker: failed to bind host port 0.0.0.0:55432/tcp: address already in use` when starting the Postgres container — 194 tests passed and the single error was the container failing to come up on the hardcoded host port. This is the e2e Postgres port-collision flake that #590 addresses; a re-run clears it.

shiv-tyagi (approved):
LGTM

PR #584, opened 2026-08-06
fix(spurctld): cool down a node that rejects a dispatch as resources-unavailable

yansun1996:
## What this fixes

When an agent rejects a batch dispatch because the controller-allocated GPUs are already in use in its local allocation table (a `ResourceExhausted` rejection), the controller's view of that node is stale. The scheduler kept selecting the same node on every tick, so the job hot-looped against it. A handful of desynced nodes could starve placement cluster-wide even while other nodes sat idle.

## Approach

On a resources-unavailable rejection, the controller now places the rejecting node on a short, self-expiring dispatch cooldown and excludes it from the scheduler's placement view until the cooldown lapses. This breaks the every-tick hot-loop while the node's actual state is re-learned.

- New config knob `controller.dispatch_reject_cooldown_secs` (default 30s, 0 disables), validated at load against an upper bound so it cannot overflow the cooldown deadline and panic the controller.
- The cooldown map is **leader-local and transient** — never replicated or persisted. Followers don't schedule, and a new leader after failover simply starts empty and re-learns on the next rejection (a one-tick re-dispatch). No wire-format or persisted-state change.
- Only the rejecting node(s) are cooled; a multi-node dispatch where one node rejects cools just that node. The cooldown affects only the scheduler's placement view — it does not drain the node, cancel running jobs, or hide it from `sinfo`/`squeue`.
- The rejection is still additionally handled by the existing job-level requeue/backoff; the node-level cooldown is complementary (it steers *other* pending jobs away from the stale node too).

## Known limitations (follow-ups)

- **Scope: batch dispatch path only.** The standalone srun/salloc allocation path returns the same rejection but is not yet wired to the cooldown (its error flows through an `anyhow` boundary that drops the status code). Threading a typed error through that path is a small follow-up; documenting the asymmetry here rather than expanding this change.
- This is a targeted stop-gap for the hot-loop. The deeper fix — reconciling the controller's per-node allocation view against the node as source of truth so the stale view is *corrected* rather than merely avoided — is a separate, larger effort.

## Tests

- Unit test covers the cooldown state machine: a node is marked on rejection, is excluded from the scheduler's node view, reappears once the cooldown expires, and the `0` setting disables it.
- A mock-agent test drives the real dispatch-confirmation path with an agent returning `ResourceExhausted` and asserts the node lands on cooldown.
- Config validation rejects an out-of-range cooldown and accepts `0`.
- All new tests were confirmed to fail against the unfixed code.

## How it was tested

Validated on an isolated deployment that a normal dispatch is unaffected (a job schedules and completes through the new node-selection path) and that an out-of-range cooldown value is rejected at config load rather than panicking. The rejection→cooldown→exclusion path is triggered by a GPU-allocation conflict, which cannot be induced on the GPU-less test nodes; that path is covered by the mock-agent unit test driving the real confirmation code with a genuine `ResourceExhausted` rejection.

### DIFF
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -1264,7 +1264,8 @@ impl SlurmConfig {
                 ),
             });
         }
-        // Bounded so `Instant::now() + cooldown` cannot overflow and panic.
+        // Capped at the same 1-day ceiling as max_launch_backoff_secs — a longer
+        // dispatch cooldown has no legitimate operational use.
         if self.controller.dispatch_reject_cooldown_secs > MAX_LAUNCH_BACKOFF_SECS {
             return Err(ConfigError::InvalidValue {
                 field: "controller.dispatch_reject_cooldown_secs".into(),
@@ -2460,8 +2461,8 @@ terminal_job_retention_secs = {MAX_TERMINAL_JOB_RETENTION_SECS}
 
     #[test]
     fn controller_config_rejects_out_of_range_dispatch_reject_cooldown_secs() {
-        // Past this bound `Instant::now() + cooldown` overflows and panics, so
-        // it must be refused at load. Zero is valid (disables the cooldown).
+        // Past this bound load must reject rather than accept an unbounded
+        // dispatch cooldown. Zero is valid (disables the cooldown).
         let toml = format!(
             r#"
 cluster_name = "test"
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -1156,6 +1156,16 @@ impl SlurmConfig {
                 ),
             });
         }
+        // Bounded so `Instant::now() + cooldown` cannot overflow and panic.
+        if self.controller.dispatch_reject_cooldown_secs > MAX_LAUNCH_BACKOFF_SECS {
+            return Err(ConfigError::InvalidValue {
+                field: "controller.dispatch_reject_cooldown_secs".into(),
+                value: format!(
+                    "{} (must be at most {})",
+                    self.controller.dispatch_reject_cooldown_secs, MAX_LAUNCH_BACKOFF_SECS
+                ),
+            });
+        }
         if self.cluster.enabled {
             if self.cluster.distro != "k0s" {
                 return Err(ConfigError::InvalidValue {
@@ -1930,6 +1940,41 @@ max_launch_backoff_secs = 0
         assert!(
             err.to_string().contains("max_launch_backoff_secs"),
             "unexpected error: {err}"
+        );
+    }
+
+    #[test]
+    fn controller_config_rejects_out_of_range_dispatch_reject_cooldown_secs() {
+        // Past this bound `Instant::now() + cooldown` overflows and panics, so
+        // it must be refused at load. Zero is valid (disables the cooldown).
+        let toml = format!(
+            r#"
+cluster_name = "test"
+
+[controller]
+dispatch_reject_cooldown_secs = {}
+"#,
+            MAX_LAUNCH_BACKOFF_SECS + 1
+        );
+        let err = SlurmConfig::load_from_str(&toml).unwrap_err();
+        assert!(
+            err.to_string().contains("dispatch_reject_cooldown_secs"),
+            "unexpected error: {err}"
+        );
+
+        let ok = r#"
+cluster_name = "test"
+
+[controller]
+dispatch_reject_cooldown_secs = 0
+"#;
+        assert_eq!(
+            SlurmConfig::load_from_str(ok)
+                .unwrap()
+                .controller
+                .dispatch_reject_cooldown_secs,
+            0,
+            "zero (disabled) must be accepted"
         );
     }
 
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -320,10 +320,19 @@ pub struct ControllerConfig {
     /// the inverse `SchedulerParameters=nohold_on_prolog_fail`.
     #[serde(default = "default_hold_on_prolog_fail")]
     pub hold_on_prolog_fail: bool,
+
+    /// Seconds a node is skipped for new dispatch after rejecting one as
+    /// resources-unavailable, so it isn't re-picked every tick (default 30, 0 disables).
+    #[serde(default = "default_dispatch_reject_cooldown_secs")]
+    pub dispatch_reject_cooldown_secs: u64,
 }
 
 fn default_max_batch_requeue() -> u32 {
     5
+}
+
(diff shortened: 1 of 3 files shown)


crates/spur-core/src/config.rs
```diff
@@ -1146,6 +1156,16 @@ impl SlurmConfig {
                 ),
             });
         }
+        // Bounded so `Instant::now() + cooldown` cannot overflow and panic.
```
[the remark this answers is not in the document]
yansun1996: Good catch — reworded to say it reuses the same operational cap as `max_launch_backoff_secs` rather than claiming an overflow, in both the validation comment and the test.

crates/spurctld/src/scheduler_loop.rs
```diff
@@ -135,7 +135,7 @@ pub async fn run(cluster: Arc<ClusterManager>, raft: Arc<RaftHandle>) {
         }
         let hit_depth_limit = pending.len() > max_jobs;
 
-        let nodes = cluster.get_nodes();
+        let nodes = cluster.schedulable_nodes();
```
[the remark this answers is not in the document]
yansun1996: Fixed the wording to "no schedulable nodes" so it doesn't claim non-registration when nodes are just cooling.

yansun1996:
Rebased onto main (no conflicts). Also fixed the failing native-host E2E test: `schedulable_nodes()` excluded a cooling node from the entire scheduling cycle with no allowance for a job that has nowhere else to go, so a `--nodelist`-pinned job (like the chronic-preemption test's single-node pin) could get starved for the whole cooldown window if a preempt-then-redispatch race triggered one reject on that same node. It now re-includes a cooling node when a pending job's nodelist names it; jobs with alternatives are unaffected, so the hot-loop prevention still holds for them.

crates/spurctld/src/cluster.rs
```diff
@@ -2054,6 +2079,28 @@ impl ClusterManager {
         self.nodes.read().values().cloned().collect()
     }
 
+    /// Nodes eligible for new placement this tick: all nodes minus those within
+    /// a dispatch cooldown after a resources-unavailable reject — except a
+    /// cooling node explicitly named by a pending job's `--nodelist`, which has
+    /// nowhere else to go and would otherwise starve for the cooldown's
+    /// duration instead of being retried, the outcome the cooldown exists to
+    /// avoid only for jobs that actually have an alternative.
+    pub fn schedulable_nodes(&self, pending: &[Job]) -> Vec<Node> {
```
shiv-tyagi: `schedulable_nodes` is misleading. It only filters the dispatch cooldown; it still returns `Down`/`Drain`/`Draining`/`Allocated` nodes (the real gate is `node.is_schedulable()` in `node_match.rs`, applied later). A reader expects this to return schedulable nodes, and the name collides conceptually with `Node::is_schedulable()`. Rename to say what it does, e.g. `nodes_off_dispatch_cooldown` or `dispatch_eligible_nodes`.

crates/spurctld/src/cluster.rs
```diff
@@ -2054,6 +2079,28 @@ impl ClusterManager {
...
+            .filter_map(|nl| spur_core::hostlist::expand(nl).ok())
+            .flatten()
+            .collect();
+        self.nodes
+            .read()
+            .values()
+            .filter(|n| !cooling.contains(&n.name) || pinned.contains(&n.name))
+            .cloned()
+            .collect()
+    }
```
shiv-tyagi: This runs every scheduler tick. In the common case nothing is on cooldown, but it still iterates all pending jobs and expands every `--nodelist` hostlist to build `pinned` before discarding it. Skip that work when nothing is cooling:

```rust
let cooling = self.nodes_on_dispatch_cooldown();
if cooling.is_empty() {
    return self.nodes.read().values().cloned().collect();
}
```

crates/spurctld/src/cluster.rs
```diff
@@ -2054,6 +2079,28 @@ impl ClusterManager {
         self.nodes.read().values().cloned().collect()
     }
 
+    /// Nodes eligible for new placement this tick: all nodes minus those within
+    /// a dispatch cooldown after a resources-unavailable reject — except a
+    /// cooling node explicitly named by a pending job's `--nodelist`, which has
+    /// nowhere else to go and would otherwise starve for the cooldown's
+    /// duration instead of being retried, the outcome the cooldown exists to
+    /// avoid only for jobs that actually have an alternative.
```
shiv-tyagi: Nit: per AGENTS.md, comments are 1-3 lines. This is a six-line run-on sentence (and uses an em dash). Keep the "why" but tighten it:

```suggestion
    /// Nodes eligible for new placement this tick: all nodes minus those on a
    /// dispatch cooldown. A cooling node explicitly pinned by a pending job's
    /// `--nodelist` is kept, since it has no alternative and would otherwise
    /// starve for the whole cooldown window.
```

crates/spurctld/src/cluster.rs
shiv-tyagi: A node cooled for a GPU-allocation conflict is excluded for all pending jobs, including CPU-only jobs that could never hit that conflict. Acceptable for a 30s stop-gap, and the PR frames it that way. Worth one line in the doc comment noting the exclusion is intentionally resource-agnostic, so the next reader doesn't file it as a bug.

shiv-tyagi (commented):
Design looks right. Keeping the `ResourceExhausted` reject as a transient, leader-local cooldown rather than reusing `Drain`/`Down` is the correct call: this is a controller-side view desync on a healthy node, not a node fault. Real faults already drain (prolog failure -> `drain_node`, heartbeat timeout -> `Down`, node-fault I/O -> self-drain, manual `scontrol`), and those states are admin-locked, sticky, and Raft-replicated, all wrong for a self-healing sub-minute blip.

Comments below are on the cooldown-layer implementation: a misleading function name, an avoidable hot-path cost, comment length, and a doc note on the resource-agnostic exclusion. None are blockers.

crates/spurctld/src/cluster.rs
yansun1996: Renamed to `nodes_off_dispatch_cooldown` — you're right, it doesn't touch node state at all.

crates/spurctld/src/cluster.rs
yansun1996: Added the short-circuit — thanks.

crates/spurctld/src/cluster.rs
yansun1996: Tightened, thanks.

crates/spurctld/src/cluster.rs
yansun1996: Added a note that the exclusion is resource-agnostic by design.

PR #589, opened 2026-08-07
fix(spur-cli): resolve first allocated node from compressed hostlist

### LINKED ISSUE #575: srun/sattach pick the wrong first node from a compressed nodelist
## What happens

`srun` and `sattach` extract the "first node" of a job by splitting the proto `nodelist` string on `,` and taking the first element:

- `crates/spur-cli/src/srun.rs:985` (`try_stream_output`): `nodelist.split(',').next()...`
- `crates/spur-cli/src/sattach.rs:76`: `nodelist.split(',').next().unwrap_or(nodelist)...`

But `JobInfo.nodelist` is a compressed Slurm-style hostlist (built via `hostlist::compress()` in `crates/spurctld/src/server.rs`). For a multi-node job the field is a bracketed pattern, not a comma-separated list of hostnames. Splitting it on `,` does not yield a hostname:

- `node[001-002]` has no top-level comma, so `split(',').next()` returns the literal `node[001-002]`.
- `node[001,003]` has a comma inside the brackets, so `split(',').next()` returns the broken fragment `node[001`.

Either way the result is used directly as a host: `sattach` builds `http://node[001-002]:6818` and fails to connect; `srun`'s `get_node()` probe fails and output streaming is silently skipped.

## Expected

The first allocated node should be resolved to a real hostname (for example, expand the hostlist and take element 0), so `srun` streaming and `sattach` connect to an actual node.

## Repro

1. Submit a job that lands on two or more nodes sharing a prefix (nodelist compresses to `node[001-002]`).
2. Run `sattach <jobid>`. It attempts to connect to `http://node[001-002]:6818` and fails.
3. With `srun`, output streaming is silently skipped because the `get_node()` probe for `node[001-002]` errors.

maybeharshit:
## What

`srun` and `sattach` connect directly to a job's first allocated node to stream
live output. They extracted that node from `JobInfo.nodelist` with
`nodelist.split(',').next()` — but the controller reports `nodelist` as a
*compressed* Slurm-style hostlist (built by `spur_core::hostlist::compress`),
not a raw comma list. For any job on 2+ nodes this yields a non-hostname:

- `node[001-002]` has no top-level comma, so the split returns the whole literal
  `node[001-002]`.
- `node[001,003]` has a comma only inside the brackets, so the split returns the
  broken fragment `node[001`.
- multi-prefix (`gpu[001-004],cpu[001-002]`) splits into the fragment `gpu[001-004`.

The result was used directly as a host: `sattach` dialed
`http://node[001-002]:6818` and failed to connect, and `srun`'s `get_node()`
probe failed so live output was silently skipped. Fixes ROCm/spur#575.

## Approach

- Add `spur_core::hostlist::expand_first`, which expands a pattern only far
  enough to yield its first hostname — the inverse of the controller's
  `compress`, and cheaper than `expand(..).next()` for large allocations.
- Add a shared `spur_cli::nodelist::first_allocated_node` helper that both call
  sites now use, replacing the duplicated `split(',')` logic. It falls back to a
  comma-split for a malformed pattern so a bad input still yields something
  connectable rather than nothing (mirroring `expand_hostlist_or_split`).
- Wire it into `sattach` (clear error if unresolvable) and `srun`
  `try_stream_output` (keeps the existing skip-on-failure behavior).

The unrelated `srun::first_node` helper (for user-typed `-w`, resolved
server-side) is intentionally left unchanged.

## Design choices

- Uses the exact inverse of what produced the string; [1] PR ROCm/spur#574
  hardened `compress()` and verified `expand(compress(x))` round-trips, so
  relying on `expand`/`expand_first` is guaranteed correct even for mixed
  zero-padding (`node[9,010-011]`) and multi-prefix lists.
- Client-side only: no proto, persisted Raft/WAL, or config change, so there is
  no upgrade/compat hazard.

## Testing

- Unit: new `expand_first` tests (range, gap, multi-prefix, mixed padding, plain
  list, single, empty, suffix bracket, unmatched-bracket error, and parity with
  `expand(..).next()`), plus `first_allocated_node` tests. `cargo clippy`
  (spur-core, spur-cli) clean; `cargo test` for spur-core and spur-cli pass.
- E2E on a 2-node LXD cluster with hostnames compressing to `spurnode[1-2]`:
  before the fix, `sattach` failed with a DNS error on `spurnode[1-2]`; after,
  `sattach` connects to `spurnode1` and streams live output, and `srun -N2`
  streams from both nodes with no controller-side connect errors.

### DIFF
--- a/crates/spur-core/src/hostlist.rs
+++ b/crates/spur-core/src/hostlist.rs
@@ -28,6 +28,21 @@ pub fn expand(pattern: &str) -> Result<Vec<String>, HostlistError> {
     Ok(results)
 }
 
+/// Expand a hostlist pattern only far enough to yield its first hostname.
+///
+/// Equivalent to `expand(pattern)?.into_iter().next()`, but stops after the
+/// first name instead of materializing every host. Useful when only the first
+/// allocated node is needed (e.g. connecting to a job's primary node) and the
+/// allocation may span thousands of nodes.
+pub fn expand_first(pattern: &str) -> Result<Option<String>, HostlistError> {
+    for part in split_top_level(pattern) {
+        if let Some(host) = first_single(part.trim())? {
+            return Ok(Some(host));
+        }
+    }
+    Ok(None)
+}
+
 /// Compress a list of hostnames into a compact hostlist pattern.
 ///
 /// Example: `["node001", "node002", "node003", "node005"]` → `"node[001-003,005]"`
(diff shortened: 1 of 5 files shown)


crates/spur-core/src/hostlist.rs
```diff
@@ -330,6 +345,52 @@ fn expand_single(pattern: &str, results: &mut Vec<String>) -> Result<(), Hostlis
...
+        let start: u64 = start_str
+            .parse()
+            .map_err(|_| HostlistError::InvalidRange(first_part.into()))?;
+        let end: u64 = end_str
+            .parse()
+            .map_err(|_| HostlistError::InvalidRange(first_part.into()))?;
+        if start > end {
+            return Err(HostlistError::InvalidRange(format!("{} > {}", start, end)));
+        }
+        format!("{:0>width$}", start, width = width)
```
shiv-tyagi: `expand_single` and `first_single` duplicate the range-parsing logic verbatim: find the dash, parse start/end, derive the zero-pad width, reject `start > end`. Only what happens after (loop the full range vs. take just `start`) differs.

A shared helper would remove the duplication without touching the early-exit behavior that makes `first_single` cheaper than `expand(..).next()`:

```suggestion
    let first_part = range_str.split(',').next().unwrap_or_default();
    let first_value = match parse_range_bounds(first_part)? {
        Some((start, _end, width)) => format!("{:0>width$}", start, width = width),
        None => first_part.to_string(),
    };
```

with `parse_range_bounds` extracted from the shared block in `expand_single`.

shiv-tyagi (approved):
LGTM

PR #593, opened 2026-08-07
fix(spurctld): enforce partition walltime limits and validate time config

shiv-tyagi:
## What this fixes

Partition `MaxTime` was never enforced, and invalid time strings in config silently became `UNLIMITED`. Jobs could exceed a partition's wall-time cap, and a typo like `max_time = "1h"` quietly removed the cap instead of applying it. Refs SPUR-122.

This is Group 1 of the fix (submit/schedule-time validation and defaulting). Group 2 (interactive allocation reaping via `InactiveLimit`) is a follow-up.

## Approach

- **`enforce_part_limits` (`NO`/`ALL`/`ANY`, default `NO`)**: gates submit-time rejection of over-limit jobs, mirroring Slurm's `EnforcePartLimits`. `ALL` rejects unless the job fits every requested partition; `ANY` rejects only when it fits none. Default `NO` keeps the prior admit-and-pend behavior. Deserialization is case-insensitive and accepts Slurm's `YES` alias.
- **Specific pending reason**: time violations now report `PartitionTimeLimit` instead of the generic `PartitionConfig`.
- **Loud config validation**: `SlurmConfig::validate()` rejects malformed partition `max_time`/`default_time` at load rather than silently dropping the cap.
- **Default-time fallback (opt-in)**: `apply_default_time_limit` follows `DefaultTime -> MaxTime -> scheduler.default_time_limit_minutes`. That cluster fallback now defaults to `0` (disabled), so on upgrade a `-t`-less job on an unlimited partition stays unbounded. Sites opt into reclamation by setting it `> 0`.
- **Suffixed durations**: time strings now accept `1h`, `90m`, `1h40m`, `2d12h`, `30s` in addition to Slurm grammar, across config, CLI `-t`, `scontrol`, and REST. Slurm grammar is tried first (disjoint, no reinterpretation of existing values) and output still renders `HH:MM:SS`.

## Backward compatibility

Non-breaking and backward compatible by default. With an unchanged config (`enforce_part_limits = NO`, `default_time_limit_minutes = 0`) no submit-time rejections are added and no job gets a wall-time it wouldn't have had before. No persisted-state, proto, or CLI/REST-output changes.

One intended behavior change worth a release note: a deployed `max_time = "1h"` was previously silently `UNLIMITED`; it now parses as a 60-minute cap (honoring intent). A genuinely invalid string (e.g. `"1 hour"`) now fails loudly at startup instead of removing the cap.

## Testing

- Unit tests for enforcement (`ALL`/`ANY` accept/reject incl. no-partition-fits), the default-time fallback branches, config validation, and suffixed-duration parsing (plus updated existing tests). Full suite green; clippy clean.
- E2E: `TestEnforcePartLimits` in `tests/native_host/e2e/test_partitions.py`.
- Bare-metal (4 nodes): confirmed `ALL` rejects `-t 2h`/`-t 90m` on a 1h partition and accepts `-t 45m`/`-t 1h`; `default_time = "10m"` applies to `-t`-less jobs; `default_time_limit_minutes = 0` leaves unlimited-partition jobs unbounded while `= 60` bounds them; `max_time = "1h"` loads as a 60-min cap; and `"1 hour"` fails startup loudly.

## Review follow-ups

Two rounds of review addressed in later commits:

- Submit-time enforcement runs only for a user-supplied `-t` (never an auto-filled default), and the `-t`-less default derives from the minimum resolved limit across all requested partitions. Error messages render at second precision.
- Submit-time enforcement checks every requested partition's configured `MaxTime` regardless of its `Up`/`Down`/`Drain` state, matching Slurm's `EnforcePartLimits` (limits are configured values, not current state). This keeps the outcome deterministic (independent of transient partition state) and avoids silently admitting an over-limit job whose only requested partition is down.
- `submit_job`'s config/partitions snapshot is threaded through all validators, so the whole submit path sees one consistent view under a concurrent `reconfigure()`.
- `validate()` rejects a partition whose finite `default_time` exceeds its finite `max_time` (which would otherwise auto-fill a `-t`-less job past its own cap and pend it forever).

Follow-up [#604](https://github.com/ROCm/spur/issues/604) tracks moving to Slurm-style per-partition default resolution at schedule time.

### DIFF
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -433,15 +433,10 @@ impl Default for AccountingConfig {
     }
 }
 
-/// When partition limits (currently the wall-time limit) are enforced against a
-/// job's request. Mirrors Slurm's `EnforcePartLimits`.
-///
-/// - `No` (default): over-limit jobs are admitted and pend with a reason
-///   (`PartitionTimeLimit`), matching Slurm's default and preserving prior
-///   behavior on upgrade.
-/// - `All`: reject at submit unless the job fits every requested partition.
-/// - `Any`: reject at submit only when the job fits none of the requested
-///   partitions.
+/// Submit-time enforcement of partition wall-time limits (Slurm's
+/// `EnforcePartLimits`). `No` (default) admits over-limit jobs to pend; `All`
+/// rejects unless the job fits every requested partition; `Any` rejects only
+/// when it fits none.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Default)]
 #[serde(rename_all = "UPPERCASE")]
 pub enum EnforcePartLimits {
@@ -1344,12 +1339,10 @@ impl SlurmConfig {
     }
 }
 
-/// Parse a partition time field, distinguishing UNLIMITED from a parse error.
-///
-/// `INFINITE`/`UNLIMITED` (case-insensitive) -> `Ok(None)` (no limit); a valid
-/// duration (Slurm grammar or a suffixed form like `"1h"`) -> `Ok(Some(minutes))`;
-/// anything else -> `Err`. Unlike [`parse_time_minutes`], a genuine typo like
-/// `"1 hour"` fails loudly instead of silently collapsing to UNLIMITED.
+/// Parse a partition time field, distinguishing UNLIMITED from a parse error:
+/// `INFINITE`/`UNLIMITED` -> `Ok(None)`, a valid duration -> `Ok(Some(minutes))`,
+/// anything else -> `Err`. Unlike [`parse_time_minutes`], a typo like `"1 hour"`
+/// fails loudly instead of collapsing to UNLIMITED.
 pub fn parse_partition_time(field: &str, s: &str) -> Result<Option<u32>, ConfigError> {
     let trimmed = s.trim();
     if trimmed.eq_ignore_ascii_case("INFINITE") || trimmed.eq_ignore_ascii_case("UNLIMITED") {
@@ -1364,12 +1357,9 @@ pub fn parse_partition_time(field: &str, s: &str) -> Result<Option<u32>, ConfigE
     }
 }
 
-/// Parse a time string to minutes.
-///
-/// Accepts Slurm grammar ("60" minutes, "H:MM", "H:MM:SS", "D-HH:MM:SS",
-/// "INFINITE"/"UNLIMITED") and, as a fallback, suffixed durations
-/// ("90m", "1h", "1h40m", "2d12h", "30s"). Sub-minute remainders round up to
-/// match the seconds-rounding of the "H:MM:SS" form.
+/// Parse a time string to minutes: Slurm grammar (`60`, `H:MM`, `H:MM:SS`,
+/// `D-HH:MM:SS`, `INFINITE`/`UNLIMITED`) or a suffixed duration (`90m`, `1h`,
+/// `2d12h`, `30s`). Sub-minute remainders round up.
 pub fn parse_time_minutes(s: &str) -> Option<u32> {
     let s = s.trim();
     if s.eq_ignore_ascii_case("INFINITE") || s.eq_ignore_ascii_case("UNLIMITED") {
@@ -1449,11 +1439,10 @@ fn parse_slurm_time_seconds(s: &str) -> Option<u64> {
 
 /// Parse a suffixed duration like "90m", "1h", "1h40m", "2d12h", "30s".
 ///
-/// Units are `d`/`h`/`m`/`s` (case-insensitive), each an integer, summed in any
-/// order. Requires at least one unit and must consume the whole string, so a
-/// bare number is rejected here (it means minutes via the Slurm grammar) and
-/// stray characters or trailing digits fail. This is the fallback that lets an
-/// operator write `max_time = "1h"` instead of `"01:00:00"`.
+/// Sum a suffixed duration like `1h40m` or `2d12h` to seconds. Units `d/h/m/s`
+/// (case-insensitive); requires at least one unit and must consume the whole
+/// string, so a bare number or trailing digits fail (the Slurm grammar owns
+/// bare numbers).
 fn parse_suffix_duration_seconds(s: &str) -> Option<u64> {
     let mut total: u64 = 0;
     let mut num: Option<u64> = None;
@@ -1530,6 +1519,26 @@ pub fn format_time(total_minutes: Option<u32>) -> String {
                 format!("{}-{:02}:{:02}:00", days, hours, minutes)
             } else {
                 format!("{:02}:{:02}:00", hours, minutes)
+            }
+        }
+    }
+}
+
+/// Format seconds as D-HH:MM:SS or HH:MM:SS. Unlike [`format_time`] this keeps
+/// second precision, so a sub-minute request is not rounded when reported.
+pub fn format_time_seconds(total_seconds: Option<i64>) -> String {
+    match total_seconds {
+        None => "UNLIMITED".into(),
+        Some(secs) => {
+            let secs = secs.max(0);
+            let days = secs / 86_400;
+            let hours = (secs % 86_400) / 3_600;
+            let minutes = (secs % 3_600) / 60;
+            let seconds = secs % 60;
+            if days > 0 {
+                format!("{days}-{hours:02}:{minutes:02}:{seconds:02}")
+            } else {
+                format!("{hours:02}:{minutes:02}:{seconds:02}")
             }
         }
     }
@@ -1941,6 +1950,39 @@ memory_mb = 1024000
         // A genuine typo still fails loudly.
         assert!(parse_partition_time("p.max_time", "1 hour").is_err());
         assert!(parse_partition_time("p.max_time", "banana").is_err());
+    }
+
+    #[test]
+    fn enforce_part_limits_parses_case_insensitively_and_yes_alias() {
+        #[derive(serde::Deserialize)]
+        struct W {
+            v: EnforcePartLimits,
+        }
+        let parse = |v: &str| toml::from_str::<W>(&format!("v = \"{v}\"")).unwrap().v;
+        assert_eq!(parse("no"), EnforcePartLimits::No);
+        assert_eq!(parse("all"), EnforcePartLimits::All);
+        assert_eq!(parse("ALL"), EnforcePartLimits::All);
+        assert_eq!(parse("Yes"), EnforcePartLimits::All);
+        assert_eq!(parse("any"), EnforcePartLimits::Any);
+        assert_eq!(parse("ANY"), EnforcePartLimits::Any);
+    }
+
+    #[test]
+    fn enforce_part_limits_rejects_unknown_value() {
+        #[derive(serde::Deserialize)]
+        struct W {
+            #[allow(dead_code)]
+            v: EnforcePartLimits,
+        }
+        assert!(toml::from_str::<W>("v = \"sometimes\"").is_err());
+    }
+
+    #[test]
+    fn format_time_seconds_keeps_second_precision() {
+        assert_eq!(format_time_seconds(Some(630)), "00:10:30");
+        assert_eq!(format_time_seconds(Some(3600)), "01:00:00");
+        assert_eq!(format_time_seconds(Some(90_061)), "1-01:01:01");
+        assert_eq!(format_time_seconds(None), "UNLIMITED");
     }
 
     #[test]
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -433,10 +433,8 @@ impl Default for AccountingConfig {
     }
 }
 
-/// Submit-time enforcement of partition wall-time limits (Slurm's
-/// `EnforcePartLimits`). `No` (default) admits over-limit jobs to pend; `All`
-/// rejects unless the job fits every requested partition; `Any` rejects only
-/// when it fits none.
+/// Submit-time enforcement of partition wall-time (Slurm `EnforcePartLimits`):
+/// `No` admits over-limit jobs to pend, `All` needs all fit, `Any` needs one.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Default)]
 #[serde(rename_all = "UPPERCASE")]
 pub enum EnforcePartLimits {
@@ -479,10 +477,8 @@ pub struct SchedulerConfig {
     /// Fair-share decay half-life (days).
     #[serde(default = "default_halflife")]
     pub fairshare_halflife_days: u32,
-    /// Cluster-wide fallback wall-time (minutes) for a job that sets no `-t`
-    /// and lands on a partition with no `DefaultTime`. `0` (the default)
-    /// disables the fallback, leaving such jobs unbounded (prior behavior).
-    /// Set > 0 to bound otherwise-unlimited jobs so they cannot run forever.
+    /// Cluster-wide fallback wall-time (minutes) for a `-t`-less job on a partition
+    /// with no `DefaultTime`. `0` (default) leaves such jobs unbounded; > 0 bounds them.
     #[serde(default = "default_time_limit")]
     pub default_time_limit_minutes: u32,
     /// Whether to reject over-limit jobs at submit (Slurm `EnforcePartLimits`).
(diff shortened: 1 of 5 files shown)


biluriuday:
Important (fix before merge)
1. Auto default-time + EnforcePartLimits=ALL can reject an under-specified job. apply_default_time_limit (crates/spurctld/src/cluster.rs:5689) resolves against only the first requested partition and, when default_time_limit_minutes > 0, assigns that partition's MaxTime (cluster.rs:5716). But ALL enforcement then checks every partition. A -t-less job to big,small (MaxTime 120 vs 60) gets auto-assigned 120 from big, then rejected against small — despite the user never requesting any wall time.

Fix: for a multi-partition request, derive the default from the minimum MaxTime across requested partitions, or skip ALL enforcement for an auto-derived (non-user-supplied) limit. Add a test pinning the chosen semantics.

Minor
2. Self-contradictory error message. The over-limit message builds the requested string at minute granularity (cluster.rs:597, format_time) even though enforcement is second-precision — so 10m30s vs a 00:10:00 cap correctly rejects but prints "00:10:00 exceeds MaxTime 00:10:00". Render request/cap at second precision.

3. Docs undersell the fallback's reach. docs/admin-guide/configuration.rst says default_time_limit_minutes bounds "otherwise-unlimited jobs," but enabling it also makes -t-less jobs on finite-MaxTime partitions (no DefaultTime) default to that MaxTime (cluster.rs:5716). Add a sentence so operators aren't surprised.

Release note
default_time_limit_minutes is now actually enforced (was inert). Untouched configs are unchanged, but any site that had set it expecting an effect will now see new behavior — worth a one-line note.

Bottom line: issue 1 is the only thing that can produce a genuinely surprising rejection and should be resolved (plus a test) before merge; 2 and 3 are cheap UX/doc wins

crates/spurctld/src/cluster.rs
```diff
@@ -5604,19 +5678,45 @@ fn apply_default_partition(spec: &mut JobSpec, partitions: &[Partition]) {
...
     if let Some(minutes) = partition.and_then(|p| p.default_time_minutes) {
         spec.time_limit = Some(chrono::Duration::minutes(minutes as i64));
+        return;
     }
+
+    // Opt-in reclamation: only bound the job when a cluster fallback is set.
+    if cluster_default_minutes == 0 {
+        return;
+    }
+    let minutes = partition
```
yansun1996, line 5716: With `EnforcePartLimits=ALL` and a multi-partition request (e.g. `-p big,small`), this only considers the first requested partition's `MaxTime` when auto-filling a `-t`-less job's time limit. If a later partition in the list has a smaller `MaxTime`, `validate_partition_time_limit` will then reject the auto-assigned value against it — even though the user never requested a wall time. Could this derive from the minimum `MaxTime` across all requested partitions instead, or otherwise avoid feeding an auto-derived limit into `ALL` enforcement?

crates/spurctld/src/cluster.rs
```diff
@@ -545,6 +553,56 @@ impl ClusterManager {
...
+        }
+        let Some(time_limit) = spec.time_limit.as_ref() else {
+            return Ok(());
+        };
+        let Some(partition_spec) = spec.partition.as_deref().filter(|p| !p.is_empty()) else {
+            return Ok(());
+        };
+
+        let partitions = self.partitions.read();
+        let requested: Vec<&Partition> = requested_partition_names(Some(partition_spec))
```
yansun1996, line 572: This re-implements partition-name resolution that already exists as `spur_core::partition::matched_partitions` (used elsewhere, e.g. in the scheduler loop). Worth calling that here instead of duplicating the `filter_map`/`find` pattern?

crates/spurctld/src/cluster.rs
```diff
@@ -545,6 +553,56 @@ impl ClusterManager {
         )))
     }
 
+    /// Reject a job whose wall-time exceeds the requested partition's `MaxTime`,
+    /// gated on `EnforcePartLimits` (Slurm semantics). When disabled (the
+    /// default), over-limit jobs are admitted and pend with `PartitionTimeLimit`.
+    fn validate_partition_time_limit(&self, spec: &JobSpec) -> Result<(), SubmitError> {
+        let mode = self.config().scheduler.enforce_part_limits;
```
yansun1996, line 560: `apply_default_time_limit` earlier in `submit_job` reuses the `config`/`partitions` snapshots taken at the top of the function, but this re-fetches both independently. It's a narrow window, but a concurrent `reconfigure()` between the two calls could make one submit's default-time computation and its enforcement check see different config/partition state. Since the snapshots are already available in `submit_job`, would it be worth threading them through instead of re-fetching here?

docs/admin-guide/configuration.rst
```diff
@@ -236,9 +236,17 @@ Scheduling loop cadence, per-cycle limits, and fairshare decay.
      - Fairshare usage decay half-life, in days.
    * - ``default_time_limit_minutes``
      - integer
-     - ``60``
-     - Default job time limit (minutes) when neither the job nor its partition sets
-       one.
+     - ``0``
+     - Cluster-wide fallback wall-time (minutes) for a job that sets no ``-t`` and
+       lands on a partition with no ``DefaultTime``. ``0`` disables the fallback,
+       leaving such jobs unbounded. Set > 0 to bound otherwise-unlimited jobs.
```
yansun1996, line 242: Two things worth calling out here: this is a default change from `60` in the current release, so a cluster that never set this explicitly will go from a 60-minute cap on `-t`-less jobs to unbounded after upgrading. Also, when set > 0, this can silently pin a `-t`-less job to a partition's `MaxTime` (not just a flat cluster-wide value) if that partition has no `DefaultTime` — could be worth a sentence on both.

yansun1996 (changes requested):
Thanks for closing this gap — the enforcement/validation/duration-parsing work looks solid and stays backward compatible by default. One correctness issue around multi-partition auto-default-time + `EnforcePartLimits=ALL` should be fixed before merge (see inline); the rest below are smaller polish items and test-coverage suggestions.

**A couple things without a good inline anchor:**
- `partition_block`'s doc comment ("limits rejected by every Up alternative return `PartitionConfig`") is now stale — the body can also return the more specific `PartitionTimeLimit`.
- Several new doc comments run 3+ lines (`EnforcePartLimits` enum doc, `parse_partition_time`/`parse_time_minutes`/`parse_suffix_duration_seconds`, `apply_default_time_limit`) — worth trimming to a line or two each.
- The `EnforcePartLimits::No` arm inside the `satisfied` match in `validate_partition_time_limit` is unreachable (the function already returns early for `No`) — harmless, just dead code.
- Agree with the existing Copilot comment on the error message's minute- vs. second-precision mismatch.

**Missing test coverage worth adding:**
- The multi-partition + `ALL` + auto-default-time scenario below.
- Case-insensitive parsing of `enforce_part_limits` (`"all"`, `"Yes"`, `"any"`) and the reject-on-garbage-value path — the `Deserialize` impl looks correct but nothing exercises it.
- A partition with both `default_time` and a smaller `max_time` (plus a nonzero cluster default), to pin that `DefaultTime` always wins in the fallback chain.

Also noting as a follow-up, not a blocker: `spec.time_limit` can be negative (no non-negativity check on the client-supplied duration), which would trivially satisfy `partition_time_allows` regardless of mode — pre-existing, adjacent to this PR's changes.

shiv-tyagi:
Correction on the `default_time_limit_minutes` point (the 60 -> 0 default change).

`default_time_limit_minutes` was never enforced before this PR. On `main` it appears only as the struct field, its `Default` of 60, and a test asserting that default. No submit or scheduler path reads it, so a `-t`-less job on a partition with no `DefaultTime` was already unbounded.

So there is no lost 60-minute cap on upgrade:

- A site that never set it: unbounded before (the 60 was inert), unbounded after (0 disables the fallback). No change.
- This PR wires the knob up for the first time. The default moves 60 -> 0 specifically so `-t`-less jobs stay unbounded exactly as they do today.
- The only new behavior is for a site that had set it > 0 expecting an effect it never had.

I kept the default at 0 for that reason. Restoring 60 would itself be the behavior change, since it would start capping `-t`-less jobs that run unbounded today. Added a release note to the config docs spelling this out.

crates/spurctld/src/cluster.rs
```diff
@@ -545,6 +553,56 @@ impl ClusterManager {
...
+            return Ok(());
+        }
+
+        let offending = requested
+            .iter()
+            .find(|p| !allows(p))
+            .unwrap_or(&requested[0]);
+        let requested_str =
+            spur_core::config::format_time(Some(time_limit.num_minutes().max(0) as u32));
+        let max_str = spur_core::config::format_time(offending.max_time_minutes);
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed in 1e08eff. Added `format_time_seconds` and render both the requested time and the cap at second precision. A `0:01:30` request against a `00:01:00` cap now reports `00:01:30 exceeds MaxTime 00:01:00` instead of the contradictory `00:01:00`.

crates/spurctld/src/cluster.rs
shiv-tyagi: Done in 1e08eff, two parts:

- `apply_default_time_limit` now derives from the minimum resolved limit across all requested partitions, so the auto-filled value fits every one.
- Submit-time enforcement now runs only for a user-supplied `-t`; an auto-filled default is never fed into ALL/ANY, so it cannot be rejected.

The single-`time_limit`-at-submit model still under-binds a job that later lands on a larger partition. Filed [604](https://github.com/ROCm/spur/issues/604) for the schedule-time per-partition resolution.

crates/spurctld/src/cluster.rs
shiv-tyagi: Done in 1e08eff. `validate_partition_time_limit` now calls `spur_core::partition::matched_partitions` instead of the hand-rolled `filter_map`/`find`.

crates/spurctld/src/cluster.rs
shiv-tyagi: Fixed in 1e08eff. `submit_job` threads its single `config`/`partitions` snapshot into both `apply_default_time_limit` and `validate_partition_time_limit`, so a concurrent `reconfigure()` cannot split state between defaulting and enforcement.

docs/admin-guide/configuration.rst
shiv-tyagi: Updated in 1e08eff, with a correction: the old `60` was inert. Nothing on main reads `default_time_limit_minutes` in the scheduler, so `-t`-less jobs already run unbounded today. Moving the default to `0` preserves that exactly; no 60-minute cap is lost on upgrade. The doc now states this, and adds that when set > 0, a `-t`-less job on a partition with a finite `MaxTime` and no `DefaultTime` is pinned to that `MaxTime` (the smallest across a multi-partition request), not a flat cluster value.

shiv-tyagi:
@yansun1996 I have addressed your comments. Can you please take a look?

crates/spurctld/src/cluster.rs
```diff
@@ -5326,38 +5358,104 @@ fn partition_block(job: &Job, partitions: &[Partition]) -> Option<spur_core::job
...
+    }
+    let Some(time_limit) = spec.time_limit.as_ref() else {
+        return Ok(());
+    };
+    let Some(partition_spec) = spec.partition.as_deref().filter(|p| !p.is_empty()) else {
+        return Ok(());
+    };
+
+    // Existence was already checked in validate_partition.
+    let requested = spur_core::partition::matched_partitions(Some(partition_spec), partitions);
```
yansun1996, line 5407: `validate_partition_time_limit` doesn't filter to `PartitionState::Up`, unlike `partition_block` just above it. With `EnforcePartLimits=ALL` and `-p up,down_smaller`, a Down partition's `MaxTime` can reject a job that should just pend. If every requested partition is inactive, this also produces a misleading "exceeds MaxTime" error instead of `PartitionInactive`. Could this filter to `Up` partitions first, same as `partition_block`?

crates/spurctld/src/cluster.rs
```diff
@@ -416,11 +416,31 @@ impl ClusterManager {
 
     /// Submit a new job. If it has an array spec, expand into individual tasks.
     pub fn submit_job(&self, mut spec: JobSpec) -> Result<SubmitOutcome, SubmitError> {
-        apply_default_partition(&mut spec, &self.partitions.read());
-        apply_default_time_limit(&mut spec, &self.partitions.read());
+        // One config/partitions snapshot for the whole submit path, so defaulting
```
yansun1996, line 419: This comment says one snapshot covers "the whole submit path," but `validate_partition` and `validate_partition_node_bounds` still call `self.partitions.read()` fresh internally, so they can still see different partition state than the time-limit checks during a concurrent `reconfigure()`. Worth threading the snapshot into those two as well, or narrowing the comment to what's actually covered.

crates/spur-core/src/config.rs
```diff
@@ -1246,10 +1285,24 @@ impl SlurmConfig {
                 }
             }
         }
+
+        // Reject malformed partition time strings at load rather than silently
+        // treating them as UNLIMITED (a typo like "1 hour" must fail, not
+        // remove the cap).
+        for pc in &self.partitions {
```
yansun1996, line 1288: Now that malformed time strings are rejected here, could this also reject a partition where `default_time > max_time`? Right now that combination silently makes `apply_default_time_limit` hand out a default exceeding the partition's own cap, and since it's auto-filled it's never enforced.

crates/spurctld/src/cluster.rs
```diff
@@ -5326,38 +5358,104 @@ fn partition_block(job: &Job, partitions: &[Partition]) -> Option<spur_core::job
...
     Some(if has_up_partition {
-        PendingReason::PartitionConfig
+        block_reason.unwrap_or(PendingReason::PartitionConfig)
     } else {
         PendingReason::PartitionInactive
     })
 }
 
-fn partition_limits_allow(job: &Job, part: &Partition) -> bool {
+/// Reject a job whose (user-requested) wall-time exceeds a requested partition's
```
yansun1996: A few of the new doc comments run 4+ lines (this one, `EnforcePartLimits`, `default_time_limit_minutes`, `parse_suffix_duration_seconds`). Worth trimming to a line or two each.

yansun1996 (commented):
Re-reviewed after the latest fixes — the 5 prior points are all genuinely addressed (verified against the code, not just the commit message). Two smaller things turned up in this pass, left as inline comments: an `Up`-state filtering gap in `validate_partition_time_limit`, and the single-snapshot fix being incomplete. Also a doc-validation gap and a couple of doc-comment nits.

crates/spurctld/src/cluster.rs
shiv-tyagi: Fixed in d4d9875. `validate_partition_time_limit` now filters to `Up` partitions before enforcing, matching `partition_block`. A Down partition's `MaxTime` no longer rejects a job that fits an Up one, and an all-inactive request returns Ok so it pends as `PartitionInactive` instead of printing "exceeds MaxTime". Tests: `validate_partition_time_limit_ignores_down_partitions`, `_ok_when_all_requested_inactive`, `_rejects_over_limit_up_partition`.

crates/spurctld/src/cluster.rs
shiv-tyagi: Fixed in d4d9875. Threaded the single snapshot into both `validate_partition` and `validate_partition_node_bounds` (they no longer call `self.partitions.read()`), so the whole submit path reads one consistent partition view. The comment is now accurate.

crates/spur-core/src/config.rs
shiv-tyagi: Done in d4d9875. `validate()` now rejects a partition whose finite `default_time` exceeds its finite `max_time` (`==` allowed, UNLIMITED `max_time` bypassed). This closes the footgun where the auto-filled default exceeds the partition's own cap and the job pends forever. Tests: `load_rejects_default_time_above_max_time`, `load_accepts_default_time_equal_to_max_time`, `load_accepts_default_time_with_unlimited_max_time`.

crates/spurctld/src/cluster.rs
shiv-tyagi: Done in d4d9875. Trimmed all four (`validate_partition_time_limit`, `EnforcePartLimits`, `default_time_limit_minutes`, `parse_suffix_duration_seconds`) to one or two lines.

yansun1996 (dismissed):
Re-reviewed at d4d9875 — all points from the last round check out (Up-partition filtering, single submit-time snapshot, default>max validation). Nice work tightening this up. Two small non-blocking follow-ups if you want to fold them in: a few new doc comments are still 3+ lines (worth trimming to 1-2), and the admin doc could mention that a malformed `max_time`/`default_time` now fails config load at startup instead of silently becoming `UNLIMITED`. Approving.

crates/spurctld/src/cluster.rs
```diff
@@ -5326,38 +5364,109 @@ fn partition_block(job: &Job, partitions: &[Partition]) -> Option<spur_core::job
...
+        return Ok(());
+    };
+
+    // Only Up partitions can schedule the job, so enforce against those alone
+    // (matching partition_block); a Down partition's MaxTime must not reject a
+    // job that would just pend. Existence was already checked in validate_partition.
+    let requested: Vec<&Partition> =
+        spur_core::partition::matched_partitions(Some(partition_spec), partitions)
+            .into_iter()
+            .filter(|p| p.state == spur_core::partition::PartitionState::Up)
```
biluriuday: New "enforce only against Up partitions" scoping diverges from Slurm

underEnforcePartLimits=ALL, filtering to onlyUppartitions means a DOWN/DRAINED partition's (possibly stricter)MaxTimeis silently ignored. Slurm'sEnforcePartLimitsvalidates against the requested partitions regardless of current state. Also, ifnorequested partition is currently Up, the job is admitted completely unchecked (requested.is_empty() => Ok).

enforcement outcome becomes time-dependent on transient node/partition state — the samesbatchis accepted or rejected depending on whether a partition happens to be down at submit. That's surprising and hard to diagnose.

shiv-tyagi: Fixed in ef2af91. Reverted the Up-only filter. `validate_partition_time_limit` now checks every requested partition's configured `MaxTime` regardless of state, matching Slurm's `EnforcePartLimits` (limits are about configured values, not current UP/DOWN).

This removes the state-dependence (same `sbatch` now yields the same result) and closes the `requested.is_empty() => Ok` hole: a request naming only a Down partition is enforced, not silently admitted. Tests updated to assert Down partitions are enforced under `ALL`.

Note this reverts a change from the previous round (Yan asked to scope enforcement to Up partitions). Your reading of Slurm semantics is the correct one, so I went with it. @yansun1996 flagging in case you disagree.

biluriuday (approved):
LGTM

crates/spurctld/src/cluster.rs
yansun1996: thanks @biluriuday @shiv-tyagi after reviewing this point and refer to the Slurm doc this makes more sense, let's keep this part as the original behavior that is not state-dependent.

yansun1996 (approved):

PR #594, opened 2026-08-07
feat(spurctld): reap abandoned interactive allocations via client keepalive

shiv-tyagi:
## What

Interactive allocations (`salloc`/`srun`) that lose their client (crash, network drop, closed laptop) previously stayed `RUNNING` forever, holding compute nodes indefinitely. This adds opt-in reaping modeled on Slurm's `InactiveLimit`.

Part of SPUR-122 (walltime/limit enforcement). Independent of [#593](https://github.com/ROCm/spur/pull/593) (partition walltime limits), which is the other half of the same effort. Both branch off `main` and touch `config.rs`/`configuration.rst` in non-overlapping spots.

## Approach

- **Client keepalive**: new `JobKeepalive` RPC on `SlurmController`. `salloc` and `srun` send it every 30s while they hold an allocation, then stop when the shell/step exits.
- **Ephemeral last-seen state**: the controller records the last keepalive time per interactive job in an in-memory map on `ClusterManager`. Never persisted to Raft/WAL (mirrors `Node::last_heartbeat`), so no log spam and no snapshot growth.
- **Leader-gated reaper**: a periodic task (10s cadence, same as time-limit enforcement) finalizes any interactive allocation idle past `scheduler.inactive_limit_secs`. It reuses the existing TIMEOUT path (`complete_job(.., Timeout)` + SIGKILL to agents) instead of a raw cancel, inheriting epoch gating and sidestepping the requeue race.
- **Failover-safe**: a newly promoted leader seeds each running allocation's last-seen to \"now\", giving clients a grace period to re-ping before they can be reaped.

## Backward compatibility

`inactive_limit_secs` defaults to `0` (disabled). No behavior change on upgrade. Reaping requires both the config knob set and keepalive-capable clients. Proto change is purely additive (new messages + RPC, no renumbering).

## Example

```toml
# spur.conf
[scheduler]
inactive_limit_secs = 60
```

1. `srun --pty bash` → allocation goes `RUNNING`, client pings every 30s.
2. Client stays alive (or user is actively working): allocation is never reaped.
3. Client dies / connection drops: no more keepalives.
4. After 60s of silence the reaper logs `interactive allocation idle past InactiveLimit — reaping`, finalizes the job as `TIMEOUT`, SIGKILLs any step on the node, and the node returns to idle.

## Testing

- Unit tests for the reaper predicate (reaps only idle allocations; disabled mode only prunes state).
- `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` clean.
- `cargo test --locked` green (610 spurctld, 455 spur-core).
- Bare-metal 4-node cluster:
  - Attended `srun sleep 600` survived well past the 60s limit (keepalives kept it alive).
  - Abandoned allocation (client killed) was reaped to `TIMEOUT` and the node freed.
  - With default `inactive_limit_secs = 0`, an abandoned allocation stayed `RUNNING` (confirms non-breaking default).

### DIFF
--- a/crates/spur-cli/src/salloc.rs
+++ b/crates/spur-cli/src/salloc.rs
@@ -5,8 +5,14 @@ use crate::env_defaults::{apply_csv, apply_flag, apply_str, apply_string};
 use anyhow::{Context, Result};
 use clap::{ArgMatches, CommandFactory, FromArgMatches, Parser};
 use spur_core::spur_env::SpurEnv;
-use spur_proto::proto::{CancelJobRequest, GetJobRequest, JobSpec, SubmitJobRequest};
+use spur_proto::proto::{
+    CancelJobRequest, GetJobRequest, JobKeepaliveRequest, JobSpec, SubmitJobRequest,
+};
 use std::collections::HashMap;
+
+/// How often an interactive client pings the controller to keep its allocation
+/// attended. Must stay well below any configured `inactive_limit_secs`.
+const KEEPALIVE_INTERVAL_SECS: u64 = 30;
 
 /// Allocate resources for an interactive job.
 #[derive(Parser, Debug)]
@@ -230,6 +236,27 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
     env.set_with_slurm_twin("SPUR_CPUS_PER_TASK", job_info.cpus_per_task);
     env.set_with_slurm_twin("SPUR_SUBMIT_DIR", &job_info.work_dir);
 
+    // Keep the allocation attended while the shell is open: absence of these
+    // pings past the controller's InactiveLimit is what lets it reap an
+    // abandoned allocation.
+    let keepalive = {
+        let mut client = client.clone();
+        let user = user.clone();
+        tokio::spawn(async move {
+            let mut tick =
+                tokio::time::interval(tokio::time::Duration::from_secs(KEEPALIVE_INTERVAL_SECS));
+            loop {
+                tick.tick().await;
+                let _ = client
+                    .job_keepalive(JobKeepaliveRequest {
+                        job_id,
+                        user: user.clone(),
+                    })
+                    .await;
+            }
+        })
+    };
+
     let shell = std::env::var("SHELL").unwrap_or_else(|_| "/bin/bash".into());
     let mut cmd = tokio::process::Command::new(&shell);
     for (k, v) in env.into_map() {
@@ -242,6 +269,8 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
         .status()
         .await
         .context("failed to spawn shell")?;
+
+    keepalive.abort();
 
     // Shell exited — cancel allocation
     eprintln!("salloc: Relinquishing job allocation {}", job_id);
--- a/crates/spur-cli/src/salloc.rs
+++ b/crates/spur-cli/src/salloc.rs
@@ -5,14 +5,8 @@ use crate::env_defaults::{apply_csv, apply_flag, apply_str, apply_string};
 use anyhow::{Context, Result};
 use clap::{ArgMatches, CommandFactory, FromArgMatches, Parser};
 use spur_core::spur_env::SpurEnv;
-use spur_proto::proto::{
-    CancelJobRequest, GetJobRequest, JobKeepaliveRequest, JobSpec, SubmitJobRequest,
-};
+use spur_proto::proto::{CancelJobRequest, GetJobRequest, JobSpec, SubmitJobRequest};
 use std::collections::HashMap;
-
-/// How often an interactive client pings the controller to keep its allocation
-/// attended. Must stay well below any configured `inactive_limit_secs`.
-const KEEPALIVE_INTERVAL_SECS: u64 = 30;
 
 /// Allocate resources for an interactive job.
 #[derive(Parser, Debug)]
@@ -238,24 +232,9 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
 
     // Keep the allocation attended while the shell is open: absence of these
     // pings past the controller's InactiveLimit is what lets it reap an
-    // abandoned allocation.
-    let keepalive = {
-        let mut client = client.clone();
-        let user = user.clone();
-        tokio::spawn(async move {
-            let mut tick =
-                tokio::time::interval(tokio::time::Duration::from_secs(KEEPALIVE_INTERVAL_SECS));
-            loop {
-                tick.tick().await;
-                let _ = client
-                    .job_keepalive(JobKeepaliveRequest {
-                        job_id,
-                        user: user.clone(),
-                    })
-                    .await;
-            }
-        })
-    };
+    // abandoned allocation. The guard stops the pings on drop.
+    let _keepalive =
+        crate::interactive::spawn_keepalive(client.clone(), job_id, user.clone(), "salloc");
 
     let shell = std::env::var("SHELL").unwrap_or_else(|_| "/bin/bash".into());
     let mut cmd = tokio::process::Command::new(&shell);
@@ -270,7 +249,7 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
         .await
         .context("failed to spawn shell")?;
 
-    keepalive.abort();
+    drop(_keepalive);
 
     // Shell exited — cancel allocation
     eprintln!("salloc: Relinquishing job allocation {}", job_id);
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -10,8 +10,7 @@ use spur_core::config::HooksConfig;
 use spur_proto::proto::slurm_controller_client::SlurmControllerClient;
 use spur_proto::proto::{
     CancelJobRequest, CompleteJobRequest, CreateJobStepRequest, GetJobRequest, GetNodeRequest,
-    JobKeepaliveRequest, JobSpec, JobState, RunStepRequest, StreamJobOutputRequest,
-    SubmitJobRequest,
+    JobSpec, JobState, RunStepRequest, StreamJobOutputRequest, SubmitJobRequest,
 };
 use std::collections::HashMap;
 use std::io::Write as _;
@@ -628,32 +627,6 @@ fn install_ctrl_c_cancel(
     })
 }
 
-/// How often srun pings the controller to keep its allocation attended. A
-/// blocking `srun <cmd>` sends no other traffic for the command's duration, so
-/// without this the InactiveLimit reaper would reclaim a live allocation.
-const KEEPALIVE_INTERVAL_SECS: u64 = 30;
-
-fn install_keepalive(
-    client: SlurmControllerClient<tonic::transport::Channel>,
-    job_id: u32,
-    user: String,
-) -> tokio::task::JoinHandle<()> {
-    tokio::spawn(async move {
-        let mut client = client;
-        let mut tick =
-            tokio::time::interval(tokio::time::Duration::from_secs(KEEPALIVE_INTERVAL_SECS));
-        loop {
-            tick.tick().await;
-            let _ = client
-                .job_keepalive(JobKeepaliveRequest {
-                    job_id,
-                    user: user.clone(),
-                })
-                .await;
-        }
-    })
-}
-
 async fn wait_for_job_running(
     client: &mut SlurmControllerClient<tonic::transport::Channel>,
     job_id: u32,
@@ -879,7 +852,9 @@ async fn run_standalone_srun(
     eprintln!("srun: Pending job allocation {}...", job_id);
 
     let ctrl_c_handle = install_ctrl_c_cancel(client.clone(), job_id, user.clone());
-    let keepalive = install_keepalive(client.clone(), job_id, user.clone());
+    // Guard drops (and stops pinging) on every return path, including `?`.
+    let _keepalive =
+        crate::interactive::spawn_keepalive(client.clone(), job_id, user.clone(), "srun");
 
     let nodelist = wait_for_job_running(&mut client, job_id).await?;
     if !nodelist.is_empty() {
@@ -959,7 +934,6 @@ async fn run_standalone_srun(
     )
     .await;
 
-    keepalive.abort();
     Ok(())
 }
 
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -10,7 +10,8 @@ use spur_core::config::HooksConfig;
 use spur_proto::proto::slurm_controller_client::SlurmControllerClient;
 use spur_proto::proto::{
     CancelJobRequest, CompleteJobRequest, CreateJobStepRequest, GetJobRequest, GetNodeRequest,
-    JobSpec, JobState, RunStepRequest, StreamJobOutputRequest, SubmitJobRequest,
+    JobKeepaliveRequest, JobSpec, JobState, RunStepRequest, StreamJobOutputRequest,
+    SubmitJobRequest,
 };
 use std::collections::HashMap;
 use std::io::Write as _;
@@ -627,6 +628,32 @@ fn install_ctrl_c_cancel(
     })
 }
 
+/// How often srun pings the controller to keep its allocation attended. A
+/// blocking `srun <cmd>` sends no other traffic for the command's duration, so
+/// without this the InactiveLimit reaper would reclaim a live allocation.
+const KEEPALIVE_INTERVAL_SECS: u64 = 30;
+
+fn install_keepalive(
+    client: SlurmControllerClient<tonic::transport::Channel>,
+    job_id: u32,
+    user: String,
+) -> tokio::task::JoinHandle<()> {
+    tokio::spawn(async move {
+        let mut client = client;
+        let mut tick =
+            tokio::time::interval(tokio::time::Duration::from_secs(KEEPALIVE_INTERVAL_SECS));
+        loop {
+            tick.tick().await;
+            let _ = client
+                .job_keepalive(JobKeepaliveRequest {
+                    job_id,
+                    user: user.clone(),
+                })
+                .await;
+        }
+    })
+}
+
 async fn wait_for_job_running(
     client: &mut SlurmControllerClient<tonic::transport::Channel>,
     job_id: u32,
@@ -852,6 +879,7 @@ async fn run_standalone_srun(
     eprintln!("srun: Pending job allocation {}...", job_id);
 
     let ctrl_c_handle = install_ctrl_c_cancel(client.clone(), job_id, user.clone());
+    let keepalive = install_keepalive(client.clone(), job_id, user.clone());
 
     let nodelist = wait_for_job_running(&mut client, job_id).await?;
     if !nodelist.is_empty() {
@@ -931,6 +959,7 @@ async fn run_standalone_srun(
     )
     .await;
 
+    keepalive.abort();
     Ok(())
 }
 
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -479,6 +479,11 @@ pub struct SchedulerConfig {
     pub inactive_limit_secs: u32,
 }
 
+/// How often an interactive client (`salloc`/`srun`) pings the controller to
+/// keep its allocation attended. Shared with the CLI so `inactive_limit_secs`
+/// can be validated against it.
+pub const KEEPALIVE_INTERVAL_SECS: u64 = 30;
+
 fn default_scheduler_plugin() -> String {
     "backfill".into()
 }
@@ -1199,6 +1204,18 @@ impl SlurmConfig {
                 ),
             });
         }
+        // A limit below the client's ping interval would reap a live client
+        // between its own keepalives. Require at least two intervals of slack.
+        let inactive = self.scheduler.inactive_limit_secs;
+        if inactive > 0 && u64::from(inactive) < 2 * KEEPALIVE_INTERVAL_SECS {
+            return Err(ConfigError::InvalidValue {
+                field: "scheduler.inactive_limit_secs".into(),
+                value: format!(
+                    "{inactive} (0 disables reaping; otherwise must be >= {})",
+                    2 * KEEPALIVE_INTERVAL_SECS
+                ),
+            });
+        }
         if self.cluster.enabled {
             if self.cluster.distro != "k0s" {
                 return Err(ConfigError::InvalidValue {
@@ -1894,6 +1911,35 @@ max_batch_requeue = 7
 "#;
         let config = SlurmConfig::load_from_str(toml).unwrap();
         assert_eq!(config.controller.max_batch_requeue, 7);
+    }
+
+    #[test]
+    fn scheduler_config_rejects_inactive_limit_below_floor() {
+        let toml = r#"
+cluster_name = "test"
+
+[scheduler]
+inactive_limit_secs = 20
+"#;
+        let err = SlurmConfig::load_from_str(toml).unwrap_err();
+        assert!(
+            err.to_string().contains("inactive_limit_secs"),
+            "unexpected error: {err}"
+        );
+    }
+
+    #[test]
+    fn scheduler_config_accepts_disabled_or_ample_inactive_limit() {
+        // 0 disables reaping and is always allowed.
+        SlurmConfig::load_from_str(
+            "cluster_name = \"test\"\n[scheduler]\ninactive_limit_secs = 0\n",
+        )
+        .expect("0 (disabled) must be accepted");
+        // A value at the floor (2x the keepalive interval) is accepted.
+        SlurmConfig::load_from_str(
+            "cluster_name = \"test\"\n[scheduler]\ninactive_limit_secs = 60\n",
+        )
+        .expect("a value at the floor must be accepted");
     }
 
     #[test]
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -472,6 +472,11 @@ pub struct SchedulerConfig {
     /// Grace minutes after a reservation ends before cancelling its running jobs.
     #[serde(default)]
     pub resv_overrun_minutes: u32,
+    /// Reap an interactive allocation (salloc/srun) whose client has sent no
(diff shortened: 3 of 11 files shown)


biluriuday:
Some comments. PTAL

C1 — Enabling the knob reaps all interactive allocations, not just keepalive-capable ones. The PR description claims otherwise.

The PR states: "Reaping requires both the config knob set and keepalive-capable clients." The code does not honor the second half. interactive_reap_candidates seeds any unseen job to now and then reaps it once now - seeded > limit:


cluster.rs
Lines 1707-1718
running_ids
    .iter()
    .copied()
    .filter(|id| now - *seen.entry(*id).or_insert(now) > limit)
    .collect()
Consequence: any interactive/srun allocation that never sends a keepalive — an older CLI, an FFI/libslurm client, or a REST-created interactive allocation — gets reaped inactive_limit_secs after the leader first sees it, even if the user is actively working. Because the reaper targets j.spec.interactive || j.spec.srun_job indiscriminately, once an admin sets the knob this becomes a silent job-killer for any non-upgraded client. That's a backward-compatibility break masked by the "disabled by default" framing.

Decide and make it explicit:

Option A (safest): only reap jobs that have sent at least one keepalive (track "ever seen"). Distinguishes keepalive-capable clients from legacy/other clients. Downside: a client that dies before its first ping is never reaped — acceptable given the 30s interval and immediate first tick.
Option B: keep current behavior but correct the PR/doc text to say plainly: "Once enabled, every interactive allocation is subject to reaping regardless of client version; upgrade all clients first."
I'd push for A, and it's a small change to interactive_reap_candidates (seed-vs-reap distinction).

Important
I1 — No guard against inactive_limit_secs being set below the keepalive interval. The client pings every 30s (hardcoded KEEPALIVE_INTERVAL_SECS), the reaper runs every 10s. If an admin sets inactive_limit_secs to, say, 20, a perfectly live client is reaped between pings. The comment in salloc.rs even says "Must stay well below any configured inactive_limit_secs" — but nothing enforces it, and the client can't see the server value. Add validation at config load (reject or warn when 0 < inactive_limit_secs < 2 * KEEPALIVE_INTERVAL_SECS), so misconfiguration fails loud instead of silently killing jobs. There's no central Config::validate today; this is a good reason to introduce one, or at minimum a warn! on the controller at startup.

I2 — Behavioral divergence from the time-limit path: no SIGTERM/grace, straight to SIGKILL. enforce_time_limits sends SIGTERM, waits GRACE_PERIOD_SECS, then SIGKILL. The reaper goes straight to complete_job(Timeout) + SIGKILL (signal 9):


scheduler_loop.rs
Lines 1594-1599
if let Err(e) = cluster.complete_job(job_id, -1, spur_core::job::JobState::Timeout) {
    warn!(job_id, error = %e, "failed to reap inactive allocation");
    continue;
}
send_cancel_to_agents(&cluster, job, 9).await; // SIGKILL
For a truly abandoned allocation (no client) SIGKILL is defensible — nothing will catch SIGTERM. But the reaper also fires when the controller↔client link drops while the job's actual workload on the node is healthy (e.g. a network partition between the login node and controller, node still fine). In that case you're SIGKILLing a live workload with no chance to flush. Consider a SIGTERM→grace→SIGKILL sequence for consistency, or explicitly document why immediate SIGKILL is correct here.

Minor
M1 — Duplicated keepalive logic across salloc.rs and srun.rs. Both define an identical KEEPALIVE_INTERVAL_SECS = 30 and the same spawn-loop (salloc inline, srun as install_keepalive). Extract one shared helper (e.g. in a cli util module) so the interval and RPC shape stay in sync. Right now a future change has to touch two places.

M2 — Keepalive send errors are fully swallowed (let _ = client.job_keepalive(...)). A persistent failure (leader unreachable, auth misconfig) means the client believes it's keeping the allocation alive while the server hears nothing — leading to a surprise reap. A debug!/trace! on error would make this diagnosable without adding noise.

M3 — record_job_keepalive accepts any job the caller owns, including non-interactive ones. Harmless (the map is pruned to running interactive/srun ids each tick), but the handler could early-return for non-interactive jobs to keep the contract tight and avoid a misleading success response.

M4 — Minor consistency nit: interactive_reap_candidates does three things under one write lock (prune, seed, evaluate). It's correct and the doc comment is good, but the dual return contract (limit_secs == 0 → prune-only) is easy to misuse from a future caller. The name doesn't hint at the pruning side effect. Consider prune_and_select_reap_candidates or splitting.

C1 and I1 are important finds. Others are good to have

crates/spurctld/src/cluster.rs
```diff
@@ -1678,6 +1683,42 @@ impl ClusterManager {
...
+        let mut seen = self.interactive_last_seen.write();
+        seen.retain(|id, _| live.contains(id));
+        if limit_secs == 0 {
+            return Vec::new();
+        }
+        let limit = chrono::Duration::seconds(i64::from(limit_secs));
+        running_ids
+            .iter()
+            .copied()
+            .filter(|id| now - *seen.entry(*id).or_insert(now) > limit)
```
yansun1996, line 1718: `seen.entry(*id).or_insert(now)` seeds a job the very first time the reaper notices it, indistinguishably from a real keepalive. That means a job whose client never calls `JobKeepalive` at all gets reaped after one `limit_secs` window rather than being exempt — the opposite of requiring a keepalive-capable client. Would it make sense to only reap ids that have received at least one *real* `record_job_keepalive` call (e.g. track that as a separate confirmed-set, or `Option<DateTime<Utc>>` where `None` means "never pinged, don't reap")?

crates/spurctld/src/scheduler_loop.rs
```diff
@@ -1539,6 +1544,60 @@ async fn enforce_time_limits(cluster: Arc<ClusterManager>, raft: Arc<RaftHandle>
...
+            let Some(job) = running.iter().find(|j| j.job_id == job_id) else {
+                continue;
+            };
+
+            info!(
+                job_id,
+                limit_secs, "interactive allocation idle past InactiveLimit — reaping"
+            );
+
+            if let Err(e) = cluster.complete_job(job_id, -1, spur_core::job::JobState::Timeout) {
```
yansun1996, line 1591: If this job independently times out and gets requeued/redispatched under the same `job_id` between this function's `Running` snapshot and this call, `complete_job`'s only guard is `is_terminal()` — which a freshly-`Running` new attempt would pass. Since `send_cancel_to_agents` below uses the `job` captured from the earlier snapshot, the SIGKILL could land on the new run. Might be worth re-checking the job's current run/generation right before finalizing, or excluding `interactive`/`srun_job` specs from requeue-on-timeout.

crates/spurctld/src/scheduler_loop.rs
```diff
@@ -1539,6 +1544,60 @@ async fn enforce_time_limits(cluster: Arc<ClusterManager>, raft: Arc<RaftHandle>
...
+                warn!(job_id, error = %e, "failed to reap inactive allocation");
+                continue;
+            }
+
+            send_cancel_to_agents(&cluster, job, 9).await; // SIGKILL
```
yansun1996, line 1596: `enforce_time_limits` does SIGTERM → grace period → SIGKILL for the same "job is being force-finished" scenario; this goes straight to SIGKILL. A dropped controller↔client link doesn't imply the workload on the node is unhealthy — would a SIGTERM+grace step (or at least a comment on why immediate SIGKILL is intentional here) make sense for consistency?

crates/spurctld/src/server.rs
```diff
@@ -550,6 +550,38 @@ impl SlurmController for ControllerService {
...
+                }
+            }
+        }
+
+        let req = request.into_inner();
+        let job = self
+            .cluster
+            .get_job(req.job_id)
+            .ok_or_else(|| Status::not_found(format!("job {} not found", req.job_id)))?;
+        spur_core::auth::check_job_owner(&req.user, &job.spec.user, "send keepalive for")
```
yansun1996, line 578: `check_job_owner` treats an empty `user` as authorized. Elsewhere (attach/exec) that's a pre-existing tradeoff, but here it has a new, concrete consequence: anyone can keep resetting an arbitrary job's keepalive timer with an empty `user`, permanently preventing that allocation from ever being reaped. Since a legitimate keepalive always carries a real username, could this handler reject an empty `user` explicitly rather than relying on the shared helper's default?

crates/spur-core/src/config.rs
```diff
@@ -456,6 +456,11 @@ pub struct SchedulerConfig {
     /// Grace minutes after a reservation ends before cancelling its running jobs.
     #[serde(default)]
     pub resv_overrun_minutes: u32,
+    /// Reap an interactive allocation (salloc/srun) whose client has sent no
+    /// keepalive for this many seconds. `0` (the default) disables reaping, so
+    /// abandoned allocations behave as before. Mirrors Slurm's `InactiveLimit`.
+    #[serde(default)]
+    pub inactive_limit_secs: u32,
```
yansun1996, line 463: Client-side `KEEPALIVE_INTERVAL_SECS` is hardcoded to 30s in both `salloc.rs`/`srun.rs`. Nothing stops an admin from setting `inactive_limit_secs` below that (e.g. 20), which would reap a perfectly live client between its own pings. `validate()` already guards similar scheduler knobs — would a floor check here (e.g. reject values under ~60s) be worth adding?

crates/spur-cli/src/salloc.rs
```diff
@@ -230,6 +236,27 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
...
+    // abandoned allocation.
+    let keepalive = {
+        let mut client = client.clone();
+        let user = user.clone();
+        tokio::spawn(async move {
+            let mut tick =
+                tokio::time::interval(tokio::time::Duration::from_secs(KEEPALIVE_INTERVAL_SECS));
+            loop {
+                tick.tick().await;
+                let _ = client
```
yansun1996, line 250: This silently drops every keepalive error, including `PermissionDenied`. If it keeps failing (leader unreachable, auth issue), the user has no idea their allocation is about to be reaped. Worth at least an `eprintln!`/log on failure? (Same pattern applies in `srun.rs`'s `install_keepalive`.)

yansun1996 (commented):
Nice addition — the design (ephemeral last-seen map, leader-gated reaper, failover reseed, additive proto) fits the existing patterns well. A few things worth resolving before merge:

**Blocking:** `interactive_reap_candidates` seeds a never-before-seen job to `now` the same way it would after a real keepalive, so a job whose client never sends one (older CLI, FFI/libslurm, or an interactive job created via a raw `SubmitJobRequest`) gets reaped exactly `inactive_limit_secs` after the leader first notices it — not because it went idle. That's the opposite of "requires keepalive-capable clients," and it's a real hazard for anyone enabling this knob on a fleet that isn't 100% upgraded. See inline note.

**Worth addressing:**
- The reaper jumps straight to SIGKILL with no SIGTERM/grace, unlike `enforce_time_limits`'s SIGTERM→grace→SIGKILL. A dropped controller↔client link doesn't necessarily mean the workload on the node is unhealthy.
- There's a narrow race between this reaper's `Running` snapshot and `complete_job`: if the same `job_id` times out and gets requeued/redispatched by an unrelated path inside the same tick window, `complete_job`'s only guard (`is_terminal()`) won't catch that it's now a different run, and the SIGKILL could land on the new attempt.
- `check_job_owner` treats an empty `user` as authorized (pre-existing helper, also used by attach/exec). For this new RPC specifically, that lets anyone reset another job's keepalive timer indefinitely — quietly defeating reaping for that job. Since a legitimate keepalive always carries a real username, worth rejecting empty `user` here even if the broader helper isn't touched.
- No validation ties `inactive_limit_secs` to the (hardcoded 30s) client ping interval — setting it below ~60s would reap live clients between their own pings.
- Both CLI keepalive loops swallow every RPC error silently (`let _ = client.job_keepalive(...)`), so a persistently failing keepalive gives the user no warning before their job is reaped.

**Nits (non-blocking):**
- `running.iter().find(...)` inside the reaper's per-stale-id loop is O(n×m); a small index would avoid the rescan (low impact at expected scale).
- `KEEPALIVE_INTERVAL_SECS` + the spawn-loop are duplicated near-identically between `salloc.rs` and `srun.rs`; could share a helper.
- The doc comments on `interactive_reap_candidates` (5 lines) and `enforce_inactive_limits` (4 lines) could trim to 2-3 lines per house style.
- `interactive_reap_candidates_disabled_only_prunes`'s first assertion always passes regardless of whether pruning worked (the `limit_secs == 0` early return is unconditional); the second assertion is the one doing real work.

**Test coverage:** the new tests exercise `ClusterManager::interactive_reap_candidates` directly, but nothing calls the `job_keepalive` RPC handler itself — the leader-forward branch, the not-found branch, and the ownership-denial branch are all untested. Worth a test or two mirroring the existing `exec_in_job_denies_non_owner` pattern.

crates/spur-cli/src/salloc.rs
```diff
@@ -243,6 +270,8 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
         .await
         .context("failed to spawn shell")?;
 
+    keepalive.abort();
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed. Extracted a shared `spawn_keepalive` helper in `interactive.rs` that returns a `KeepaliveGuard` aborting the task on `Drop`, so an early `?` return can't leak it. salloc holds it as `_keepalive` for the shell's lifetime.

crates/spur-cli/src/srun.rs
```diff
@@ -627,6 +628,32 @@ fn install_ctrl_c_cancel(
...
+            tick.tick().await;
+            let _ = client
+                .job_keepalive(JobKeepaliveRequest {
+                    job_id,
+                    user: user.clone(),
+                })
+                .await;
+        }
+    })
+}
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed by the same `KeepaliveGuard`. This also closes a real leak on srun's pty and error return paths, where the old explicit `keepalive.abort()` at the end of the function was skipped.

crates/spurctld/src/scheduler_loop.rs
```diff
@@ -1539,6 +1544,60 @@ async fn enforce_time_limits(cluster: Arc<ClusterManager>, raft: Arc<RaftHandle>
...
+            .collect();
+        let ids: Vec<spur_core::job::JobId> = running.iter().map(|j| j.job_id).collect();
+
+        // Prunes the keepalive map every tick; returns candidates only when
+        // the limit is enabled.
+        let stale = cluster.interactive_reap_candidates(&ids, now, limit_secs);
+        for job_id in stale {
+            let Some(job) = running.iter().find(|j| j.job_id == job_id) else {
+                continue;
+            };
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed. The reaper now builds a `HashMap<JobId, &Job>` once per tick and looks up stale ids by key instead of scanning per id.

crates/spurctld/src/cluster.rs
shiv-tyagi: Kept intentionally, with the docs corrected. The seed-to-now is the failover grace: a newly promoted leader must not instantly reap live allocations it has never seen. Only the spur CLI sets `interactive`/`srun_job` (REST and FFI never do), so the only clients that create reapable allocations already send keepalives. The real contract is: once enabled, every interactive allocation is subject to reaping regardless of client version, so upgrade all CLI clients before enabling. That is now stated in `configuration.rst`, and a config floor (see the config.rs thread) guarantees the window can never be shorter than the ping interval.

crates/spurctld/src/scheduler_loop.rs
shiv-tyagi: Fixed. Before finalizing, the reaper re-fetches the job and skips unless it is still `Running` with the same `run_attempt` as the snapshot, then sends SIGKILL to the fresh `allocated_nodes`. Verified on the cluster: a salloc allocation that reached a terminal state within the grace window was correctly skipped by this guard rather than double-killed.

crates/spurctld/src/scheduler_loop.rs
shiv-tyagi: Fixed. The reaper now mirrors `enforce_time_limits`: SIGTERM, a 30s grace, then SIGKILL and TIMEOUT finalize. The grace state is a task-local map (ephemeral like `interactive_last_seen`), pruned each tick so a client that resumes pinging during the grace aborts the kill. Confirmed on a 3-peer Raft cluster: SIGTERM logged at T, force-kill at exactly T+30s.

crates/spurctld/src/server.rs
shiv-tyagi: Fixed. `job_keepalive` now rejects an empty `user` with `PermissionDenied` before the ownership check, so it cannot be used to hold an allocation open. Unit test added (`job_keepalive_rejects_empty_user`), alongside not-found and non-owner cases.

crates/spur-core/src/config.rs
shiv-tyagi: Fixed. `KEEPALIVE_INTERVAL_SECS` moved to `spur-core::config` and is now shared by the CLI. `validate()` rejects a non-zero `inactive_limit_secs` below `2 * KEEPALIVE_INTERVAL_SECS` (60s). Tests added; verified spurctld exits with `invalid value for scheduler.inactive_limit_secs: 20 (0 disables reaping; otherwise must be >= 60)`.

crates/spur-cli/src/salloc.rs
shiv-tyagi: Fixed in the shared helper. Keepalive send failures now print a warning, once per failure streak and reset on the next success, so a persistent problem (leader unreachable, auth) is visible to the user without spamming a line every tick.

shiv-tyagi:
Pushed a follow-up commit addressing all review threads. Summary:

**Substantive fixes**
- Config floor: `KEEPALIVE_INTERVAL_SECS` moved to `spur-core::config` and shared with the CLI; `validate()` rejects a non-zero `inactive_limit_secs` below 60s (2x the ping interval).
- Grace sequence: the reaper now does SIGTERM, 30s grace, then SIGKILL + TIMEOUT, mirroring `enforce_time_limits`. Grace state is task-local and pruned each tick, so a client that resumes pinging during the grace aborts the kill.
- Requeue race: re-fetch the job and verify it is still `Running` with the same `run_attempt` before finalizing; SIGKILL targets the fresh `allocated_nodes`.
- Empty-user keepalive rejected with `PermissionDenied` before the ownership check.
- Keepalive send failures now warn (once per failure streak) instead of being swallowed.
- Seed-to-now kept by design (failover grace); docs corrected to state the real contract: once enabled, every interactive allocation is subject to reaping regardless of client version, so upgrade all CLI clients before enabling.

**Nits**
- Reaper builds a `HashMap<JobId, &Job>` index once per tick.
- Keepalive handler skips recording for non-interactive jobs.
- Shared `spawn_keepalive` helper with an RAII `KeepaliveGuard` (aborts on drop) used by both salloc and srun. This also fixed a latent task leak on srun's pty/error paths.

**Tests:** added keepalive handler tests (not-found, non-owner, empty-user) and config-floor tests; tightened the disabled-only-prunes assertion. `cargo clippy` and `cargo test --locked` are clean.

**Cluster validation (3-peer bare-metal Raft):**
- Regular batch jobs and steps unaffected; the reaper never touches non-interactive jobs.
- Default (`inactive_limit_secs=0`): abandoned allocation stays running, no reaping. Backward compatible.
- Enabled: abandoned srun reaped via SIGTERM -> +30s -> SIGKILL -> TIMEOUT; abandoned salloc cleaned up by SIGTERM within the grace (the re-fetch guard correctly skipped the redundant force-kill).
- Config floor: spurctld exits with a clear error on `inactive_limit_secs=20`.
- HA failover: a live salloc session survived a leader kill without being reaped (keepalives followed the new leader via the multi-address channel), and the newly promoted leader did not instantly reap (seed-to-now grace held for a full window).

**Known follow-up (pre-existing, not from this PR):** srun clients do not survive a controller failover today because the step-dispatch/output stream is pinned to the connected controller and the client exits on transport error. When that happens the allocation is genuinely orphaned and the reaper then cleans it up correctly. Making srun reconnect its stream across failover is worth a separate issue.

Rebased onto the latest `main`.

crates/spurctld/src/server.rs
```diff
@@ -603,6 +603,51 @@ impl SlurmController for ControllerService {
         Ok(Response::new(()))
     }
 
+    async fn job_keepalive(
```
yansun1996, line 606: The rejection tests here (unknown job, non-owner, empty user) are solid, but is there a test that a keepalive on an interactive job succeeds and actually updates its last-seen state through this RPC? Right now only the underlying cluster method is unit-tested directly.

crates/spurctld/src/scheduler_loop.rs
```diff
@@ -1711,6 +1716,99 @@ async fn enforce_time_limits(cluster: Arc<ClusterManager>, raft: Arc<RaftHandle>
...
 }
 
+/// Reap interactive allocations (salloc/srun) whose client stopped sending
+/// keepalives, mirroring Slurm's `InactiveLimit`. Idle allocations get a
+/// SIGTERM -> grace -> SIGKILL sequence (like `enforce_time_limits`) and are
+/// finalized via the TIMEOUT path. Disabled when `inactive_limit_secs == 0`.
+async fn enforce_inactive_limits(cluster: Arc<ClusterManager>, raft: Arc<RaftHandle>) {
+    use spur_core::job::{JobId, JobState};
+
+    const GRACE_PERIOD_SECS: i64 = 30;
```
yansun1996: `GRACE_PERIOD_SECS = 30` is now defined in two places (here and in `enforce_time_limits` above). Worth hoisting to a shared constant so the two grace windows can't drift apart.

yansun1996 (commented):
Re-reviewed after the latest fixes — all 6 prior points are genuinely addressed. Two things worth a look, without a good inline anchor since they're in code this PR doesn't touch:

- `complete_job` returns `Ok(())` even when its WAL proposal was a no-op because the job was already finalized by a concurrent proposal (the dedup in `apply_operation` returns an empty `jobs_finalized` for that case, but callers never check it). Now that a job can be reaped by either `enforce_time_limits` or this new `enforce_inactive_limits` in the same window, the loser still sees `Ok` and proceeds to SIGKILL nodes the winner already tore down. Could `complete_job` signal back whether it actually finalized, so a losing caller can skip the kill?
- `enforce_time_limits` has the same requeue-race the new reaper fixes below — it calls `complete_job` then SIGKILLs using the loop's original job snapshot, with no re-fetch or `run_attempt` check. Since the fix pattern already exists a few lines away, might be worth applying it there too.

Two smaller things left as inline comments.

crates/spurctld/src/scheduler_loop.rs
shiv-tyagi: Fixed. Hoisted `GRACE_PERIOD_SECS` to a single module-level constant shared by `enforce_time_limits` and `enforce_inactive_limits`, so the two grace windows can't drift.

crates/spurctld/src/server.rs
shiv-tyagi: Added. `job_keepalive_records_interactive` drives the RPC on an interactive job and asserts last-seen is recorded; `job_keepalive_skips_non_interactive` asserts a non-interactive job returns `Ok` without being tracked. Both go through the real handler, observed via a small `cfg(test)` accessor over the last-seen map.

crates/spurctld/src/scheduler_loop.rs
```diff
@@ -1711,6 +1718,98 @@ async fn enforce_time_limits(cluster: Arc<ClusterManager>, raft: Arc<RaftHandle>
...
+
+    // Jobs that were sent SIGTERM and are in their grace window, keyed by
+    // job id -> when SIGTERM was sent. Local to this task (only the reaper
+    // touches it); lost on restart, which merely restarts the grace.
+    let mut signaled: HashMap<JobId, DateTime<Utc>> = HashMap::new();
+
+    loop {
+        interval.tick().await;
+
+        if !raft.is_leader() {
```
yansun1996, line 1742: Neither `signaled` (this function's local grace-tracking map) nor `ClusterManager::interactive_last_seen` gets cleared when a node's Raft role changes — both are scoped to the process/`ClusterManager` lifetime, not to a leadership term, and there's no leadership-change hook anywhere that resets them.

That's fine for a node's *first-ever* promotion (both maps start empty), but not for a node regaining leadership after a prior stint: while it's a follower, `interactive_last_seen` isn't touched at all (only the current leader records keepalives), so its copy of a job's last-seen time stays frozen at whatever it was when it lost leadership. `interactive_reap_candidates`'s `or_insert(now)` only seeds a *missing* key, so it won't correct an entry that's already present but stale.

Concretely: node A leads, has fresh keepalive data for job X → leadership moves to B → X keeps pinging, now landing on B → leadership moves back to A → A's `interactive_last_seen[X]` is still frozen from before, so X looks idle past `limit_secs` on A's very first tick as leader again, even though it's been alive the whole time. And if `signaled` already held an entry for X from before A lost leadership, A skips straight to the grace-expired branch — no fresh SIGTERM is even observed, and X gets SIGKILLed immediately.

Since leader flapping is routine here (Raft is always-on), would it make sense to track `is_leader()` transitions in this loop and reseed/clear both maps whenever a node goes from follower back to leader?

shiv-tyagi: Good catch, fixed. Confirmed the failure: on a follower to leader transition both ephemeral maps (`signaled` here and `ClusterManager::interactive_last_seen`) carry stale entries from a prior stint, and `interactive_reap_candidates`'s `or_insert(now)` only seeds missing keys, so a still-present frozen entry makes a live allocation look idle on the first tick as leader again (and a leftover grace timer skips straight to SIGKILL).

The reaper loop now tracks the follower->leader transition and, on regain, clears `signaled` and calls a new `reset_interactive_last_seen()`. The reap check then reseeds every running allocation to `now`, restoring a full grace window under the new term. Added `reset_interactive_last_seen_reseeds_on_leadership_regain`, which asserts a carried-over stale entry would reap a live allocation and that the reset exempts it.

My earlier cluster testing missed this because I always restarted the controller process on failover (fresh empty maps); a pure Raft role flap on a surviving process is the trigger.

biluriuday (approved):
LGTM

yansun1996 (approved):
LGTM

PR #595, opened 2026-08-07
feat(spur-cli)!: restrict reservation mgmt to root and sudo/wheel users

biluriuday:
Reservation create/update/delete via scontrol previously succeeded for any user. Gate these commands behind a local privilege check: allow only effective root (euid 0) or members of the 'sudo'/'wheel' group, denying everyone else before any RPC is sent. Covers both the explicit subcommands and the Slurm-inline "create/delete ReservationName=" paths; "show reservation" stays open. The check fails closed if local identity cannot be resolved.

This is a client-side guardrail, not a security boundary: the controller does not authenticate callers, so server-side enforcement behind an authenticated identity remains the RBAC follow-up.

BREAKING CHANGE: non-privileged users can no longer manage reservations via the CLI; scripts relying on the old permissive behavior will fail.

Tested with cargo clippy (workspace, no warnings), cargo test (all pass, including new privilege unit tests), and new e2e coverage in test_reservations.py exercising the root-allowed / nobody-denied paths.

### DIFF
--- a/crates/spur-cli/src/privilege.rs
+++ b/crates/spur-cli/src/privilege.rs
@@ -1,0 +1,96 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Local privilege gate for administrative CLI actions.
+//!
+//! This is a client-side guardrail, not a security boundary: the controller
+//! does not authenticate callers, so it only stops ordinary users from running
+//! privileged commands through the CLI. Real enforcement belongs server-side
+//! behind an authenticated identity (future RBAC work).
+
+use anyhow::{bail, Context, Result};
+
+/// Unix groups whose members are treated as privileged, covering the
+/// Debian/Ubuntu (`sudo`) and RHEL/CentOS (`wheel`) conventions.
+const PRIVILEGED_GROUPS: &[&str] = &["sudo", "wheel"];
+
+/// Pure decision core, split from the syscalls so it is unit-testable.
+fn is_privileged(euid: u32, caller_groups: &[String], privileged_groups: &[&str]) -> bool {
+    euid == 0
+        || caller_groups
+            .iter()
+            .any(|g| privileged_groups.contains(&g.as_str()))
+}
+
+/// Resolve the invoking process's group names (supplementary groups plus the
+/// effective gid). Returns `Err` on any lookup failure so callers fail closed.
+fn caller_group_names() -> Result<Vec<String>> {
+    use nix::unistd::{getegid, getgroups, Group};
+
+    let mut gids = getgroups().context("failed to read process groups")?;
+    gids.push(getegid());
+
+    let mut names = Vec::new();
+    for gid in gids {
+        // A gid without a matching group entry cannot be privileged; skip it.
+        if let Some(group) = Group::from_gid(gid).context("failed to resolve group name")? {
+            names.push(group.name);
+        }
+    }
+    Ok(names)
+}
+
+/// Gate a privileged CLI action. Fails closed: any identity/group lookup error
+/// denies the action rather than allowing it.
+pub fn require_privileged(action: &str) -> Result<()> {
+    let euid = nix::unistd::geteuid().as_raw();
+    // Root is always privileged; short-circuit so a group-lookup failure can
+    // never deny root.
+    if euid == 0 {
+        return Ok(());
+    }
+    let groups = caller_group_names()?;
+    if is_privileged(euid, &groups, PRIVILEGED_GROUPS) {
+        return Ok(());
+    }
+    bail!(
+        "insufficient privileges to {action}: requires root or membership in the 'sudo' or 'wheel' group"
+    );
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    const PRIV: &[&str] = &["sudo", "wheel"];
+
+    #[test]
+    fn root_is_privileged_regardless_of_groups() {
+        assert!(is_privileged(0, &[], PRIV));
+    }
+
+    #[test]
+    fn sudo_member_is_privileged() {
+        assert!(is_privileged(1000, &["users".into(), "sudo".into()], PRIV));
+    }
+
+    #[test]
+    fn wheel_member_is_privileged() {
+        assert!(is_privileged(1000, &["wheel".into()], PRIV));
+    }
+
+    #[test]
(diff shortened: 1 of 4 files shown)


crates/spur-cli/src/privilege.rs
```diff
@@ -0,0 +1,96 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Local privilege gate for administrative CLI actions.
+//!
+//! This is a client-side guardrail, not a security boundary: the controller
+//! does not authenticate callers, so it only stops ordinary users from running
+//! privileged commands through the CLI. Real enforcement belongs server-side
+//! behind an authenticated identity (future RBAC work).
```
yansun1996, line 9: This module doc comment runs 6 lines (4-9). Could trim to 1-2 lines, e.g. "Client-side privilege gate for reservation commands; not a security boundary since the controller doesn't authenticate callers (real enforcement is future server-side RBAC)." Might also be worth being explicit that this checks *group membership*, not that the caller actually elevated via `sudo` — a user merely listed in the `sudo`/`wheel` group passes even if they never ran `sudo`.

tests/native_host/e2e/test_reservations.py
```diff
@@ -216,19 +247,95 @@ def test_create_rejects_busy_node_without_ignore_jobs(self, cluster):
...
+            ],
         )
         msg = out.lower()
         assert "busy" in msg or "until after reservation start" in msg, f"unexpected: {out}"
 
+    def test_reservation_management_requires_privileged_user(self, cluster):
+        """Reservation create/update/delete via the CLI is restricted to root or
+        members of the 'sudo'/'wheel' group. An unprivileged caller is denied
+        locally before any RPC is sent. Uses the always-present, deterministically
+        unprivileged 'nobody' account so no user provisioning is required."""
```
yansun1996, line 268: Nit: this docstring is 4 lines (265-268) — could be trimmed to 1-2 lines to match the rest of the file's style.

tests/native_host/e2e/test_reservations.py
```diff
@@ -216,19 +247,95 @@ def test_create_rejects_busy_node_without_ignore_jobs(self, cluster):
...
+        # Probe the exact identity under test: `show` is unguarded, so running
+        # it as `nobody` verifies both that `sudo -u nobody` is permitted and
+        # that `nobody` can exec the binary. Skip (not fail) if it cannot.
```
yansun1996, line 271: Nit: 3-line comment block above the probe call — could be tightened.

Separately: `probe`/`low` here are only checked for the skip condition; nothing asserts the probe actually succeeded (i.e. that `show reservation` is *not* gated for an unprivileged caller). Worth adding a positive assertion so a future accidental gating of the read path would actually fail this test instead of just changing which branch it takes.

tests/native_host/e2e/test_reservations.py
```diff
@@ -245,6 +352,16 @@ def test_non_owner_cannot_delete_or_update_reservation(self, cluster):
         ):
             pytest.skip(f"sudo -u unavailable in this environment: {probe.strip()}")
 
+        # The non-owner must be privileged enough to clear the client-side
+        # reservation gate (root or sudo/wheel), or its delete/update is denied
+        # locally before the server ownership check this test targets.
```
yansun1996, line 357: Nit: 3-line comment block — could be tightened.

More substantively: this skips the test whenever `submit_user` isn't in the `sudo`/`wheel` unix *group*. Since CI runs as a fixed `--ssh-user ci` (per `.github/workflows/e2e.yml`), and cloud/VM base images commonly grant a default user passwordless sudo via a per-user `sudoers.d` entry rather than group membership, it seems plausible this pre-existing ownership-authz test (SPUR-69 coverage) could end up silently skipping in CI from now on. Could you confirm whether the `ci` user is actually in `sudo`/`wheel` on the CI image? If not, might be worth having the test add itself to the group for the duration (with cleanup) rather than depending on incidental provisioning.

yansun1996 (commented):
Reviewed the privilege gate itself and it looks correct: `require_privileged()` fails closed on every error path (a `getgroups()`/group-lookup failure denies rather than allows), and it's called before any RPC at all four call sites (the three dedicated subcommands plus the Slurm-inline `create`/`delete ReservationName=` paths, which funnel through the same guarded helpers). The "not a security boundary" framing also matches reality — the controller doesn't authenticate the `user` field on any reservation RPC, so the caveat isn't overstated.

A handful of smaller things worth a look before merge:

- A few comment blocks run longer than the rest of the codebase's style (see inline notes) — could be tightened to a line or two.
- `test_non_owner_cannot_delete_or_update_reservation` now skips unless the SSH test user is in the `sudo`/`wheel` *group*. Worth double-checking against the actual CI image: cloud base images often grant a default user passwordless sudo via a `sudoers.d` entry rather than group membership, which would mean this (pre-existing) ownership-authz test quietly skips in CI going forward.
- The unprivileged `show reservation` probe in the new test only gets used to decide skip/no-skip — nothing asserts it actually succeeded, so it wouldn't catch a future accidental gating of the read path.
- No e2e coverage for the *denial* path through the Slurm-inline `ReservationName=` create/delete syntax — it's correctly gated in code, just untested end-to-end.
- Worth a one-line note somewhere (PR description or module doc) that since reservation creation now requires privilege, a non-privileged user can no longer become an owner of a *new* reservation via the CLI going forward — the existing server-side ownership check (`owner == user`, no privilege requirement) still allows non-privileged owners in principle, just not reachable via CLI creation anymore. Not a bug, just worth documenting as a known limitation.

None of these are blockers — nice, tightly-scoped change overall.

crates/spur-cli/src/privilege.rs
biluriuday: done

tests/native_host/e2e/test_reservations.py
biluriuday: done

tests/native_host/e2e/test_reservations.py
biluriuday: done

tests/native_host/e2e/test_reservations.py
biluriuday: done

yansun1996 (approved):
LGTM, discussed offline that this would be a temporary change to block users from inappropriate holding too many reservations. In the long term we'll still support each user to manage their own reservations under the RBAC and quota limit.

PR #598, opened 2026-08-08
fix(spurctld): return unavailable when accounting startup fails

### LINKED ISSUE #417: Accounting silently disables on PostgreSQL connection failure
When `[accounting] database_url` is configured but the database is unreachable, `spurctld` logs a message and continues without the accounting gRPC service registered. Users then get cryptic `UNIMPLEMENTED` gRPC errors from `sacct`, `sacctmgr`, and `sbatch --qos=...` with no indication that accounting failed to initialize.

**Current behavior:**
- Controller starts, fails to connect to PostgreSQL (5s timeout), silently falls back to `None`
- Any accounting RPC returns: `code: 'Operation is not implemented or not supported'`
- `sbatch --qos=X` returns: `QOS 'X' does not exist` (misleading — QoS data simply isn't loaded)

**Expected behavior:**
- The gRPC error should clearly state: "accounting service is not available (database connection failed at startup)"
- Or: register a stub accounting service that returns a descriptive `Status::unavailable(...)` instead of the default tonic `UNIMPLEMENTED`
- Controller startup logs should be more prominent (WARN/ERROR level) when accounting is configured but fails to connect

joshkmartinez:
Closes #417.

When accounting is configured but PostgreSQL fails to connect or run migrations, `spurctld` currently starts without registering the accounting RPCs. That leaves `sacct` and `sacctmgr` with tonic's generic `UNIMPLEMENTED` response, which is indistinguishable from accounting being intentionally disabled.

This keeps the accounting service registered after a startup failure and returns `UNAVAILABLE` with a short, sanitized reason from every accounting RPC. If no `database_url` is configured, accounting remains disabled as before. Scheduling stays available in either case.

Database recovery is still startup-scoped: after PostgreSQL comes back, `spurctld` must be restarted

## Testing

- Covered all 14 accounting RPCs and the loopback gRPC path
- `cargo test -p spurctld --locked` — 619 passed, 20 PostgreSQL-backed tests ignored
- `cargo clippy -p spurctld --all-targets --locked -- -D warnings`
- Ran `spurctld` locally against an unreachable database and confirmed accounting commands return `UNAVAILABLE` while scheduling and job submission continue to work
- Confirmed that omitting `database_url` preserves the existing `UNIMPLEMENTED` behavior
- Ran formatting, SPDX, and the cross-platform workspace checks available on macOS

### DIFF
--- a/crates/spurctld/src/accounting/grpc.rs
+++ b/crates/spurctld/src/accounting/grpc.rs
@@ -55,19 +55,35 @@ fn fairshare_to_i32(v: f64) -> Result<i32, Status> {
     Ok(v as i32)
 }
 
-pub(crate) struct AccountingService {
-    pool: PgPool,
+pub(crate) enum AccountingService {
+    Available(PgPool),
+    Unavailable { reason: &'static str },
 }
 
 impl AccountingService {
-    pub(super) fn new(pool: PgPool) -> Self {
-        Self { pool }
+    pub(crate) fn available(pool: PgPool) -> Self {
+        Self::Available(pool)
+    }
+
+    pub(crate) fn unavailable(reason: &'static str) -> Self {
+        Self::Unavailable { reason }
+    }
+
+    fn pool(&self) -> Result<&PgPool, Status> {
+        match self {
+            Self::Available(pool) => Ok(pool),
+            Self::Unavailable { reason } => Err(Status::unavailable(format!(
+                "accounting service is not available ({reason})"
+            ))),
+        }
     }
 }
 
 /// Build a ready-to-register tonic service for embedding in another server.
-pub fn accounting_server(pool: PgPool) -> SlurmAccountingServer<AccountingService> {
-    spur_proto::accounting_server(AccountingService::new(pool))
+pub(crate) fn accounting_server(
+    service: AccountingService,
+) -> SlurmAccountingServer<AccountingService> {
+    spur_proto::accounting_server(service)
 }
 
 #[tonic::async_trait]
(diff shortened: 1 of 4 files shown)


yansun1996 (approved):
Nice fix — this closes a real gap where a broken accounting DB was indistinguishable from accounting being intentionally disabled. The startup-failure handling is well scoped, and the new tests exercise the real RPC path (including a genuine loopback gRPC call), not just internals.

A few optional nits, none blocking:

- `main.rs` — the comment "Accounting stays best-effort..." carries over wording from the old silently-skip-registration behavior. Since the new behavior always registers the service and returns an explicit `Unavailable` status, might be worth a small reword so it doesn't read as "may still not register."
- `pool()` in `grpc.rs` — this covers startup-time failures. If Postgres drops mid-runtime after a successful start, in-flight RPCs will still surface `Status::internal` rather than `Status::unavailable`, since that path isn't touched here. Given the PR description already scopes recovery to "restart required," a one-line doc comment on `pool()` noting this is startup-only would make that clear from the code itself.
- `unavailable_service_reports_migration_failure` looks like it exercises the same path as the RPC-sweep test above it (just a different reason string on `list_accounts`) — could probably be dropped or folded in without losing coverage.
- The actual `main.rs` wiring (a real connect/migrate failure producing these `AccountingService::unavailable(...)` calls) isn't covered by a test — only the enum's own behavior with an arbitrary string is. Understandably hard to unit test without a real Postgres given the repo's test rules; flagging in case a lightweight e2e case is easy to add later, otherwise fine as-is.
- The two reason strings are hardcoded in both `main.rs` and the test module — small optional cleanup to hoist them into shared consts, though the tests already assert exact messages so drift would be caught immediately either way.

None of these block merging — approving as is.

PR #599, opened 2026-08-08
fix(srun): honor step node selection

### LINKED ISSUE #597: srun step mode: -N is a task count overridden by inherited SPUR_NTASKS, and -w is silently discarded
## Summary

Inside an existing allocation, `spur run -N 1` runs on **every** allocated node. Two separate defects combine:

1. `-N` in step mode is not a node count — it only seeds a *task* count, and `-n` takes precedence over it. `spur alloc` exports `SPUR_NTASKS`, so an inherited `-n` silently overrides the `-N` typed on the command line.
2. `-w`/`--nodelist` is discarded outright on the step path — the field is hardcoded empty before the request is sent.

Both fail silently: no warning, no error, and `srun: dispatched to node …` reports the fan-out as if it were intended.

## Reproduction

`spur 0.7.0 (4fb4a38f)`, bare-metal, 2-node allocation on `crsuse2-m2m-[032,060]`:

```bash
spur alloc -p <partition> -A <account> -N 2 -G 16 -t 1:00:00
# Nodes crsuse2-m2m-[032,060] are ready for job 49846

spur run -N 1 hostname
# srun: dispatched to node crsuse2-m2m-060,crsuse2-m2m-032
# crsuse2-m2m-060
# crsuse2-m2m-032          <-- asked for 1 node, got 2
```

| command (inside the allocation) | nodes used |
|---|---|
| `spur run -N 1 hostname` | **both** (060, 032) |
| `spur run -n 1 hostname` | one (060) |
| `spur run -N 1 hostname`, with `SPUR_NTASKS`/`SLURM_NTASKS` unset | one (060) |
| `spur run -N 3 hostname` (2-node allocation) | 2 — silently clamped, no warning |
| `spur run -n 1 -w crsuse2-m2m-032 hostname` | **060** — `-w` ignored, wrong node |

Real Slurm treats `-N` as a node count in step mode and honours `-w`.

## Root cause 1 — `-N` is a task count, and `-n` wins

`-N` feeds `effective_ntasks`, which gives `ntasks` precedence and only falls back to the node count:

```rust
// crates/spur-cli/src/sbatch.rs:799-812
pub(crate) fn effective_ntasks(ntasks: Option<u32>, ntasks_per_node: Option<u32>, nodes: u32) -> u32 {
    if let Some(n) = ntasks {
        return n;
    }
    ...
    None => nodes,
}
```

`spur alloc` exports the allocation's task count into the interactive shell:

```rust
// crates/spur-cli/src/salloc.rs:228-229
env.set_with_slurm_twin("SPUR_NTASKS", job_info.num_tasks);
env.set_with_slurm_twin("SPUR_NPROCS", job_info.num_tasks);
```

and `srun` resolves `ntasks` from those variables (`crates/spur-cli/src/srun.rs:341-345`). So in a 2-node allocation the shell carries `SPUR_NTASKS=2`; a subsequent `-N 1` loses to it and the step gets 2 tasks.

The env-default helpers implement CLI-over-env precedence correctly *per flag* (`was_cli_set`, `env_defaults.rs:60-63`) — the problem is cross-flag: an explicitly typed `-N` is beaten by an *inherited* `-n`. Real Slurm's option verifier resolves `--nodes` against `--ntasks` at this point (see #570, which added `effective_num_nodes` on the submit path); step dispatch has no equivalent.

The controller then ignores the requested count entirely and spreads tasks over all allocated nodes:

```rust
// crates/spurctld/src/server.rs:1964-1965
let num_nodes = job.allocated_nodes.len() as u32;
let plan = build_step_task_plan(step.num_tasks, num_nodes, step.distribution);
```

`build_step_task_plan` emits one entry per node that received a task, so 2 tasks over 2 nodes = both nodes. This also explains the silent clamp at `-N 3`.

## Root cause 2 — `-w` is dropped on the step path

`CreateJobStepRequest.node` is documented as the `-w` target ("empty = first allocated node", `proto/slurm.proto:1272`), and `create_job_step` honours it via `select_step_node`. But `dispatch_step` never populates it:

```rust
// crates/spur-cli/src/srun.rs:719-729
.create_job_step(CreateJobStepRequest {
    job_id,
    command: args.command.clone(),
    num_tasks: ntasks,
    cpus_per_task: args.cpus_per_task,
    overlap: false,
    pty: false,
    winsize: None,
    node: String::new(),          // <-- args.nodelist never read
    user: crate::interactive::current_user(),
})
```

`args.nodelist` is parsed and validated but never reaches the request, so `-w` is inert in step mode. It does work on the `--jobid --overlap` path, which sets the field (verified: `-w crsuse2-m2m-032` lands on 032, and an unallocated name is correctly rejected with `node ... is not allocated to this job`).

## Expected

- `-N <n>` in step mode limits the step to `n` of the allocated nodes.
- An explicitly typed `-N` should not be silently overridden by an inherited `SPUR_NTASKS`/`SLURM_NTASKS`.
- `-w` should either target the named node(s) or be rejected as unsupported — not silently ignored.
- Requesting more nodes than allocated should warn rather than silently clamp (#570 added exactly this kind of warning at submit time).

One constraint worth noting for whoever picks this up: `CreateJobStepRequest` carries no node count today (`proto/slurm.proto:1264-1274`), so the controller has no requested-node figure to honour even in principle.

## Workaround

Use `-n` rather than `-N` inside an allocation (`spur run -n 1 …` gets one node), and `spur run --jobid <J> --overlap -w <node> …` to target a specific node. Note that `--overlap` currently loses supplementary groups, so it cannot be used for GPU work — filed separately.

## Related

- #460 — the opposite symptom (steps did not fan out at all); its fix is why they fan out now
- #553 / #535 — per-node vs per-job execution of the *batch script*, a different axis
- #570 — added `effective_num_nodes` and a shrink warning at submit; step dispatch has no equivalent
- #377 — made `-w` additive at allocation time, not step time

joshkmartinez:
Closes #597.

Inside an allocation, `srun -N 1` still fanned out across every allocated node, and `-w` was dropped before step creation.

This keeps step node count separate from task count through the CLI and controller. The controller validates the requested count and hostlist against the parent allocation, stores the selected nodes on the step, and dispatches only to that subset. Slurm-style node-count warnings are returned to `srun` instead of silently clamping.

Agents now receive both the allocation and step nodelists. `SLURM_JOB_*` stays allocation-scoped, while `SLURM_STEP_*` and `SLURM_NNODES` describe the step. empty step nodelist falls back to the allocation for compatibility with older controllers.

Tested locally:

- `cargo test -p spur-cli --locked`
- `cargo test -p spurctld --locked`, including a live two-agent loopback dispatch test
- `cargo test -p spur-core spur_env::tests --locked`
- `cargo clippy -p spur-core -p spur-cli -p spurctld --all-targets --locked -- -D warnings`
- Ruff, Python compilation, formatting, license headers, and diff checks

I could not run two node native-host E2E coverage b/c i'm on macos

### DIFF
--- a/crates/spur-cli/src/mock_controller.rs
+++ b/crates/spur-cli/src/mock_controller.rs
@@ -31,6 +31,8 @@ pub(crate) const MOCK_EXIT_CODE: i32 = 7;
 #[derive(Clone, Default)]
 pub(crate) struct StepCapture {
     create_step_num_tasks: Arc<AtomicU32>,
+    create_step_num_nodes: Arc<Mutex<Option<u32>>>,
+    create_step_node: Arc<Mutex<String>>,
     run_step_step_id: Arc<AtomicU32>,
     run_step_calls: Arc<AtomicU32>,
     update_node_names: Arc<Mutex<Vec<String>>>,
@@ -42,6 +44,14 @@ impl StepCapture {
     /// Task count carried by the most recent `CreateJobStep`.
     pub(crate) fn create_step_num_tasks(&self) -> u32 {
         self.create_step_num_tasks.load(Ordering::SeqCst)
+    }
+
+    pub(crate) fn create_step_num_nodes(&self) -> Option<u32> {
+        *self.create_step_num_nodes.lock().unwrap()
+    }
+
+    pub(crate) fn create_step_node(&self) -> String {
+        self.create_step_node.lock().unwrap().clone()
     }
 
     /// Step id carried by the most recent `RunStep`.
@@ -97,12 +107,16 @@ mock_controller_impl! {
             &self,
             request: tonic::Request<proto::CreateJobStepRequest>,
         ) -> Result<tonic::Response<proto::CreateJobStepResponse>, tonic::Status> {
+            let request = request.into_inner();
             self.capture
                 .create_step_num_tasks
-                .store(request.into_inner().num_tasks, Ordering::SeqCst);
+                .store(request.num_tasks, Ordering::SeqCst);
+            *self.capture.create_step_num_nodes.lock().unwrap() = request.num_nodes;
+            *self.capture.create_step_node.lock().unwrap() = request.node;
             Ok(tonic::Response::new(proto::CreateJobStepResponse {
                 step_id: MOCK_STEP_ID,
                 node_addr: String::new(),
+                warnings: Vec::new(),
             }))
         }
 
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -522,8 +522,7 @@ struct StepLayout {
 }
 
 fn resolve_step_layout(matches: &ArgMatches, args: &SrunArgs) -> StepLayout {
-    let node_count_requested = was_cli_set(matches, "nodes")
-        || crate::env_defaults::env_first(NODE_COUNT_ENV_NAMES).is_some();
+    let node_count_requested = was_cli_set(matches, "nodes");
     StepLayout {
         num_tasks: crate::sbatch::effective_ntasks(args.ntasks, args.ntasks_per_node, args.nodes),
         num_nodes: node_count_requested.then_some(args.nodes),
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -15,6 +15,15 @@ use spur_proto::proto::{
 use std::collections::HashMap;
 use std::io::Write as _;
 
+const NODE_COUNT_ENV_NAMES: &[&str] = &[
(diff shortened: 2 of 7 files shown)


yansun1996:
Hi @sgopinath1 , this PR is trying to address the issue https://github.com/ROCm/spur/issues/597 which was self-assigned to you, can you check whether this is overlapping with what you are working on ?

sgopinath1:
> Hi @sgopinath1 , this PR is trying to address the issue #597 which was self-assigned to you, can you check whether this is overlapping with what you are working on ?

Alright. @joshkmartinez please close #597 once this PR is merged.

PR #609, opened 2026-08-11
fix(spur-cli): reject unresolved k8s callers

### LINKED ISSUE #603: fix(spur-cli): reject unresolved k8s callers
`crates/spur-cli/src/k8s.rs` still falls back to an `\"unknown\"` caller when username resolution fails.

This was intentionally left out of #569 because it is outside #567's job-owner paths and the server rejects `\"unknown\"` via `is_k0s_admin`, so it is not a privilege bypass.

Follow up by reusing the fail-closed username resolver introduced in #569.

Context: https://github.com/ROCm/spur/pull/569#discussion_r3724294886

joshkmartinez:
Closes #603.

`spur k8s up`, `down`, and `kubeconfig` used `unknown` when the local username could not be resolved. The controller rejects that caller, but the CLI still sent an ambiguous identity.

These commands now reuse the existing fail-closed username resolver and return its error before dispatching an RPC. `status` and the local `install-k0s` command do not perform a username lookup.

## Testing

- 14 focused k8s command tests, including controller-backed caller and no-dispatch checks
- full `spur-cli` test suite: 360 unit tests and 1 integration test
- `cargo clippy -p spur-cli --all-targets --locked -- -D warnings`
- formatting, SPDX, and diff checks

The `spur-cli` checks required a temporary macOS-only compile shim for the pre-existing `nix::unistd::getgroups` failure on Apple targets. The shim was reverted and is not part of this PR. The repository-wide run then reached the existing Linux-only `spur-mpi-pmix` linker boundary.

### DIFF
--- a/crates/spur-cli/src/k8s.rs
+++ b/crates/spur-cli/src/k8s.rs
@@ -119,6 +119,13 @@ pub async fn main() -> Result<()> {
 }
 
 pub async fn main_with_args(args: Vec<String>) -> Result<()> {
+    main_with_args_and_user_resolver(args, crate::interactive::current_user).await
+}
+
+async fn main_with_args_and_user_resolver(
+    args: Vec<String>,
+    current_user: impl Fn() -> Result<String>,
+) -> Result<()> {
     let parsed = K8sArgs::try_parse_from(args)?;
     let controller = parsed.controller;
     match parsed.command {
@@ -130,8 +137,10 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
             partition,
             selector,
         } => {
+            let caller = current_user()?;
             cmd_up(
                 &controller,
+                caller,
                 control_plane_node,
                 replicas,
                 control_plane_nodes,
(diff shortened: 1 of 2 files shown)


crates/spur-cli/src/mock_controller.rs
```diff
@@ -7,9 +7,9 @@
 //! Follows the same shape as the `MockAgent` harness in `spurctld`: bind an
 //! ephemeral localhost port, serve a hand-written service on it, and hand the
 //! caller back the address plus a shared record of what the server observed.
-//! Only a handful of RPCs are implemented (`CreateJobStep`, `RunStep`,
-//! `UpdateNode`); every other RPC reports `unimplemented` so an unexpected call
-//! fails loudly instead of silently returning a default.
+//! Only the RPCs exercised by CLI tests are implemented; every other RPC
+//! reports `unimplemented` so an unexpected call fails loudly instead of
```
yansun1996: Small nit: this doc comment is now 3 lines, one over the usual 2-line comment-block convention here. Mind trimming it, e.g.:

```
//! Only the RPCs exercised by CLI tests are implemented; every other RPC
//! reports `unimplemented` so an unexpected call fails loudly instead of silently returning a default.
```

yansun1996 (commented):
Nice fix — the fail-closed caller resolution and the ordering fix (selectors validated before the username lookup) both look correct, and the new tests exercise the real dispatch path well.

Also, as a possible follow-up (not blocking this PR): `AddNodes`/`RemoveNodes` still go through `effective_user()` and the uncredentialed channel, so they don't get the same authenticated-identity benefit this PR gives `up`/`down`/`kubeconfig`. Might be worth a tracked follow-up to bring them in line.

crates/spur-cli/src/k8s.rs
```diff
@@ -102,8 +109,10 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
             partition,
             selector,
         } => {
+            let caller = current_user()?;
             cmd_up(
```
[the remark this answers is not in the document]
yansun1996: Fixed in the follow-up commit — selectors are now validated (`selector_map`) before `current_user()` is called in the Up path.

crates/spur-cli/src/mock_controller.rs
joshkmartinez: Good catch. I trimmed this to the two-line form.

PR #611, opened 2026-08-11
fix: fail rejected k8s jobs and enforce swap, burst-buffer, and scancel semantics

biluriuday:
## Summary

The expanded e2e suite (companion PR #579) ran the daemons, CLI, and operator
against real clusters and turned up a set of genuine product defects. This PR
carries only those fixes and their unit tests; the e2e tests that caught them
live in #579.

## Fixes

- **Burst buffer directives were silently dropped** (`spurd`) — `stage_in:`/
  `stage_out:` were read from the daemon's own environment instead of the job's,
  so they never ran (a job with `stage_in:exit 7` still completed with exit 0).
  Now read from the job config's environment.
- **`scancel --partition` / `--account` never selected anything** (`spur-cli`) —
  both flags were declared and passed through, but the "did the caller name a
  selection?" guard only checked job IDs, `--user`, and `--name`, so they always
  failed with *"no job IDs or filters specified."* Replaced with a
  `has_selection()` predicate (unit-tested); `--state` stays a narrower, not a
  selector.
- **`--mem` overruns were paged to swap instead of OOM-killed** (`spurd`,
  `spur-core`) — `memory.max` was set but swap was uncapped, so a job that
  outgrew its limit kept running on swap and never reached `OutOfMemory`. A new
  `[cgroup]` config section caps swap, mirroring Slurm's `ConstrainSwapSpace` /
  `AllowedSwapSpace`. It defaults **off** so deployed clusters keep today's
  behavior; enforcement is opt-in.
- **Compressed nodelists were mis-parsed** (`spur-cli`) — `sattach`/`srun` took
  the first comma-separated field of the nodelist, yielding `node[1` once the
  controller compresses a multi-node allocation. Both now expand the hostlist
  first.
- **The k8s operator wedged on permanent rejections and retried too slowly**
  (`spur-k8s`) — every failed submit was requeued, so a SpurJob the controller
  permanently rejected (unknown partition, denied account) sat without status
  forever; and reconcile failures backed off a flat 60s despite a comment
  claiming exponential backoff. Permanent rejections are now marked `Failed`
  (only transient RPC codes requeue, matching the agent's rule), and retries
  double from 1s to a 60s cap, counted per SpurJob and reset on a clean pass.

## Compatibility

No breaking changes. `[cgroup]` is additive with serde defaults; there are no
proto, persisted-state, config, or CLI removals/renames/retypes.

## Testing

- Unit tests: `scancel` selection predicate; the `[cgroup]` swap policy;
  `spur-k8s` operator (`cargo test -p spur-k8s`, 152 passing), clippy clean.
- Validated on the 3-node bare-metal cluster: resource limits pass 17/17
  including the OOM cases that depend on the new swap knob.
- End-to-end validation is the companion suite in #579.

## Merge order

Merge this first, then rebase #579 onto `main` so its e2e tests run against the
fixed code (those tests are red without these fixes — that's the point of the
split).

### DIFF
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -1315,10 +1315,8 @@ impl MpiConfig {
     }
 }
 
-/// cgroup constraints applied to a job's processes (Slurm's `cgroup.conf`).
-///
-/// The limits themselves come from the allocation; this section only sets how
-/// they are enforced.
+/// cgroup enforcement settings for a job's processes (Slurm's `cgroup.conf`).
+/// See `docs/admin-guide/configuration.rst` for the semantics.
 #[derive(Debug, Clone, Default, Serialize, Deserialize)]
 pub struct CgroupConfig {
     /// Cap a job's swap alongside its memory (Slurm's `ConstrainSwapSpace`).
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -112,6 +112,10 @@ pub struct SlurmConfig {
     #[serde(default)]
     pub rlimits: RlimitsConfig,
 
+    /// cgroup enforcement policy for job processes.
+    #[serde(default)]
+    pub cgroup: CgroupConfig,
+
     /// MPI plugin settings for `--mpi=pmix` job steps.
     #[serde(default)]
     pub mpi: MpiConfig,
@@ -1308,6 +1312,81 @@ impl MpiConfig {
             return std::path::PathBuf::from(&self.pmix_plugin);
         }
         std::path::Path::new(&self.plugin_dir).join("spur_mpi_pmix.so")
+    }
+}
+
+/// cgroup constraints applied to a job's processes (Slurm's `cgroup.conf`).
+///
+/// The limits themselves come from the allocation; this section only sets how
+/// they are enforced.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct CgroupConfig {
+    /// Cap a job's swap alongside its memory (Slurm's `ConstrainSwapSpace`).
+    ///
+    /// Off by default, as in Slurm: while off, `memory.max` bounds resident
+    /// memory only and a job that outgrows `--mem` swaps instead of being killed.
+    #[serde(default)]
+    pub constrain_swap_space: bool,
+
+    /// Swap a job may use, as a percentage of its memory allocation (Slurm's
+    /// `AllowedSwapSpace`). Only consulted when `constrain_swap_space` is set.
+    #[serde(default)]
+    pub allowed_swap_space: u32,
+}
+
+/// Swap budget for a job's cgroup.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SwapLimit {
+    /// Leave `memory.swap.max` alone; the job may swap without bound.
+    Unconstrained,
+    /// Cap swap at this percentage of the job's memory allocation.
+    Percent(u32),
+}
+
+impl SwapLimit {
+    /// Swap budget in bytes for a job capped at `memory_bytes`, or `None` when
+    /// swap is unconstrained.
+    pub fn bytes_for(&self, memory_bytes: u64) -> Option<u64> {
+        match self {
+            Self::Unconstrained => None,
+            Self::Percent(percent) => Some(memory_bytes.saturating_mul(u64::from(*percent)) / 100),
+        }
+    }
+}
+
+impl CgroupConfig {
+    pub fn swap_limit(&self) -> Result<SwapLimit, ConfigError> {
+        if !self.constrain_swap_space {
+            return Ok(SwapLimit::Unconstrained);
+        }
+        if self.allowed_swap_space > 100 {
+            return Err(ConfigError::InvalidValue {
+                field: "cgroup.allowed_swap_space".into(),
+                value: format!(
+                    "{} (expected a percentage of the memory allocation, 0 to 100)",
+                    self.allowed_swap_space
+                ),
+            });
+        }
+        Ok(SwapLimit::Percent(self.allowed_swap_space))
+    }
+}
+
+/// Resolved limits an agent applies to every job it launches, as opposed to
+/// the per-job sizes that come from the allocation.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct JobLimits {
+    /// Applied before exec, while still privileged.
+    pub memlock: MemlockLimit,
+    pub swap: SwapLimit,
+}
+
+impl Default for JobLimits {
+    fn default() -> Self {
+        Self {
+            memlock: MemlockLimit::Unlimited,
+            swap: SwapLimit::Unconstrained,
+        }
     }
 }
 
@@ -2938,6 +3017,76 @@ cluster_name = "test"
         assert_eq!(
             config.rlimits.memlock_limit().unwrap(),
             MemlockLimit::Unlimited
+        );
+    }
+
+    #[test]
+    fn cgroup_swap_is_unconstrained_by_default() {
+        let cfg = CgroupConfig::default();
+        assert_eq!(cfg.swap_limit().unwrap(), SwapLimit::Unconstrained);
+        assert_eq!(SwapLimit::Unconstrained.bytes_for(64 * 1024 * 1024), None);
+    }
+
+    #[test]
+    fn cgroup_constrained_swap_defaults_to_denying_swap() {
+        let cfg = CgroupConfig {
+            constrain_swap_space: true,
+            allowed_swap_space: 0,
+        };
+        let limit = cfg.swap_limit().unwrap();
+        assert_eq!(limit, SwapLimit::Percent(0));
+        assert_eq!(limit.bytes_for(64 * 1024 * 1024), Some(0));
+    }
+
+    #[test]
+    fn cgroup_allowed_swap_is_a_percentage_of_the_allocation() {
+        let cfg = CgroupConfig {
+            constrain_swap_space: true,
+            allowed_swap_space: 50,
+        };
+        assert_eq!(
+            cfg.swap_limit().unwrap().bytes_for(64 * 1024 * 1024),
+            Some(32 * 1024 * 1024)
+        );
+    }
+
+    #[test]
+    fn cgroup_allowed_swap_percentage_above_100_errors() {
+        let cfg = CgroupConfig {
+            constrain_swap_space: true,
+            allowed_swap_space: 101,
+        };
+        assert!(cfg.swap_limit().is_err());
+    }
+
+    #[test]
+    fn cgroup_allowed_swap_is_ignored_while_unconstrained() {
+        let cfg = CgroupConfig {
+            constrain_swap_space: false,
+            allowed_swap_space: 999,
+        };
+        assert_eq!(cfg.swap_limit().unwrap(), SwapLimit::Unconstrained);
+    }
+
+    #[test]
+    fn cgroup_from_toml_explicit() {
+        let toml = r#"
+cluster_name = "test"
+
+[cgroup]
+constrain_swap_space = true
+allowed_swap_space = 25
+"#;
+        let config = SlurmConfig::load_from_str(toml).unwrap();
+        assert_eq!(config.cgroup.swap_limit().unwrap(), SwapLimit::Percent(25));
+    }
+
+    #[test]
+    fn cgroup_from_toml_default() {
+        let config = SlurmConfig::load_from_str("cluster_name = \"test\"\n").unwrap();
+        assert_eq!(
+            config.cgroup.swap_limit().unwrap(),
+            SwapLimit::Unconstrained
         );
     }
 
--- a/crates/spur-k8s/src/job_controller.rs
+++ b/crates/spur-k8s/src/job_controller.rs
@@ -108,22 +108,41 @@ fn should_submit(status: &SpurJobStatus) -> bool {
     status.spur_job_id.is_none()
 }
 
+/// Whether a failed submit is worth another attempt. Only "no controller
+/// answered" codes qualify; anything else faults the spec itself and would be
+/// rejected identically forever. Matches the agent's controller-RPC rule.
+fn submit_is_retryable(status: &tonic::Status) -> bool {
+    use tonic::Code;
+    matches!(
+        status.code(),
+        Code::Unavailable | Code::Internal | Code::DeadlineExceeded | Code::Unknown
+    )
+}
+
+/// Result of the submit phase.
+enum SubmitOutcome {
+    Submitted,
+    /// A prior reconcile already submitted this SpurJob.
+    AlreadySubmitted,
+    /// The controller refused the spec; the SpurJob is now Failed.
+    Rejected,
+}
+
 /// Submit to spurctld, apply job-id label, and patch CRD status.
 /// Re-reads from the API server first to guard against stale informer cache.
-/// Returns `Ok(None)` if a prior reconcile already submitted.
 async fn submit_to_controller(
     api: &Api<SpurJob>,
     ctx: &JobControllerCtx,
     name: &str,
     ns: &str,
     job: &SpurJob,
-) -> Result<Option<u32>, ReconcileError> {
+) -> Result<SubmitOutcome, ReconcileError> {
     // Fresh read from API server — informer cache may be stale after finalizer patch
     let fresh = api.get(name).await.map_err(ReconcileError::Kube)?;
     let fresh_status = fresh.status.clone().unwrap_or_default();
     if !should_submit(&fresh_status) {
         debug!(spurjob = %name, "already submitted by prior reconcile");
-        return Ok(None);
+        return Ok(SubmitOutcome::AlreadySubmitted);
     }
 
     let user = job
@@ -145,9 +164,19 @@ async fn submit_to_controller(
         .await
     {
         Ok(resp) => resp.into_inner().job_id,
-        Err(e) => {
+        Err(e) if submit_is_retryable(&e) => {
             error!(spurjob = %name, error = %e, "failed to submit SpurJob");
             return Err(ReconcileError::Grpc(e));
+        }
+        Err(e) => {
+            error!(spurjob = %name, error = %e, "SpurJob rejected by the controller");
+            let new_status = SpurJobStatus {
+                state: "Failed".into(),
+                message: Some(format!("rejected by spurctld: {}", e.message())),
+                ..fresh_status.clone()
+            };
+            patch_status(api, name, &new_status).await;
+            return Ok(SubmitOutcome::Rejected);
         }
     };
     drop(ctrl);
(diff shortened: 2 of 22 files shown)


crates/spur-k8s/src/job_controller.rs
```diff
@@ -108,22 +161,40 @@ fn should_submit(status: &SpurJobStatus) -> bool {
     status.spur_job_id.is_none()
 }
 
+/// Whether a failed submit is worth another attempt. Only "no controller
+/// answered" codes qualify; anything else faults the spec itself and would be
+/// rejected identically forever. Matches the agent's controller-RPC rule.
+fn submit_is_retryable(status: &tonic::Status) -> bool {
+    use tonic::Code;
+    matches!(
```
yansun1996: This doc comment is 3 lines — repo convention keeps these to ~2. Maybe trim to something like "Only transient RPC failures are retried; a spec rejection would repeat forever."

crates/spur-k8s/src/job_controller.rs
```diff
@@ -53,6 +54,50 @@ pub struct JobControllerCtx {
     pub ctrl_client: Mutex<SlurmControllerClient<Channel>>,
     /// Track multi-pod completion: job_id → (expected_count, completed_count, any_failed)
     pub(crate) pod_tracker: Mutex<HashMap<u32, PodTracker>>,
+    /// Consecutive reconcile failures per SpurJob, driving the retry delay.
+    /// Std mutex because `error_policy` is synchronous.
+    pub(crate) failures: StdMutex<HashMap<String, u32>>,
+}
+
+fn object_key(job: &SpurJob) -> String {
```
yansun1996, line 62: `object_key()` is namespace/name only, and `failures` entries are only cleared on a successful reconcile. Two edge cases: a delete+recreate with the same name inherits the old failure count/backoff, and if a SpurJob is ever removed without a clean reconcile pass (finalizer stripped externally, etc.), the entry leaks for the operator's lifetime. Might be worth keying on UID too, and/or reaping on the delete event.

crates/spur-k8s/src/job_controller.rs
```diff
@@ -108,22 +161,40 @@ fn should_submit(status: &SpurJobStatus) -> bool {
     status.spur_job_id.is_none()
 }
 
+/// Whether a failed submit is worth another attempt. Only "no controller
+/// answered" codes qualify; anything else faults the spec itself and would be
+/// rejected identically forever. Matches the agent's controller-RPC rule.
+fn submit_is_retryable(status: &tonic::Status) -> bool {
```
yansun1996: This and `agent_server.rs`'s `controller_rpc_retryable` have identical match arms, tied together only by a comment saying they match. Since they live in different crates, a future change to one could silently drift from the other — might be worth a shared helper in `spur-core`.

yansun1996 (commented):
Nice cleanup — these are clear, well-isolated fixes for real e2e-caught defects, and the swap-cap default-off compatibility story is sound. Two things worth a look before merge, plus a few smaller suggestions inline.

**Container-launch path still has the fork race this PR fixes elsewhere.** `spawn_job_process` now joins the cgroup pre-exec inside the forked child specifically to avoid a race with `unshare --fork`, but `launch_container_job`'s cgroup placement (`executor.rs:1584`, `cgroup.join("cgroup.procs")`) still happens from the *parent* after `fork()` returns. The container's own `execve` can run before that write lands, so `memory.max`/`memory.swap.max` (this PR's new feature) aren't guaranteed to be in place before a container job starts. Worth applying the same pre-exec-in-child pattern here.

**`scancel --account`/`--partition` without `-u` silently narrows to the caller's own jobs.** Now that these are real selectors (`scancel.rs`, `has_selection`), `user` still defaults to `current_user()` when `-u` is omitted (`scancel.rs:75`), and that default is sent on `GetJobsRequest` regardless of the other selectors. So `scancel --account gpu_team` without `-u` cancels "my jobs in that account," not all of them — probably not what an operator using this newly-working flag expects. Would it make sense to only default to the current user when no other selector was given?

A few smaller items below on comment length and duplicated retry-classification logic.

**Missing test coverage:** the actual regression this PR fixes in `spur-k8s` — a permanently-rejected SpurJob getting stuck retrying forever — has no test that drives `reconcile`/`handle_job` end-to-end; only the pure `submit_is_retryable`/`backoff_delay` helpers are unit-tested. A test with a fake `Api<SpurJob>` + fake controller client asserting "one permanent rejection → `Failed` status → no further reconciliation" would be the real regression guard.

crates/spur-k8s/src/job_controller.rs
biluriuday: done

docs/admin-guide/configuration.rst
```diff
@@ -512,6 +513,43 @@ POSIX ``RLIMIT_*`` values ``spurd`` applies to job steps at launch.
...
+     - ``false``
+     - Cap a job's swap along with its memory. While ``false``, ``memory.max``
+       bounds resident memory only: on a node with swap, a job that outgrows
+       ``--mem`` keeps running on swap instead of being killed.
+   * - ``allowed_swap_space``
+     - integer
+     - ``0``
+     - Swap a job may use, as a percentage (0-100) of its memory allocation.
+       Only consulted when ``constrain_swap_space`` is set; ``0`` denies swap
+       outright. A value above 100 errors at parse time.
```
yansun1996: "A value above 100 errors at parse time" doesn't quite match the implementation — the bound is only checked in `swap_limit()`, which is called from `spurd`'s startup path, not from config parsing/validation. A bad value would pass through silently until `spurd` starts. Maybe reword to "validated at `spurd` startup"?

crates/spur-k8s/src/job_controller.rs
```diff
@@ -145,10 +217,24 @@ async fn submit_to_controller(
...
+                state: "Failed".into(),
+                message: Some(format!("rejected by spurctld: {}", e.message())),
+                ..fresh_status.clone()
+            };
+            // Rejected returns await_change(), so a dropped write here can't
+            // self-heal; propagate it and let error_policy retry with backoff.
+            patch_status(api, name, &new_status)
+                .await
+                .map_err(ReconcileError::Kube)?;
+            return Ok(SubmitOutcome::Rejected);
```
yansun1996, line 239: If the submit is rejected and this `patch_status` call itself fails, the whole function returns `Err` and `error_policy` requeues — but the next reconcile re-enters `submit_to_controller` from the top (since `should_submit` still sees no status), which issues a fresh real `submit_job` RPC rather than just retrying the status write. It's bounded by backoff so not corruption-level, but a permanently-rejected job could trigger several duplicate submits before the `Failed` status finally lands. Might be worth a status that only retries the patch in this case.

crates/spur-k8s/src/job_controller.rs
```diff
@@ -53,6 +54,60 @@ pub struct JobControllerCtx {
     pub ctrl_client: Mutex<SlurmControllerClient<Channel>>,
     /// Track multi-pod completion: job_id → (expected_count, completed_count, any_failed)
     pub(crate) pod_tracker: Mutex<HashMap<u32, PodTracker>>,
+    /// Consecutive reconcile failures per SpurJob (keyed by [`failure_key`]),
+    /// driving the retry delay. Std mutex because `error_policy` is synchronous.
+    pub(crate) failures: StdMutex<HashMap<String, u32>>,
```
yansun1996, line 59: With UID keying, delete+recreate under the same name is handled well now. One residual gap: if a SpurJob's finalizer is ever stripped externally (or the object is force-deleted) rather than going through this controller's own cleanup path, its entry here is never reaped and stays for the operator's process lifetime — there's no TTL/cap on this map. Given it's already called out as a known limitation, maybe just worth a tracking issue rather than fixing here.

crates/spurd/src/executor.rs
```diff
@@ -961,18 +982,33 @@ fn setup_cgroup(
...
-            "failed to move process to cgroup — job runs without isolation"
-        );
-        false
-    } else {
-        true
+/// Move the calling process into a cgroup by writing its own pid to
+/// `cgroup.procs`. Runs in the post-fork child before exec, so it must be
+/// async-signal-safe: no heap allocation, only raw syscalls. Best-effort —
+/// on failure the job stays in the parent cgroup rather than aborting launch.
+fn join_cgroup_self(procs_path: &std::ffi::CStr) {
```
yansun1996: `join_cgroup_self` swallows failures silently (understandable given the async-signal-safety constraint here), but the old `move_to_cgroup` it replaces logged a `warn!` on failure. That means a cgroup-join failure — and thus an unenforced swap cap — is now completely invisible. Would a raw `libc::write` to stderr on the failure path be worth adding, to keep some visibility without breaking the safety contract?

crates/spur-core/src/config.rs
```diff
@@ -1148,6 +1152,81 @@ impl MpiConfig {
     }
 }
 
+/// cgroup constraints applied to a job's processes (Slurm's `cgroup.conf`).
+///
+/// The limits themselves come from the allocation; this section only sets how
+/// they are enforced.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct CgroupConfig {
```
yansun1996, line 1321: This doc comment nearly duplicates the prose just added to `configuration.rst` for the same field. Not wrong, just a bit redundant — could trim one down to a one-liner pointing at the other.

yansun1996 (commented):
Re-reviewed after the latest round of fixes — this looks solid.

All three prior review points check out:
- The `requester`/`filter_user` split in `scancel` correctly keeps cancel authorization on the real caller while only widening the *listing* filter, so `scancel --account X` without `-u` no longer silently narrows to the caller's own jobs, and there's no fail-open.
- `controller_rpc_retryable` is genuinely consolidated into `spur-proto` with all three crates calling the shared helper.
- The failure/backoff map is now keyed by UID, so delete+recreate under the same name starts a fresh backoff sequence.

A few smaller things worth a look, none blocking:

docs/admin-guide/configuration.rst
biluriuday: fixed the documentation

crates/spur-k8s/src/job_controller.rs
biluriuday: On a permanent rejection, the Failed status write now retries in place within a budget instead of propagating the error, so a transient API blip no longer causes the next reconcile to re-run a real submit_job for an already-doomed spec.

Added patch_terminal_status, mirroring the existing ensure_job_id_label retry pattern (backon exponential retry on transient kube errors, bounded by a 3s STATUS_PATCH_BUDGET):

The rejection branch calls it instead of the single-shot patch_status. Behavior:

- Transient API error → retried in-place until it lands (no re-submit). This is the common case the reviewer flagged.
- Non-transient error (e.g. object deleted) → returns immediately (not retried), falls through to error_policy as before.
- Persistent outage exceeding the 3s budget → times out and falls back to error_policy (which may re-submit), but only in that rare pathological case, bounded by backoff.

So a permanently-rejected job now reliably lands Failed in a single reconcile under normal conditions, eliminating the duplicate-submit churn.

crates/spurd/src/executor.rs
biluriuday: Good catch. A raw stderr write from the child wouldn't actually help though — by the time the pre_exec closure runs, fd 2 is already redirected to /dev/null (batch) or the job's own pty/output (interactive), so it'd be invisible or would corrupt job output. Instead I dup spurd's stderr to a CLOEXEC fd before spawn/fork and hand that to join_cgroup_self; on failure it does an async-signal-safe libc::write of a fixed message there, then the fd closes at exec so it never leaks into the job. Warning now lands in spurd's log/journal. Added a pipe-backed test for the failure path.

crates/spur-core/src/config.rs
biluriuday: Agreed, trimmed the struct doc to a one-liner pointing at configuration.rst, which now owns the prose. Kept the per-field docs since they're useful inline and aren't a wholesale duplicate.

crates/spur-k8s/src/job_controller.rs
biluriuday: Opened an issue https://github.com/ROCm/spur/issues/632 to track and follow-up

crates/spur-proto/src/lib.rs
```diff
@@ -22,6 +22,15 @@ pub const MAX_GRPC_MESSAGE_SIZE: usize = 32 * 1024 * 1024;
 /// while allowing large outbound responses.
 pub const MAX_GRPC_REQUEST_SIZE: usize = 8 * 1024 * 1024;
 
+/// Only transient RPC failures are retried; a spec rejection would repeat forever.
+pub fn controller_rpc_retryable(status: &tonic::Status) -> bool {
```
yansun1996, line 26: This classifier decides whether spur-k8s marks a SpurJob permanently `Failed`. One case slips through: `cluster.rs`'s account-membership check returns `InvalidArgument` for "account associations are temporarily unavailable (accounting cache not loaded)" — which happens for a real window right after `spurctld` (re)starts with accounting enabled, since the association-cache refresh loop runs in the background after the RPC server is already serving. A SpurJob submitted in that window would get permanently failed instead of retried. Might be worth having spurctld return `Unavailable` for that specific case, or having this classifier check for it explicitly.

crates/spur-k8s/src/job_controller.rs
```diff
@@ -732,12 +832,50 @@ async fn ensure_job_id_label(
...
-    if let Err(e) = api.patch_status(name, &pp, &Patch::Merge(&patch)).await {
-        error!(spurjob = %name, error = %e, "failed to patch SpurJob status");
-    }
+    api.patch_status(name, &pp, &Patch::Merge(&patch))
+        .await
+        .map(|_| ())
+}
+
+/// Write a terminal status, retrying transient API errors within a budget. The
+/// submit that produced this state is already spent, so a dropped write must not
```
yansun1996: This comment ("a dropped write must not fall back to re-running submit") is only true within `STATUS_PATCH_BUDGET` (3s). If the status patch times out past that budget, `patch_terminal_status` still returns `Err`, which propagates up through `submit_to_controller` → `handle_job` → `error_policy`, which requeues — and the next reconcile finds no persisted terminal state and resubmits. It's a much narrower window than before this PR (needs the apiserver to stay unavailable for 3+ seconds rather than just any blip), but it's the same failure mode the PR is fixing, just pushed further out. Might be worth an unbounded retry here (or bounded by the reconcile's own timeout) rather than a fixed budget, since `await_change()` means there's no other trigger to retry the write.

crates/spurd/src/executor.rs
```diff
@@ -756,6 +755,26 @@ async fn spawn_job_process(
         });
     }
 
+    // Join the cgroup pre-exec, not parent-side after spawn: under `unshare
```
yansun1996: Small thing: a few of the new comments in this PR run 3-4 lines (here, `join_cgroup_self`'s doc comment around line 999, and the `launch_container_job` one around line 1292) — repo convention keeps these to ~2 lines, moving the extra rationale into the PR description if needed.

yansun1996 (commented):
Took another pass over this before merge. The earlier round's points all check out as genuinely fixed — nothing to re-litigate there. Two more things worth a look on the new permanent-rejection path, plus a couple of small comment-length nits. None of this should block merge on its own.

One thing worth being aware of rather than fixing here: `authoritative_user` (pre-existing, not touched by this PR) overwrites `get_jobs`'s `user` filter with the authenticated identity whenever auth is enforced. That means the new `scancel --account`/`--partition` "span the whole account" behavior only takes effect when auth is off — under `auth.mode = required` it still narrows to the caller, since the server always fills in the filter with the caller's own identity. Cancellation itself stays correctly scoped either way (the per-cancel owner check doesn't use this filter), so this is a UX limitation, not a security issue — but worth a follow-up if the account-wide use case matters under enforced auth.

crates/spur-proto/src/lib.rs
biluriuday: Good catch — fixed at the source instead of in the classifier. spurctld now returns Unavailable (gRPC) / 503 (REST) for the cold-association-cache window rather than InvalidArgument, so it's retryable for every client, not just spur-k8s. String-matching the message in controller_rpc_retryable would be fragile and wrong-layer. While there I fixed the same transient-as-permanent issue in the sibling require_association "no account resolved" branch (a default account may not be cached yet right after start), mapped the new variant in both the gRPC and REST error paths, added tests, and corrected accounting.rst, which still described this path as fail-open.

crates/spur-k8s/src/job_controller.rs
biluriuday: Fixed: patch_terminal_status now retries transient API errors unbounded (bailing only on permanent errors) instead of a fixed 3s budget, so the Rejected terminal write lands whenever the apiserver is reachable — no more resubmit after a >3s blip. Also dropped the controller-client lock before the write; holding it across an unbounded retry would have stalled every other reconcile's RPCs during an API outage.

crates/spurd/src/executor.rs
biluriuday: Trimmed all three to 2 lines. Kept the async-signal-safety constraint in join_cgroup_self's doc since it's a hard contract for anyone editing that function (not just PR rationale), and dropped the redundant/extra detail from the other two.

yansun1996 (approved):
Approving. Re-reviewed the latest commit (`689835e`) on top of the three earlier rounds — everything I raised previously is genuinely addressed, all threads are resolved, and CI is green across all 17 checks.

The new cold-cache work checks out:

- `SubmitError::Unavailable` is wired through both transports (gRPC `Unavailable`, REST 503) and the test asserts `spur_proto::controller_rpc_retryable` agrees, so the classifier and the error surface can't drift apart. This closes a hole this PR itself opened: without it, a submit during the association-cache warm-up window returned `InvalidArgument`, which the operator's new permanent-rejection path would turn into a `Failed` SpurJob for a condition that self-heals in seconds. Both paths are covered (explicit account and the `require_association` default-account path).
- `patch_terminal_status` dropping the 3s budget for unbounded retry is the right call given `Rejected` returns `await_change()` — a dropped write can't self-heal, and a bounded one would fall back to resubmitting an already-doomed spec. Releasing the controller lock before the write is a good catch; it keeps an API outage from stalling other reconciles.
- Docs corrected too — the accounting guide had described this path as fail-open when it has fenced closed since the cold-cache guard landed.

One thing to be aware of rather than change: the unbounded retry means that object's reconcile won't return during a prolonged API-server outage. kube-rs reconciles per object, so it doesn't wedge the controller, and little else works in that situation anyway — just noting the bounded-failure/unbounded-wait trade.

Two follow-ups worth tickets, neither blocking:

1. The `spur-k8s` regression this PR fixes — a permanently-rejected SpurJob stuck retrying forever — still has no test that drives `reconcile`/`handle_job` end to end; only the pure helpers (`submit_is_retryable`, `backoff_delay`, `failure_key`) are unit-tested. A fake `Api<SpurJob>` + fake controller client asserting "one permanent rejection → `Failed` → no further reconciliation" would be the real guard.
2. The `authoritative_user` limitation noted last round: under `auth.mode = required`, `scancel --account`/`--partition` still narrows to the caller because the server fills the filter with the authenticated identity. UX gap, not a security issue.

Merge order per the description still applies — this first, then rebase #579 so its e2e tests run against the fixed code.

sgopinath1 (approved):
LGTM

PR #613, opened 2026-08-11
fix(spurctld): preserve node allocations during recovery

jamesETsmith:
## Motivation

Prevent nodes with active CPU or GPU allocations from being reported as  IDLE  after heartbeat recovery. This inconsistency causes new jobs to remain pending with Reason=Resources  even though  sinfo  reports the node as available.

## Technical Details

Heartbeat recovery unconditionally transitioned recovered nodes to  NodeState::Idle, ignoring the controller’s existing  alloc_resources . The fix derives the recovered state from current allocations, preserving Allocated  or Mixed as appropriate and using  Idle only when no resources remain allocated.

A regression test verifies that a fully allocated node recovers as  Allocated .

## Test Plan

•  cargo fmt --all -- --check
•  git diff --check
• cargo test -p spurctld health_ --locked

## Test Result

All tests passed

## Submission Checklist

[X] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

### DIFF
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -2340,11 +2340,15 @@ impl ClusterManager {
                     }
                 }
                 HealthAction::Recover { name, old_state } => {
-                    info!(node = %name, "node recovered (heartbeat resumed)");
+                    let recovered_state = self
+                        .get_node(&name)
+                        .map(|node| recovered_node_state(&node))
+                        .unwrap_or(NodeState::Idle);
+                    info!(node = %name, state = ?recovered_state, "node recovered (heartbeat resumed)");
                     if let Err(e) = self.propose(WalOperation::NodeStateChange {
                         name,
                         old_state,
-                        new_state: NodeState::Idle,
+                        new_state: recovered_state,
                         reason: None,
                         admin_locked: false,
                     }) {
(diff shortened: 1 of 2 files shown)


crates/spurctld/src/cluster.rs
```diff
@@ -5832,6 +5833,16 @@ pub(crate) enum HealthAction {
     },
 }
 
+pub(crate) fn recovered_node_state(node: Option<&Node>) -> NodeState {
+    let Some(node) = node else {
+        return NodeState::Idle;
+    };
+    let mut recovered = node.clone();
+    recovered.state = NodeState::Idle;
+    recovered.update_state_from_alloc();
```
yansun1996, line 5842: Small nit: forcing `state = Idle` right before this call means `update_state_from_alloc`'s `is_admin_hold()` guard can never trigger here. That's fine today (`Recover` only fires for `Down`/`Error` nodes), but it's not obvious just from reading this function. A one-line comment on why the reset happens first would help future readers.

crates/spurctld/src/cluster.rs
```diff
@@ -14389,6 +14400,27 @@ mod tests {
         );
     }
 
+    #[test]
+    fn recovered_allocated_node_stays_allocated() {
```
yansun1996, line 14404: These pure-function tests are a good start. Would it be worth also adding one that goes through the real path (`check_node_health` → `apply_health_actions` → `propose`), similar to `check_node_health_marks_stale_down` further down? That would cover the fresh `get_node()` read this PR adds, not just the state-derivation logic in isolation.

yansun1996 (commented):
Nice fix — this closes a real gap where a node recovering from a heartbeat blip would always be reported `Idle` even if it still had live allocations, which is confusing for anyone reading `sinfo`. Two small suggestions below, otherwise LGTM.

yansun1996 (approved):
LGTM, the above comments are nit or test case suggestion

PR #617, opened 2026-08-12
chore: resolve semantic merge conflict due to #593 and #565

shiv-tyagi:
## What

Fixes a build break on `main` from a semantic merge conflict between two independently-merged PRs.

- #593 changed `validate_partition` and `validate_partition_node_bounds` to take a `partitions: &[Partition]` snapshot argument.
- #565 added new call sites for both methods in `apply_submit_outcome` (the job-submit hook re-validation path), still calling them with only `spec`.

Both merged without a textual conflict, so `main` failed to compile:

```
error[E0061]: this method takes 2 arguments but 1 argument was supplied
  --> crates/spurctld/src/cluster.rs  self.validate_partition_node_bounds(spec)?
  --> crates/spurctld/src/cluster.rs  self.validate_partition(spec)?
```

## Approach

`submit_job` already takes one `partitions` snapshot at the top of the submit path so defaulting and enforcement can't observe a concurrent `reconfigure()` mid-submit. Thread that same snapshot down through `run_job_submit_hook` into `apply_submit_outcome`, rather than re-reading the lock. Hook re-validation now runs against the same snapshot as the rest of the submit path.

## Testing

- `cargo build --workspace`
- `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` (no warnings)
- `cargo test -p spurctld --locked` (727 passed, 0 failed), including the hook re-validation tests

### DIFF
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -574,7 +574,7 @@ impl ClusterManager {
         // bounded whitelist fields, so this still measures what gets persisted.
         check_submission_size(&spec)?;
 
-        self.run_job_submit_hook(&mut spec)?;
+        self.run_job_submit_hook(&mut spec, &partitions)?;
 
         // Reject unknown/malformed dependency types up front so users get a
         // clear error instead of a silently-deadlocked job (e.g. `expand:N`).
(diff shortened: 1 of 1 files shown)


sgopinath1 (approved):
LGTM

Issue #619, opened 2026-08-12
Jobs with a future --begin time show pending Reason=None on submit


shiv-tyagi:
## What happens

A job submitted with a future `--begin=<time>` is correctly held until that time (`is_begin_held` gates it out of scheduling), but `PendingReason::BeginTime` is only assigned on the preempt/backoff paths, not at submit. Between submit and the begin time, the job shows `Reason=None` in `squeue`/`scontrol show job`.

## Expected

A job waiting on its `--begin` time should show `BeginTime` (Slurm displays `(BeginTime)`) from submit onward, not `None`.

## Repro

1. `sbatch --begin=now+1hour job.sh` (or any future time).
2. `squeue` / `scontrol show job <id>` immediately.
3. Reason is `None` until the begin window passes, instead of `BeginTime`.

## Notes

Related to the broader `Reason=None` gap. The in-pass aggregate-gate drops are fixed; the runtime backstop deliberately leaves begin-held jobs alone (so it does not mislabel this case as `Priority`), which is why this needs its own tagging at submit. Related umbrella: [307].

shiv-tyagi:
Fixed on main

Issue #618, opened 2026-08-12
Array-throttled tasks show pending Reason=None


shiv-tyagi:
## What happens

When a job array is throttled by its per-array concurrency limit (`--array=...%N`), the tasks held back above the limit are marked `scheduling_eligible = false` in `classify_pending_jobs`, but no pending reason is assigned. They remain at `PendingReason::None`, so `squeue`/`scontrol show job` display `Reason=None` for a job that is actually waiting on the array throttle.

## Expected

A task held back purely by the array concurrency limit should carry a reason that explains the hold. Slurm reports `JobArrayTaskLimit` for this case.

## Repro

1. Submit an array with a low concurrency cap and more tasks than the cap, e.g. `sbatch --array=1-4%1 job.sh`.
2. `squeue` / `scontrol show job <id>` immediately.
3. The throttled tasks show `Reason=None` instead of a throttle reason.

## Notes

Same class of gap as the in-pass aggregate-gate drops that leave `Reason=None` (fixed for the account/QOS/license/burst-buffer gates). This one is on the ineligible (`scheduling_eligible = false`) path rather than the aggregate gates, so it is out of scope for that fix and tracked here. Related umbrella: [307].

PR #620, opened 2026-08-12
fix(spurctld): tag in-pass-blocked jobs so pending reason is never None

shiv-tyagi:
## What this fixes

Pending jobs could sit indefinitely with `Reason=None` in `squeue`/`scontrol`. Observed on a production cluster: a job blocked purely by a higher-priority sibling reserving QOS group headroom showed no reason at all.

## Root cause

`classify_pending_jobs` filters candidates through four aggregate gates (account, QOS, license, burst-buffer). Each gate dropped a candidate from the schedulable set **without recording a reason** when the block came only from *in-pass* reservation (a higher-priority job earlier in the same pass claimed the headroom), not from the persistent limit. Such a job never reached `update_pending_reasons` and was never added to the blocked set, so it kept the default `PendingReason::None`. It became permanent whenever a higher-priority job perpetually reserved the aggregate headroom in-pass but was itself never placed.

The gates split the drop decision (`*_block_with`, in-pass) from the displayed reason (`*_block_for`, persistent). When `_with` tripped but `_for` did not, the job was dropped and left untagged. This split was introduced by [478], which added the in-pass `*_block_with` checks and `PassReservations` to stop a single pass over-subscribing aggregate caps, but kept `*_block_for` for the displayed reason.

## Approach

Restore the invariant: any candidate removed from the schedulable set carries a reason.

- Add `retain_eligible`, mirroring `retain_unblocked` but for the consumable gates. It keeps not-yet-eligible candidates untouched, lets the gate reserve headroom on the keep path, and records a reason whenever it drops an eligible candidate.
- Route the account+QOS gate through it and drop the `_for` guard, so any in-pass block is tagged.
- Tag the license and burst-buffer in-pass exhaustion (`Licenses` / `BurstBufferResources`).
- Delete the now-dead `account_block_for` / `qos_block_for`. This removes the `_with`/`_for` drift that caused the split, so the bug class cannot recur.

## How it was tested

- New in-pass regression tests per gate (QOS, account, burst-buffer) asserting the specific reason instead of `None`. Updated the existing license single-pass test to assert `Licenses`.
- `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` clean; `cargo test --locked` green.
- Bare-metal repro on the QOS path: same submission sequence, the in-pass-blocked job flipped from `Reason=None` (before) to `QOSGrpNodeLimit` (after).

## Follow-ups (same bug class, separate scope)

- [1] Array-throttled tasks show `Reason=None`: [618].
- [2] Jobs with a future `--begin` show `Reason=None` on submit: [619].

[478]: https://github.com/ROCm/spur/pull/478
[618]: https://github.com/ROCm/spur/issues/618
[619]: https://github.com/ROCm/spur/issues/619

### DIFF
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -2854,15 +2854,15 @@ impl ClusterManager {
                 if let Some(reason) =
                     account_block_with(job, &self.association_cache, &jobs, &reserved)
                 {
-                    return Some(reason);
+                    return GateOutcome::Block(reason);
                 }
                 if let Some(reason) =
                     qos_block_with(job, &qos_by_job[&job.job_id], &jobs, &reserved)
                 {
-                    return Some(reason);
+                    return GateOutcome::Block(reason);
                 }
                 reserved.reserve(job);
-                None
+                GateOutcome::Keep
             });
         }
 
@@ -2871,7 +2871,7 @@ impl ClusterManager {
             let mut remaining = available.clone();
             retain_eligible(&mut candidates, &mut blocked, |job| {
                 if let Some(reason) = license_block(job, &available) {
-                    return Some(reason);
+                    return GateOutcome::Block(reason);
                 }
                 let req = extract_license_requirements(&job.spec);
                 // In-pass contention counts as a license wait, same as a shortage.
@@ -2879,14 +2879,14 @@ impl ClusterManager {
                     .iter()
                     .any(|(lic, n)| remaining.get(lic).copied().unwrap_or(0) < *n)
                 {
-                    return Some(PendingReason::Licenses);
+                    return GateOutcome::Block(PendingReason::Licenses);
                 }
                 for (lic, n) in &req {
                     if let Some(avail) = remaining.get_mut(lic) {
                         *avail = avail.saturating_sub(*n);
                     }
                 }
-                None
+                GateOutcome::Keep
             });
         }
 
@@ -2894,36 +2894,29 @@ impl ClusterManager {
         {
             let available = self.available_bb_with(&jobs);
             let mut remaining = available;
-            candidates.retain(|candidate| {
-                if !candidate.scheduling_eligible {
-                    return true;
-                }
-                let job = &candidate.job;
+            retain_eligible(&mut candidates, &mut blocked, |job| {
                 if job.bb_stage_state == BbStageState::Staging {
-                    record_blocked(&mut blocked, candidate, PendingReason::BurstBufferStageIn);
-                    return false;
+                    return GateOutcome::Block(PendingReason::BurstBufferStageIn);
                 }
                 if let Some(reason) = burst_buffer_block(job, available) {
-                    record_blocked(&mut blocked, candidate, reason);
-                    return false;
+                    return GateOutcome::Block(reason);
                 }
                 let req = extract_bb_requirement(&job.spec);
                 if req == 0 {
-                    return true;
+                    return GateOutcome::Keep;
                 }
                 if job.bb_stage_state == BbStageState::Ready {
-                    return true;
+                    return GateOutcome::Keep;
                 }
                 // In-pass contention counts as a BB wait, same as a shortage.
                 if req > remaining {
-                    record_blocked(&mut blocked, candidate, PendingReason::BurstBufferResources);
-                    return false;
+                    return GateOutcome::Block(PendingReason::BurstBufferResources);
                 }
-                // Not a block: capacity is reserved and stage-in scheduled, so the
-                // job leaves the schedulable set and shows BurstBufferStageIn next.
+                // Capacity reserved and stage-in scheduled, so the job leaves the
+                // schedulable set and shows BurstBufferStageIn next cycle.
                 remaining = remaining.saturating_sub(req);
                 bb_stage_candidates.push(job.job_id);
-                false
+                GateOutcome::DropUntagged
             });
         }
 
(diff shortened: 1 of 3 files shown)


crates/spurctld/src/scheduler_loop.rs
```diff
@@ -225,6 +225,10 @@ pub async fn run(cluster: Arc<ClusterManager>, raft: Arc<RaftHandle>) {
             }
         }
 
+        // Invariant check: warn loudly if any pending job the passes above left
+        // at Reason=None slipped through, so a classifier gap surfaces in logs.
+        cluster.warn_untagged_pending();
```
yansun1996: This call sits after the `if pending.is_empty() { continue; }` above, so it never runs on a cycle where every pending job is already blocked or ineligible — i.e. exactly the saturated-queue state where a classifier gap would be most likely to exist and least likely to be noticed otherwise. Would it make sense to move this call up so it runs right after `pending_jobs_and_tag_reasons()`, independent of whether anything is currently schedulable?

crates/spurctld/src/cluster.rs
```diff
@@ -4037,6 +4027,30 @@ impl ClusterManager {
...
+    /// tasks are legitimately None and skipped.
+    pub(crate) fn warn_untagged_pending(&self) -> Vec<JobId> {
+        let now = Utc::now();
+        let jobs = self.jobs.read();
+        let mut untagged = Vec::new();
+        for job in jobs.values() {
+            if job.state != JobState::Pending
+                || job.pending_reason != PendingReason::None
+                || job.is_begin_held(now)
+                || job.spec.array_max_concurrent.is_some()
```
yansun1996: This skips every task belonging to a concurrency-limited array, not just tasks currently at their concurrency cap. A task in a limited array that's pending for some unrelated reason would be silently excluded from this check forever. Might be worth checking the array's actual running-count against its limit here (similar to the array-count logic in `classify_pending_jobs`) rather than gating on the field simply being set.

crates/spurctld/src/cluster.rs
```diff
@@ -4037,6 +4027,30 @@ impl ClusterManager {
         }
     }
 
+    /// Logs (never rewrites, which would mask the gap) any pending job stuck at
+    /// `Reason=None`. Holds, deadlines, begin-time holds, and array-throttled
```
yansun1996: Small nit: "Holds, deadlines" reads like they get a dedicated check here, but the only checks below are `is_begin_held` and `array_max_concurrent` — held/deadline jobs are actually excluded a few lines up via the `pending_reason != None` guard, since they already carry a concrete reason. Might be worth rewording slightly so it doesn't imply special-casing that isn't there.

yansun1996 (changes requested):
Nice fix — the `retain_eligible` refactor cleanly closes the gap between the in-pass and persistent checks, and the new regression tests do a good job pinning the specific reason for each gate. Left a few notes below.

One item I'd suggest addressing before merge: the new `warn_untagged_pending()` invariant check looks like it can be skipped on exactly the cycles where it matters most (see inline comment). The other two are smaller wording/coverage suggestions on the same function.

crates/spurctld/src/cluster.rs
```diff
@@ -4037,6 +4027,30 @@ impl ClusterManager {
...
+    /// tasks are legitimately None and skipped.
+    pub(crate) fn warn_untagged_pending(&self) -> Vec<JobId> {
+        let now = Utc::now();
+        let jobs = self.jobs.read();
+        let mut untagged = Vec::new();
+        for job in jobs.values() {
```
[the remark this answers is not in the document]
shiv-tyagi: Removed `warn_untagged_pending` entirely, so this is moot. The check asserted "every dropped pending job is tagged," which isn't universally true: BB stage candidates, array-throttled tasks, and begin-held jobs are legitimately untagged, so it needed a growing exclusion list. The fix is now structural: `retain_eligible` drops a candidate iff it returns a reason, backed by per-gate tests. On the cross-reference: line 5636 (`retain_eligible`) has no array check, so that pointer was incorrect.

crates/spurctld/src/scheduler_loop.rs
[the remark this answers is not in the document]
shiv-tyagi: Moot: the call and the function are removed, so there is no per-cycle scan to guard.

shiv-tyagi: Good catch on the placement gap. Rather than reposition the check, I removed it. On a saturated cycle the schedulable-but-unplaced jobs are tagged later by `update_pending_reasons`; moving the check above that would false-positive on them, and leaving it below misses exactly the saturated case you describe. The invariant is now enforced structurally in `retain_eligible` (a candidate is dropped iff a reason is returned), so there is no cycle-timing to get right.

crates/spurctld/src/cluster.rs
shiv-tyagi: Moot: the function is removed. The over-broad array exclusion was a symptom of the check asserting an invariant that isn't universally true. Structural tagging in `retain_eligible` plus the per-gate tests replace it.

crates/spurctld/src/cluster.rs
shiv-tyagi: Moot: the doc comment and the function are removed.

crates/spurctld/src/cluster.rs
```diff
@@ -2928,9 +2914,13 @@ impl ClusterManager {
                 if job.bb_stage_state == BbStageState::Ready {
                     return true;
                 }
+                // In-pass contention counts as a BB wait, same as a shortage.
```
biluriuday, line 2911: BB is still a hand-rolled candidates.retain with three manual record_blocked calls plus one intentional unrecorded drop into bb_stage_candidates. It's the gate with the most drop branches (historically the most bug-prone), and it's the only one with no structural coupling. A future return false without a matching record reintroduces the exact silent-None bug — and the net that would have surfaced it is now deleted. Pick one:

(best) Make the gate return an exhaustive enum Outcome { Keep, Block(PendingReason), Stage } and route BB through the same helper — the compiler then forces every BB branch to name an outcome.

(cheap) Add a debug_assert! after the BB block that every eligible candidate removed there lands in blocked ∪ bb_stage_candidates.

Secondary (pre-existing, non-blocking): the classify → advance_bb_staging_for handoff recomputes capacity under a separate jobs.write() from the read-lock snapshot, so a shrinking-capacity window can leave a stage candidate at None for one cycle. Self-heals next cycle — a transient, not a persistent leak, but another symptom of the hand-rolled BB path.

yansun1996 (dismissed):
my part is good now, PTAL at the comments from @biluriuday

crates/spurctld/src/cluster.rs
shiv-tyagi: Done, part 1 addressed in this PR. Went with the enum route (your "best" option) rather than the debug_assert.

Gate closures now return an exhaustive `GateOutcome { Keep, Block(PendingReason), DropUntagged }`, and the burst-buffer gate goes through the same `retain_eligible` helper as account/QOS/license. A bare `return false` no longer type-checks, so a future BB branch cannot drop a candidate without naming an outcome: `Block` records the reason, and the one sanctioned unrecorded drop (the stage-in handoff) is explicitly `DropUntagged`. Behavior is unchanged; the existing per-gate tests (`in_pass_bb_exhaustion_tags_reason_not_none`, the license single-pass test, the QOS/account in-pass tests, and the BB stage-in tests) stay green.

On naming: I used `DropUntagged` instead of `Stage` so it doesn't collide with the existing BB vocabulary (`BbStageState`, `bb_stage_candidates`) and so the untagged drop is greppable and obvious in review.

Part 2 (the classify -> `advance_bb_staging_for` capacity recompute under a separate write lock, the self-healing one-cycle transient) is pre-existing and non-blocking as you noted, so I'll track it as a separate follow-up rather than expanding this PR.

biluriuday (approved):
LGTM

PR #623, opened 2026-08-12
fix(spurctld): reclaim agent allocations the controller no longer tracks on that node

yansun1996:
## What this fixes

A node agent can keep holding a job's CPU/GPU reservation after the controller has finished with that run. The node then rejects **every** subsequent GPU dispatch with `controller-allocated GPUs unavailable on this node`, while the controller still believes those devices are free and keeps selecting it. Jobs bounce off it until they exhaust `max_batch_requeue` and are held permanently.

The agent has a launch-path self-heal, but it only releases conflicting owners **absent** from its running map — correct, since releasing a job it believes is alive could double-book live work. The controller-side backstop is the heartbeat reclaim: the agent reports what it holds, and the controller re-sends a cancel for anything it considers finished. That backstop had two holes, fixed here.

**1. A job aged out of the controller's job map was never reclaimed.**

```rust
cluster.job_state(job_id).is_some_and(|s| s.is_terminal())
```

`is_some_and` yields `false` for an id the controller has no record of. Since terminal jobs are aged out of the in-memory map to bound memory, a leaked job eventually reads as *unknown* rather than *finished*, and the cancel was never sent. The longer a leak survived, the less likely it became to ever be reclaimed.

Now an untracked id is reclaimable when it is below the next id this controller would assign. Ids at or above that watermark were never issued here — an agent pointed at a reset or different controller — and stay spared.

**2. A job the controller had restarted on a different node was never reclaimed.**

The predicate reasoned only about job *state*, never about *which node* the job belongs to. When a node is evicted mid-run, the controller requeues the job elsewhere and — correctly, for a node it believes is down — sends no cancel. If that node was in fact alive, it keeps running the job and holding its devices. The job is then `Running` (somewhere else), so the reclaim spared it on the original node forever.

Now an **active** job is releasable by any node absent from its allocation. Only an active job has an authoritative nodelist; a job that has not started yet may be mid-dispatch to the reporting node, so it stays spared. That guard is pinned by its own test.

Observed in production: a controller restart caused a large batch of healthy nodes to be marked DOWN on stale heartbeats within one minute of the new leader taking over. Their jobs were requeued elsewhere while the nodes kept running them. Each affected node was then permanently unable to accept GPU work, with the underlying job visible as `RUNNING` on a different node. Left alone, the strand does not clear: nodes were observed still holding devices more than 24 hours later, across a controller restart and a leader election.

The upstream trigger — a newly elected leader evaluating nodes against heartbeat timestamps that no non-leader could refresh — is a separate defect and is not addressed here.

## Relationship to the dispatch-reject cooldown change

This is the companion to the open PR that cools down a node returning a resources-unavailable rejection [1]. They solve different halves and neither subsumes the other:

- **[1] addresses the rate** — the scheduler stops re-selecting a desynced node every tick, so one bad node cannot hot-loop and starve placement cluster-wide.
- **This PR addresses recovery** — the node actually heals, instead of staying poisoned until someone restarts its agent.

[1]'s own description names this as the follow-up: *"reconciling the controller's allocation view against the node remains a separate follow-up."* With only [1], leaked nodes accumulate quietly and never come back. With only this, a node still hot-loops during the window before its next heartbeat. Recommend landing both.

## Known limitations

Deliberately out of scope here, surfaced by review and left for follow-ups:

- **Controller-side phantom allocations are not addressed.** Nodes can hold `alloc_resources` with no corresponding job, which makes them invisible to the scheduler entirely (they look busy, so no dispatch is ever attempted and no error is logged). That needs a reconcile of the controller's allocation view and is a separate, larger change.
- **A cancel can be sent for an id the controller no longer knows.** This is the deliberate trade: the fix acts on absence rather than treating it as a reason to do nothing. It is safe as long as an id below the watermark is never simultaneously absent from the controller and live on an agent. Two ways that can be violated, both accepted for now:
  - *Raft state wiped while agents keep running.* The watermark restarts at `controller.first_job_id` and climbs; once it passes a still-running pre-wipe id, that job becomes reclaimable and is cancelled. The damage is deferred and non-deterministic — it lands whenever submission volume crosses the old id, not at wipe time. **Operationally: after wiping controller state, recycle the agents before resuming submissions.**
  - *Duplicate ids from the `next_job_id` lost-update race.* `apply_operation` does a non-atomic read-modify-write on the counter while the submit path uses `fetch_add`, so it can regress and reissue an id. Pre-existing, worth filing separately. Note the regression direction is safe for this predicate specifically — a lower watermark only spares more ids, it can never manufacture a reclaim.

  An epoch (`run_attempt`) on the cancel would close in-lifetime id reuse and remove the re-check tautology below, but it would **not** close the wipe case: the agent's reported epoch is its live epoch, so echoing it back proves only that the controller's snapshot is current, not that the job should die. Closing that properly needs a controller-incarnation id, which is a larger change and left as a follow-up.

- **Ids below the watermark were not necessarily issued.** Array parent ids are consumed but never stored, as are ids burned when submission fails after the counter bump. These read as reclaimable. Not reachable today because agents only report dispatched task ids; pinned by a test so the behaviour is explicit rather than accidental.

- **A job aged out while `Preempted` becomes reclaimable.** `is_terminal()` deliberately excludes `Preempted` (it may requeue), but eviction filters on `is_finalized()`, which includes it. Once evicted the id cannot be requeued under that id anyway, so the reclaim targets a run that provably ended.

- **The pre-send re-check is inert for untracked ids.** It exists to catch a requeue landing after the snapshot; an absent id can never un-become absent, so it adds nothing on the new branch. Retained because it still guards the terminal-but-present case.

- **Re-sends are unbounded.** A permanently-reclaimable id produces one signal-0 cancel per heartbeat with no dedup or backoff. A healthy agent drops the id on the first cancel.

- **Controller-side phantom allocations are still not addressed** (see above) — that remains the larger reconcile follow-up.

## Testing

- New unit test drives the real eviction WAL operation and asserts the evicted id is reclaimed while ids at and above the watermark are spared. Verified it fails against the unfixed predicate (`left: []`, expected `[20]`) and passes with it.
- New unit test pins the consumed-but-unstored id behaviour described above.
- Existing requeue-race gate retained and still passing under the rename.
- `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` clean; `cargo fmt --all --check` clean; 729 `spurctld` tests pass.
- Not exercised on a live cluster: reproducing it faithfully requires an agent whose local allocation table has desynced from the controller, which is impractical to stage deliberately. The failure mode and the fix are established from the production trace above plus the code path.

### DIFF
--- a/crates/spurctld/src/server.rs
+++ b/crates/spurctld/src/server.rs
@@ -228,8 +228,9 @@ impl ControllerService {
         Err(self.not_leader_status())
     }
 
-    /// Re-send a cancel to a node still reporting a job the controller considers
-    /// finished, freeing an allocation whose cancel never landed. Best-effort.
+    /// Re-send a cancel to a node still holding an allocation the controller no
+    /// longer believes belongs to it there — job finished, job moved to
+    /// another node, or an id older than any this controller issued. Best-effort.
     fn reclaim_stale_agent_jobs(&self, node: &str, reported: &[RunningJobStatus]) {
         let stale = stale_reported_jobs(&self.cluster, node, reported);
         if stale.is_empty() {
@@ -247,7 +248,7 @@ impl ControllerService {
                 warn!(
                     job_id,
                     node = %node,
-                    "agent still holds a job the controller no longer tracks — re-sending cancel to reclaim its allocation"
+                    "agent still holds an allocation the controller no longer believes belongs to it there — re-sending cancel to reclaim it"
                 );
                 // Signal 0 = graceful release, no-op on an unknown id. Not
                 // epoch-gated — a requeue racing this send is still possible.
@@ -407,11 +408,17 @@ fn is_k0s_admin(cache: &crate::association_cache::AssociationCache, caller: &str
 /// Whether `node` may release what it holds for `job_id`: the run is over, the
 /// job is active elsewhere, or the id is untracked but was issued by us.
 fn is_reclaimable(cluster: &ClusterManager, node: &str, job_id: u32) -> bool {
-    match cluster.job_state(job_id) {
-        Some(state) if state.is_terminal() => true,
-        // Only an active job has an authoritative nodelist; anything earlier may
-        // be mid-dispatch to this very node, so spare it.
-        Some(state) => state.is_active() && !cluster.job_holds_node(job_id, node),
+    match cluster.get_job(job_id) {
+        Some(job) if job.state.is_terminal() => true,
+        // An active job's nodelist is authoritative only once populated: state
+        // and allocation commit as two separate WAL entries, so a job can be
+        // observed as Running with `allocated_nodes` still empty. Spare it, same
+        // as a job earlier than active that may be mid-dispatch to this node.
+        Some(job) => {
+            job.state.is_active()
+                && !job.allocated_nodes.is_empty()
+                && !job.allocated_nodes.iter().any(|n| n == node)
+        }
         None => job_id < cluster.peek_next_job_id(),
     }
 }
@@ -3394,6 +3401,10 @@ mod tests {
     use spur_core::reservation::ReservationFlags;
     use tonic::Code;
 
+    fn job_state(cluster: &crate::cluster::ClusterManager, job_id: u32) -> Option<JobState> {
+        cluster.get_job(job_id).map(|j| j.state)
+    }
+
     #[test]
     fn resolve_registration_comm_addr_normalizes_explicit_ip() {
         assert_eq!(
@@ -3699,12 +3710,12 @@ mod tests {
             JobState::Preempted,
         ));
 
-        assert_eq!(cluster.job_state(10), Some(JobState::Pending));
-        assert_eq!(cluster.job_state(11), Some(JobState::Running));
-        assert!(cluster.job_state(12).unwrap().is_terminal());
-        assert_eq!(cluster.job_state(13), Some(JobState::Completing));
-        assert_eq!(cluster.job_state(14), Some(JobState::Suspended));
-        assert_eq!(cluster.job_state(15), Some(JobState::Preempted));
+        assert_eq!(job_state(&cluster, 10), Some(JobState::Pending));
+        assert_eq!(job_state(&cluster, 11), Some(JobState::Running));
+        assert!(job_state(&cluster, 12).unwrap().is_terminal());
+        assert_eq!(job_state(&cluster, 13), Some(JobState::Completing));
+        assert_eq!(job_state(&cluster, 14), Some(JobState::Suspended));
+        assert_eq!(job_state(&cluster, 15), Some(JobState::Preempted));
 
         let reported: Vec<RunningJobStatus> = [10, 11, 12, 13, 14, 15, 999]
             .into_iter()
@@ -3759,7 +3770,7 @@ mod tests {
         ));
         apply(&WalOperation::EvictTerminalJobs { job_ids: vec![20] });
 
-        assert_eq!(cluster.job_state(20), None, "eviction drops the record");
+        assert_eq!(job_state(&cluster, 20), None, "eviction drops the record");
         let issued = cluster.peek_next_job_id();
         assert!(20 < issued, "id 20 was issued by this controller");
 
@@ -3809,7 +3820,7 @@ mod tests {
             );
         }
 
-        assert_eq!(cluster.job_state(30), None, "parent id is never stored");
+        assert_eq!(job_state(&cluster, 30), None, "parent id is never stored");
         assert!(30 < cluster.peek_next_job_id());
         assert!(
             is_reclaimable(&cluster, "n1", 30),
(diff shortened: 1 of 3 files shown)


crates/spurctld/src/server.rs
```diff
@@ -228,10 +228,10 @@ impl ControllerService {
         Err(self.not_leader_status())
     }
 
-    /// Re-send a terminal-job cancel to a node still reporting it on heartbeat,
-    /// freeing an allocation whose terminal cancel never landed. Spawned, best-effort.
+    /// Re-send a cancel to a node still reporting a job the controller considers
+    /// finished, freeing an allocation whose cancel never landed. Best-effort.
```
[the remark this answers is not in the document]
yansun1996: Reworded — the doc comment now covers all three reclaim reasons (finished, moved to another node, untracked id) instead of just "finished".

crates/spurctld/src/server.rs
```diff
@@ -394,18 +394,28 @@ fn is_k0s_admin(cache: &crate::association_cache::AssociationCache, caller: &str
...
-fn is_still_terminal(cluster: &ClusterManager, job_id: u32) -> bool {
-    cluster.job_state(job_id).is_some_and(|s| s.is_terminal())
+/// Whether `node` may release what it holds for `job_id`: the run is over, the
+/// job is active elsewhere, or the id is untracked but was issued by us.
+fn is_reclaimable(cluster: &ClusterManager, node: &str, job_id: u32) -> bool {
+    match cluster.job_state(job_id) {
+        Some(state) if state.is_terminal() => true,
+        // Only an active job has an authoritative nodelist; anything earlier may
+        // be mid-dispatch to this very node, so spare it.
+        Some(state) => state.is_active() && !cluster.job_holds_node(job_id, node),
```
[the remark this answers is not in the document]
yansun1996: Good catch — is_reclaimable now reads state and allocated_nodes from a single job snapshot and treats an active job with an empty nodelist as non-authoritative (spared), so a heartbeat landing between the Running-state commit and the JobStart commit won't cancel a legitimate new run. Added a test that applies the WAL ops in that exact order and asserts the node is spared.

crates/spurctld/src/server.rs
```diff
@@ -241,13 +242,13 @@ impl ControllerService {
...
                     continue;
                 }
                 warn!(
                     job_id,
                     node = %node,
-                    "agent still holds a terminal job — re-sending cancel to reclaim its allocation"
+                    "agent still holds an allocation the controller no longer believes belongs to it there — re-sending cancel to reclaim it"
                 );
                 // Signal 0 = graceful release, no-op on an unknown id. Not
                 // epoch-gated — a requeue racing this send is still possible.
```
shiv-tyagi: The comment says signal 0 is a "graceful release, no-op on an unknown id". That's only true for the finished-job case. The new active-elsewhere branch sends this to a node still running the job, where the agent's `graceful_cancel` does SIGTERM then a delayed SIGKILL. It terminates a live process (the intended heal for a stranded run). Worth updating the comment so this isn't read as release-only.

crates/spurctld/src/server.rs
```diff
@@ -394,18 +395,34 @@ fn is_k0s_admin(cache: &crate::association_cache::AssociationCache, caller: &str
...
-    cluster.job_state(job_id).is_some_and(|s| s.is_terminal())
+/// Whether `node` may release what it holds for `job_id`: the run is over, the
+/// job is active elsewhere, or the id is untracked but was issued by us.
+fn is_reclaimable(cluster: &ClusterManager, node: &str, job_id: u32) -> bool {
+    match cluster.get_job(job_id) {
+        Some(job) if job.state.is_terminal() => true,
+        // An active job's nodelist is authoritative only once populated: state
+        // and allocation commit as two separate WAL entries, so a job can be
+        // observed as Running with `allocated_nodes` still empty. Spare it, same
+        // as a job earlier than active that may be mid-dispatch to this node.
```
shiv-tyagi, line 416: Q: this compares `allocated_nodes` entries against the heartbeat hostname by exact string equality. If those two ever diverge in representation (FQDN vs short name, casing), a node legitimately running the job would not match its own allocation and get a SIGKILL for a live run. Are both guaranteed to be the same canonical node name here?

shiv-tyagi (dismissed):
LGTM. Solid fix and thorough tests. Two non-blocking comments inline (a stale signal-0 comment and a Q on node-name matching).

crates/spurctld/src/server.rs
biluriuday: agree to this concern

Actionable: don't compare against the raw heartbeat hostname. Resolve the reporting node to its canonical node name (the key allocated_nodes uses) before the comparison, or route the comparison through a single helper that owns node identity.

yansun1996: Traced it through — spurd sends the same `hostname` string for both registration and every heartbeat (set once at startup, never re-resolved), and `register_agent` sets NodeName/NodeHostname to that identical value by design (`// NodeName and NodeHostname are the same until agents can supply both`). `allocated_nodes` is populated from that same key, so there's no path today where the two diverge — this is the same identity assumption `update_heartbeat` and job-completion matching already rely on elsewhere in the file. I'd rather not bolt on a normalization helper against an assumption that's baked in everywhere else; happy to revisit if NodeName/NodeHostname are ever allowed to split.

crates/spurctld/src/server.rs
yansun1996: Good catch, updated the comment — signal 0 is only a no-op when the agent has no record of the job; for the active-elsewhere case it's a real SIGTERM/SIGKILL against a live process.

PR #626, opened 2026-08-13
fix(spur-cli): accept slurm-format reservation durations

biluriuday:
Reservation duration was parsed as a raw integer with a silent `unwrap_or(0)` fallback (and `--duration` was a `u32`), so a Slurm-style value like `Duration=30-00:00:00`, or an omitted duration, became 0 minutes. That yields `end_time == start_time`, so the reservation was created but immediately expired and purged by the scheduler within a cycle -- invisible in `scontrol show reservation` and `sinfo` (SPUR-182).

Parse reservation duration through the shared `parse_time_minutes` grammar (whole minutes, HH:MM:SS, D-HH:MM:SS, 90m), matching the `--time` flag on sbatch/srun/salloc; `--duration` becomes a String (backward compatible: `60` still means 60 minutes). Reject a non-positive span in the controller's create/update reservation paths as defense-in-depth for CLI/REST/FFI.

Tested with cargo clippy (no warnings), cargo test (new cluster + CLI unit tests), and a new e2e case covering both entry points, all accepted formats, purge survival, and invalid/zero/omitted rejection.

### DIFF
--- a/crates/spur-cli/src/scontrol.rs
+++ b/crates/spur-cli/src/scontrol.rs
@@ -236,9 +236,9 @@ pub enum ScontrolCommand {
         /// Start time (ISO 8601 or "now")
         #[arg(long, default_value = "now")]
         start_time: String,
-        /// Duration in minutes
+        /// Duration: whole minutes, Slurm time (H:MM, H:MM:SS, D-HH:MM:SS), or suffixed (90m, 1h30m, 30s)
         #[arg(long)]
-        duration: u32,
+        duration: String,
         /// Comma-separated node names
         #[arg(long)]
         nodes: String,
@@ -258,9 +258,9 @@ pub enum ScontrolCommand {
         /// Reservation name
         #[arg(long)]
         name: String,
-        /// New duration in minutes (0 = no change)
+        /// New duration: whole minutes, Slurm time (01:00:00, 30-00:00:00), or suffixed (90m, 1h30m); any zero-length value (e.g. 0, 00:00:00) leaves it unchanged
         #[arg(long, default_value = "0")]
-        duration: u32,
+        duration: String,
         /// Comma-separated nodes to add
         #[arg(long, default_value = "")]
         add_nodes: String,
@@ -488,6 +488,8 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
             users,
             flags,
         } => {
+            crate::privilege::require_privileged("manage reservations")?;
+            let duration = parse_reservation_duration(&duration)?;
             create_reservation(
                 &args.controller,
                 &name,
@@ -512,6 +514,10 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
         } => {
             crate::privilege::require_privileged("manage reservations")?;
 
+            // Any zero-length value (default "0", or an explicit "00:00:00" etc.)
+            // resolves to 0, which the controller treats as "leave duration
+            // unchanged"; there is no zero-length reservation to set.
+            let duration = parse_reservation_duration(&duration)?;
             let channel = spur_client::connect_channel(&args.controller)
                 .await
                 .context("failed to connect to spurctld")?;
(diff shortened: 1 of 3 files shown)


crates/spur-cli/src/scontrol.rs
```diff
@@ -1488,7 +1540,11 @@ async fn create_reservation(
     users: &str,
     flags: &str,
 ) -> Result<()> {
-    crate::privilege::require_privileged("manage reservations")?;
+    // Callers gate privilege before parsing so format/privilege errors are
+    // reported in a consistent order; this stays as the last input guard.
+    if duration == 0 {
```
shiv-tyagi: Nit: this leans into narrating the change rationale ("reported in a consistent order") rather than a constraint the code needs. The "why privilege isn't here" part is the useful bit.

```suggestion
    // Privilege is gated by callers; this is the last input guard.
    if duration == 0 {
```

crates/spur-cli/src/scontrol.rs
```diff
@@ -1476,6 +1516,18 @@ async fn reconfigure(controller: &str) -> Result<()> {
...
+/// Parse a reservation duration into minutes via the shared `--time` grammar
+/// (whole minutes, `HH:MM:SS`, `D-HH:MM:SS`, `90m`); rejects INFINITE. Quirk of
+/// that grammar: a bare `MM:SS` is read as `HH:MM` and bare `days-hours` is
+/// rejected, so prefer the unambiguous colon forms.
+fn parse_reservation_duration(s: &str) -> Result<u32> {
+    spur_core::config::parse_time_minutes(s).ok_or_else(|| {
+        anyhow::anyhow!(
+            "invalid reservation duration '{s}'; use whole minutes, Slurm time (01:00:00, 30-00:00:00), or suffixed (90m, 1h30m)"
+        )
+    })
```
shiv-tyagi, line 1528: Q: Is rejecting `Duration=UNLIMITED`/`INFINITE` intentional? Slurm accepts unbounded reservation durations, and `parse_time_minutes` returns `None` for those, so they land in this error path. Not a regression (the old `u32` didn't support it either), so fine to punt, but worth a follow-up or a note in the help text if unbounded reservations are expected to work.

tests/native_host/e2e/test_reservations.py
```diff
@@ -358,6 +358,134 @@ def test_reservation_management_requires_privileged_user(self, cluster):
...
+            deadline = time.time() + window
+            seen = False
+            while time.time() < deadline:
+                state = classify(name)
+                if state == "present":
+                    seen = True
+                elif state == "absent":
+                    raise AssertionError(f"reservation {name} was purged unexpectedly")
+                time.sleep(step)
+            assert seen, f"reservation {name} was never observed present within {window}s"
```
shiv-tyagi, line 416: Q: Any deterministic hook to force a purge cycle instead of polling a fixed 5s window against the ~1s purge loop? The check is defensively written (tolerates transient errors, requires one real sighting), so it won't pass vacuously, but the window is the only thing separating "survived" from "we stopped looking too early" under a loaded runner. Acceptable for e2e, just flagging the timing dependence.

shiv-tyagi (commented):
Solid fix. Core approach is right: route duration through the shared `parse_time_minutes` grammar and reject non-positive spans at both CLI and controller layers, with the server computing `end_time == start_time` for a zero duration so the controller guard covers REST/FFI. Privilege gating moved cleanly to both call sites (no bypass). A few nits and questions inline, nothing blocking.

crates/spur-cli/src/scontrol.rs
biluriuday: On unbounded reservations (Duration=UNLIMITED/INFINITE):

Rejecting these is intentional, not an oversight. Spur models a reservation as a concrete [start_time, end_time] interval — Reservation.end_time is a real DateTime<Utc>, and the whole expiry/purge path is built on now >= end_time. There's no sentinel for "never ends", so parse_time_minutes returning None for UNLIMITED/INFINITE correctly lands in the error path rather than being coerced into some arbitrary end time.

As you noted, this isn't a regression — the old u32 --duration couldn't express it either. Supporting it properly is a real feature: it needs an optional/sentinel end_time on Reservation (a persisted Raft WAL + snapshot type, so a serialization-compat change), a proto signal for "unbounded", and updates to is_expired/state_label/purge_expired_reservations to skip such reservations. That's out of scope for this bug fix.

To avoid surprising users, this change now makes the rejection explicit instead of a bare parse error:

The error message now ends with ; UNLIMITED/INFINITE not supported.
Both create-reservation and update-reservation --duration help text call out UNLIMITED/INFINITE not supported.
Added a unit test (parse_reservation_duration_rejects_unbounded) to lock the behavior.
If unbounded reservations are wanted, we'll track it as a separate follow-up.

tests/native_host/e2e/test_reservations.py
biluriuday: - The timing dependence is inherent to e2e integration coverage here; there's no clean deterministic trigger without adding a test-only RPC to production surface, which isn't worth it.
- The deterministic guarantee actually lives in the unit test create_reservation_rejects_non_positive_span, which proves the guard rejects a zero-length span with zero timing dependence. The e2e assert_survives_purge is complementary: it exercises the real scheduler and, with scheduler.interval_secs=1 in the e2e config, the 5s window spans ~5 purge iterations — comfortable margin, and the check is written to fail (not pass vacuously) if the reservation is ever observed absent or never observed present.
- A regression (bad duration → zero span → purged) would be caught: the reservation would flip to absent within ~1s, well inside the window.

crates/spur-cli/src/scontrol.rs
biluriuday: done

shiv-tyagi (approved):
LGTM

PR #629, opened 2026-08-13
feat(spurctld): implement scontrol requeue and requeuehold

biluriuday:
Fixes [SPUR-159](https://amd.atlassian.net/browse/SPUR-159).

### What this fixes
`scontrol requeue $JOBID` was a cancel-only stub — it killed the job and left it gone, despite printing "requeued". It now returns a job to `PENDING` with the same `job_id` and spec, matching Slurm. Adds `scontrol requeuehold`, which re-pends the job into a held state.

### Scope
- Works on running/suspended jobs (kill + re-pend) and on any terminal job (completed, failed, timeout, node_fail, cancelled, deadline, OOM).
- Manual requeue is owner-or-root, uncapped (tracked via a separate `user_requeue_count` that never trips `max_batch_requeue`), rejects interactive/`srun` allocations and already-pending jobs.
- A bare array-parent id fans out to every task.

### Approach
A new additive `RequeueJob` RPC drives a `JobUserRequeue` WAL op that reuses the atomic machinery behind `JobPreemptRequeue`. A live job is finalized exactly once — routed through a new first-class `JobState::Requeued` so steps/accounting see a finished run, allocations and GPUs are freed, epilog fires — then re-pended. An already-terminal job re-pends with **no** second finalization, avoiding double-counted accounting/epilog. The server kills the old processes via `send_cancel_to_agents`, and a leader-computed `begin_time` hold on the live path prevents the scheduler from re-dispatching a job into its own in-flight cancel (same guard as preemption requeue).

### Compatibility
All changes are additive (new RPC/message, appended `JOB_REQUEUED` enum value, new WAL variant and `Job` field with serde defaults, widened state transitions), so a forward rolling upgrade is safe. Note: like any new WAL op, a `scontrol requeue` issued mid-upgrade can't be applied by a not-yet-upgraded follower — finish the rolling upgrade before using it.

### Known limitations / follow-ups
- Accounting keeps one row per `job_id` (the `REQUEUED` end record is overwritten by the next run), same as the existing preempt-requeue path — Slurm keeps per-run history. Follow-up: expose restart count in `scontrol show job` and consider per-run accounting rows.
- Requeuing a completed job does not re-trigger `afterok` dependents that already ran (same behavior as Slurm).

### Testing
`cargo clippy --workspace --exclude spur-ffi --all-targets` and `cargo test --locked` pass. Added unit tests (running/suspended/all terminal states/array fan-out/hold/authz/non-batch/replay-idempotency/batch-cap guard/live-vs-terminal finalization), a WAL round-trip test, and an e2e suite (`tests/native_host/e2e/test_requeue.py`).

### DIFF
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -1185,6 +1185,135 @@ impl ClusterManager {
         Ok(())
     }
 
+    /// Admin requeue (`scontrol requeue` / `requeuehold`), keeping the `job_id`.
+    /// A bare array-parent id (no job record of its own) fans out to every task.
+    /// Returns snapshots of the jobs that were Running/Suspended so the caller
+    /// can kill their old processes.
+    pub fn requeue_job_by_user(
+        &self,
+        job_id: JobId,
+        user: &str,
+        hold: bool,
+    ) -> anyhow::Result<Vec<Job>> {
+        let (targets, is_array) = {
+            let jobs = self.jobs.read();
+            if jobs.contains_key(&job_id) {
+                (vec![job_id], false)
+            } else {
+                let mut tasks: Vec<JobId> = jobs
+                    .values()
+                    .filter(|j| j.spec.array_job_id == Some(job_id))
+                    .map(|j| j.job_id)
+                    .collect();
+                tasks.sort_unstable();
+                (tasks, true)
+            }
+        };
+        if targets.is_empty() {
+            anyhow::bail!("job {} not found", job_id);
+        }
+
+        // A single, explicitly named job propagates its error verbatim so the
+        // user sees exactly why (not owner, interactive, already pending, ...).
+        if !is_array {
+            return Ok(self
+                .requeue_one_job(targets[0], user, hold)?
+                .into_iter()
+                .collect());
+        }
+
+        // Array fan-out: requeue every task that can be, collecting per-task
+        // failures. Only fail outright if no task could be requeued at all.
+        let total = targets.len();
+        let mut killed = Vec::new();
+        let mut errors = Vec::new();
+        for id in targets {
+            match self.requeue_one_job(id, user, hold) {
+                Ok(Some(job)) => killed.push(job),
+                Ok(None) => {}
+                Err(e) => errors.push(format!("task {id}: {e}")),
+            }
+        }
+        if killed.is_empty() && errors.len() == total {
+            anyhow::bail!(
+                "no tasks of array {} could be requeued: {}",
+                job_id,
+                errors.join("; ")
+            );
+        }
+        // Partial success is reported as success (some tasks requeued), so make
+        // the skipped tasks visible rather than swallowing them silently.
+        if !errors.is_empty() {
+            warn!(
+                array_job_id = job_id,
+                "some array tasks were not requeued: {}",
+                errors.join("; ")
+            );
+        }
+        Ok(killed)
+    }
+
+    /// Requeue exactly one job/task. Returns its pre-requeue snapshot when the
+    /// job was Running/Suspended (so its processes must be killed on the nodes).
+    fn requeue_one_job(
+        &self,
+        job_id: JobId,
+        user: &str,
+        hold: bool,
+    ) -> anyhow::Result<Option<Job>> {
+        let snapshot = {
+            let jobs = self.jobs.read();
+            let job = jobs
+                .get(&job_id)
+                .ok_or_else(|| anyhow::anyhow!("job {} not found", job_id))?;
+            Self::check_job_owner(user, &job.spec.user, "requeue")?;
+            // Interactive/srun allocations have no batch script to re-run.
+            if job.spec.interactive || job.spec.srun_job || job.spec.pty {
+                anyhow::bail!("job {} is interactive and cannot be requeued", job_id);
+            }
+            match job.state {
+                JobState::Pending => anyhow::bail!("job {} is already pending", job_id),
+                JobState::Completing => {
+                    anyhow::bail!("job {} is completing; cannot requeue", job_id)
+                }
+                JobState::Preempted | JobState::Requeued => {
+                    anyhow::bail!("job {} is already being requeued", job_id)
+                }
+                _ => {}
+            }
+            matches!(job.state, JobState::Running | JobState::Suspended).then(|| job.clone())
+        };
+
+        // Defer a live, non-held requeue's eligibility so the scheduler can't
+        // re-dispatch it into its own in-flight cancel (same guard as preemption
+        // requeue). Held/terminal requeues can't hit that race. Computed on the
+        // leader so every replica applies one instant.
+        let begin_time = match (&snapshot, hold) {
+            (Some(job), false) => {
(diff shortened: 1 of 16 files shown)


tests/native_host/e2e/test_requeue.py
```diff
@@ -0,0 +1,178 @@
...
+            body = "#!/bin/bash\necho STARTED\nsleep 600\n"
+            script = cluster.write_file("rq-array.sh", body)
+            parent = parse_job_id(
+                cluster.sbatch(["-J", "rq-array", "-N", "1", "-a", "0-1", script])
+            )
+            assert parent is not None
+            # At least one task should reach RUNNING before we requeue the array.
+            assert _wait_state(cluster, parent, "R", timeout=90) or _wait_state(
+                cluster, parent, "PD", timeout=1
+            ), "no array task became active"
```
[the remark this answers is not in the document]
biluriuday: addressed the issue

crates/spurctld/src/cluster.rs
```diff
@@ -4531,6 +4668,109 @@ impl ClusterManager {
...
+                            node.update_state_from_alloc();
+                            if node.state == NodeState::Draining
+                                && node.alloc_resources.cpus == 0
+                                && !node.alloc_resources.has_devices()
+                            {
+                                node.state = NodeState::Drain;
+                            }
+                        }
+                    }
+                }
```
shiv-tyagi: This double-frees node resources for a terminal requeue. The comment is wrong: `JobComplete` (L4988-5027) frees the node slice but never nulls `job.allocated_resources`/`allocated_nodes`, so a Completed/Failed/etc. job still carries `allocated_resources = Some(...)`. On the terminal path (`was_live = false`) that value is cloned at L4711 before `clear_run_state_for_requeue` nulls it, so this loop runs and subtracts the slice a second time.

Tests pass only because `subtract` is `saturating_sub` and they use one job per node (already at 0). On a node still running other jobs, the second subtract decrements below the true allocation, under-counting and oversubscribing on the next schedule.

Gate the dealloc on `was_live` (terminal jobs already released their nodes at completion), and add a regression test: requeue a terminal job sharing a node with a live job, assert node alloc is unchanged.

crates/spurctld/src/cluster.rs
```diff
@@ -1147,6 +1147,135 @@ impl ClusterManager {
...
+        // requeue). Held/terminal requeues can't hit that race. Computed on the
+        // leader so every replica applies one instant.
+        let begin_time = match (&snapshot, hold) {
+            (Some(job), false) => {
+                let hold_secs = (self.scheduler_interval_secs as i64 * 2 + 3).max(5);
+                let deferred = Utc::now() + chrono::Duration::seconds(hold_secs);
+                Some(
+                    job.spec
+                        .begin_time
+                        .map_or(deferred, |user| user.max(deferred)),
```
shiv-tyagi: Nit: `user` shadows the `user: &str` auth arg and actually binds `spec.begin_time` (a DateTime). The preempt path calls it `user_begin`.

```suggestion
                        .map_or(deferred, |user_begin| user_begin.max(deferred)),
```

crates/spurctld/src/cluster.rs
```diff
@@ -1147,6 +1147,135 @@ impl ClusterManager {
...
+        // Partial success is reported as success (some tasks requeued), so make
+        // the skipped tasks visible rather than swallowing them silently.
+        if !errors.is_empty() {
+            warn!(
+                array_job_id = job_id,
+                "some array tasks were not requeued: {}",
+                errors.join("; ")
+            );
+        }
+        Ok(killed)
```
shiv-tyagi: Partial array fan-out is reported to the caller as success, so the CLI prints "job X requeued" while skipped tasks only land in this server-side warn. The operator gets no signal the fan-out was partial. Fully fixing this needs the RPC to return a skipped count (it returns `()` today), so fair as a scoped follow-up, but worth surfacing rather than logging only.

shiv-tyagi (commented):
Solid PR. The core design is right: a single atomic `JobUserRequeue` WAL op that finalizes the run, frees allocations, and re-pends in one apply, reusing the `JobPreemptRequeue` machinery. That avoids stranding a job in an intermediate state across a leadership change, and the changes are additive (enum tail, appended proto field/RPC, WAL variant with serde defaults). Test coverage is thorough.

One blocking correctness issue (terminal requeue double-frees node resources), plus a few smaller items and refactoring suggestions below.

### Design notes (not blocking)

- The live and terminal paths share one apply arm gated on `was_live`. They have different resource ownership (a live job still holds allocations; a terminal one released them at completion), and conflating them is what produced the double-free. Consider splitting them, or better, fixing the root cause: `JobComplete` frees the node slice but leaves `job.allocated_resources`/`allocated_nodes` populated, so every requeue path has to defensively reason about stale allocations.
- The transition table now allows `Completed/Cancelled/Deadline/OutOfMemory -> Pending` for every caller of `transition`, not just requeue. That removes the state-machine guard that a finished job can't be resurrected. A dedicated requeue transition (or a guarded re-pend only the requeue op can call) keeps that invariant.
- `JobState::Requeued` never escapes the write lock, so `squeue`/`scontrol` never show `RQ` and `spur_jobs_requeued` is always 0. Fine (same as `Preempted`), just noting the display/metrics surface is inert. Its one real use is the `JobFinalized` end-record.

### Refactoring worth doing

- Unify the `JobPreemptRequeue` and `JobUserRequeue` apply arms (~80 near-identical lines each). A shared `finalize_run_and_repend` plus a requeue-kind param (counter + hold vs begin_time) would collapse them and give one tested dealloc path.
- Extract the node-dealloc loop, now copied in four places (`evict_job_locked`, `JobComplete`, `JobPreemptRequeue`, `JobUserRequeue`). evict/JobComplete carry an `already_deallocated` skip the requeue pair don't.
- Extract array-parent fan-out (`array_job_id == Some(id)` at L926, L1167, L2920, L3266) into one helper.

Nit: `requeuehold` re-pends into PENDING+held rather than Slurm's distinct REQUEUE_HOLD (`RH`) state. Functionally fine; a script keying on `squeue -t RH` wouldn't catch it.

crates/spurctld/src/cluster.rs
biluriuday: addressed this gap

crates/spurctld/src/cluster.rs
biluriuday: applied suggestion

crates/spurctld/src/cluster.rs
biluriuday: addressed the issue

shiv-tyagi (dismissed):
LGTM. The double-free on terminal requeue is fixed and covered by a shared-node regression test, the array fan-out now surfaces requeued/skipped counts through to the CLI, and the dealloc loop is deduped into one helper. clippy and the full test suite pass locally and in CI (including both e2e jobs).

Non-blocking follow-up: the transition table now allows Completed/Cancelled/Deadline/OutOfMemory -> Pending for every transition caller, not just requeue. Consider a guarded requeue-only transition later so the state machine keeps that invariant.

shiv-tyagi (approved):
LGTM

PR #630, opened 2026-08-13
test(spur-cli): cover the hostlist first-host path srun depends on

nikhilsk:
This PR originally carried the `srun`/`sattach` compressed-nodelist fix for #575. That fix landed in #589, so everything overlapping it is dropped, including this branch's `hostlist` rewrite, which covered the same ground as the `parse_range_bounds` refactor that merged with it. What is left is the coverage that #589 did not bring, and the branch is rebased onto main so it applies against `expand_first` rather than this branch's old helper.

## What this adds

**A test through the `srun` call path.** Main tests `nodelist::first_allocated_node` directly, but nothing asserts that `try_stream_output` actually calls it. Reintroducing a comma split at that call site would pass every test in the tree while streaming to `node[001-002]`, which is not a host and is exactly the original bug. This drives `try_stream_output` against the mock controller and asserts the name it asks `GetNode` about is `node001`.

That needs the mock to record the request, so `GetNode` moves out of its unimplemented list into a handler that records the name and returns `NotFound`. Returning `NotFound` matters: the caller treats a failed lookup as "node unreachable" and stops, so the test never dials an agent that is not running.

**T51 coverage for `expand_first`.** The tier suite covers `expand`, `compress` and `count` but stopped before the first-host path. The new cases cover ranges, a comma inside brackets, multiple prefixes, plain lists, a single host, the empty pattern, invalid patterns, and agreement with full expansion, which is the property that keeps the two functions from drifting.

## Testing

The srun test is mutation-checked rather than assumed: swapping `first_allocated_node` back to `nodelist.split(',').next()` fails it on the requested name, `left: ["node[001-002]"]` against `right: ["node001"]`. `cargo fmt --check` and `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` are clean, and `cargo test --locked` is green.

No production code changes here, so there is no upgrade or compatibility surface.

### DIFF
--- a/crates/spur-cli/src/mock_controller.rs
+++ b/crates/spur-cli/src/mock_controller.rs
@@ -8,9 +8,9 @@
 //! ephemeral localhost port, serve a hand-written service on it, and hand the
 //! caller back the address plus a shared record of what the server observed.
 //! Only a handful of RPCs are implemented (`CreateJobStep`, `RunStep`,
-//! `GetNodes`, `UpdateNode`, `DrainNode`, `DeregisterNode`); every other RPC
-//! reports `unimplemented` so an unexpected call fails loudly instead of
-//! silently returning a default.
+//! `GetNode`, `GetNodes`, `UpdateNode`, `DrainNode`, `DeregisterNode`); every
+//! other RPC reports `unimplemented` so an unexpected call fails loudly instead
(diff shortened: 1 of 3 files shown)


shiv-tyagi (approved):
LGTM

PR #635, opened 2026-08-15
fix(spurctld): fsync raft state so a crash cannot roll it back

drmingdrmer:
## Problem

`SpurStore` in `crates/spurctld/src/raft.rs` persists every piece of raft state with `std::fs::write`:

```rust
fn persist_vote(&self, vote: &Vote<NodeId>) -> Result<(), std::io::Error> {
    let path = self.raft_dir.join("vote.json");
    let data = serde_json::to_vec(vote)?;
    std::fs::write(&path, &data)   // truncate, write, close
}
```

`std::fs::write` truncates the target first and returns once the bytes are in the OS page cache. Openraft requires the opposite of both — [`RaftStorage::save_vote`](https://docs.rs/openraft/0.9/openraft/storage/trait.RaftStorage.html#tymethod.save_vote) is documented as *"The vote must be persisted on disk before returning"*, and an append must be on disk before it reports success, because the leader commits on the strength of that report.

Three consequences, none of them visible on a graceful shutdown:

**1. A crash mid-write bricks the node.** `std::fs::write` truncates before writing, so a crash in between leaves a half-length JSON file. `SpurStore::new` in the default strict mode then does:

```
anyhow::bail!("failed to parse vote.json: {e}")
```

and `spurctld` will not start. `--allow-partial-wal-recovery` gets it running by dropping the record, which is worse — see (2). `persist_last_purged` already used write-then-rename with the comment *"a crash mid-write must never corrupt purged.json"*; `persist_vote` and `persist_snapshot` did not, though the same reasoning applies to them.

**2. A power loss can roll the vote back.** `save_vote` returns `Ok` with the vote only in the page cache. After the crash the node reads back an older vote and grants its vote a second time in the same term — two candidates can each collect a majority and both become leader for that term.

**3. New files can be missing entirely.** A file's name lives in its parent directory. Flushing the file's contents does not flush that directory entry, so a newly created `log/00000000000000000042.json` can be absent after a crash even though its bytes were written.

## Fix

One helper, used by all four persist paths:

```rust
fn atomic_write_json<T: Serialize>(path: &Path, value: &T) -> Result<(), std::io::Error> {
    // ... write to `<path>.tmp`
    file.write_all(&data)?;
    file.sync_all()?;      // contents durable
    drop(file);
    std::fs::rename(&tmp, path)   // atomic swap
}
```

- **Single-file state** (`vote.json`, `purged.json`, `snapshot.json`) — `atomic_write_json` + `sync_dir(raft_dir)`.
- **Log entry files** — `persist_log_entry` writes contents only; `append_to_log`, `delete_conflict_logs_since` and `purge_logs_upto` each call `sync_log_dir()` **once per call** rather than once per entry, so the directory flush is amortised over the batch.

`sync_dir` is `#[cfg(unix)]`; Windows has no directory-fsync equivalent, so it is a no-op there.

The two delete paths need the directory flush for the same reason as the appends: if a crash loses a truncation, the conflicting tail reappears on disk and `get_log_state` reports a `last_log_id` this node never accepted from the current leader.

## Cost, and what this PR deliberately does not change

This makes an existing cost visible rather than creating it. `persist_log_entry` writes **one file per raft log entry**, so an append of *K* entries is already *K* file creations; with this PR it is *K* fsyncs plus one for the directory.

The fix for that is a single append-only log file (framed records) instead of a file per entry, which would bring an append down to one fsync per batch. That is a much larger change and belongs in its own PR — correctness first. If throughput turns out to matter, the other lever is openraft's storage-v2 `append`, whose completion callback may fire *after* the method returns, allowing several batches to share one fsync. `SpurStore` is currently on the v1 `RaftStorage` path via `Adaptor`, which has no such callback.

## Verification

```
cargo check -p spurctld
cargo test  -p spurctld --bins raft::     # 47 passed, 0 failed
```

The existing recovery tests (`store_corrupt_vote_file_hard_fails_by_default`, `store_persists_vote`, `store_persists_snapshot`, `store_persists_last_purged_across_restart`, `store_removes_log_entries`, …) all still pass. Durability itself cannot be asserted from inside the process — a test would have to kill the machine, not the process — so the change is instead scoped so that every write goes through one helper.

---

Found while reviewing how projects use openraft (I maintain [openraft](https://github.com/databendlabs/openraft)). Unrelated to this PR, two things in `raft.rs` were the best treatment of their kind I found across the projects I looked at: `append_entries` returning `RPCError::PayloadTooLarge` with `new_entries_hint` (nobody else uses that feedback path — everyone else just lets the oversized append fail), and the `4 * RAFT_SNAPSHOT_MAX_CHUNK_SIZE < RAFT_MAX_MESSAGE_SIZE` invariant with tests on both sides of it.

### DIFF
--- a/crates/spurctld/src/raft.rs
+++ b/crates/spurctld/src/raft.rs
@@ -280,9 +280,8 @@ impl SpurStore {
         self.raft_dir.join("log")
     }
 
-    /// Flush the log directory so that entry files created or removed since the
-    /// last call survive a crash. Called once per append/truncate/purge rather
-    /// than once per entry file.
+    /// Flushes the log dir once per append/truncate/purge rather than once per
+    /// entry file.
     // The error type is openraft's, and every caller is a `RaftStorage` method
     // that must return it anyway; boxing it here would only unbox at the call.
     #[allow(clippy::result_large_err)]
@@ -473,9 +472,7 @@ impl openraft::RaftStorage<SpurTypeConfig> for Arc<SpurStore> {
             })?;
             inner.log.insert(entry.log_id.index, entry);
         }
-        // One directory flush for the whole batch: each entry's contents are
-        // already on disk, but a newly created file is not durable until the
-        // directory holding its name is flushed too.
+        // A newly created file is not durable until its directory is flushed.
         self.sync_log_dir()?;
         Ok(())
     }
@@ -490,9 +487,7 @@ impl openraft::RaftStorage<SpurTypeConfig> for Arc<SpurStore> {
             inner.log.remove(&k);
             self.remove_log_entry(k);
         }
-        // The deletions must be durable before this returns: a crash that loses
-        // them resurrects the conflicting tail, and `get_log_state` would then
-        // report a last_log_id this node never accepted from the leader.
+        // A crash that loses these deletions resurrects the conflicting tail.
         self.sync_log_dir()?;
         Ok(())
     }
--- a/crates/spurctld/src/raft.rs
+++ b/crates/spurctld/src/raft.rs
@@ -9,7 +9,7 @@
 //! `ClusterManager::propose()` and applied through `StateMachineApply`.
 
 use std::collections::BTreeMap;
-use std::io::Cursor;
+use std::io::{Cursor, Write};
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
@@ -275,42 +275,40 @@ impl SpurStore {
         })
     }
 
+    fn log_dir(&self) -> PathBuf {
+        self.raft_dir.join("log")
+    }
+
+    /// Flush the log directory so that entry files created or removed since the
+    /// last call survive a crash. Called once per append/truncate/purge rather
+    /// than once per entry file.
+    fn sync_log_dir(&self) -> Result<(), StorageError<NodeId>> {
+        sync_dir(&self.log_dir()).map_err(|e| {
+            StorageError::from_io_error(openraft::ErrorSubject::Logs, openraft::ErrorVerb::Write, e)
+        })
+    }
+
     fn persist_last_purged(&self, log_id: &LogId<NodeId>) -> Result<(), std::io::Error> {
-        let path = self.raft_dir.join("purged.json");
-        let tmp = self.raft_dir.join("purged.json.tmp");
-        let data = serde_json::to_vec(log_id)
-            .map_err(|e| std::io::Error::new(std::io::ErrorKind::InvalidData, e))?;
-        // Write-then-rename: a crash mid-write must never corrupt purged.json.
-        std::fs::write(&tmp, &data)
-            .map_err(|e| std::io::Error::new(e.kind(), format!("{tmp:?}: {e}")))?;
-        std::fs::rename(&tmp, &path)
-            .map_err(|e| std::io::Error::new(e.kind(), format!("{path:?}: {e}")))
+        atomic_write_json(&self.raft_dir.join("purged.json"), log_id)?;
+        sync_dir(&self.raft_dir)
     }
 
     fn persist_vote(&self, vote: &Vote<NodeId>) -> Result<(), std::io::Error> {
-        let path = self.raft_dir.join("vote.json");
-        let data = serde_json::to_vec(vote)
-            .map_err(|e| std::io::Error::new(std::io::ErrorKind::InvalidData, e))?;
-        std::fs::write(&path, &data)
-            .map_err(|e| std::io::Error::new(e.kind(), format!("{path:?}: {e}")))
+        atomic_write_json(&self.raft_dir.join("vote.json"), vote)?;
+        sync_dir(&self.raft_dir)
     }
 
     fn persist_log_entry(&self, entry: &Entry<SpurTypeConfig>) -> Result<(), std::io::Error> {
+        // Contents only: the caller flushes the log directory once per append
+        // batch rather than once per entry.
         let path = self
-            .raft_dir
-            .join("log")
+            .log_dir()
             .join(format!("{:020}.json", entry.log_id.index));
-        let data = serde_json::to_vec(entry)
-            .map_err(|e| std::io::Error::new(std::io::ErrorKind::InvalidData, e))?;
-        std::fs::write(&path, &data)
-            .map_err(|e| std::io::Error::new(e.kind(), format!("{path:?}: {e}")))
+        atomic_write_json(&path, entry)
     }
 
     fn remove_log_entry(&self, index: u64) {
-        let path = self
-            .raft_dir
-            .join("log")
-            .join(format!("{:020}.json", index));
+        let path = self.log_dir().join(format!("{:020}.json", index));
         let _ = std::fs::remove_file(&path);
     }
 
@@ -323,11 +321,8 @@ impl SpurStore {
             meta: meta.clone(),
             data: data.to_vec(),
         };
-        let path = self.raft_dir.join("snapshot.json");
-        let json = serde_json::to_vec(&ps)
-            .map_err(|e| std::io::Error::new(std::io::ErrorKind::InvalidData, e))?;
-        std::fs::write(&path, &json)
-            .map_err(|e| std::io::Error::new(e.kind(), format!("{path:?}: {e}")))
+        atomic_write_json(&self.raft_dir.join("snapshot.json"), &ps)?;
+        sync_dir(&self.raft_dir)
     }
 
     fn load_persisted_snapshot(&self) -> Option<PersistedSnapshot> {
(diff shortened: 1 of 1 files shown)


crates/spurctld/src/raft.rs
```diff
@@ -341,6 +336,54 @@ impl SpurStore {
...
+        .and_then(|f| f.sync_all())
+        .map_err(|e| std::io::Error::new(e.kind(), format!("{dir:?}: {e}")))
+}
+
+/// Windows cannot fsync a directory handle; `MoveFileEx` is ordered enough there
+/// once the file's own contents have been flushed.
+#[cfg(not(unix))]
+fn sync_dir(_dir: &Path) -> Result<(), std::io::Error> {
+    Ok(())
+}
```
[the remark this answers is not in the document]
drmingdrmer: `std::fs::rename` replaces an existing destination on Windows: std maps it to `MoveFileExW` with `MOVEFILE_REPLACE_EXISTING`, or to `SetFileInformationByHandle` with `FILE_RENAME_FLAG_REPLACE_IF_EXISTS` on filesystems that support `FileRenameInfoEx`. Repeated persists of `vote.json`/`purged.json`/`snapshot.json` overwrite the previous file on every platform.

crates/spurctld/src/raft.rs
```diff
@@ -275,42 +275,40 @@ impl SpurStore {
         })
     }
 
+    fn log_dir(&self) -> PathBuf {
+        self.raft_dir.join("log")
+    }
+
+    /// Flush the log directory so that entry files created or removed since the
+    /// last call survive a crash. Called once per append/truncate/purge rather
+    /// than once per entry file.
```
yansun1996: Nit: this is a 3-line doc comment; repo style caps these at ≤2 lines. Could trim to something like `// Flushes the log dir once per batch rather than once per entry file.`

crates/spurctld/src/raft.rs
```diff
@@ -341,6 +336,54 @@ impl SpurStore {
...
+/// returns.
+///
+/// Both halves are required, and `std::fs::write` provides neither. It truncates
+/// the target first, so a crash mid-write leaves a half-length JSON file that
+/// `SpurStore::new` then refuses to parse — the node will not start at all in
+/// the default strict recovery mode. And it returns once the bytes are in the
+/// page cache, which a power loss or kernel panic discards even though openraft
+/// was told the write succeeded: `RaftStorage::save_vote` is documented as "the
+/// vote must be persisted on disk before returning", and losing a vote lets this
+/// node grant its vote twice in the same term.
```
yansun1996: This doc comment runs 12 lines. The mechanism is clear from the code itself -- could this collapse to 1-2 lines on the "why" (crash-safety + openraft's vote-durability requirement), with the rest moved to the PR description?

crates/spurctld/src/raft.rs
```diff
@@ -341,6 +336,54 @@ impl SpurStore {
...
+    let mut file = std::fs::File::create(&tmp)
+        .map_err(|e| std::io::Error::new(e.kind(), format!("{tmp:?}: {e}")))?;
+    file.write_all(&data)
+        .map_err(|e| std::io::Error::new(e.kind(), format!("{tmp:?}: {e}")))?;
+    file.sync_all()
+        .map_err(|e| std::io::Error::new(e.kind(), format!("{tmp:?}: {e}")))?;
+    drop(file);
+
+    std::fs::rename(&tmp, path)
+        .map_err(|e| std::io::Error::new(e.kind(), format!("{path:?}: {e}")))
```
yansun1996: If `sync_all()` or this `rename` fails, `tmp` is never cleaned up. Low severity today -- `SpurStore::new`'s recovery scan filters strictly on `.json` extension so a leftover `.json.tmp` is ignored, and the next successful write to this path truncates it via `File::create` -- but under a sustained disk-full condition these could pile up one per failed index. A `let _ = std::fs::remove_file(&tmp);` on the error paths would keep things tidy.

crates/spurctld/src/raft.rs
```diff
@@ -341,6 +336,54 @@ impl SpurStore {
...
+}
+
+/// Flush a directory so that files created or renamed inside it survive a crash.
+///
+/// Flushing a file's contents is not enough for a file that was just created or
+/// renamed into place: its directory entry lives in the parent directory, and
+/// until that is flushed the file can be missing entirely after a power loss.
```
yansun1996: Nit: 5-line doc comment, could trim to 1-2 lines per repo comment style.

crates/spurctld/src/raft.rs
```diff
@@ -341,6 +336,54 @@ impl SpurStore {
...
+#[cfg(unix)]
+fn sync_dir(dir: &Path) -> Result<(), std::io::Error> {
+    std::fs::File::open(dir)
+        .and_then(|f| f.sync_all())
+        .map_err(|e| std::io::Error::new(e.kind(), format!("{dir:?}: {e}")))
+}
+
+/// Windows cannot fsync a directory handle; `MoveFileEx` is ordered enough there
+/// once the file's own contents have been flushed.
```
yansun1996: Minor: this states "`MoveFileEx` is ordered enough" as a settled durability guarantee, but it's not cited and Windows NTFS metadata-ordering semantics without an explicit directory flush are genuinely murky. Since spurctld doesn't ship on Windows in practice, might be worth softening to "best-effort; Windows is not a supported deployment target" rather than asserting a guarantee.

crates/spurctld/src/raft.rs
```diff
@@ -434,6 +477,10 @@ impl openraft::RaftStorage<SpurTypeConfig> for Arc<SpurStore> {
             })?;
             inner.log.insert(entry.log_id.index, entry);
         }
+        // One directory flush for the whole batch: each entry's contents are
+        // already on disk, but a newly created file is not durable until the
+        // directory holding its name is flushed too.
```
yansun1996: Nit: 3-line comment, could trim to 1-2 lines.

Separately: `inner` (from `self.inner.write()` above) is still held here across `sync_log_dir()`'s blocking fsync syscall -- readers like `get_log_state`/`try_get_log_entries` will block for the fsync's duration. The lock-held-across-blocking-I/O shape is pre-existing, but swapping a page-cache `write` for a real `fsync` meaningfully increases how long the lock is held. Not a correctness issue, but might be worth a follow-up to release the lock before the fsync (the in-memory map mutation is already done by this point).

crates/spurctld/src/raft.rs
```diff
@@ -447,6 +494,10 @@ impl openraft::RaftStorage<SpurTypeConfig> for Arc<SpurStore> {
             inner.log.remove(&k);
             self.remove_log_entry(k);
         }
+        // The deletions must be durable before this returns: a crash that loses
+        // them resurrects the conflicting tail, and `get_log_state` would then
+        // report a last_log_id this node never accepted from the leader.
```
yansun1996: Nit: 3-line comment, could trim to 1-2 lines. Same lock-held-across-fsync note as in `append_to_log` applies here too.

crates/spurctld/src/raft.rs
```diff
@@ -460,6 +511,7 @@ impl openraft::RaftStorage<SpurTypeConfig> for Arc<SpurStore> {
             inner.log.remove(&k);
             self.remove_log_entry(k);
         }
+        self.sync_log_dir()?;
```
yansun1996, line 503: Same lock-held-across-fsync note as `append_to_log`/`delete_conflict_logs_since` applies here as well.

crates/spurctld/src/raft.rs
```diff
@@ -341,6 +336,54 @@ impl SpurStore {
     }
```
yansun1996, line 339: Since this durability guarantee is hard to assert end-to-end (as the PR description notes), a couple of cheap, deterministic unit tests would still add real coverage for the new atomicity behavior: (1) leave a stale `<path>.json.tmp` on disk before calling `SpurStore::new` and assert recovery isn't confused by it, (2) assert the `.tmp` sibling is gone after a successful `persist_vote`/`persist_snapshot`/etc. Neither needs crash simulation.

yansun1996 (commented):
Nice, well-reasoned fix -- the write-fsync-rename-then-fsync-dir pattern is correctly applied across all four persist paths, and the crash-mode analysis in the PR description holds up (traced each crash point: before/during create, after write before sync, after sync before rename, after rename -- all leave a recoverable state). A few small suggestions below, nothing blocking.

yansun1996 (changes requested):

crates/spurctld/src/raft.rs
drmingdrmer: `AGENTS.md` puts the cap at 3 lines, not 2: "Keep comments short — 1-3 lines is the norm." This one is 3, so it stays.

crates/spurctld/src/raft.rs
drmingdrmer: fixed — down to 2 lines.

crates/spurctld/src/raft.rs
drmingdrmer: fixed — every error path after the temp file is created now runs `let _ = std::fs::remove_file(&tmp)` before returning.

crates/spurctld/src/raft.rs
drmingdrmer: fixed — down to 2 lines.

crates/spurctld/src/raft.rs
drmingdrmer: fixed — the comment now says a rename there is best-effort and that spurctld is built and tested on unix only, in 2 lines.

crates/spurctld/src/raft.rs
drmingdrmer: On the lock: holding the write lock through `sync_log_dir()` is what keeps `try_get_log_entries` and `get_log_state` from ever observing a log entry whose directory entry is not yet on disk. Releasing it after the map insert would break that ordering, so shortening the hold is a separate change with its own reasoning, not part of this one.

crates/spurctld/src/raft.rs
drmingdrmer: Same answer on the lock as in `append_to_log`: the hold is what keeps a reader from seeing a removal that is not yet durable, and trading that for a shorter hold belongs in its own change.

crates/spurctld/src/raft.rs
drmingdrmer: Same as `append_to_log`: the lock hold is what orders reader visibility behind the directory flush. `purge_logs_upto` also sets `inner.last_purged` after the flush, so releasing early would need that assignment reordered against the durability point — more reason to do it as its own change.

crates/spurctld/src/raft.rs
drmingdrmer: fixed — both tests added in `test: spurctld: cover temp-file handling in the raft store's atomic writes`.

`store_recovery_ignores_stale_tmp_files` plants a garbage `.json.tmp` beside each of the four persisted records and asserts strict recovery still returns the real vote and exactly the one real log index. The log-directory case is the one with teeth, since that scan matches by extension rather than exact filename: widening the filter to `path.is_file()` makes the test fail.

`atomic_write_leaves_no_tmp_sibling` calls each `persist_*` helper and asserts the target exists while its `.json.tmp` sibling does not.

crates/spurctld/src/raft.rs
drmingdrmer: Trimmed to two lines, using your wording.

crates/spurctld/src/raft.rs
drmingdrmer: Trimmed to one line: `// A newly created file is not durable until its directory is flushed.`

crates/spurctld/src/raft.rs
drmingdrmer: Trimmed to one line: `// A crash that loses these deletions resurrects the conflicting tail.`

PR #637, opened 2026-08-15
fix(security): refuse root execution, default the REST API off, gate the admin kubeconfig

### LINKED ISSUE #636: security: unauthenticated remote code execution as root via job submission (uid 0 skips the privilege drop)
## Summary

An unauthenticated peer that can reach spurctld can execute arbitrary code **as root** on any compute node, and can obtain the **k0s cluster-admin kubeconfig**. No credential of any kind is required.

The root cause is that no RPC authenticates its caller: the identity used for authorization (`user`, `uid`, `caller`) is supplied by the client, and `uid: 0` does not fail the privilege drop — it *skips* it.

Fix: #637 (raised alongside this issue).

---

## ⚠️ Immediate mitigation (before patching)

If you run spurctld today, do these now:

1. **Restrict network access to the control-plane ports** — 6817 (gRPC), 6820 (REST), 6822 (metrics) — to trusted hosts only. Until user authentication is enforced, **the listening port is the trust boundary.**
2. **Disable the REST API** unless you need it: `[rest_api] enabled = false`. It performs no authentication and is currently **on by default**.
3. Treat anyone with network reach to 6817 as having root on every node in the cluster, and audit accordingly.

---

## Chain 1 — unauthenticated remote root execution

| # | Step | Code |
|---|---|---|
| 1 | The REST API is **enabled by default** and binds all interfaces | `spur-core/src/config.rs` (`RestApiConfig::default() { enabled: true }`), default addr `[::]:6820` |
| 2 | The submit handler builds a `JobSpec` directly from the request body | `spurctld/src/rest/handlers.rs` (`user: body.job.user.unwrap_or_default()`, `..Default::default()`) |
| 3 | …and `JobSpec::default()` carries **`uid: 0`** | `spur-core/src/job.rs` |
| 4 | The uid is never validated; it is forwarded to the agent untouched | `spurctld/src/server.rs` (`let uid = req.uid;`) |
| 5 | On the agent, **uid 0 skips the privilege drop rather than failing it** | `spurd/src/privdrop.rs`: `if uid == 0 \|\| !geteuid().is_root() { return None; }` — `None` means "no drop", so the job inherits spurd's root. Same shape in `executor.rs` (`should_run_as_user(uid) = uid > 0 && …`) |

Net effect: an unauthenticated HTTP request containing a job script runs **as root on a GPU node**.

The same primitive over gRPC (which is also unauthenticated) lets a caller submit as *any* uid, i.e. run code as an arbitrary user. This is the mechanism behind the previously-observed "either user could cancel the other's job" behaviour — the `check_job_owner` string comparison added for that stops a typo, not an attacker, because the string it compares is supplied by the caller.

## Chain 2 — self-promotion to admin, then cluster-admin credential

- The accounting gRPC service is mounted on the **same plaintext port 6817** and performs **no authorization at all**, including `add_user`, which takes `admin_level`.
- `is_k0s_admin` is `caller.is_empty() || caller == "root" || cache.is_admin(caller)` (`spurctld/src/server.rs`) against a client-supplied string — so an omitted or forged `caller` is admin.
- That check gates `cluster_kubeconfig`, which returns the **k0s cluster-admin kubeconfig**.

So an anonymous peer can either forge `caller`, or promote itself to Spur Admin through the unauthenticated accounting API, and then take over the Kubernetes cluster.

## Related

- Node admission (`spur-core/src/admission.rs`) — the one properly authenticated path (hashed token, constant-time compare, JWT node identity) — defaults to `AdmissionMode::Open`.
- `[auth] plugin = "jwt"` is documented as requiring signed tokens, but the selector is never read and `spur_core::auth::verify_token` has no caller outside tests. The documentation currently describes a control that does not execute.

## What the fix does — and does not do

The PR closes the paths that turn *reachability* into *privilege*:

- spurd refuses to execute as uid 0 unless `[auth] allow_root_jobs` is explicitly set (gated at both wire entry points that carry a uid, before anything is spawned).
- `[rest_api] enabled` now defaults to **false**, and warns loudly if enabled on a non-loopback address.
- The cluster-admin kubeconfig requires a new `[cluster] allow_admin_kubeconfig` opt-in (default false) *in addition to* the admin check — because that check compares an unauthenticated string, the opt-in is what actually protects the credential.
- `is_k0s_admin` no longer treats an empty caller as admin.
- spurctld warns at startup when listening on a non-loopback address.

**It does not fix the root cause.** Callers are still unauthenticated, so an anonymous peer with network reach can still submit jobs as an arbitrary non-root user. Closing that requires deriving identity from a verified credential instead of a request field:

- [ ] Wire `verify_token` into a tonic interceptor; derive `Identity` from it; stop trusting `req.user` / `req.uid` / `req.caller` (the Tower layer, the HS256 implementation, and a good token-issuance pattern all already exist in-tree — they are simply not connected).
- [ ] Resolve `uid`/`gid` server-side from the authenticated username rather than accepting them from the wire.
- [ ] Add authorization to the accounting service (it has no caller field today — needs a proto change).
- [ ] TLS on 6817 / 6820 / 6822 and the Raft transport.
- [ ] Make `[auth] plugin` select the mode it advertises, or fail loudly at startup that it is unimplemented; correct the docs meanwhile.

Separately (lower severity, same theme): the per-account `Role` created by the k8s quota reconciler grants write on `deployments`/`daemonsets` plus `pods/exec` into a namespace carrying **no Pod Security Admission labels**, and `spur k8s kubeconfig` mints 1-year ServiceAccount tokens with no revocation path.

powderluv:
Fixes #636.

Closes the paths that turn *reachability* of the control plane into *privilege*: root code execution on a compute node, and hand-out of the k0s cluster-admin kubeconfig. See #636 for the full chain and for operator mitigations.

## What changed

**1. spurd refuses to execute as uid 0** (`spurd/src/privdrop.rs`, `agent_server.rs`, `main.rs`)

`PrivDrop::resolve_if_needed` returns `None` for uid 0 — and `None` means "no privilege drop", so a request carrying `uid: 0` did not fail, it ran with spurd's own root credentials. Since the uid arrives on the wire and no RPC authenticates its caller, that made agent reachability equivalent to root.

New `check_root_execution_allowed(uid, allow_root_jobs)` is called at **both wire entry points that carry a uid** — `launch_job` (batch) and `run_command` (steps) — *before* anything is spawned. `exec_in_job` / `spawn_pty_in_job` operate on an already-tracked job whose uid was validated at launch.

Escape hatch: `[auth] allow_root_jobs = true` (default **false**), with a startup warning when enabled. Deliberately a no-op when spurd is not root, so unprivileged dev/test runs are unaffected.

**2. `[rest_api] enabled` now defaults to `false`** (`spur-core/src/config.rs`, `spurctld/src/main.rs`)

The REST surface performs no authentication and builds a job spec straight from the request body, so on a reachable address it was unauthenticated job submission. It now also warns loudly if enabled on a non-loopback address.

⚠️ **Behaviour change:** deployments relying on the REST API must now set `[rest_api] enabled = true` explicitly. This felt like the right default for a surface with no authentication; happy to switch to "enabled but loopback-only" if maintainers prefer.

**3. The cluster-admin kubeconfig requires an explicit opt-in** (`spur-core/src/config.rs`, `spurctld/src/server.rs`)

`spur k8s kubeconfig --admin` now additionally requires `[cluster] allow_admin_kubeconfig = true` (default false). The admin check it sat behind compares a client-supplied `caller` string, so while user auth is unenforced, the opt-in — not the check — is what actually protects that credential. With it off, get the admin kubeconfig on the control-plane node (`k0s kubeconfig admin`). Per-user scoped kubeconfigs are unaffected.

**4. `is_k0s_admin` no longer treats an empty caller as admin** (`spurctld/src/server.rs`)

Nothing in-tree sends an empty caller (the CLI always fills it, falling back to `"unknown"`), so the bypass only granted admin to an *omitted field*. `root` is retained deliberately: with accounting disabled the cache reports no admins at all, and removing it would strand `spur k8s up` on every cluster that doesn't run accounting.

**5. Startup posture warning** (`spurctld/src/main.rs`)

spurctld warns when it listens on a non-loopback address, because until user auth lands the listening port *is* the trust boundary — and `[auth] plugin = "jwt"` currently implies a control that does not run.

## What this does NOT fix

Callers are still unauthenticated, so an anonymous peer with network reach can still submit jobs as an arbitrary **non-root** user. That is the root cause and needs identity derived from a verified credential rather than a request field. #636 tracks the follow-up list (interceptor + `verify_token`, server-side uid resolution, accounting authz, TLS, making `[auth] plugin` mean something).

I kept this PR to the escalation paths so it can be reviewed and shipped quickly; happy to follow up with the interceptor work.

## Testing

Verified on Linux (x86-64):

- `cargo check --workspace` ✅
- `cargo clippy --workspace --all-targets -- -D warnings` ✅
- `cargo fmt --all --check` ✅
- `cargo test -p spurd -p spur-core -p spurctld -p spur-cli` — **1840 passed, 0 failed**

New tests:
- 3 covering the root-execution guard (non-root uid allowed; opt-in allows root; uid 0 refused by default when spurd is root — and correctly a no-op when it isn't).
- 2 covering the admin-kubeconfig gate (denied by default even for an Admin-level caller; reaches the agent when enabled) plus one asserting a non-admin is still refused when enabled.
- `is_k0s_admin` test updated to assert the new intent (empty denied, `root` still allowed).

Five existing `cluster_up`/`cluster_down` tests relied on the empty-caller bypass and now pass an explicit `caller: "root"`.

### DIFF
--- a/crates/spurctld/src/rest/handlers.rs
+++ b/crates/spurctld/src/rest/handlers.rs
@@ -114,6 +114,21 @@ pub async fn submit_job(
         ..Default::default()
     };
 
+    // The REST body carries no uid, so the spec inherits `JobSpec::default()`'s uid 0. The agent
+    // refuses to execute as root, but that rejection lands only after the job has been accepted,
+    // queued and dispatched — the caller would get a job_id back and discover the failure late, or
+    // not at all. Reject here, while we can still answer 400.
+    //
+    // This is a stopgap: the class disappears once callers are authenticated and uid is derived
+    // server-side from the verified identity instead of defaulted (#636).
+    if spec.uid == 0 {
+        return Err(bad_request_response(
+            "REST job submission cannot attribute a uid to the caller, so the job would run as \
+             root and is refused. Submit via `sbatch`/gRPC, which carries the submitting user's \
+             credentials, until authenticated REST submission is available (see #636).",
+        ));
+    }
+
     // GPU demand is validated in submit_job after node-count normalization.
     let outcome = state.cluster.submit_job(spec).map_err(submit_rest_error)?;
 
--- a/crates/spurctld/src/rest/handlers.rs
+++ b/crates/spurctld/src/rest/handlers.rs
@@ -114,18 +114,20 @@ pub async fn submit_job(
         ..Default::default()
     };
 
-    // The REST body carries no uid, so the spec inherits `JobSpec::default()`'s uid 0. The agent
-    // refuses to execute as root, but that rejection lands only after the job has been accepted,
-    // queued and dispatched — the caller would get a job_id back and discover the failure late, or
-    // not at all. Reject here, while we can still answer 400.
+    // REST job submission is not supported yet, and this is where that becomes visible.
     //
-    // This is a stopgap: the class disappears once callers are authenticated and uid is derived
-    // server-side from the verified identity instead of defaulted (#636).
+    // The request body carries no uid, so the spec always inherits `JobSpec::default()`'s uid 0 —
+    // i.e. every REST submission would ask to run as root, which the agent refuses. Without this
+    // check the refusal lands only after the job has been accepted, queued and dispatched, so the
+    // caller gets a job_id back and discovers the failure late, or never. Reject up front instead.
+    //
+    // Supporting it means deriving the uid server-side from an authenticated caller rather than
+    // defaulting it; until then the honest answer is that this endpoint cannot submit work.
     if spec.uid == 0 {
         return Err(bad_request_response(
-            "REST job submission cannot attribute a uid to the caller, so the job would run as \
-             root and is refused. Submit via `sbatch`/gRPC, which carries the submitting user's \
-             credentials, until authenticated REST submission is available (see #636).",
+            "REST job submission is not supported yet: the request carries no authenticated \
+             caller, so a job could only be attributed to uid 0 (root), which is refused. Submit \
+             via `sbatch`/`srun`, which carry the submitting user's credentials.",
         ));
     }
 
--- a/crates/spurd/src/privdrop.rs
+++ b/crates/spurd/src/privdrop.rs
@@ -11,9 +11,33 @@ pub(crate) struct PrivDrop {
     groups: Vec<Gid>,
 }
 
+/// Refuse to execute work as uid 0 unless the operator opted in.
+///
+/// `PrivDrop::resolve_if_needed` returns `None` for uid 0, which means "no privilege drop" — so a
+/// request carrying `uid: 0` does not fail, it runs with spurd's own (root) credentials. Since the
+/// uid arrives on the wire as part of the job spec and no RPC authenticates its caller, that turns
+/// reachability of the agent into root on the node. Callers must run this check BEFORE spawning any
+/// child, on every execution path (batch jobs, steps, container jobs).
+///
+/// Returns `Err` with an operator-facing message when the request must be refused.
+pub(crate) fn check_root_execution_allowed(uid: u32, allow_root_jobs: bool) -> Result<(), String> {
+    if uid != 0 || allow_root_jobs || !nix::unistd::geteuid().is_root() {
+        return Ok(());
+    }
+    Err(
+        "refusing to execute as uid 0: the job requested root and spurd is running as root, but \
+         [auth] allow_root_jobs is false. Submit as an unprivileged user, or set \
+         allow_root_jobs = true if every submitter on this cluster is trusted with root."
+            .to_string(),
+    )
+}
+
 impl PrivDrop {
     /// Resolve credentials if privilege drop is needed. Returns None if
     /// spurd is not root or the job user is already root.
+    ///
+    /// SECURITY: `None` here means "run with spurd's credentials". For uid 0 that is root, so every
+    /// caller must first gate on [`check_root_execution_allowed`].
     pub fn resolve_if_needed(uid: u32, gid: u32) -> Option<Self> {
         if uid == 0 || !nix::unistd::geteuid().is_root() {
             return None;
@@ -54,5 +78,37 @@ impl PrivDrop {
             format!("--setuid={}", self.uid),
             format!("--setgid={}", self.gid),
         ]
+    }
+}
+
+#[cfg(test)]
+mod root_execution_tests {
+    use super::check_root_execution_allowed;
+
+    #[test]
+    fn non_root_uid_is_always_allowed() {
+        assert!(check_root_execution_allowed(1000, false).is_ok());
+        assert!(check_root_execution_allowed(1000, true).is_ok());
+    }
+
+    #[test]
+    fn opt_in_allows_root() {
+        assert!(check_root_execution_allowed(0, true).is_ok());
+    }
+
+    #[test]
+    fn uid_zero_is_refused_by_default_when_spurd_is_root() {
+        // Only meaningful as root: a non-root spurd cannot grant root anyway, and the guard
+        // deliberately stays quiet there so unprivileged test/dev runs are unaffected.
+        let res = check_root_execution_allowed(0, false);
+        if nix::unistd::geteuid().is_root() {
+            let msg = res.expect_err("uid 0 must be refused by default on a root spurd");
+            assert!(msg.contains("allow_root_jobs"), "actionable message: {msg}");
+        } else {
+            assert!(
+                res.is_ok(),
+                "non-root spurd cannot escalate; guard is a no-op"
+            );
+        }
     }
 }
--- a/crates/spurd/src/privdrop.rs
+++ b/crates/spurd/src/privdrop.rs
@@ -11,6 +11,12 @@ pub(crate) struct PrivDrop {
     groups: Vec<Gid>,
 }
 
+/// Whether this spurd process runs as root — the production input to
+/// [`check_root_execution_allowed`].
+pub(crate) fn spurd_runs_as_root() -> bool {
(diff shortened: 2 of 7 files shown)


crates/spurctld/src/rest/handlers.rs
```diff
@@ -114,6 +114,21 @@ pub async fn submit_job(
...
     };
 
+    // The REST body carries no uid, so the spec inherits `JobSpec::default()`'s uid 0. The agent
+    // refuses to execute as root, but that rejection lands only after the job has been accepted,
+    // queued and dispatched — the caller would get a job_id back and discover the failure late, or
+    // not at all. Reject here, while we can still answer 400.
+    //
+    // This is a stopgap: the class disappears once callers are authenticated and uid is derived
+    // server-side from the verified identity instead of defaulted (#636).
+    if spec.uid == 0 {
```
yansun1996, line 126: Since `SubmitJobFields` has no `uid` field, `spec.uid` is always `0` here (from `JobSpec::default()`), so this rejects *every* REST submission, not just root attempts — REST submission is effectively disabled now, not gated on root. If that's the intended stance until #636 lands, maybe word the check/message as "REST submission isn't supported yet" so it doesn't read as narrower than it is.

crates/spurctld/src/rest/handlers.rs
```diff
@@ -114,6 +114,21 @@ pub async fn submit_job(
         ..Default::default()
     };
 
+    // The REST body carries no uid, so the spec inherits `JobSpec::default()`'s uid 0. The agent
+    // refuses to execute as root, but that rejection lands only after the job has been accepted,
+    // queued and dispatched — the caller would get a job_id back and discover the failure late, or
+    // not at all. Reject here, while we can still answer 400.
+    //
+    // This is a stopgap: the class disappears once callers are authenticated and uid is derived
+    // server-side from the verified identity instead of defaulted (#636).
```
yansun1996: Small one: this comment references an issue number (`#636`) — repo convention keeps issue refs out of code comments (they belong in commit history/PR description). Worth dropping the `(#636)` here; the user-facing message below pointing at #636 is fine to keep.

crates/spurd/src/privdrop.rs
```diff
@@ -56,3 +80,35 @@ impl PrivDrop {
...
+        assert!(check_root_execution_allowed(1000, true).is_ok());
+    }
+
+    #[test]
+    fn opt_in_allows_root() {
+        assert!(check_root_execution_allowed(0, true).is_ok());
+    }
+
+    #[test]
+    fn uid_zero_is_refused_by_default_when_spurd_is_root() {
```
yansun1996, line 115: This only exercises the refusal (`Err`) branch when the test runner is root — in typical non-root CI it falls into the `else` and just asserts `Ok`, so the guard has no executable coverage there. Might be worth making "is spurd root" injectable so the refusal path is always testable, plus a test through a real RPC entry point (e.g. `launch_job` with `uid: 0`) asserting `PermissionDenied`.

yansun1996 (commented):
Thanks for tackling this — closing off the uid-0 and REST-default holes is the right move, and scoping it to the escalation paths rather than full RPC auth in one PR makes sense.

One thing I think needs to land before merge: `spurd`'s own `get_kubeconfig` RPC (`crates/spurd/src/agent_server.rs`, not touched by this diff) still hands out the cluster-admin kubeconfig unconditionally whenever `user` is empty — no `allow_admin_kubeconfig` check, no admin check:
```rust
let result = if req.user.is_empty() {
    self.k0s.admin_kubeconfig().await
} else { ... }
```
`fetch_admin_kubeconfig` calls this with an empty `user` by construction, so the new gate in `cluster_kubeconfig` protects the path through the controller, but anyone reaching a control-plane node's spurd gRPC port directly still gets the admin credential. Since item [3] is specifically about closing that hand-out, I think `allow_admin_kubeconfig` needs to be threaded into `AgentService` too (same pattern as `allow_root_jobs`) and checked there.

A couple of inline notes on the REST path below. Also some lower-priority follow-ups, not blockers:
- `run_command` (steps) still trusts `req.uid`/`req.gid` off the wire without checking them against the job/step's real owner — pre-existing, but worth a tracked follow-up given this PR is closing exactly this class of gap.
- `allow_admin_kubeconfig`'s doc could use a one-line note that `[auth] plugin` isn't enforced yet, so it's not misread as making `caller` trustworthy.
- A few new doc comments (config.rs, privdrop.rs, server.rs) run 4-8 lines — might be worth trimming to a line or two per the repo's usual style.

Happy to take another look once the kubeconfig path is closed.

sajmera-pensando (dismissed):
Thanks for addressing both items from my review.

**[1]** `exec_in_job` and `interactive_session` are now guarded — and fixing rather than deferring was the right call. The total invariant ("spurd never executes as uid 0 unless opted in") is much easier to maintain going forward.

**[2]** The early 400 is in place, and the fix correctly captures the broader problem: every REST submission was silently broken, not just root ones.

One thing still needs to land before merge: @yansun1996's blocker on `spurd`'s own `get_kubeconfig` RPC (`crates/spurd/src/agent_server.rs` around line 2399). The `allow_admin_kubeconfig` gate was added to `spurctld`, but `AgentService::get_kubeconfig` still hands out the admin kubeconfig unconditionally to any caller with an empty `user` field — no `allow_admin_kubeconfig` check, no admin check. Anyone reaching spurd's gRPC port directly bypasses the new gate entirely. That needs to be threaded in before this merges.

Approving on my end.

sajmera-pensando (approved):
LGTM

PR #639, opened 2026-08-15
fix(spur-core): reject hostlist patterns with ']' before '['

### LINKED ISSUE #638: [Issue]: hostlist::expand panics on a ']' that precedes '[' (spur-core)
### Root cause

`expand_single` locates the two brackets **independently**:

```rust
let bracket_start = pattern.find('[')?;                     // first '['
let bracket_end   = pattern.find(']').ok_or_else(|| ...)?;  // first ']'
let range_str = &pattern[bracket_start + 1..bracket_end];   // assumes end > start
```

`pattern.find(']')` returns the position of the *first* `]` in the string
without ever checking that it actually comes **after** the `[` it is meant to
close. When a `]` appears before the `[` (e.g. `a]b[1-2]`), `bracket_end`
lands to the left of `bracket_start`, the range slice becomes reversed
(`start > end`), and Rust aborts the process.

### Why it's a bug (a gap in existing validation)

`expand` already validates and rejects every other class of malformed pattern
by returning a `HostlistError`:

- unmatched `[` (`node[1-3`) -> `InvalidPattern`
- reversed numeric range (`node[5-3]`) -> `InvalidRange`

The `]`-before-`[` case is simply an **ordering check that was never added**:
the code trusts the first `]` to be the closing bracket instead of verifying
its position relative to `[`. As a result, one malformed input aborts the
process while its siblings return clean errors — same category of bad input,
two different failure modes. The fix closes this gap by treating a
`]` positioned before `[` as `InvalidPattern`, consistent with the checks
already in place.

### Why it matters (reachable on real surfaces)

`hostlist::expand` runs on strings that arrive from user and config input, so a
single malformed value aborts the process instead of producing an error:

- `squeue -w '<pattern>'` — `spur-cli/src/squeue.rs:393` (expanded before any daemon connection)
- `scontrol update NodeName='<pattern>'` — `spur-cli/src/scontrol.rs:1250`
- `spur node <pattern>` — `spur-cli/src/node.rs:249`
- `sinfo` node expansion — `spur-cli/src/sinfo.rs:317`
- reservation `nodes=` — `spur-core/src/reservation.rs:200` (via `normalize_node_list`)
- controller partition/topology config — `spurctld/src/cluster.rs:3549/6136/6156`, `topology.rs:53`

The `match` / `.map_err` / `.unwrap_or_default` at these call sites only handle
the `Err` variant — a panic unwinds past all of them, so none of them protect
against it. Callers (and hand-edited config lines) should receive a clear
"invalid pattern" error, not a process abort with a cryptic `begin > end`
message.

### Expected

`expand("a]b[1-2]")` returns `Err(HostlistError::InvalidPattern(...))`,
matching the existing "unmatched `[`" and "reversed range" handling. No panic,
and no behavior change for any valid pattern.

Verified on commit `9697ed6` (current tip of `main`) by building `spur-core`
and the `spur` CLI and running both against the real crate — full steps and
output in the reproduction section.

### Operating System

Ubuntu 24.04 LTS (reproduced on the test host; bug is OS-independent)

### CPU

N/A — hardware-independent logic bug (occurs on any CPU)

### GPU

N/A — hardware-independent logic bug (no GPU involved)

### ROCm Version

N/A — Spur scheduler bug, independent of ROCm runtime version

### ROCm Component

_No response_

### Steps to Reproduce

Reproduced from a fresh clone of `main` (commit `9697ed6`).

**Option A — CLI (no running controller needed; `-w` is expanded before any daemon connect):**

```bash
git clone https://github.com/ROCm/spur.git
cd spur
cargo build -p spur-cli
./target/debug/spur squeue -w 'a]b[1-2]'
```

Actual output:

```
thread 'main' panicked at crates/spur-core/src/hostlist.rs:291:33:
begin > end (4 > 1) when slicing `a]b[1-2]`
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Exit code 101. With the fix it instead prints a clean error and exits 1:

```
Error: Invalid node name specified: invalid hostlist pattern: ']' before '[': a]b[1-2]
```

**Option B — library unit test (no build of the CLI, no daemon):**

```bash
cat >> crates/spur-core/src/hostlist.rs <<'EOF'

#[cfg(test)]
mod repro {
    use super::expand;
    #[test]
    fn repro_panic() { let _ = expand("a]b[1-2]"); }
}
EOF

cargo test -p spur-core repro_panic
```

Actual output:

```
test hostlist::repro::repro_panic ... FAILED
thread 'hostlist::repro::repro_panic' panicked at crates/spur-core/src/hostlist.rs:291:33:
begin > end (4 > 1) when slicing `a]b[1-2]`
test result: FAILED. 0 passed; 1 failed
```

Equivalent minimal assertion (panics today, should pass):

```rust
assert!(spur_core::hostlist::expand("a]b[1-2]").is_err());
```

<details>

```
Paste output here
```

</details>

### Additional Information

Suggested one-line guard in `expand_single`, right after `bracket_end` is resolved and before the slice:

```rust
if bracket_end < bracket_start {
    return Err(HostlistError::InvalidPattern(format!("']' before '[': {pattern}")));
}
```

With the guard applied, all existing `spur-core` hostlist tests pass (33/33) plus a new regression test covering the malformed input. Confirmed unaffected:

- `node]1` -> `Ok` (plain hostname, no `[`)
- `node[1-2]`, `node[001-003]`, `rack[1-2]-node[1-2]` -> `Ok`
- `node[5-3]` -> `InvalidRange`
- `node[1-3` -> unmatched `[` error

I'll submit a PR with the guard plus regression tests in `hostlist.rs` (and `spur-tests/t51_hostlist.rs`).

SumonAMD:
Fixes #638

## Summary

`hostlist::expand` panicked (reversed slice, process abort exit 101) on
patterns where `]` appears before `[`. `find(']')` took the first `]` without
checking it comes after `[`, producing a reversed range slice. This adds the
missing ordering guard so the input is rejected as `InvalidPattern`, consistent
with existing handling for unmatched `[` and reversed ranges.

## Approach

- Guard in `expand_single`: return `InvalidPattern` when `bracket_end < bracket_start`.
- Regression tests in `hostlist.rs` and `spur-tests/t51_hostlist.rs`.

## Test plan

- [x] `cargo clippy --workspace --exclude spur-ffi --all-targets --locked -- -D warnings`
- [x] `cargo test -p spur-core --locked` (hostlist: pass)
- [x] `cargo test -p spur-tests --locked` (366 passed, incl. t51_19)
- [x] `spur squeue -w 'a]b[1-2]'` now exits 1 with a clean error instead of panicking (exit 101)

### DIFF
--- a/crates/spur-core/src/hostlist.rs
+++ b/crates/spur-core/src/hostlist.rs
@@ -287,6 +287,13 @@ fn expand_single(pattern: &str, results: &mut Vec<String>) -> Result<(), Hostlis
             .find(']')
             .ok_or_else(|| HostlistError::InvalidPattern("unmatched [".into()))?;
 
+        // ']' before '[' would make the slice below reversed (start > end) and panic.
+        if bracket_end < bracket_start {
+            return Err(HostlistError::InvalidPattern(format!(
+                "']' before '[': {pattern}"
+            )));
+        }
+
         let prefix = &pattern[..bracket_start];
         let range_str = &pattern[bracket_start + 1..bracket_end];
         let suffix = &pattern[bracket_end + 1..];
@@ -588,5 +595,13 @@ mod tests {
         round_tripped.sort();
 
         assert_eq!(round_tripped, expected);
+    }
+
+    #[test]
+    fn test_closing_bracket_before_opening() {
+        assert!(expand("n]ode[1-2]").is_err());
+        assert!(expand("a]b[1-2]").is_err());
+        assert!(expand("][").is_err());
+        assert!(expand("node]1").is_ok()); // ']' with no '[' is a plain hostname
     }
 }
--- a/crates/spur-tests/src/t51_hostlist.rs
+++ b/crates/spur-tests/src/t51_hostlist.rs
@@ -136,4 +136,9 @@ mod tests {
     fn t51_18_invalid_reversed_range() {
         assert!(hostlist::expand("node[5-1]").is_err());
     }
+
+    #[test]
+    fn t51_19_invalid_closing_before_opening_bracket() {
+        assert!(hostlist::expand("a]b[1-2]").is_err());
+    }
 }


sajmera-pensando (approved):
Reviewed the fix and the approach is correct. The guard catches the real panic path — `find('[')` and `find(']')` scan independently, so a `]` before `[` produces a reversed slice range (`bracket_start+1 > bracket_end`) that would panic. Placing the guard after both `find` calls and before any slice operation is the right spot.

Slurm compatibility is maintained — `hostlist_create()` returns `NULL` for this input, and Spur now returns `Err(InvalidPattern(...))` consistently.

A couple of minor observations for follow-up (not blocking):
- The test suite covers the direct case well, but a test for `"node[1-2]a]b[3-4]"` would document that the recursive suffix path also catches the error correctly.
- `split_top_level` has a pre-existing depth underflow when a stray `]` appears before a comma — it's contained by this guard, but worth a separate cleanup.

Overall the fix is minimal, correct, and well-scoped.

PR #641, opened 2026-08-16
feat(auth): authenticate callers of the agent (spurd) surface

powderluv:
Closes the last path by which **reaching a machine was enough to run work on it**.

Until now spurd verified nothing. An attacker who could reach a node's agent port could ask it to launch jobs directly, stepping around every control the controller gained in #637 and #640 — which is why I flagged this as the reason that work was not sufficient on its own.

> **Stacked on #640** (which is stacked on #637). Base branch is `users/powderluv/auth-identity-spine`; re-target down the chain as each merges.

## What changed

- **spurd gates its gRPC surface** with a Tower layer that verifies a credential signed with the cluster's `[auth] jwt_key`, honouring the same `disabled` / `permissive` / `required` modes as the controller.
- **The controller presents a credential** — minted short-lived per agent connection — through a single `agent_client::connect` used by all **14** call sites, so a new call site cannot silently become an unauthenticated one.
- **The ruling moved to `spur_core::auth::authenticate_bearer`**, shared by both daemons so the controller and the agent cannot drift apart on a security decision. The controller's middleware now delegates to it; its seven tests pass unchanged.
- **spurd refuses to start in `required` without a key**, rather than booting into a state where it would reject every launch.

## Rollout order

`permissive` is the default, and the order matters: **upgrade controllers first** (they begin presenting a credential), then flip agents to `required`. Doing it the other way round would refuse every launch from a controller that has not yet been upgraded. Permissive logs each uncredentialed call, so the migration is observable rather than guessed at.

## The honest limitation

The signing key is a **cluster-wide shared secret**, like Slurm's `munge.key`. Any holder can mint, so a compromised node can present itself to another node as the controller.

That is the accepted HPC model and a large improvement over "any peer at all" — but it is not the same as asymmetric signing, where nodes could verify and only the controller could sign. That is the natural hardening step; it is called out in the module docs rather than left for a reader to discover.

Two consequences worth naming: `[auth] jwt_key` must now be distributed to nodes (it previously only needed to exist on the controller), and it is a genuine secret on every node.

## Testing

Verified on Linux — clippy `-D warnings`, fmt, full suites green (1861 tests; spurd gains 4).

New agent-side coverage: `required` refuses an uncredentialed caller; `permissive` still accepts one (the migration window); a controller credential is accepted; and a credential signed with **another key is refused even in `permissive`** — permissive tolerates the absence of a credential, never a forged one.

## After this

The remaining items on #636 are TLS (a bearer token on plaintext is replayable), per-RPC authorization on the accounting service (needs a proto caller field), and a munge plugin.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3881,6 +3881,7 @@ dependencies = [
  "chrono",
  "clap",
  "hostname",
+ "http",
  "libc",
  "libloading",
  "nix",
@@ -3900,6 +3901,7 @@ dependencies = [
  "tokio",
  "tokio-stream",
  "tonic",
+ "tower",
  "tracing",
  "tracing-subscriber",
 ]
--- a/crates/spur-cli/src/authclient.rs
+++ b/crates/spur-cli/src/authclient.rs
@@ -79,9 +79,10 @@ fn interceptor() -> AuthInterceptor {
     AuthInterceptor { header }
 }
 
-/// Wrap an already-established channel, for callers that build one directly (a lazily connected
-/// test channel) instead of going through [`connect`]. Test-only today.
-#[cfg(test)]
+/// Wrap an already-established channel with the caller's credential.
+///
+/// Used when the channel is built by the caller (e.g. agent connections, test channels) rather
+/// than going through [`connect`].
 pub fn wrap(channel: Channel) -> AuthChannel {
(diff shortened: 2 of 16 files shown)


sajmera-pensando (dismissed):
Gaps addressed and fixes pushed directly to this PR. LGTM.

PR #643, opened 2026-08-16
fix(spur-core): prevent stray ']' from merging hosts in split_top_level

### LINKED ISSUE #642: [Issue]: hostlist::split_top_level depth underflow silently accepts stray ']' patterns (spur-core)
### Problem Description

`split_top_level` (used by `hostlist::expand`) tracks bracket nesting with a signed `depth` counter and only splits on a comma when `depth == 0`. A stray `]` that appears before any matching `[` decrements `depth` from `0` to `-1`, so every following top-level comma is treated as "inside brackets" and is not split. The whole string is then handed to `expand_single` as a single term and returned as one merged, bogus hostname.

### Concrete symptom on current `main` (after the #639 panic fix)

Verified by running `hostlist::expand` in a spur-core unit test:

    expand("x],y")            =>  Ok(["x],y"])              // BUG: comma never split — one merged fake host
    expand("node01],node02")  =>  Ok(["node01],node02"])   // BUG: comma never split — one merged fake host

Scope note: a lone `]` with no matching `[` (e.g. `node]1`, `x]`) is **intentionally** a valid literal hostname — #639 added `assert!(expand("node]1").is_ok()); // ']' with no '[' is a plain hostname`. So the bug here is strictly the *comma-split* corruption, not the presence of a `]`.

### Root cause

`split_top_level` decrements `depth` on every `]` without clamping it at zero:

    fn split_top_level(s: &str) -> Vec<&str> {
        let mut depth = 0;
        for (i, c) in s.char_indices() {
            match c {
                '[' => depth += 1,
                ']' => depth -= 1,          // no floor — 0 becomes -1
                ',' if depth == 0 => { /* split here */ }
                _ => {}
            }
        }
    }

For `x],y`:

1. `]` -> `depth = -1`
2. `,` -> `depth != 0` -> comma ignored
3. the whole string stays one part: `["x],y"]`
4. `expand_single` finds no `[`, so it treats it as a plain hostname -> `Ok(["x],y"])`

### Why it's a bug

`expand` splits on top-level commas to produce separate hostnames. When a stray `]` drives `depth` negative, a legitimate top-level comma is silently swallowed and two hosts are merged into one fake host containing a `,` and `]`. Since a lone `]` is a valid literal (per #639), `x],y` should expand to two hosts `["x]", "y"]`, not one merged string.

### Why it matters (reachable on real surfaces)

Every path that expands a hostlist from user/config input runs through `split_top_level` first (`expand -> split_top_level -> expand_single`):

* `squeue -w '<pattern>'`
* `scontrol` / `spur node` / `sinfo` node expansion
* reservation `nodes=` / controller nodelist handling

A typo like `node01],node02` is silently accepted as a single bogus hostname instead of two nodes, which is hard to diagnose in pending/filter behavior. Callers should get the correct multi-host split.

### Expected

Clamp the depth counter at zero so a stray `]` never suppresses a top-level comma split:

    ']' => depth = (depth - 1).max(0),

Result: `expand("x],y")` -> `["x]", "y"]` and `expand("node01],node02")` -> `["node01]", "node02"]`. This preserves #639's convention that a lone `]` is a literal hostname (`expand("node]1")` stays `Ok(["node]1"])`) and leaves every valid pattern unchanged (`expand("gpu[01-02],cpu[01-02]")` -> `["gpu01","gpu02","cpu01","cpu02"]`).

Verified against the full spur-core suite on current `main` (includes #639 / bdd2528): 517 passed, 0 failed, plus new regression tests — full steps in **Steps to Reproduce**.

### Operating System

Ubuntu 24.04 LTS (Noble Numbat) (reproduced on the test host; bug is OS-independent)

### CPU

AMD EPYC 9575F 64-Core Processor (N/A for root cause — hardware-independent logic bug)

### GPU

N/A — hardware-independent logic bug (no GPU involved)

### ROCm Version

N/A — Spur scheduler bug, independent of ROCm runtime version

### ROCm Component

_No response_

### Steps to Reproduce

Reproduced from `main` after #639 merged. Library unit test — no running daemon needed:

    git clone https://github.com/ROCm/spur.git
    cd spur

    cat > crates/spur-core/tests/repro_split_underflow.rs << 'EOF'
    use spur_core::hostlist;

    #[test]
    fn good_comma_split_still_works() {
        // Control: valid patterns are unaffected.
        let r = hostlist::expand("gpu[01-02],cpu[01-02]").unwrap();
        assert_eq!(r, vec!["gpu01", "gpu02", "cpu01", "cpu02"]);
    }

    #[test]
    fn repro_stray_bracket_before_comma() {
        // BUG today: one merged fake host instead of a clean 2-way split.
        // After the depth-floor fix: ["x]", "y"] and ["node01]", "node02"].
        assert_eq!(hostlist::expand("x],y").unwrap(), vec!["x],y"]);
        assert_eq!(hostlist::expand("node01],node02").unwrap(), vec!["node01],node02"]);
    }

    #[test]
    fn lone_stray_bracket_is_a_literal_host() {
        // Intended per #639 — must stay Ok after the fix.
        assert_eq!(hostlist::expand("node]1").unwrap(), vec!["node]1"]);
    }
    EOF

    cargo test -p spur-core --locked --test repro_split_underflow -- --nocapture

Actual output (bug present):

    "x],y"           => Ok(["x],y"])            // should be ["x]", "y"]
    "node01],node02" => Ok(["node01],node02"])  // should be ["node01]", "node02"]
    "node]1"         => Ok(["node]1"])          // correct (lone ']' is a literal)

Valid control (unchanged):

    expand("gpu[01-02],cpu[01-02]") => Ok(["gpu01", "gpu02", "cpu01", "cpu02"])

<details>

```
Paste output here
```

</details>

### Related

* #638 / #639 — fixed the `]`-before-`[` **panic** in `expand_single` (for terms that still contain a later `[`), and established that a lone `]` with no `[` is a valid literal hostname (`assert!(expand("node]1").is_ok())`).
* This issue — the remaining `split_top_level` depth underflow that merges hosts across a top-level comma after a stray `]`. This is the reviewer follow-up on #639: *"split_top_level has a pre-existing depth underflow when a stray `]` appears before a comma — it's contained by this guard, but worth a separate cleanup."*

### Suggested fix (depth floor)

Clamp the depth counter in `split_top_level` so it never goes negative:

    '[' => depth += 1,
    ']' => depth = (depth - 1).max(0),   // was: depth -= 1
    ',' if depth == 0 => { /* split */ }

This fixes the comma-split corruption while preserving #639's rule that a lone `]` is a valid literal hostname.

Note: rejecting on negative depth (returning `InvalidPattern`) is intentionally **not** proposed — it breaks #639's existing `assert!(expand("node]1").is_ok())` test. The depth floor is the minimal change that keeps the whole spur-core suite green (517 passed, 0 failed).

### Tests to add with the fix

* `expand("x],y")` -> `["x]", "y"]`
* `expand("node01],node02")` -> `["node01]", "node02"]`
* `expand("node]1")` -> `["node]1"]` (unchanged; guards #639's convention)
* `expand("gpu[01-02],cpu[01-02]")` unchanged (valid-pattern regression guard)
* `expand("node[1-2]a]b[3-4]").is_err()` (recursive-suffix case, already correct via #639)

I'll submit a PR with this fix and tests.

SumonAMD:
## Summary
`split_top_level` decremented its bracket-depth counter on every `]` without a floor. A stray `]` before a top-level comma drove `depth` negative, so the comma was treated as "inside brackets" and the split was suppressed — `expand("x],y")` returned `Ok(["x],y"])` (one merged, bogus host) instead of `["x]", "y"]`.

This clamps the counter with `(depth - 1).max(0)` so a top-level comma always splits. It preserves #639's convention that a lone `]` with no `[` is a valid literal hostname (`expand("node]1").is_ok()`) and leaves every valid pattern unchanged.

**Scope:** this only fixes the comma-split merge. It intentionally does **not** change how a
lone `]` with no `[` is handled — that stays a literal hostname per #639
(`expand("node]1").is_ok()`). Whether a lone `]` should be rejected is a separate question
I'll raise in a Discussion.

Fixes #642.

## Test plan
- `expand("x],y")` → `["x]", "y"]`, `expand("node01],node02")` → `["node01]", "node02"]`
- `expand("node[1-2]a]b[3-4]")` → `Err` (recursive-suffix bracket order still rejected)
- `expand("gpu[01-02],cpu[01-02]")` unchanged
- Full spur-core suite: 520 passed, 0 failed; `cargo fmt --all --check` clean

### DIFF
--- a/crates/spur-core/src/hostlist.rs
+++ b/crates/spur-core/src/hostlist.rs
@@ -268,7 +268,7 @@ fn split_top_level(s: &str) -> Vec<&str> {
     for (i, c) in s.char_indices() {
         match c {
             '[' => depth += 1,
-            ']' => depth -= 1,
+            ']' => depth = (depth - 1).max(0),
             ',' if depth == 0 => {
                 parts.push(&s[start..i]);
                 start = i + 1;
@@ -603,5 +603,21 @@ mod tests {
         assert!(expand("a]b[1-2]").is_err());
         assert!(expand("][").is_err());
         assert!(expand("node]1").is_ok()); // ']' with no '[' is a plain hostname
+    }
+
+    #[test]
+    fn stray_bracket_before_comma_now_splits() {
+        assert_eq!(expand("x],y").unwrap(), vec!["x]", "y"]);
+        assert_eq!(expand("node01],node02").unwrap(), vec!["node01]", "node02"]);
+    }
+
+    #[test]
+    fn lone_stray_bracket_stays_literal() {
+        assert_eq!(expand("x]").unwrap(), vec!["x]"]);
+    }
+
+    #[test]
+    fn recursive_suffix_bracket_order_rejected() {
+        assert!(expand("node[1-2]a]b[3-4]").is_err());
     }
 }
--- a/crates/spur-core/src/hostlist.rs
+++ b/crates/spur-core/src/hostlist.rs
@@ -609,6 +609,9 @@ mod tests {
     fn stray_bracket_before_comma_now_splits() {
         assert_eq!(expand("x],y").unwrap(), vec!["x]", "y"]);
         assert_eq!(expand("node01],node02").unwrap(), vec!["node01]", "node02"]);
+        assert_eq!(expand("],y").unwrap(), vec!["]", "y"]);
+        assert_eq!(expand("x,]y").unwrap(), vec!["x", "]y"]);
+        assert_eq!(expand("x]],y").unwrap(), vec!["x]]", "y"]);
     }
 
     #[test]


sajmera-pensando (dismissed):
Fix is correct and minimal — clamping depth at zero is the right call. The layering between `split_top_level` and `expand_single` is sound and all existing tests still pass.

sajmera-pensando (approved):
lgtm

PR #646, opened 2026-08-16
fix(spur-core): bound array size before expansion and use checked step increment

### LINKED ISSUE #645: [Issue]:  parse_array_spec integer overflow + late size check → controller DoS on sbatch --array (spur-core)
### Problem Description

`spur_core::array::parse_array_spec` (called from the controller's job-submission path in `crates/spurctld/src/cluster.rs:6452`) expands the **entire** array-task range into a `Vec<u32>` before validating its size, and increments the loop counter with unchecked `u32` arithmetic. An unprivileged user can crash or hang the controller with a single `sbatch --array=...`.

Two distinct defects live in the same loop:

```rust
// crates/spur-core/src/array.rs
let mut i = start;
while i <= end {
    task_ids.push(i);
    i += step;   // line 127  ← unchecked u32 arithmetic
}
// ...
if task_ids.len() > MAX_ARRAY_SIZE {   // line 137  ← checked AFTER full materialization
    return Err(ArrayError::TooLarge { count: task_ids.len(), max: MAX_ARRAY_SIZE });
}
```

### Bug 1 — integer overflow on `i += step` (array.rs:127)
When `end` is at/near `u32::MAX`, `i += step` overflows. Both endpoints are legal input — `4294967295` is exactly `u32::MAX`, so parsing succeeds and the overflow happens *inside the loop* on the increment:
- **Debug builds (overflow-checks on):** panics — `attempt to add with overflow` → controller thread dies.
- **Release builds:** silently wraps `i` back to a small value, so `i <= end` stays true forever → **infinite loop** that keeps pushing into `task_ids` until memory is exhausted.

### Bug 2 — size limit checked too late (array.rs:137)
`MAX_ARRAY_SIZE` (100_000) is only checked *after* the whole range is expanded. A spec like `--array=0-4000000000` allocates ~16 GB (`4e9 × 4 bytes`) **before** the guard runs — OOM long before `TooLarge` can be returned. The guard is meant to bound resource use, but it runs after the allocation it's supposed to prevent.

### Why it matters
- **Reachable by any unprivileged user:** `sbatch --array=<payload>` → `expand_job_specs()` → `parse_array_spec()` (`cluster.rs:6452`). No special permissions required.
- **Availability (DoS):** a single malformed/oversized submission can panic the controller thread (debug) or drive it into an unbounded loop + OOM (release), taking down scheduling for the whole cluster.
- **Silent in release:** the release-build wrap-around is the dangerous case — no panic, just a hang and memory blowup.

### Root cause
The range is expanded eagerly with no up-front bound on element count, and the counter uses unchecked arithmetic on attacker-controlled `u32` values.

### Suggested fix
Compute the element count up front and reject before allocating; use checked arithmetic for the increment:

```rust
let count = ((end - start) / step) as usize + 1;
if count > MAX_ARRAY_SIZE {
    return Err(ArrayError::TooLarge { count, max: MAX_ARRAY_SIZE });
}
task_ids.reserve(count);

let mut i = start;
while i <= end {
    task_ids.push(i);
    match i.checked_add(step) {
        Some(n) => i = n,
        None => break,   // reached top of u32 range
    }
}
```

### Operating System

OS-independent (spur controller logic). Reproduced on Ubuntu 22.04 LTS.

### CPU

N/A — hardware-independent controller-side logic bug (reproduced on an AMD EPYC dev host).

### GPU

N/A — controller job-submission parsing; no GPU involved.

### ROCm Version

N/A — pure spur (spur-core) logic, independent of ROCm version.

### ROCm Component

_No response_

### Steps to Reproduce

Throwaway integration test in `crates/spur-core`, run on a **debug** build:

```rust
#[test]
fn overflow_at_top_of_range() {
    // 4294967295 == u32::MAX; step 2 pushes i past the max
    let _ = spur_core::array::parse_array_spec("4294967293-4294967295:2");
}

#[test]
fn late_size_check_materializes_first() {
    // 150_001 > MAX_ARRAY_SIZE (100_000): still allocates all 150k
    // entries BEFORE returning TooLarge
    let r = spur_core::array::parse_array_spec("0-150000");
    assert!(r.is_err(), "expected TooLarge, got {:?}", r);
}
```

```
cargo test -p spur-core --test repro_array_overflow overflow_at_top_of_range -- --nocapture
```

Observed on a debug build (verified):
```
thread 'overflow_at_top_of_range' panicked at crates/spur-core/src/array.rs:127:17:
attempt to add with overflow
```

Real trigger (any user, no test needed):
```
sbatch --array=0-4000000000 job.sh              # OOM before the size guard
sbatch --array=4294967294-4294967295:2 job.sh   # overflow
```

<details>

```
Paste output here
```

</details>

### Additional Information

- **Affected code:** `crates/spur-core/src/array.rs:124-141` (`parse_array_spec`), reachable via `crates/spurctld/src/cluster.rs:6452`.
- Latent since job arrays landed (`592380b`, "Implement Phase 7: job arrays"); the loop has never been changed since.

SumonAMD:
## Summary
`parse_array_spec` had two DoS defects reachable from any user's `sbatch --array=...` (via `expand_job_specs()` → `cluster.rs:6452`):

1. **Integer overflow on the loop counter.** `i += step` used unchecked `u32` arithmetic. A range near `u32::MAX` (e.g. `4294967293-4294967295:2`) panicked the controller thread in debug builds and silently wrapped into an **infinite loop + memory blowup** in release builds.
2. **Size limit checked too late.** `MAX_ARRAY_SIZE` was only enforced *after* the whole range was materialized into a `Vec<u32>`, so `--array=0-4000000000` allocated ~16 GB before the guard could reject it — OOM before the check.

## Fix
- Compute the element count in `u64` and reject **before** allocating, tracking the running total across comma-separated ranges. Same `ArrayError::TooLarge`, just raised early.
- Replace `i += step` with `i.checked_add(step)`, breaking cleanly at the top of the `u32` range.
- The existing post-expansion `TooLarge` check is kept as a backstop.

No behavior change for valid input.

Fixes #645

## Test plan
- `cargo test -p spur-core array` — 28 passed, 0 failed, including 3 new regression tests:
  - `rejects_range_overflow_at_u32_max` — `4294967293-4294967295:2` now returns `[4294967293, 4294967295]` instead of panicking/looping
  - `rejects_oversized_range_before_alloc` — `0-4000000000` returns `TooLarge` without allocating
  - `max_size_boundary_unchanged` — `0-99999` ok (exactly 100_000), `0-100000` errors
- Existing array/dependency tests unchanged (no regressions)
- `cargo fmt -p spur-core` clean

### DIFF
--- a/crates/spur-core/src/array.rs
+++ b/crates/spur-core/src/array.rs
@@ -122,10 +122,23 @@ pub fn parse_array_spec(spec: &str) -> Result<ArraySpec, ArrayError> {
             if start > end {
                 return Err(ArrayError::InvalidSpec(format!("{} > {}", start, end)));
             }
+
+            // Bound element count BEFORE materializing (u64 avoids overflow in the count itself).
+            let count = (end as u64 - start as u64) / step as u64 + 1;
+            if task_ids.len() as u64 + count > MAX_ARRAY_SIZE as u64 {
+                return Err(ArrayError::TooLarge {
+                    count: (task_ids.len() as u64 + count) as usize,
+                    max: MAX_ARRAY_SIZE,
+                });
+            }
+
             let mut i = start;
             while i <= end {
                 task_ids.push(i);
-                i += step;
+                match i.checked_add(step) {
+                    Some(n) => i = n,
+                    None => break, // reached top of u32 range
+                }
             }
         } else {
             let id: u32 = range_str
@@ -304,5 +317,25 @@ mod tests {
             aggregate_array_state(&[JobState::OutOfMemory, JobState::Failed]),
             Some(JobState::OutOfMemory)
         );
+    }
+
+    #[test]
+    fn rejects_range_overflow_at_u32_max() {
+        let spec = parse_array_spec("4294967293-4294967295:2").unwrap();
+        assert_eq!(spec.task_ids, vec![4294967293, 4294967295]);
+    }
+
+    #[test]
+    fn rejects_oversized_range_before_alloc() {
+        assert!(matches!(
+            parse_array_spec("0-4000000000"),
+            Err(ArrayError::TooLarge { .. })
+        ));
+    }
+
+    #[test]
+    fn max_size_boundary_unchanged() {
+        assert!(parse_array_spec("0-99999").is_ok());
+        assert!(parse_array_spec("0-100000").is_err());
     }
 }


sajmera-pensando (approved):
The two bugs are real and the fix is correct.

**Math:** The count formula `(end - start) / step + 1` is correct — it's `floor((end-start)/step)+1`, which matches exactly what the loop pushes. Verified all edge cases including step > 1, step doesn't divide the range evenly, and the u32-max boundary. The `checked_add` break fires *after* the last valid element is already pushed, so the loop and formula always agree. ✓

**One subtle note (not a blocker):** The pre-check counts elements before `dedup` runs (which deduplicates overlapping comma-separated ranges). So a pathological spec like `0-99999,0-99999` would be false-rejected even though its deduped result is valid. Not a practical problem since no real user writes overlapping ranges, but the PR description's claim of "no behavior change for valid input" is slightly inaccurate if you consider overlapping ranges valid.

**Nits:**
- `rejects_range_overflow_at_u32_max` is a misleading name — the test asserts *success*, not rejection. Suggest `parses_range_at_u32_boundary`.
- The comment "Bound element count BEFORE materializing" describes *what* the code does rather than *why*. The second half ("u64 avoids overflow in the count itself") is the useful part.

Both nits are minor. The fix is sound and the regression tests cover the right cases.

PR #647, opened 2026-08-17
fix(spur-cli): accept Slurm's --noconvert on squeue and sinfo

nikhilsk:
## Summary

`squeue` and `sinfo` did not define Slurm's `--noconvert`, so `clap` rejected it as an
unknown argument and the command exited non-zero having done nothing:

```
$ squeue --noconvert
Error: error: unexpected argument '--noconvert' found

Usage: squeue [OPTIONS]
```

`--noconvert` is widely scripted in Slurm to keep output machine-parseable, so a wrapper
carried over from Slurm fails at the argument parser rather than at anything real. The user
sees an argument error, not a scheduler problem.

## Approach

The flag is accepted and deliberately does nothing, because Spur already produces what it
asks for. In Slurm, `--noconvert` means "do not humanize units" — print `2048` rather than
`2G`. Spur never humanizes units in these commands:

- `sinfo` renders memory as `r.memory_mb.to_string()` and `node.free_memory_mb.to_string()`
  — raw integers, no suffix.
- `format_engine.rs` contains no unit conversion of any kind.
- The CLI's only `format_bytes` lives in `sdiag.rs`, an unrelated command.

So accepting the flag without branching on it is the correct implementation rather than a
stub, and output is byte-identical whether or not it is passed. Confirmed on a running
cluster:

```
$ sinfo --noconvert -N -o '%n %m %e'
HOSTNAMES MEMORY FREE_MEM
node01    24035  11077
```

Long-only, matching Slurm, which gives `--noconvert` no short form.

## Design notes

**Inertness is pinned by a test, not a comment.** `noconvert_changes_nothing_else` asserts
that parsing with the flag leaves every other output-affecting argument identical. If unit
conversion is ever added to these commands, that test is the thing that should force someone
to decide whether the flag must now be honoured. A comment would not.

**No unit conversion was added.** Implementing conversion so the flag had something to
disable would change the output of every existing user to make a compatibility flag
meaningful — a far larger change than the incompatibility it fixes.

**Not swallowed by a catch-all.** Accepting arbitrary unknown arguments would fix this
invocation and hide genuine typos and future incompatibilities, turning a loud failure into
a silent one. The flag is declared explicitly.

## Compatibility

No change to persisted state, the Raft log, proto, the config schema, or any existing flag,
default, or output format. Purely additive: an invocation that used to fail now succeeds, and
every other invocation behaves exactly as before.

## Testing

`cargo test --locked` → **2894 passed, 0 failed, 25 ignored** (ignored tests need
PostgreSQL). Four new tests, two per command, following each file's existing arg-test idiom.

Gates green: `cargo fmt --all --check`; `cargo clippy --workspace --exclude spur-ffi
--all-targets --locked` with `RUSTFLAGS="-D warnings"`; `cargo test --locked`;
`cargo deny check` (`advisories ok, bans ok, licenses ok, sources ok`).

The failure was reproduced against a built binary before the fix, and the fix verified
against a live single-node cluster afterwards:

| Check | Result |
|---|---|
| `squeue --noconvert` before the change | `unexpected argument '--noconvert' found`, exit 1 |
| `sinfo --noconvert` before the change | `unexpected argument '--noconvert' found`, exit 1 |
| `diff` of `squeue` vs `squeue --noconvert` | **empty**, both exit 0 |
| `diff` of `sinfo` vs `sinfo --noconvert` | **empty**, both exit 0 |
| `squeue --noconvert -h -o '%i %m'` | exit 0 |
| `squeue --help`, `sinfo --help` | `--noconvert` documented |

The two empty diffs are the checks that matter: they are what guarantee no existing user's
output moved.

## Scope: the other four commands

Slurm also defines `--noconvert` on `sacct`, `sstat`, `sshare` and `sreport`, and none of
them accept it after this change, so a migrated script that uses it there still hits the
same clap error one command later. That is deliberately left out of this PR and tracked as
a follow-up, for one reason: on `sacct`, `sshare` and `sreport` the flag would be inert as
it is here, but on `sstat` it would not. `sstat` renders `format!("{}M", res.memory_mb)`,
so that output is genuinely unit-suffixed and the flag has something to suppress. Honouring
it there means changing what an existing command prints, which is a different kind of
change from accepting an inert flag and deserves its own review rather than riding along
with this one.

`docs/migration-from-slurm.rst` records both halves: that the flag is accepted and inert on
`squeue` and `sinfo`, and that the other four do not accept it yet.

## Two unrelated rows removed from that table

While adding the row: the `squeue --sort` and `sinfo --states` entries claimed both flags
are accepted but not applied, and both have since been implemented. Sorting is applied at
`squeue.rs:150` after `parse_sort_arg`, and node states are filtered server-side at
`server.rs:846` from the states the client puts in the request. A limitations table that
lists working features teaches migrating admins the wrong thing, so those two rows are
gone. No code changed for this.

### DIFF
--- a/crates/spur-cli/src/sinfo.rs
+++ b/crates/spur-cli/src/sinfo.rs
@@ -45,8 +45,7 @@ pub struct SinfoArgs {
     #[arg(short = 'h', long)]
     pub noheader: bool,
 
-    /// Report values without unit conversion. Accepted for Slurm compatibility:
-    /// Spur already emits unconverted values, so there is nothing to suppress.
+    /// Accepted for Slurm compatibility; has no effect
     #[arg(long)]
     pub noconvert: bool,
 
@@ -352,7 +351,7 @@ mod tests {
     use super::*;
     use spur_proto::proto as pb;
     use spur_proto::proto::slurm_controller_server::SlurmController;
-    use spur_proto::proto::NodeState;
+    use spur_proto::proto::{NodeState, ResourceSet};
     use std::sync::{Arc, Mutex};
     use tonic::{Request, Response, Status};
 
@@ -402,16 +401,37 @@ mod tests {
     }
 
     #[test]
-    fn noconvert_changes_nothing_else() {
-        // Spur never humanizes units, so the flag has nothing to suppress. If unit
-        // conversion is ever added, this is what should force the flag to be honoured.
-        let plain = parse_sinfo_args(&["sinfo", "-N", "-h", "-o", "%n %m"]);
-        let flagged = parse_sinfo_args(&["sinfo", "-N", "-h", "-o", "%n %m", "--noconvert"]);
-        assert_eq!(plain.format, flagged.format);
-        assert_eq!(plain.noheader, flagged.noheader);
-        assert_eq!(plain.node_oriented, flagged.node_oriented);
-        assert!(!plain.noconvert);
-        assert!(flagged.noconvert);
+    fn memory_columns_render_as_raw_megabytes_with_and_without_noconvert() {
+        // The flag is inert only because these cells are already raw MB. Driving the
+        // real render path pins that: humanizing memory_mb or free_memory_mb later
+        // turns this red, which a parse-only test would not.
+        let mut node = make_node("gpu001", NodeState::NodeIdle, "gpu");
+        node.total_resources = Some(ResourceSet {
+            memory_mb: 2_321_924,
+            ..Default::default()
+        });
+        node.free_memory_mb = 11_077;
+        let partitions = vec![make_partition("gpu", true)];
+
+        for argv in [
+            ["sinfo", "-N", "-h", "-o", "%m %e"].as_slice(),
+            ["sinfo", "-N", "-h", "-o", "%m %e", "--noconvert"].as_slice(),
+        ] {
+            let args = parse_sinfo_args(argv);
+            let fields = format_engine::parse_format(
+                args.format.as_deref().expect("-o sets format"),
+                &format_engine::sinfo_header,
+            );
+
+            let lines = render_sinfo_output(
+                &fields,
+                &partitions,
+                std::slice::from_ref(&node),
+                args.node_oriented,
+            );
+
+            assert_eq!(lines, ["2321924 11077"], "argv: {argv:?}");
+        }
     }
 
     #[test]
--- a/crates/spur-cli/src/sinfo.rs
+++ b/crates/spur-cli/src/sinfo.rs
@@ -402,9 +402,7 @@ mod tests {
 
     #[test]
     fn memory_columns_render_as_raw_megabytes_with_and_without_noconvert() {
-        // The flag is inert only because these cells are already raw MB. Driving the
-        // real render path pins that: humanizing memory_mb or free_memory_mb later
-        // turns this red, which a parse-only test would not.
+        // Pins that %m/%e are raw MB today; humanizing them later turns this red.
         let mut node = make_node("gpu001", NodeState::NodeIdle, "gpu");
         node.total_resources = Some(ResourceSet {
             memory_mb: 2_321_924,
--- a/crates/spur-cli/src/sinfo.rs
+++ b/crates/spur-cli/src/sinfo.rs
@@ -45,6 +45,11 @@ pub struct SinfoArgs {
     #[arg(short = 'h', long)]
     pub noheader: bool,
 
+    /// Report values without unit conversion. Accepted for Slurm compatibility:
+    /// Spur already emits unconverted values, so there is nothing to suppress.
+    #[arg(long)]
+    pub noconvert: bool,
+
     /// Controller address
     #[arg(
         long,
@@ -387,6 +392,26 @@ mod tests {
         let args = parse_sinfo_args(&["sinfo", "-t", "idle"]);
         let req = build_get_nodes_request(&args).unwrap();
         assert_eq!(req.states, vec![NodeState::NodeIdle as i32]);
+    }
+
+    #[test]
+    fn noconvert_is_accepted() {
+        // Slurm scripts pass --noconvert to keep output machine-parseable.
+        let args = parse_sinfo_args(&["sinfo", "--noconvert"]);
+        assert!(args.noconvert);
+    }
+
+    #[test]
+    fn noconvert_changes_nothing_else() {
+        // Spur never humanizes units, so the flag has nothing to suppress. If unit
+        // conversion is ever added, this is what should force the flag to be honoured.
+        let plain = parse_sinfo_args(&["sinfo", "-N", "-h", "-o", "%n %m"]);
+        let flagged = parse_sinfo_args(&["sinfo", "-N", "-h", "-o", "%n %m", "--noconvert"]);
+        assert_eq!(plain.format, flagged.format);
+        assert_eq!(plain.noheader, flagged.noheader);
+        assert_eq!(plain.node_oriented, flagged.node_oriented);
+        assert!(!plain.noconvert);
+        assert!(flagged.noconvert);
     }
 
     #[test]
--- a/crates/spur-cli/src/squeue.rs
+++ b/crates/spur-cli/src/squeue.rs
@@ -57,6 +57,11 @@ pub struct SqueueArgs {
     #[arg(short = 'h', long)]
     pub noheader: bool,
 
+    /// Report values without unit conversion. Accepted for Slurm compatibility:
+    /// Spur already emits unconverted values, so there is nothing to suppress.
+    #[arg(long)]
+    pub noconvert: bool,
+
     /// Print help
     #[arg(long, action = clap::ArgAction::Help)]
     pub help: Option<bool>,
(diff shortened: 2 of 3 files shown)


crates/spur-cli/src/squeue.rs
```diff
@@ -57,6 +57,11 @@ pub struct SqueueArgs {
     #[arg(short = 'h', long)]
     pub noheader: bool,
 
+    /// Report values without unit conversion. Accepted for Slurm compatibility:
+    /// Spur already emits unconverted values, so there is nothing to suppress.
```
yansun1996: Both structs set `disable_help_flag = true` and re-add a long `--help`, and clap treats a single-paragraph doc comment as both short and long help — so this whole two-sentence rationale is what a user sees in `squeue --help`, next to terse neighbours like "Don't print header" and "Long format (more columns)".

Also, "Spur already emits unconverted values" is accurate for `sinfo` (`%m`/`%e` render raw `memory_mb`/`free_memory_mb`), but on `squeue` it is vacuous — `resolve_job_field` has no size- or memory-valued specifier at all, so there is nothing being emitted either way.

Would a terse one-liner work better here, with the rationale living in the commit message? e.g. `/// Accepted for Slurm compatibility; has no effect.` (Same for the identical block at `sinfo.rs:48`.)

crates/spur-cli/src/squeue.rs
```diff
@@ -460,6 +465,27 @@ mod tests {
...
+        // Slurm scripts pass --noconvert to keep output machine-parseable.
+        let args = SqueueArgs::try_parse_from(["squeue", "--noconvert"]).unwrap();
+        assert!(args.noconvert);
+    }
+
+    #[test]
+    fn noconvert_changes_nothing_else() {
+        // Spur never humanizes units, so the flag has nothing to suppress. If unit
+        // conversion is ever added, this is what should force the flag to be honoured.
+        let plain = SqueueArgs::try_parse_from(["squeue", "-h", "-o", "%i %m"]).unwrap();
```
yansun1996: `%m` is not a squeue specifier in this repo — there is no `'m'` arm in `resolve_job_field` (it falls through to `"?"` at line 207) and none in `format_engine::squeue_header`, so this renders header `?` / value `?`. The format string reads as though it exercises a memory column that does not exist.

Could we use a real specifier (`%i %C`) or drop the `-o` entirely? (`%n %m` on the sinfo side is fine — sinfo does implement `%m`.)

crates/spur-cli/src/squeue.rs
```diff
@@ -460,6 +465,27 @@ mod tests {
...
 
+    #[test]
+    fn noconvert_is_accepted() {
+        // Slurm scripts pass --noconvert to keep output machine-parseable.
+        let args = SqueueArgs::try_parse_from(["squeue", "--noconvert"]).unwrap();
+        assert!(args.noconvert);
+    }
+
+    #[test]
+    fn noconvert_changes_nothing_else() {
```
yansun1996: I think this test may be worth dropping rather than keeping. The two argv vectors share an identical prefix and differ only by `--noconvert`, so `format`/`noheader`/`long` cannot diverge — the assertions cannot fail — and the last two lines restate `noconvert_is_accepted` directly above.

The comment also promises more than the test delivers: "if unit conversion is ever added, this is what should force the flag to be honoured" — but the test stops at the parser and never reaches the render path, so humanizing `memory_mb` at `sinfo.rs:241` tomorrow would leave it green.

Two options: delete it and keep `noconvert_is_accepted` (which is the one that actually goes red pre-fix), or make it load-bearing by rendering a `NodeInfo` fixture through `resolve_node_field` with `-o "%m"` and asserting the cell is exactly the raw MB value with and without the flag. The second is the honest version of the stated intent; the first is fine too. (Same applies to `sinfo.rs:405`.)

yansun1996 (commented):
Thanks for this — the diagnosis is right, and the "accept, don't implement conversion, don't add a catch-all" reasoning in the PR body is the correct call. A few things worth tightening before merge:

**1. `docs/migration-from-slurm.rst` should get a row.** That file's "Behavioral Differences and Current Limitations" table is the canonical list of accepted-but-inert flags — `docs/user-guide/slurm-compatibility.rst` points at it as "the full list of behavioral differences and current limitations", and it already carries rows for `squeue --sort` and `sinfo --states`. Adding a `squeue`/`sinfo` `--noconvert` row there keeps that claim true and gives the flag a documented home. Something like: "Accepted for compatibility; Spur already reports raw values (memory in MB), so there is nothing to suppress."

**2. Only two of the affected commands are covered.** Slurm also accepts `--noconvert` on `sacct`, `sstat`, `sshare` and `sreport`, and none of them take it after this change — so a migrated script still hits the same clap error one command later. `sstat` is the interesting one: `sstat.rs:208` renders `format!("{}M", res.memory_mb)`, which is genuinely unit-suffixed output, so there the flag would actually *mean* something rather than being inert. Happy for that to be a follow-up, but it would be good to say so in the PR body so the scope is explicit.

**3. Small doc/test nits inline below.**

On the test side: `noconvert_is_accepted` is the test that matters and it goes red against the unfixed parser — good. The second test in each pair is the one I would reconsider (see inline). No concerns on the structural side: neither arg struct is serialized or persisted, nothing constructs them as a struct literal, and the flag reaches the parser identically via both the `squeue`/`sinfo` symlinks and the `spur queue`/`spur nodes` verbs.

One optional extra while you are in that docs table: the `squeue --sort` and `sinfo --states` rows are now stale — both are implemented (`squeue.rs:146`, `sinfo.rs:118-124`). Pre-existing, so entirely your call whether to fix them here.

yansun1996 (changes requested):

crates/spur-cli/src/squeue.rs
nikhilsk: Both are now the one-liner you suggested: `/// Accepted for Slurm compatibility; has no effect`, on `squeue` and `sinfo` alike, with the rationale living in the commit message. You were right that the old text was the long help too, and right that "Spur already emits unconverted values" was vacuous on `squeue`, which has no size-valued specifier to emit either way.

Two things from your review summary, both now done:

**The docs row** is in `8cb56cb`, using your wording. It also records that `sacct`, `sstat`, `sshare` and `sreport` do not accept the flag yet, since that is the part a migrated script actually trips over.

**The stale rows** are gone in the same commit. Both claims are false now: sorting is applied after `parse_sort_arg`, and node states are filtered server-side from the states the client puts in the request. A limitations table listing working features misleads the reader it exists for.

crates/spur-cli/src/squeue.rs
nikhilsk: Correct, and the test that used it is gone. `resolve_job_field` has no `'m'` arm, so it fell through to `?` and the format string read as though it exercised a memory column that does not exist on `squeue`.

The remaining `squeue` test is the parse-level one that goes red against the unfixed parser. The render-path assertion now lives only on the `sinfo` side, where `%m` is real.

crates/spur-cli/src/squeue.rs
nikhilsk: Agreed on both counts: the assertions could not diverge, and the comment promised a render-path guarantee the test never reached.

Took the second option on the `sinfo` side and the first on `squeue`. `sinfo` now has one test that renders a `NodeInfo` fixture through `resolve_node_field` with `-o "%m %e"` and asserts the cells are exactly the raw megabyte values, with and without the flag, so humanizing `memory_mb` later does turn it red. On `squeue` there is nothing to render, so the duplicate test is simply deleted and `noconvert_is_accepted` stands alone.

On the wider scope point from your summary: this PR stays at `squeue` and `sinfo`, and the reason is your `sstat` observation. On `sacct`, `sshare` and `sreport` the flag would be inert exactly as it is here, but `sstat` prints memory with an `M` suffix, so there it has something real to suppress and honouring it changes what an existing command prints. That deserves its own review rather than riding along with an inert-flag change. Filed as #695, and the PR body now says so.

Also rebased onto main, which clears the `deny` failure: the branch predated main's h2 bump.

crates/spur-cli/src/sinfo.rs
```diff
@@ -389,6 +393,47 @@ mod tests {
...
+    #[test]
+    fn noconvert_is_accepted() {
+        // Slurm scripts pass --noconvert to keep output machine-parseable.
+        let args = parse_sinfo_args(&["sinfo", "--noconvert"]);
+        assert!(args.noconvert);
+    }
+
+    #[test]
+    fn memory_columns_render_as_raw_megabytes_with_and_without_noconvert() {
+        // The flag is inert only because these cells are already raw MB. Driving the
```
yansun1996: Nit: this comment block is 3 lines — could it collapse to one, e.g. "Pins that %m/%e are raw MB today; humanizing them later turns this red."? The extra rationale reads fine in the commit message.

docs/migration-from-slurm.rst
```diff
@@ -55,10 +55,8 @@ noted as "accepted for compatibility" parse without error but have no effect yet
      - Difference
    * - Partitions
      - Defined in ``spur.conf``; there is no runtime ``scontrol create/update/delete partition``. Edit the config and reload the controller to change a partition.
-   * - ``squeue --sort``
-     - Accepted for compatibility; result ordering is not yet applied.
-   * - ``sinfo --states``
-     - Accepted for compatibility; the state filter is not yet applied (all node states are returned).
+   * - ``squeue``/``sinfo`` ``--noconvert``
```
yansun1996, line 58: Small wording nit: "Spur already reports raw values (memory in MB)" is accurate for `sinfo`, but `squeue` doesn't have a memory-valued format specifier at all, so the flag is inert there for a different reason (nothing to render, not "already raw"). Maybe: "Spur already reports raw values where applicable (sinfo's memory columns are MB integers; squeue has no memory column yet), so there is nothing to suppress."

yansun1996 (dismissed):
This looks solid — nice, tight fix with the reasoning for scope (leaving `sstat`/etc. for a follow-up) clearly laid out, and the render-path test on `sinfo` is a good regression pin. Two small things below, neither blocking.

sgopinath1 (approved):
LGTM

PR #652, opened 2026-08-17
fix(spurctld): QOS priority not applied to job priority; preemption never triggers

sajmera-pensando:
Fixes [SPUR-199](https://amd.atlassian.net/browse/SPUR-199).

## What broke

Three independent bugs combined to make burst preemption a no-op:

**[1] QOS priority ignored at submit time.** Every job got `priority = 1000` (DEFAULT_PRIORITY) regardless of its QOS tier. With partition tier 100 and any age, the effective priority was already ~100K+, making the QOS delta negligible — nowhere near the 2× threshold preemption requires.

**[2] Cancel preemption recorded the wrong terminal state.** When the scheduler did fire preemption (rarely), the `Cancel` arm called `complete_job(..., JobState::Cancelled)`. The victim appeared as `CANCELLED ExitCode -1:0` — identical to a `scancel` — and the `PREEMPTED` accounting counter stayed at 0 forever.

**[3] `sprio` EFFECTIVE column misleading.** `sprio -l` computed EFFECTIVE without the QOS adjustment and approximated fairshare as 1.0, so every job showed `Priority=1000 × age × tier`. Operators could not see any QOS signal.

## Approach

**Bug [1] — QOS priority as the base seed.** At submit time, if the user did not set `--priority` explicitly, `spec.priority` is set to `DEFAULT_PRIORITY.saturating_add_signed(qos.priority).max(1)` before the WAL proposal. This bakes it into the replicated log entry so all replicas produce the same result. The QOS priority now seeds the multiplicative formula (`base × fairshare × age × partition_tier`) — boost tiers rank above DEFAULT_PRIORITY, penalty tiers below it.

**Bug [2] — `JobPreemptCancel` WAL operation.** A new `WalOperation::JobPreemptCancel` atomically transitions `Running → Preempted → Cancelled`, frees node allocations, and fires `jobs_finalized` with `Preempted` state in a single committed entry. `Preempted` is the accounting record (PREEMPTED counter increments); `Cancelled` is the terminal live state so `afterok`/`AfterAny` dependents and array aggregation resolve immediately.

**Bug [3] — fixed as a corollary of [1].** `sprio` already uses `job.priority` as the EFFECTIVE base. Now that `job.priority` is the QOS-derived value, the display is correct without any changes to `sprio.rs`.

## Important: explicit `--priority` opt-out

When a user sets `--priority N` explicitly, QOS priority is not applied — `N` is used as-is. Two jobs with identical explicit priorities will not preempt each other regardless of QOS tier. This matches Slurm's semantics: `--priority` is a manual override. Note: there is currently no admin guard preventing regular users from setting `--priority`; fixing that requires a coherent operator role model across the whole API.

## Known limitations not fixed in this PR

- **`scontrol update JobId=N QOS=...` / `sacctmgr modify qos set priority=`** no longer affects an already-submitted job's base priority — the value is baked in at submit time and stored in the WAL entry. Operators who change a QOS priority will only see the effect on new submissions.
- **`release_job` unconditionally proposes `DEFAULT_PRIORITY`**, so hold+release will overwrite a job's QOS-derived base with 1000. Pre-existing behaviour, but this change makes it consequential. Filing as a follow-up.
- **Duplicated drain logic** (`alloc_resources.subtract → update_state_from_alloc → Draining → Drain`) appears five times across the codebase; `JobPreemptCancel` adds the fifth copy. Extracting a shared helper is the right move — leaving for a follow-up to keep this PR tight.

## WAL compatibility

`WalOperation` is externally tagged with no unknown-variant fallback. Once a `JobPreemptCancel` entry is written to the Raft log, an older controller cannot replay it and will refuse to start under strict recovery. All controllers must be upgraded before the first cancel-mode preemption fires; there is no rollback after that point.

## Upgrade edge case

WAL entries are immutable. Burst jobs submitted before this fix have `spec.priority = None` stored, so they keep their replayed base of 1000 (DEFAULT_PRIORITY) for their lifetime. Only newly submitted jobs get QOS-derived priorities. A burst job with very high age and fairshare could temporarily resist preemption by a fresh primus job during the transition window. Once all pre-upgrade burst jobs cycle out, the fix is fully effective.

## Testing

Six new tests added:

- `apply_preempt_cancel_is_atomic_and_replay_deterministic` — WAL-level atomicity and replay idempotency for `JobPreemptCancel`; verifies accounting sees `Preempted`, live state is `Cancelled`, nodes freed, no-op on replay
- `preempt_cancel_op_round_trips` — serde round-trip for the new WAL variant
- `submit_uses_qos_priority_as_base_when_user_omits_priority` — verifies base = `DEFAULT_PRIORITY.saturating_add_signed(qos.priority)`
- `submit_explicit_priority_overrides_qos_priority` — explicit `--priority` wins over QOS
- `high_qos_job_preempts_low_qos_job_end_to_end` — primus (QOS=10000) preempts burst (QOS=100); asserts effective ratio ≥ 2×, victim lands in `Cancelled`, allocation freed
- `explicit_priority_disables_qos_based_preemption` — documents the opt-out behaviour

```
cargo clippy --workspace --exclude spur-ffi --all-targets --locked  # clean
cargo test --locked                                                  # 758+ passed, 0 failed
```

### DIFF
--- a/crates/spur-core/src/wal.rs
+++ b/crates/spur-core/src/wal.rs
@@ -947,6 +947,17 @@ mod suspend_wal_tests {
     use super::*;
 
     #[test]
+    fn preempt_cancel_op_round_trips() {
+        let op = WalOperation::JobPreemptCancel { job_id: 7 };
+        let json = serde_json::to_string(&op).unwrap();
+        let back: WalOperation = serde_json::from_str(&json).unwrap();
+        match back {
+            WalOperation::JobPreemptCancel { job_id } => assert_eq!(job_id, 7),
+            _ => panic!("wrong variant"),
+        }
+    }
+
+    #[test]
     fn preempt_requeue_op_round_trips() {
         let begin_time = chrono::Utc::now();
         let op = WalOperation::JobPreemptRequeue {
--- a/crates/spur-core/src/wal.rs
+++ b/crates/spur-core/src/wal.rs
@@ -127,9 +127,11 @@ pub enum WalOperation {
         job_id: JobId,
         begin_time: chrono::DateTime<chrono::Utc>,
     },
-    /// Preempt a running job with cancel (job ends in PREEMPTED, not requeued).
-    /// Single atomic entry so a leadership change cannot strand the job running
-    /// with its allocations held.
+    /// Preempt a running job with cancel. The job transitions Running →
+    /// Preempted → Cancelled: Preempted is reported to accounting so the
+    /// PREEMPTED counter increments; Cancelled is the terminal live state so
+    /// dependents and arrays resolve. Single atomic entry so a leadership
+    /// change cannot strand the job running with its allocations held.
     JobPreemptCancel {
         job_id: JobId,
     },
--- a/crates/spur-core/src/wal.rs
+++ b/crates/spur-core/src/wal.rs
@@ -127,6 +127,12 @@ pub enum WalOperation {
         job_id: JobId,
         begin_time: chrono::DateTime<chrono::Utc>,
     },
+    /// Preempt a running job with cancel (job ends in PREEMPTED, not requeued).
+    /// Single atomic entry so a leadership change cannot strand the job running
+    /// with its allocations held.
+    JobPreemptCancel {
+        job_id: JobId,
+    },
     JobSuspend {
         job_id: JobId,
         /// Controller-stamped instant of suspension (for replay-deterministic accounting).
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -4653,10 +4653,8 @@ impl ClusterManager {
                     allocated_resources = job.allocated_resources.clone();
                     per_node_map = job.per_node_alloc.clone();
                     job.node_completions.clear();
-                    // Preempted → Cancelled so the job is terminal: dependents
-                    // (afterok/AfterAny) unblock and array aggregation resolves.
-                    // Preempted is reported to accounting via jobs_finalized so
-                    // the PREEMPTED counter still increments correctly.
+                    // Cancelled is terminal (dependents/arrays unblock); Preempted
+                    // is reported to accounting via jobs_finalized.
                     if let Err(e) = job.transition(JobState::Cancelled) {
                         warn!(job_id = *job_id, error = %e, "invalid preempt-cancel final transition in WAL apply");
                         return ClientResponse::default();
@@ -12121,6 +12119,45 @@ mod tests {
             burst_job.state,
             JobState::Running,
             "burst job must keep running when both jobs have identical explicit --priority"
+        );
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn preempt_cancel_unblocks_afterok_and_afterany_dependents() {
+        // A cancel-preempted job lands in Cancelled (terminal), so afterok
+        // and afterany dependents must stop waiting and resolve their outcome.
+        let dir = TempDir::new().unwrap();
+        let cm = test_cluster(&dir).await;
+        register_node(&cm, "worker1", 8, 16000);
+
+        let parent_id = run_job_on(&cm, "parent", "worker1");
+        cm.preempt_job(parent_id, PreemptMode::Cancel).unwrap();
+        settle(&cm, parent_id, JobState::Cancelled);
+
+        // afterok: parent exited non-zero (preempted) → unsatisfiable; watcher cancels it.
+        let mut afterok_child = basic_spec("afterok-child");
+        afterok_child.dependency = vec![format!("afterok:{parent_id}")];
+        let afterok_id = submit_and_wait(&cm, afterok_child);
+
+        // afterany: fires regardless of exit status → satisfied once parent is terminal.
+        let mut afterany_child = basic_spec("afterany-child");
+        afterany_child.dependency = vec![format!("afterany:{parent_id}")];
+        let afterany_id = submit_and_wait(&cm, afterany_child);
+
+        cm.cancel_unsatisfiable_dependency_jobs();
+
+        // afterok cancelled: parent's non-zero exit makes it permanently unsatisfiable.
+        assert_eq!(
+            cm.get_job(afterok_id).unwrap().state,
+            JobState::Cancelled,
+            "afterok dependent must be cancelled after parent was preempted"
+        );
+
+        // afterany unblocked: parent is terminal, dependency is satisfied.
+        // pending_jobs() returns only schedulable (non-blocked) jobs.
+        assert!(
+            cm.pending_jobs().iter().any(|j| j.job_id == afterany_id),
+            "afterany dependent must be schedulable once parent is terminal"
         );
     }
 
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -2922,12 +2922,8 @@ impl ClusterManager {
             let fair_share = self
                 .fairshare_cache
                 .get(&job.spec.user, job.spec.account.as_deref().unwrap_or(""));
-            job.priority = compute_effective_priority(
-                job.priority,
-                fair_share,
-                age_minutes,
-                partition_tier,
-            );
+            job.priority =
+                compute_effective_priority(job.priority, fair_share, age_minutes, partition_tier);
         }
         drop(partitions);
 
@@ -4297,11 +4293,7 @@ impl ClusterManager {
     /// `priority` is stale). Takes `qos` pre-resolved so it can be reused,
     /// and `partitions` so callers iterating over multiple jobs don't pay
     /// for a separate lock acquisition per call.
-    pub(crate) fn current_effective_priority(
-        &self,
-        job: &Job,
-        partitions: &[Partition],
-    ) -> u32 {
+    pub(crate) fn current_effective_priority(&self, job: &Job, partitions: &[Partition]) -> u32 {
         let now = Utc::now();
         let age_minutes = (now - job.submit_time).num_minutes().max(0);
         let partition_tier =
@@ -10255,7 +10247,10 @@ mod tests {
 
         let burst_job = cm.get_job(burst_id).unwrap();
         let primus_job = cm.get_job(primus_id).unwrap();
-        assert_eq!(burst_job.priority, 100, "burst job base priority must equal qos.priority");
+        assert_eq!(
+            burst_job.priority, 100,
+            "burst job base priority must equal qos.priority"
+        );
         assert_eq!(
             primus_job.priority, 10000,
             "primus job base priority must equal qos.priority"
@@ -11837,8 +11832,7 @@ mod tests {
             },
         );
 
-        let priority =
-            cm.current_effective_priority(&job, &cm.get_partitions());
+        let priority = cm.current_effective_priority(&job, &cm.get_partitions());
         assert_eq!(
             priority, 9000,
             "multi-partition job should use the highest matched priority_tier (9), \
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -21,7 +21,7 @@ use spur_core::job::{
 };
 use spur_core::node::{Node, NodeEvent, NodeSource, NodeState};
 use spur_core::partition::{requested_partition_names, Partition, PreemptMode};
-use spur_core::qos::{check_qos_limits, qos_adjusted_priority, QosCheckResult};
+use spur_core::qos::{check_qos_limits, QosCheckResult};
 use spur_core::reservation::{self, normalize_node_list, running_jobs_overlap_start, Reservation};
 use spur_core::resource::{ResourceAllocations, ResourceSet};
 use spur_core::step::{JobStep, StepState, STEP_BATCH, STEP_RESERVED_MIN};
@@ -631,12 +631,29 @@ impl ClusterManager {
         let specs =
             expand_job_specs(spec, job_id).map_err(|e| SubmitError::invalid(e.to_string()))?;
 
-        for task_spec in specs {
+        for mut task_spec in specs {
             let task_id = if task_spec.array_job_id.is_some() {
                 self.next_job_id.fetch_add(1, Ordering::SeqCst)
             } else {
                 job_id
             };
+
+            // When the user didn't request an explicit --priority, use the QOS
+            // priority as the base so it seeds the multiplicative formula
+            // (fair-share × age × partition-tier). Without this, every job
+            // starts at DEFAULT_PRIORITY (1000) and the QOS delta is negligible
+            // compared to the age-driven effective value, so preemption never
+            // fires between tiers.
+            if task_spec.priority.is_none() {
+                if let Some(qos_name) = task_spec.qos.as_deref() {
+                    if let Some(qos) = self.qos_cache.get(qos_name) {
+                        if qos.priority > 0 {
+                            task_spec.priority = Some(qos.priority as u32);
+                        }
+                    }
+                }
+            }
+
             self.propose(WalOperation::JobSubmit {
                 job_id: task_id,
                 spec: Box::new(task_spec),
@@ -1521,7 +1538,8 @@ impl ClusterManager {
                 Ok(PreemptOutcome::Suspended)
             }
             PreemptMode::Cancel => {
-                self.complete_job(job_id, -1, JobState::Cancelled)?;
+                let resp = self.propose(WalOperation::JobPreemptCancel { job_id })?;
+                self.run_all_finalized_side_effects(&resp);
                 info!(job_id, "job preempted (cancel)");
                 Ok(PreemptOutcome::Killed)
             }
@@ -2909,7 +2927,6 @@ impl ClusterManager {
                 fair_share,
                 age_minutes,
                 partition_tier,
-                &qos_by_job[&job.job_id],
             );
         }
         drop(partitions);
@@ -4280,10 +4297,9 @@ impl ClusterManager {
     /// `priority` is stale). Takes `qos` pre-resolved so it can be reused,
     /// and `partitions` so callers iterating over multiple jobs don't pay
     /// for a separate lock acquisition per call.
-    pub(crate) fn current_effective_priority_with_qos(
+    pub(crate) fn current_effective_priority(
         &self,
         job: &Job,
-        qos: &Qos,
         partitions: &[Partition],
     ) -> u32 {
         let now = Utc::now();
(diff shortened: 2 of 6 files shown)


crates/spurctld/src/cluster.rs
```diff
@@ -631,12 +631,29 @@ impl ClusterManager {
...
+            // When the user didn't request an explicit --priority, use the QOS
+            // priority as the base so it seeds the multiplicative formula
+            // (fair-share × age × partition-tier). Without this, every job
+            // starts at DEFAULT_PRIORITY (1000) and the QOS delta is negligible
+            // compared to the age-driven effective value, so preemption never
+            // fires between tiers.
+            if task_spec.priority.is_none() {
+                if let Some(qos_name) = task_spec.qos.as_deref() {
+                    if let Some(qos) = self.qos_cache.get(qos_name) {
+                        if qos.priority > 0 {
```
yansun1996: `Qos::priority` is `i32`, and the previous `qos_adjusted_priority` applied it as a signed delta clamped to `>= 1`, so penalty tiers really did penalise. With the `> 0` guard, a QOS at `-500` leaves `spec.priority` as `None` and falls back to `DEFAULT_PRIORITY` (1000), while a QOS at `100` seeds base 100 — the penalty tier ends up ranking 10x above the boost tier. Any QOS with `0 < priority < 1000` also now ranks below jobs with no QOS at all.

Since operators configure these as deltas today, this would silently reorder existing clusters on upgrade. Would something like `task_spec.priority = Some(DEFAULT_PRIORITY.saturating_add_signed(qos.priority).max(1))` work? That keeps the ordering monotonic across the full `i32` range and preserves the meaning of already-deployed values.

crates/spurctld/src/cluster.rs
```diff
@@ -631,12 +631,29 @@ impl ClusterManager {
...
             } else {
                 job_id
             };
+
+            // When the user didn't request an explicit --priority, use the QOS
+            // priority as the base so it seeds the multiplicative formula
+            // (fair-share × age × partition-tier). Without this, every job
+            // starts at DEFAULT_PRIORITY (1000) and the QOS delta is negligible
+            // compared to the age-driven effective value, so preemption never
+            // fires between tiers.
```
yansun1996: Minor: this is a six-line comment block — the repo leans towards one or two lines for the *why*, with the longer rationale living in the commit message or PR description. Something like `// Seed the base from QOS so it feeds the multiplicative formula rather than
// being a negligible delta on top of it.` would cover it.

yansun1996 (commented):
Thanks for digging into this — the diagnosis reads right, and separating the QOS seeding from the preemption terminal-state fix is a sensible decomposition. A few things I think need another pass before this lands.

**1. Negative and zero QOS priorities are now silently dropped, which inverts the tiers.** `Qos::priority` is `i32`, and the old `qos_adjusted_priority` applied it as a signed delta (`base + qos.priority`, clamped to >= 1). The new seed only fires on `qos.priority > 0`, so a penalty tier at `-500` keeps base 1000 while a boost tier at `100` gets base 100 — the penalty tier now outranks the boost tier 10x. More broadly, any QOS with `0 < priority < 1000` now ranks *below* jobs with no QOS at all. Since operators set these values as deltas today, this would silently reorder existing clusters. Would seeding across the full signed range work, e.g. `DEFAULT_PRIORITY.saturating_add_signed(qos.priority).max(1)`?

**2. `Preempted` is `is_finalized()` but not `is_terminal()`, and the cancel path parks the job there permanently.** The requeue path passes through `Preempted` transiently on its way back to `Pending`; the new cancel path has no such exit. `check_dependencies` treats a non-terminal target as `Waiting` for `afterok`/`afternotok`/`aftercorr`, and `AfterAny` gates on `is_terminal()` — so every dependent of a preempt-cancelled job would wait indefinitely. `aggregate_array_state` has the same issue (it bails while any task is non-terminal). Previously the victim landed in `Cancelled` and dependents resolved immediately. Either land the job in a terminal state after the `Preempted` hop, or teach the dependency/array predicates that a cancel-preempted job is finished.

**3. Clearing the allocation fields diverges from `JobComplete`** — details inline.

**4. `pending_jobs_applies_qos_priority_adjustment` now passes trivially** — its low tier is `-500`, which the new code discards, so the assertion holds for the wrong reason. It would stay green even if the guard were `> 999`.

**5. `cargo fmt --check` looks like it would fail** in two spots (inline).

**On scope:** the branch also carries five documentation commits adding `docs/ARCHITECTURE_AND_ROADMAP.md` and `docs/spur-for-helios.md` (~954 lines, roughly 2.5x the size of the fix). Since PRs here are squash-merged, they would land under a `fix(spurctld):` commit title. Could those move to their own PR? Two reasons beyond scope hygiene: neither file is listed in `docs/sphinx/_toc.yml.in`, so Sphinx will pick them up as orphaned documents (a failure under `-W`); and `spur-for-helios.md` reads more like an internal design memo than user documentation — it documents `[fabric]` config sections and `sinfo`/`sbatch` flags that do not exist in `config.rs` or the CLI yet, and goes into unreleased fabric and key-distribution detail. Probably worth an accuracy and confidentiality pass before it sits in a public docs tree. `ARCHITECTURE_AND_ROADMAP.md` also references `spurdbd`/`spurrestd`, which Spur does not ship (AGENTS.md is explicit that accounting and REST are in-process in `spurctld`), and states "no external database required" while the same doc later lists PostgreSQL.

**Two things worth calling out in the PR description rather than fixing here:** `scontrol update JobId=N QOS=...` and `sacctmgr modify qos ... set priority=` no longer affect an already-submitted job now that the value is baked in at submit time; and `release_job` unconditionally proposes `DEFAULT_PRIORITY`, so hold+release will wipe a job's QOS-derived base. The second is pre-existing, but this change makes it consequential — probably a follow-up issue rather than work for this PR.

**Test suggestions:** a submit-path test for negative and zero `qos.priority` (that gap is what lets [1] through); a dependency test asserting an `afterok` dependent resolves after its target is preempt-cancelled; and a `JobPreemptCancel` serde round-trip test alongside `preempt_requeue_op_round_trips`.

crates/spurctld/src/cluster.rs
```diff
@@ -4280,20 +4293,15 @@ impl ClusterManager {
     /// `priority` is stale). Takes `qos` pre-resolved so it can be reused,
     /// and `partitions` so callers iterating over multiple jobs don't pay
     /// for a separate lock acquisition per call.
-    pub(crate) fn current_effective_priority_with_qos(
-        &self,
-        job: &Job,
-        qos: &Qos,
-        partitions: &[Partition],
-    ) -> u32 {
+    pub(crate) fn current_effective_priority(&self, job: &Job, partitions: &[Partition]) -> u32 {
```
yansun1996, line 4289: Small drift: the doc still says "Takes `qos` pre-resolved so it can be reused", but the `qos` parameter is gone in this change. Worth dropping that clause.

Relatedly, `qos_adjusted_priority` in `spur-core/src/qos.rs` now has no production callers left (only its own unit tests and the parity tests), and its doc comment states the opposite of the new behaviour — it explains that QOS is applied *after* the formula because "applying it earlier would let those factors amplify it". Worth either removing it or updating that doc so it does not mislead the next reader.

crates/spurctld/src/cluster.rs
```diff
@@ -4628,6 +4636,73 @@ impl ClusterManager {
...
+                        return ClientResponse::default();
+                    }
+                    if let Err(e) = job.transition(JobState::Preempted) {
+                        warn!(job_id = *job_id, error = %e, "invalid preempt-cancel transition in WAL apply");
+                        return ClientResponse::default();
+                    }
+                    job.exit_code = Some(-1);
+                    job.end_time = Some(timestamp);
+                    if let Some(since) = job.suspended_at.take() {
+                        job.suspended_secs += (timestamp - since).num_seconds().max(0);
```
yansun1996, line 4650: `Preempted` is included in `is_finalized()` but deliberately excluded from `is_terminal()` — the comment on `is_finalized` notes it "may still requeue". The requeue arm relies on that: it passes through `Preempted` and then transitions back to `Pending`. This cancel arm has no such exit, so the job stays in a non-terminal state for the rest of its life.

That has some reach beyond this arm:

- `check_dependencies` returns `Waiting` for a non-terminal target on `afterok`/`afternotok`/`aftercorr`, and `AfterAny` gates on `s.is_terminal()` — so dependents of a cancel-preempted job would never be released. Previously the victim became `Cancelled` and they resolved immediately.
- `aggregate_array_state` returns `None` while any task is non-terminal, so an array containing a cancel-preempted task never aggregates.
- `evict_expired_terminal_jobs` filters on `is_finalized()` so it would normally reap the job, but it spares anything a live dependent still references — which is exactly the dependent that is now stuck.

Two options I can see: transition on to a terminal state after the `Preempted` hop (`Preempted -> Cancelled` is already legal) and report `Preempted` to accounting only, or extend the dependency/array predicates to treat a cancel-preempted job as a finished run. Either way a test covering an `afterok` dependent of a preempt-cancelled target would be good insurance.

crates/spurctld/src/cluster.rs
```diff
@@ -4628,6 +4636,73 @@ impl ClusterManager {
...
+                    allocated_resources = job.allocated_resources.clone();
+                    per_node_map = job.per_node_alloc.clone();
+                    job.node_completions.clear();
+                    job.allocated_nodes.clear();
+                    job.allocated_resources = None;
+                    job.per_node_alloc.clear();
+                }
+                if let Some(ref total) = allocated_resources {
+                    let node_count = freed_nodes.len().max(1) as u32;
+                    for name in &freed_nodes {
```
yansun1996, line 4665: These three clears follow `JobPreemptRequeue`, but I think `JobComplete` is the closer model here. `JobPreemptRequeue` clears `allocated_nodes`/`allocated_resources`/`per_node_alloc` because the job re-runs and must not carry stale placement; `JobComplete` deliberately preserves them (it only clears `node_completions`) so the finished job keeps a NodeList.

Since cancel-mode preemption is end-of-life, clearing them means `scontrol show job` and accounting lose the NodeList, and `run_job_finalized_side_effects` -> `run_epilog_slurmctld` reads `j.allocated_nodes.join(",")` *after* this apply, so `epilog_slurmctld` would fire with `nodelist=""`.

The local `freed_nodes` / `allocated_resources` / `per_node_map` copies are already taken above, so the dealloc below still works if these three clears are dropped.

crates/spurctld/src/cluster.rs
```diff
@@ -4628,6 +4636,73 @@ impl ClusterManager {
...
+                            node.alloc_resources.subtract(&slice);
+                            node.update_state_from_alloc();
+                            if node.state == NodeState::Draining
+                                && node.alloc_resources.cpus == 0
+                                && !node.alloc_resources.has_devices()
+                            {
+                                node.state = NodeState::Drain;
+                            }
+                        }
+                    }
```
yansun1996, line 4687: This makes the fifth verbatim copy of the "subtract the per-node slice, `update_state_from_alloc`, then `Draining && cpus == 0 && !has_devices -> Drain`" block, each with its own `per_node_alloc`-missing scalar fallback (the others are in `evict_job_locked`, `JobPreemptRequeue`, `JobNodeComplete`, and `JobComplete`).

Not introduced by this PR, but adding a fifth copy seems like the moment to extract a shared helper — five hand-maintained copies of the drain rule will drift. Happy for it to be a follow-up if you would rather keep this change tight.

crates/spurctld/src/cluster.rs
```diff
@@ -11540,13 +11722,14 @@ mod tests {
             ..Default::default()
         });
 
+        // No explicit --priority: QOS priority becomes the base seed for the
+        // multiplicative formula. high-QoS (5000) should outrank low-QoS
+        // (negative penalty → falls back to DEFAULT_PRIORITY 1000).
         let mut low = basic_spec("low");
-        low.priority = Some(1000);
         low.qos = Some("low".into());
```
yansun1996, line 11732: This test no longer exercises the thing its name describes. The low tier is configured at `-500`, which the new `> 0` guard discards entirely, so the low job just gets `DEFAULT_PRIORITY` and `5000 > 1000` holds trivially — it would stay green even if the guard were `> 999`. The updated comment says as much ("negative penalty -> falls back to DEFAULT_PRIORITY 1000"), which is really the bug in [1] rather than intended behaviour.

Could this assert on the absolute value too (e.g. that the low tier lands *below* `DEFAULT_PRIORITY`)? That would make it fail against the current seeding and pass once negatives are handled.

crates/spurctld/src/cluster.rs
```diff
@@ -11649,8 +11832,7 @@ mod tests {
             },
         );
 
-        let priority =
-            cm.current_effective_priority_with_qos(&job, &Qos::default(), &cm.get_partitions());
+        let priority = cm.current_effective_priority(&job, &cm.get_partitions());
         assert_eq!(
             priority, 9000,
```
yansun1996, line 11843: Now that the `qos` argument is gone this fits on one line, and `cargo fmt` would collapse it — looks like it would trip `cargo fmt --check` in CI. Same in `scheduler_loop.rs` around line 581, where the `.map(|j| { (j.job_id, ...) })` closure collapses to `.map(|j| (j.job_id, cluster.current_effective_priority(j, partitions)))` after the shorter call.

crates/spur-core/src/wal.rs
```diff
@@ -127,6 +127,12 @@ pub enum WalOperation {
         job_id: JobId,
         begin_time: chrono::DateTime<chrono::Utc>,
     },
+    /// Preempt a running job with cancel (job ends in PREEMPTED, not requeued).
+    /// Single atomic entry so a leadership change cannot strand the job running
+    /// with its allocations held.
+    JobPreemptCancel {
```
yansun1996, line 135: Every recent variant here has a serde round-trip test (`preempt_requeue_op_round_trips`, and the `JobSuspend`/`JobResume`/`JobTimeLimitSignaled` ones next to it) — worth adding a matching one for `JobPreemptCancel` so it is covered by the same convention.

Also worth a line in the PR description: `WalOperation` is externally tagged with no unknown-variant fallback, so once a `JobPreemptCancel` entry is in the log an older controller cannot replay it and will refuse to start under strict recovery. That is inherent to adding a variant rather than something to fix here, but it does mean all controllers need upgrading before the first cancel-mode preemption fires, and no rollback after that point.

yansun1996 (dismissed):
Follow-up commit addresses everything from the previous round — signed QOS-priority seeding, the `Preempted → Cancelled` terminal-state fix (so dependents/arrays unblock), allocation-field handling now matching `JobComplete`, the trivially-passing test tightened, `qos_adjusted_priority` removed cleanly, the WAL round-trip test added, and formatting fixed. Ran clippy/fmt/tests locally against the branch tip and everything is green.

A few small non-blocking things worth a look, none of which affect correctness:

- `crates/spur-core/src/wal.rs` — the doc comment on `JobPreemptCancel` says the job "ends in PREEMPTED," but it actually ends in `Cancelled` (`Preempted` is transient, reported to accounting via `jobs_finalized`). Worth a quick reword so it doesn't mislead a future reader.
- `docs/admin-guide/accounting.rst` — same wording in the `preemptmode` table ("cancel (job ends in PREEMPTED)"); since this is user-facing, might be worth clarifying that the job's terminal/queryable state is `CANCELLED` and `PREEMPTED` is the accounting record.
- The new comment in the `JobPreemptCancel` WAL-apply arm runs a bit long (4 lines) relative to the rest of the file's style — could probably be trimmed to one or two.
- `t21_14_qos_priority_boost` / `t21_15_qos_priority_penalty` / `t21_16_qos_priority_floor` in `t21_acctmgr.rs` now just assert `DEFAULT_PRIORITY.saturating_add_signed(N).max(1)` against literals directly, with no call into the actual submit/QOS-seeding path — they'd stay green even if that logic broke. Might be worth pointing them at the real code path (or dropping them, since the new `cluster.rs` tests already cover the seeding behavior).
- Since dependents unblocking is one of the fix's explicit claims, a small test that submits an `afterok`/`afterany` dependent, preempt-cancels the parent, and asserts the dependent resolves would be a nice addition alongside the existing coverage.

None of these block — approving.

crates/spurctld/src/cluster.rs
sajmera-pensando: Fixed — switched to `DEFAULT_PRIORITY.saturating_add_signed(qos.priority).max(1)`. Penalty tiers (negative priority) now correctly rank below DEFAULT_PRIORITY rather than falling back to 1000 and outranking boost tiers. The t21_14–16 tests in t21_acctmgr.rs have been rewritten to exercise the new formula, and `pending_jobs_applies_qos_priority_adjustment` now asserts the penalty tier lands below DEFAULT_PRIORITY.

crates/spurctld/src/cluster.rs
sajmera-pensando: Trimmed to two lines.

crates/spurctld/src/cluster.rs
sajmera-pensando: Both done. Updated the doc comment on `current_effective_priority` to drop the stale 'Takes `qos` pre-resolved' clause. Removed `qos_adjusted_priority` entirely — the function, its three tests in `qos.rs`, and the three dependent tests in `t21_acctmgr.rs` (rewritten to cover the new seeding formula instead).

crates/spurctld/src/cluster.rs
sajmera-pensando: Fixed. `JobPreemptCancel` now does `Running → Preempted → Cancelled` in the same WAL apply. `jobs_finalized` still reports `Preempted` so accounting and the PREEMPTED counter work correctly, but the job's live state is `Cancelled` — terminal — so `afterok`/`AfterAny` dependents and array aggregation resolve immediately.

crates/spurctld/src/cluster.rs
sajmera-pensando: Fixed — dropped the three `.clear()` calls. Followed `JobComplete`'s model: the local `freed_nodes` / `allocated_resources` / `per_node_map` copies still drive node deallocation, and the originals stay on the job so `scontrol show job` retains NodeList and `epilog_slurmctld` fires with the correct nodelist.

crates/spurctld/src/cluster.rs
sajmera-pensando: Agreed — this is the right moment to extract it. Leaving as a follow-up to keep this PR focused on the SPUR-199 fixes.

crates/spurctld/src/cluster.rs
sajmera-pensando: Fixed. Added `assert!(low_priority < DEFAULT_PRIORITY)` so the test actually fails if the penalty case is broken. The assertion would have caught the original `> 0` guard regression.

crates/spurctld/src/cluster.rs
sajmera-pensando: Fixed in the previous rustfmt commit.

crates/spur-core/src/wal.rs
sajmera-pensando: Added `preempt_cancel_op_round_trips` next to the existing requeue one. On the rollback caveat: added a 'WAL compatibility' section to the PR description — once a `JobPreemptCancel` entry is in the log, all controllers must be at this version before the first cancel-mode preemption fires; no rollback after that point.

PR #655, opened 2026-08-17
feat(spurctld): k0s partial-Ready and online worker add/remove

spraveenio:
## What this does

Three related improvements to the SPUR-managed k0s cluster so a running cluster
survives a bad node and can be resized without a destructive teardown.

**1. Partial-Ready.** One slow or unreachable worker used to pin the whole
cluster in `Provisioning` until a timeout flipped it to `Degraded`. Now the
cluster reaches `Ready` as soon as the control-plane etcd quorum is active.
Stragglers keep converging in the background and each carries its own per-node
reason in `spur k8s status`. `Degraded` is now reserved for an actual loss of
control-plane quorum, not a single down worker.

**2. Add a worker to a running cluster** — `spur k8s add-nodes`:
```
spur k8s add-nodes --nodes gpu[09-12]      # or --partition / --selector
```
Grows a scoped cluster's member set on the fly; the new nodes are assigned a
role/IP/CIDR, joined, and meshed by the existing reconcile loop. No
`down --reset` needed.

**3. Remove a worker from a running cluster** — `spur k8s remove-nodes`:
```
spur k8s remove-nodes --nodes gpu09 [--drain-timeout 120] [--force]
```
Graceful, k8s-aware: cordon → drain (PDB-aware) → delete the k8s node → stop and
`k0s reset` the component → clear its role and drop it from the member set. The
departing node's k0s unit + join token are purged so it can't restart and
rejoin. Re-adding the same node later rejoins it fresh.

## Why

A single node that never came up could stall an entire cluster with no clear
signal, and the only way to change cluster size was to tear the whole thing
down. This makes the cluster tolerant of partial failure and lets operators
grow/shrink it online.

## Design choices / trade-offs

- **Readiness is gated on control-plane quorum, not every worker** — mirrors how
  Slurm and Kubernetes treat availability (control plane up = usable; unhealthy
  nodes are simply excluded). No new cluster phase, so no client has to learn a
  new value.
- **`remove-nodes` is destructive on purpose.** It runs `k0s reset` on the
  departing node (wipes its k0s state) so no stale etcd/cert state is stranded.
  Re-adding re-pulls images / re-seeds state. For a temporary "stop scheduling
  here", use `spur node drain` instead (that's a separate, non-destructive
  scheduling-layer operation).
- **Refuses to yank a busy node.** `remove-nodes` rejects a node with running
  jobs unless `--force`, matching `scontrol delete` semantics. It also refuses
  control-plane nodes (changing the control plane / etcd quorum is out of scope
  here and tracked separately).

## Compatibility

- **Proto**: appends `ClusterAddNodes` / `ClusterRemoveNodes` (controller) and
  `DrainK8sNode` (agent) with new tags only — no existing field is renumbered or
  retyped, so FFI/REST/wire compat is preserved.
- **Persisted state**: two new append-only WAL ops fold into the existing k0s
  member set; old snapshots and log entries replay unchanged.

## Test plan / results

Unit tests (deterministic, no external services); the async drain/reset path is
mocked at the RPC boundary and verified on hardware separately:

- [x] `cargo test --locked` across the touched crates:
  - `spur-core`: 520 passed
  - `spurctld`: 775 passed (20 ignored)
  - `spurd`: 224 passed (2 ignored)
  - `spur-k8s`: 147 passed
  - `spur-cli`: passed
- [x] `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` — no warnings
- [x] `cargo fmt --check` — clean

New coverage includes: control-plane quorum helper (1/1, 2/3, 3/5), the
Ready/Degraded edge decisions, member-set add/subtract helpers, the add/remove
RPC guard paths (control-plane reject, unregistered reject, running-jobs reject,
whole-inventory reject, down-cluster reject), WAL round-trip for the new ops,
unit+token purge on reset, and CLI parsing for the new subcommands.

Manually verified end-to-end on a 2-node cluster:
- partial-Ready: cluster reaches `ready` on control-plane quorum with a worker
  down; the worker self-heals to active once it recovers, with no re-up.
- add: a node added to a running scoped cluster joins and becomes a Ready k8s
  worker.
- remove: the node is drained, its k8s node object deleted, its k0s unit + token
  purged (stays down), and it is dropped from the member set; re-adding rejoins
  it fresh.

### DIFF
--- a/crates/spur-cli/src/k8s.rs
+++ b/crates/spur-cli/src/k8s.rs
@@ -8,7 +8,8 @@ use clap::{Parser, Subcommand};
 
 use spur_proto::proto::slurm_controller_client::SlurmControllerClient;
 use spur_proto::proto::{
-    ClusterDownRequest, ClusterKubeconfigRequest, ClusterStatusRequest, ClusterUpRequest,
+    ClusterAddNodesRequest, ClusterDownRequest, ClusterKubeconfigRequest,
+    ClusterRemoveNodesRequest, ClusterStatusRequest, ClusterUpRequest,
 };
 
 /// Manage the SPUR-provisioned k0s cluster.
@@ -52,6 +53,33 @@ pub enum K8sCommand {
         /// Scope the cluster to nodes matching every key=value label (repeatable).
         #[arg(long = "selector", value_parser = parse_key_val)]
         selector: Vec<(String, String)>,
+    },
+    /// Add worker nodes to a running cluster (scoped clusters only; no down/reset needed).
+    AddNodes {
+        /// Nodes to add (hostlist, e.g. "gpu[09-12]"), unioned with --partition/--selector.
+        #[arg(long)]
+        nodes: Option<String>,
+        /// Add a partition's nodes.
+        #[arg(long)]
+        partition: Option<String>,
+        /// Add nodes matching every key=value label (repeatable).
+        #[arg(long = "selector", value_parser = parse_key_val)]
+        selector: Vec<(String, String)>,
+    },
+    /// Remove worker nodes from a running cluster: cordon + drain, then `k0s reset` the node
+    /// (destructive — the node re-downloads/re-seeds on a later add). Use `spur node drain` instead
+    /// for a temporary "stop scheduling here".
+    RemoveNodes {
+        /// Worker nodes to remove (hostlist, e.g. "gpu[09-12]").
+        #[arg(long)]
+        nodes: String,
+        /// Max seconds to wait for the k8s drain per node (0 = server default).
+        #[arg(long)]
+        drain_timeout: Option<u32>,
+        /// Proceed even if a node has running jobs (they are left running — this only skips the
+        /// busy-node check) or its drain does not complete (bypasses PodDisruptionBudgets).
+        #[arg(long)]
+        force: bool,
     },
     /// Tear the k0s cluster down.
     Down {
@@ -113,6 +141,16 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
             )
             .await
         }
+        K8sCommand::AddNodes {
+            nodes,
+            partition,
+            selector,
+        } => cmd_add_nodes(&controller, nodes, partition, selector).await,
+        K8sCommand::RemoveNodes {
+            nodes,
+            drain_timeout,
+            force,
+        } => cmd_remove_nodes(&controller, nodes, drain_timeout, force).await,
         K8sCommand::Down { reset } => cmd_down(&controller, reset).await,
         K8sCommand::Status => cmd_status(&controller).await,
         K8sCommand::Kubeconfig { user, admin } => cmd_kubeconfig(&controller, user, admin).await,
@@ -201,6 +239,61 @@ async fn cmd_up(
         println!("k0s cluster up requested: {}", resp.message);
     } else {
         eprintln!("k0s cluster up NOT accepted: {}", resp.message);
+    }
+    for n in resp.nodes {
+        println!("  {} [{}] {}", n.node, n.role, n.component_state);
+    }
+    Ok(())
+}
+
+async fn cmd_add_nodes(
+    controller: &str,
+    nodes: Option<String>,
+    partition: Option<String>,
+    selector: Vec<(String, String)>,
+) -> Result<()> {
+    let selector = selector_map(selector)?;
+    let mut client = SlurmControllerClient::new(spur_client::connect_channel(controller).await?);
+    let resp = client
+        .cluster_add_nodes(ClusterAddNodesRequest {
+            nodes: nodes.unwrap_or_default(),
+            partition: partition.unwrap_or_default(),
+            selector,
+            caller: effective_user(),
+        })
+        .await?
+        .into_inner();
+    if resp.accepted {
+        println!("k0s add-nodes requested: {}", resp.message);
+    } else {
+        eprintln!("k0s add-nodes NOT accepted: {}", resp.message);
+    }
+    for n in resp.nodes {
+        println!("  {} [{}] {}", n.node, n.role, n.component_state);
+    }
+    Ok(())
+}
+
+async fn cmd_remove_nodes(
+    controller: &str,
+    nodes: String,
+    drain_timeout: Option<u32>,
+    force: bool,
+) -> Result<()> {
+    let mut client = SlurmControllerClient::new(spur_client::connect_channel(controller).await?);
+    let resp = client
+        .cluster_remove_nodes(ClusterRemoveNodesRequest {
+            nodes,
+            caller: effective_user(),
+            drain_timeout_secs: drain_timeout,
+            force: Some(force),
+        })
+        .await?
+        .into_inner();
+    if resp.accepted {
+        println!("k0s remove-nodes requested: {}", resp.message);
+    } else {
+        eprintln!("k0s remove-nodes NOT accepted: {}", resp.message);
     }
     for n in resp.nodes {
         println!("  {} [{}] {}", n.node, n.role, n.component_state);
@@ -384,6 +477,47 @@ mod tests {
         assert!(matches!(args.command, K8sCommand::Down { reset: true }));
         let args = K8sArgs::try_parse_from(["k8s", "status"]).unwrap();
         assert!(matches!(args.command, K8sCommand::Status));
+    }
+
+    #[test]
+    fn parses_add_nodes_scope_flags() {
+        let args = K8sArgs::try_parse_from(["k8s", "add-nodes", "--nodes", "gpu[09-12]"]).unwrap();
+        match args.command {
+            K8sCommand::AddNodes { nodes, .. } => assert_eq!(nodes.as_deref(), Some("gpu[09-12]")),
+            _ => panic!("wrong command"),
+        }
+    }
+
+    #[test]
+    fn parses_remove_nodes_with_flags() {
+        let args = K8sArgs::try_parse_from([
+            "k8s",
+            "remove-nodes",
+            "--nodes",
+            "gpu[09-12]",
+            "--drain-timeout",
+            "300",
+            "--force",
+        ])
+        .unwrap();
+        match args.command {
+            K8sCommand::RemoveNodes {
+                nodes,
+                drain_timeout,
+                force,
+            } => {
+                assert_eq!(nodes, "gpu[09-12]");
+                assert_eq!(drain_timeout, Some(300));
+                assert!(force);
+            }
+            _ => panic!("wrong command"),
+        }
+    }
+
+    #[test]
+    fn remove_nodes_requires_nodes() {
+        // --nodes is mandatory for remove-nodes.
+        assert!(K8sArgs::try_parse_from(["k8s", "remove-nodes"]).is_err());
     }
 
     #[test]
--- a/crates/spur-core/src/k0s.rs
+++ b/crates/spur-core/src/k0s.rs
@@ -218,6 +218,33 @@ pub fn validate_control_plane_replicas(replicas: u32) -> Result<(), String> {
     ))
 }
 
+/// The majority of `n` (`n/2 + 1`) — the number of control planes whose k0s unit must be active for
+/// the reconcile loop to gate `Ready` (a proxy for etcd quorum, not a direct etcd health check).
+/// `quorum(0)` is 0 (no control plane assigned yet — never a satisfiable quorum).
+pub fn quorum(n: usize) -> usize {
+    if n == 0 {
+        return 0;
+    }
+    n / 2 + 1
+}
+
+#[cfg(test)]
+mod quorum_tests {
+    use super::quorum;
+
+    #[test]
+    fn quorum_is_majority_for_valid_cp_counts() {
+        assert_eq!(quorum(1), 1);
+        assert_eq!(quorum(3), 2);
+        assert_eq!(quorum(5), 3);
+    }
+
+    #[test]
+    fn quorum_of_zero_is_zero() {
+        assert_eq!(quorum(0), 0);
+    }
+}
+
 /// Which k0s role a node's spurd-owned systemd unit runs.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "snake_case")]
--- a/crates/spurctld/src/cluster_k8s.rs
+++ b/crates/spurctld/src/cluster_k8s.rs
@@ -22,8 +22,9 @@ use spur_net::address::AddressPool;
 use spur_net::mesh::{MeshMembership, MeshNode};
 use spur_proto::proto::slurm_agent_client::SlurmAgentClient;
 use spur_proto::proto::{
-    ClusterNodeStatus, CreateK0sJoinTokenRequest, GetClusterComponentStatusRequest,
-    GetKubeconfigRequest, StartClusterComponentRequest, StopClusterComponentRequest,
+    ClusterNodeStatus, CreateK0sJoinTokenRequest, DeleteK8sNodeRequest, DrainK8sNodeRequest,
+    GetClusterComponentStatusRequest, GetKubeconfigRequest, StartClusterComponentRequest,
+    StopClusterComponentRequest,
 };
 
 use crate::cluster::ClusterManager;
@@ -151,11 +152,12 @@ pub(crate) async fn reconcile_phase(
                 warn!(error = %e, "k0s provisioning assignment failed; will retry next tick");
                 return true;
             }
-            let errors = converge_provisioning(cluster, net, join_tokens).await;
+            let (errors, active_cp) = converge_provisioning(cluster, net, join_tokens).await;
             // converge may have flipped us to Ready this tick; only degrade a still-provisioning
-            // cluster that has blown its deadline.
+            // cluster that has blown its deadline. Reuse converge's control-plane liveness (no second
+            // per-node status sweep).
             if timed_out && cluster.k0s_state().phase == K0sPhase::Provisioning {
-                degrade_stuck_cluster(cluster, net, join_tokens).await;
+                degrade_stuck_cluster(cluster, net, join_tokens, &active_cp).await;
             }
             errors > 0
         }
@@ -673,15 +675,17 @@ fn controller_k0s_config(net: &ClusterNetworking, node: &spur_core::node::Node) 
     )
 }
 
-/// Start any assigned component that is not yet active; when all are active, mark the cluster Ready.
+/// One convergence pass: start any assigned component not yet active and, once the control-plane
+/// quorum is up, mark the cluster Ready (partial-Ready — workers keep converging in the background).
 /// `join_tokens` caches each worker's minted join token across ticks so we mint once per join.
-/// Returns the number of errors encountered this iteration (currently join-token mint failures),
-/// so the reconcile loop can surface them via `spur_k8s_reconcile_errors_total`.
+/// Returns the error count for this pass (currently join-token mint failures, surfaced via
+/// `spur_k8s_reconcile_errors_total`) plus the set of control-plane nodes observed `active` this
+/// pass, so the caller can decide the degrade/Ready edge without a second per-node status sweep.
 async fn converge_provisioning(
     cluster: &ClusterManager,
     net: &ClusterNetworking,
     join_tokens: &mut HashMap<String, String>,
-) -> u64 {
+) -> (u64, HashSet<String>) {
     let mut errors = 0u64;
     let assigned: Vec<_> = cluster
         .get_nodes()
@@ -690,11 +694,13 @@ async fn converge_provisioning(
         .collect();
     if assigned.is_empty() {
         join_tokens.clear();
-        return errors;
+        return (errors, HashSet::new());
     }
     let bootstrap = cluster.k0s_state().bootstrap();
-    let mut all_active = true;
     let mut bootstrap_active = false;
+    // Active control-plane nodes this tick: Ready is gated on their quorum (partial-Ready), not on
+    // every worker being up.
+    let mut active_cp: HashSet<String> = HashSet::new();
     // Bootstrap control-plane first: it seeds etcd (tokenless) and its k0s API must answer before any
     // secondary control-plane or worker can mint a join token. A Single node is always the bootstrap.
     for node in &assigned {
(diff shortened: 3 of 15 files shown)


crates/spurctld/src/server.rs
```diff
@@ -2572,6 +2572,189 @@ impl SlurmController for ControllerService {
...
+                     (tear down with `spur k8s down --reset` to change the control plane)"
+                )));
+            }
+            // Refuse a busy node unless forced (Slurm `scontrol delete` semantics).
+            if !req.force.unwrap_or(false) && self.cluster.node_has_running_jobs(n) {
+                return Err(Status::failed_precondition(format!(
+                    "node {n} has running jobs; drain them or pass --force to evict"
+                )));
+            }
+        }
```
yansun1996, line 2749: This validation block is missing three preconditions that `cluster_add_nodes` has (or that its docs claim):

- **Whole-inventory cluster.** `cluster_add_nodes` rejects when `member_nodes.is_empty()`; there's no equivalent here. On such a cluster the subtraction is a no-op, `is_member` still returns true, and `provision_assignments` re-assigns a role next tick — so the node gets drained and `k0s reset` for nothing, then re-enrolls.
- **Membership.** Nothing checks `state.is_member(n)`, so an out-of-scope but registered node can be drained and destructively reset (reachable under `--force`, which is exactly the flag an operator reaches for here).
- **Would-empty.** Nothing stops the set shrinking to empty, which flips the cluster to whole-inventory. The control-plane rejection above only prevents this incidentally.

Would it be worth adding all three here, failing closed with a message pointing at `spur k8s down`? That would also make the doc comments on `subtract_member_nodes` and `remove_k0s_member_nodes` accurate.

yansun1996 (commented):
Nice work — partial-Ready is the right call, and the add/remove flows are cleanly structured. A few things I'd like to sort out before this merges, mostly clustered in `cluster_remove_nodes`.

**1. `cluster_remove_nodes` is missing the precondition block that `cluster_add_nodes` has.** There's no whole-inventory guard, no `is_member` check, and nothing stops the member set shrinking to empty. Because an empty `member_nodes` means "whole inventory", all three paths end the same way: the node is drained and `k0s reset`, then `provision_assignments` re-enrolls it on the next tick. The destructive half is permanent; the removal isn't. Three doc comments currently assert that the handler guards this — adding the guard would make them true.

**2. The member-set subtraction is proposed after the whole per-node loop**, while `clear_node_k0s` runs inside each `remove_worker`. In between, the node is still a member with no role — exactly what `provision_assignments` picks up. With `RECONCILE_INTERVAL` at 30s and the default drain at 120s per node, a multi-node removal will hit this rather than race it, and a leader flip mid-loop makes it permanent.

**3. `--force` is documented in three places as evicting/failing running jobs**, but it only skips the busy-node check — the jobs keep running on a node whose k0s state is about to be wiped. Either wire in the eviction the way `ClusterManager::remove_node` does, or reword.

**4. Both new responses document `nodes` as the resulting per-node status**, but `node_statuses()` filters to nodes that already have a role and hardcodes `component_state: "unknown"`, so a just-added node never appears in the add response.

**Testing.** The new RPC tests are all rejection paths that return before `remove_worker`, so the removal loop, the member-set subtraction, and the `--force` branch are uncovered — a mock-agent-driven happy path would have surfaced [1] and [2]. Similarly `control_plane_quorum_met` is covered as a pure function, but `degrade_stuck_cluster`'s new early return isn't, and that's the branch where a regression tears down a healthy cluster. The `K0sMemberNodesRemove` apply arm also has no test (the add side does, via `cluster_add_nodes_unions_into_scoped_member_set`).

Also, outside the diff: `cluster_k8s.rs:677` ("when all are active, mark the cluster Ready") is now stale. And as a possible follow-up — add-nodes onto a `Single`-role control plane looks like it would produce an unjoinable worker, since `provision_assignments` never re-rolls an already-assigned CP. That hazard is pre-existing, but this widens its reach.

One style note: there are several 4–6 line comment blocks in the new code (`cluster_k8s.rs:848` and `spurd/cluster.rs:956` are the longest). The repo norm is 1–3 lines — worth trimming those to the one-line *why* and moving the rest into the PR description.

crates/spurctld/src/server.rs
```diff
@@ -2572,6 +2572,189 @@ impl SlurmController for ControllerService {
...
+        let force = req.force.unwrap_or(false);
+        let mut removed: Vec<String> = Vec::new();
+        let mut failures: Vec<String> = Vec::new();
+        for n in &requested {
+            match crate::cluster_k8s::remove_worker(&self.cluster, n, timeout, force).await {
+                Ok(()) => removed.push(n.clone()),
+                Err(e) => failures.push(format!("{n}: {e}")),
+            }
+        }
+        if !removed.is_empty() {
```
yansun1996: The member-set subtraction lands only after the whole loop, but `remove_worker` calls `clear_node_k0s` per node. In that window the node is still `is_member()` with `k0s_role == None`, which is precisely the shape `provision_assignments` picks up — it retains on `is_member` and skips only when `k0s_role.is_some()` — so the reconcile loop hands it a fresh role, mesh IP and pod /24 and `converge_provisioning` restarts k0s on the node that was just drained and reset.

With `RECONCILE_INTERVAL` at 30s and the default drain at 120s per node, for any multi-node removal this is the expected outcome rather than a narrow race. A leader flip mid-loop is worse: the `remove_k0s_member_nodes` propose never lands, so the reset node silently re-enrolls permanently.

Could the subtraction move to per node, before `remove_worker` (re-adding on failure)? That way the reconciler never observes an in-scope role-less node.

crates/spurctld/src/server.rs
```diff
@@ -2572,6 +2572,189 @@ impl SlurmController for ControllerService {
...
+                "removed {} node(s); {} failed: {}",
+                removed.len(),
+                failures.len(),
+                failures.join("; ")
+            )
+        };
+        Ok(Response::new(ClusterRemoveNodesResponse {
+            accepted: failures.is_empty(),
+            message,
+            nodes: crate::cluster_k8s::node_statuses(&self.cluster),
```
yansun1996: `node_statuses()` does `let role = n.k0s_role?`, so it drops any node without a role, and it hardcodes `component_state: "unknown"`. For `cluster_add_nodes` that means the nodes you just added are absent from the response entirely (they have no role until the reconcile loop assigns one), which doesn't match the proto's "resulting per-node status after the add".

`cluster_status` uses `live_node_statuses` — would that be a better fit for both new handlers? Otherwise it'd be good to soften the proto comments to say the added nodes appear once a role is assigned.

proto/slurm.proto
```diff
@@ -906,6 +912,37 @@ message ClusterNodeStatus {
...
+}
+
+message ClusterRemoveNodesRequest {
+  // Worker nodes to remove (hostlist, e.g. "gpu[09-12]").
+  string nodes = 1;
+  string caller = 2;   // For authorization (cluster admin required).
+  // Max seconds to wait for the k8s drain per node. 0/absent = server default.
+  optional uint32 drain_timeout_secs = 3;
+  // Proceed even if the node has running SPUR jobs or the drain does not complete
+  // (jobs are failed; drain is best-effort). Destructive.
```
yansun1996: Small accuracy point (related to the existing bot comment on this field): "jobs are failed" doesn't match the handler, which only *skips* the `node_has_running_jobs` check — nothing evicts or fails them, unlike `ClusterManager::remove_node`, which really does evict via `NodeRemove`. So the jobs keep running on a node whose k0s state is about to be wiped.

Either implement the eviction, or reword to something like "proceeds despite running SPUR jobs (they are left running); drain is best-effort". The CLI help and the handler's own error text ("pass --force to evict") would want the same treatment.

Also worth considering: one flag currently conflates "ignore a busy node" with "bypass PodDisruptionBudgets" — quite different blast radii.

crates/spur-cli/src/k8s.rs
```diff
@@ -53,6 +54,33 @@ pub enum K8sCommand {
...
+    /// (destructive — the node re-downloads/re-seeds on a later add). Use `spur node drain` instead
+    /// for a temporary "stop scheduling here".
+    RemoveNodes {
+        /// Worker nodes to remove (hostlist, e.g. "gpu[09-12]").
+        #[arg(long)]
+        nodes: String,
+        /// Max seconds to wait for the k8s drain per node (0 = server default).
+        #[arg(long)]
+        drain_timeout: Option<u32>,
+        /// Proceed even if a node has running jobs or its drain does not complete (evicts jobs;
```
yansun1996: Same wording drift as the proto field — `--force` doesn't evict jobs, it only skips the busy-node check, so this help text over-promises. Suggest "proceed even if a node has running jobs (they are left running) or its drain does not complete; bypasses PodDisruptionBudgets".

crates/spurctld/src/cluster_k8s.rs
```diff
@@ -771,17 +776,64 @@ async fn converge_provisioning(
...
+    set.extend(add.iter().cloned());
+    let mut out: Vec<String> = set.into_iter().collect();
+    out.sort();
+    out
+}
+
+/// Remove `drop` from the current member set, keeping it sorted. Used to shrink a scoped cluster's
+/// `member_nodes` online (`spur k8s remove-nodes`). Removing an absent node is a no-op. Note: this
+/// can return an empty vec — the caller must not let a scoped cluster shrink to empty, since empty
+/// `member_nodes` means "whole inventory" (the RPC handler guards that).
```
yansun1996, line 837: "the RPC handler guards that" isn't true today — `cluster_remove_nodes` has no would-empty check. The same claim appears on `remove_k0s_member_nodes` and on the `K0sMemberNodesRemove` WAL variant. Adding the guard is the better fix; failing that, these three comments should say what actually protects the invariant.

crates/spurctld/src/cluster_k8s.rs
```diff
@@ -815,18 +878,113 @@ async fn degrade_stuck_cluster(
         if let Err(e) = cluster.set_node_k0s_error(&node.name, Some(reason)) {
             warn!(node = %node.name, error = %e, "failed to record k0s node error");
         }
+        stuck.push(node);
+    }
+
+    // Partial-Ready: a control-plane quorum means the cluster is usable. Don't degrade — leave the
+    // stragglers converging; `converge_provisioning` flips the cluster Ready on the next tick.
```
yansun1996: This comment says `converge_provisioning` flips the cluster Ready on the next tick, but in the path that actually reaches here it never will. `converge_provisioning` early-returns on `!bootstrap_active` before it gets to its own quorum check, whereas the `active_cp` computed here is broader (`Controller || Single || bootstrap`). So with the bootstrap CP down but 2-of-3 secondary controllers up, converge keeps early-returning and this branch keeps returning — the cluster pins in `Provisioning` indefinitely, neither Ready nor Degraded, and (since only `degrade_stuck_cluster` writes them) with no further per-node reasons.

Worth either flipping to Ready here, or relaxing converge's bootstrap gate so quorum is honoured in one place. Separately, recomputing `active_cp` costs a second full per-node `fetch_component_state` sweep that converge just did — could that be threaded out of `converge_provisioning` instead?

crates/spurctld/src/cluster_k8s.rs
```diff
@@ -815,18 +878,113 @@ async fn degrade_stuck_cluster(
...
+    node: &str,
+    timeout_secs: u32,
+    force: bool,
+) -> anyhow::Result<()> {
+    let mut client = connect_control_plane(cluster).await?;
+    let resp = client
+        .drain_k8s_node(DrainK8sNodeRequest {
+            node: node.to_string(),
+            timeout_secs,
+            force,
```
yansun1996, line 943: `timeout_secs` is forwarded raw and the RPC itself isn't wrapped in `tokio::time::timeout` — `connect_control_plane` bounds only the dial. Since the handler loops over nodes serially, a large `--drain-timeout` pins the controller's request handler for N × timeout. Would a clamp plus a `tokio::time::timeout(timeout + slack, …)` be worth adding here?

crates/spurctld/src/cluster_k8s.rs
```diff
@@ -815,18 +878,113 @@ async fn degrade_stuck_cluster(
...
+async fn stop_component_now(
+    cluster: &ClusterManager,
+    node: &str,
+    reset: bool,
+) -> anyhow::Result<()> {
+    let endpoint = agent_endpoint(cluster, node)
+        .ok_or_else(|| anyhow::anyhow!("{node} has no agent address"))?;
+    let mut client = SlurmAgentClient::connect(endpoint)
+        .await
+        .map_err(|e| anyhow::anyhow!("connect to agent {node} failed: {e}"))?;
```
yansun1996: Minor: this dial isn't wrapped in `tokio::time::timeout(AGENT_TIMEOUT, …)`. The other bare dials in this file (`spawn_start_component`, `spawn_stop_component`) are detached so it doesn't matter there, but this one is awaited inline inside a gRPC handler, so a black-holed node address hangs the removal.

crates/spurctld/src/cluster_k8s.rs
```diff
@@ -815,18 +878,113 @@ async fn degrade_stuck_cluster(
...
+        anyhow::bail!("stop/reset of {node} did not complete: {}", r.message);
+    }
+    Ok(())
+}
+
+/// Gracefully remove a worker from a running cluster: cordon+drain its pods (via a control-plane
+/// agent), then stop + `k0s reset` its component (wiping the node's k0s state), then declassify it
+/// (clear role/IP/CIDR). The caller subtracts it from `member_nodes`; the mesh prunes on the next
+/// ApplyMesh tick. `force` proceeds past a blocked/timed-out drain. Errors leave the node cordoned so
+/// the operator can retry.
```
yansun1996: "Errors leave the node cordoned so the operator can retry" only holds when the drain itself fails. `drain_node` ends with `kubectl delete node`, so a failure in `stop_component_now` or `clear_node_k0s` leaves the Node object gone rather than cordoned — and the node still roled and still a member, so the reconcile loop will restart its component. Might be worth rewording, and recording a `set_node_k0s_error` reason so `spur k8s status` explains that state.

crates/spurd/src/cluster.rs
```diff
@@ -929,6 +953,68 @@ impl K0sAgent {
...
+                "k0s kubectl drain {node} failed: {}",
+                String::from_utf8_lossy(&drain.stderr).trim()
+            );
+        }
+        // Drain leaves the Node object behind (as NotReady); delete it so removal is complete. Best
+        // effort + tolerant of "not found" so a re-run of a partly-removed node still succeeds.
+        let del = Command::new(&self.k0s_binary)
+            .args(["kubectl", "delete", "node", node, "--ignore-not-found"])
+            .output()
+            .await?;
```
yansun1996, line 1016: Deleting the Node object here happens while the kubelet is still running — the component isn't stopped until the controller calls `stop_component_now` afterwards. In that gap the kubelet re-registers a fresh, *uncordoned* Node, so the scheduler can place pods back onto a node that is about to be `k0s reset`.

Would it work to stop the component before the delete, or to re-issue the delete once the stop succeeds?

crates/spur-core/src/k0s.rs
```diff
@@ -218,6 +218,33 @@ pub fn validate_control_plane_replicas(replicas: u32) -> Result<(), String> {
     ))
 }
 
+/// The majority of `n` (`n/2 + 1`) — the etcd quorum size for a control-plane set of `n`. A cluster
+/// is usable once this many control planes are active, so the k0s reconcile loop gates `Ready` on it.
```
yansun1996: Nit on wording: `active_cp` is populated from `fetch_component_state(...) == Some("active")`, i.e. systemd unit liveness, so a controller whose unit is up but whose etcd never joined still counts. Calling it "the etcd quorum size" here (and "active etcd quorum" on `control_plane_quorum_met`) reads stronger than what's measured — maybe "control planes whose k0s unit is active, as a proxy for etcd quorum"?

crates/spurctld/src/cluster_k8s.rs
```diff
@@ -989,6 +1147,88 @@ pub async fn live_node_statuses(cluster: &ClusterManager) -> Vec<ClusterNodeStat
...
+    }
+
+    #[test]
+    fn subtract_member_nodes_can_empty_the_set() {
+        let cur = names(&["a"]);
+        assert_eq!(
+            subtract_member_nodes(&cur, &names(&["a"])),
+            Vec::<String>::new()
+        );
+    }
```
yansun1996, line 1293: This test codifies the empty-set outcome as expected behaviour while the function's own doc four lines above says the handler prevents it. Once the would-empty guard exists, it'd be good for this to assert the guard rather than the empty vec (the pure-function contract is fine to keep testing, just worth pairing with a handler-level test).

crates/spurctld/src/server.rs
```diff
@@ -2572,6 +2572,189 @@ impl SlurmController for ControllerService {
...
+            return Err(Status::invalid_argument("no nodes selected; pass --nodes"));
+        }
+        let registered: std::collections::HashSet<String> = self
+            .cluster
+            .get_nodes()
+            .into_iter()
+            .map(|n| n.name)
+            .collect();
+        let controller_set = state.controllers();
+        let controllers: std::collections::HashSet<&String> = controller_set.iter().collect();
```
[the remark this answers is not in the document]
spraveenio: Fixed. Added guards to `cluster_remove_nodes`: rejects a whole-inventory cluster (empty `member_nodes`) and a request that would empty the member set — both would otherwise flip scope to whole-inventory and re-enroll. Also now subtracts the member *before* draining (re-add on failure) so the reconcile loop never observes an in-scope role-less node.

proto/slurm.proto
```diff
@@ -962,6 +999,16 @@ message CreateK0sJoinTokenResponse {
   string join_token = 1;
 }
 
+message DrainK8sNodeRequest {
+  string node = 1;                  // k8s node name to cordon + drain
+  uint32 timeout_secs = 2;          // drain timeout; 0 = agent default
+  bool force = 3;                   // pass kubectl drain --force / --disable-eviction on timeout
```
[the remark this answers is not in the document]
spraveenio: Fixed. Reworded to `when set, pass kubectl drain --force --disable-eviction (bypass PDBs)` — no longer implies the flags are timeout-gated.

crates/spurctld/src/server.rs
spraveenio: Added all three guards to `cluster_remove_nodes`, all failing closed: `state.is_member(n)` (rejects out-of-scope registered nodes), empty `member_nodes` (whole-inventory), and would-empty (all requested == all members). Messages point at `spur node remove` / `spur k8s down`. The doc comments on `subtract_member_nodes` / `remove_k0s_member_nodes` are now accurate.

crates/spurctld/src/server.rs
spraveenio: Fixed exactly as suggested. The subtraction is now per-node and happens *before* `remove_worker`, re-adding the node on failure. So the reconcile loop never sees an in-scope role-less node, and a leader flip mid-loop leaves the (already-durable) member removal in place rather than a re-enrolling orphan.

crates/spurctld/src/server.rs
spraveenio: Both new handlers now return `live_node_statuses` (real component_state), and the add-response proto comment is softened to note a just-added node appears only once the reconcile loop assigns it a role.

proto/slurm.proto
spraveenio: Reworded across proto/CLI/handler: `--force` only skips the busy-node check — the jobs are left running, it does not evict (unlike `remove_node` via `NodeRemove`). Kept the drain-force (`--disable-eviction`) under the same flag for now; splitting "ignore busy node" from "bypass PDBs" into two flags is a reasonable follow-up if you want distinct blast radii.

crates/spur-cli/src/k8s.rs
spraveenio: Fixed. CLI help now reads `proceed even if a node has running jobs (they are left running — this only skips the busy-node check) or its drain does not complete (bypasses PodDisruptionBudgets)`.

crates/spurctld/src/cluster_k8s.rs
spraveenio: True at the time — now fixed: the would-empty guard exists in `cluster_remove_nodes`, so the three doc comments (`subtract_member_nodes`, `remove_k0s_member_nodes`, the WAL variant) that reference the handler guard are accurate.

crates/spurctld/src/cluster_k8s.rs
spraveenio: Good catch — fixed. The quorum decision now lives in one shared helper (`maybe_mark_ready`) called from both the normal converge path and the `!bootstrap_active` early return, so bootstrap-down + secondary-quorum-up flips Ready instead of pinning Provisioning. `degrade_stuck_cluster` also flips Ready when quorum holds rather than just logging. And the double sweep is gone: `converge_provisioning` now returns the `active_cp` set it computed and threads it into the degrade path.

crates/spurctld/src/cluster_k8s.rs
spraveenio: Added. `drain_via_control_plane` now wraps the RPC in `tokio::time::timeout(drain_deadline + 30s slack)`, so a large `--drain-timeout` bounds the handler at N × (timeout + slack) with a hard ceiling per node instead of hanging indefinitely.

crates/spurctld/src/cluster_k8s.rs
spraveenio: Fixed. `stop_component_now` now wraps connect+RPC in `tokio::time::timeout(AGENT_TIMEOUT)`, and the new `delete_node_via_control_plane` is bounded the same way — no bare inline dials in the awaited remove path.

crates/spurctld/src/cluster_k8s.rs
spraveenio: Reworded and hardened: `remove_worker` now records a `set_node_k0s_error` reason at every failure point (drain/stop/delete/declassify) and the handler re-adds the node to `member_nodes` on failure, so a failure after the Node delete leaves a managed, in-scope node with a visible reason in `spur k8s status` rather than a silently-restarting orphan.

crates/spurd/src/cluster.rs
spraveenio: Fixed — delete now happens after the stop. Split the old combined drain+delete into two RPCs: `DrainK8sNode` (cordon+drain) and `DeleteK8sNode`. `remove_worker` runs drain → stop+reset (kubelet gone) → delete-node → declassify, so the kubelet can no longer re-register a fresh uncordoned node in the gap.

crates/spur-core/src/k0s.rs
spraveenio: Fixed the wording. `quorum` and `control_plane_quorum_met` docs now say "control planes whose k0s unit is active, as a proxy for etcd quorum" rather than claiming a direct etcd check.

crates/spurctld/src/cluster_k8s.rs
spraveenio: Done. The pure-function test now carries a comment noting the handler prevents the empty case, and there is a paired handler-level test `cluster_remove_nodes_rejects_emptying_the_member_set` (plus an E2E for the same) asserting the guard.

yansun1996 (approved):
LGTM thanks

PR #656, opened 2026-08-17
fix(spurd): close container sync-pipe fds exactly once

### LINKED ISSUE #653: [Issue]: spurd double-closes the container-launch sync-pipe fds — debug builds abort and lose the node daemon, release builds close descriptors they no longer own
### Problem Description

`launch_container_job()` in `crates/spurd/src/executor.rs` closes both ends of its fork sync pipe **twice**: once by hand with `libc::close()`, and again when the `OwnedFd` that still owns each descriptor is dropped at the end of the function. This happens on every containerized job launch, inside the long-lived `spurd` daemon.

Verified against `spur 0.8.0`, upstream `main` at `3f85b50` (current head, 2026-08-17), where all six lines below are unchanged: `1428`, `1429`, `1448`, `1485`, `1532`, `1554`. The pattern was introduced in `0a52afb` ("Harden native container runtime with direct syscalls and integration tests", #154, 2026-05-01) and is present in every release since: `v0.4.0`, `v0.4.1`, `v0.5.0`, `v0.5.1`, `v0.6.0`, `v0.7.0`, `v0.8.0`, and `nightly`.

### Root cause

`nix::unistd::pipe()` returns two `OwnedFd`. The function takes the raw numbers with `as_raw_fd()` — which is only a **borrow**, ownership stays with the `OwnedFd` — and then deliberately keeps both owners alive:

```rust
// crates/spurd/src/executor.rs:1416-1429
// Sync pipe: child writes status, parent reads.
// Convert OwnedFd to raw fds for manual lifecycle management across fork.
let (pipe_r, pipe_w) = nix::unistd::pipe().context("create sync pipe")?;
nix::fcntl::fcntl(&pipe_r, nix::fcntl::FcntlArg::F_SETFD(nix::fcntl::FdFlag::FD_CLOEXEC)).ok();
let ready_r = pipe_r.as_raw_fd();   // borrow — pipe_r still OWNS this fd
let ready_w = pipe_w.as_raw_fd();   // borrow — pipe_w still OWNS this fd
// Keep OwnedFd alive so the fds aren't closed prematurely
let _pipe_r_owner = pipe_r;
let _pipe_w_owner = pipe_w;
```

The two comments state opposite intentions. "Manual lifecycle management" (line 1417) requires giving up ownership; "keep `OwnedFd` alive" (line 1427) retains it. Both were done, so each descriptor gets closed once by hand and once by its owner's destructor. There is no `into_raw_fd`, `mem::forget`, or `ManuallyDrop` anywhere in the file to suppress the destructors. (A named `let _pipe_r_owner = …` binding holds the value to end of scope and *does* run `Drop`, unlike a bare `let _ = …`, which drops immediately.)

In the **parent** — the `spurd` daemon itself — the sequence is unconditional:

| Line | Action |
|---|---|
| 1532 | `libc::close(ready_w)` |
| 1541 | `std::fs::write(cgroup.join("cgroup.procs"), …)` — opens a descriptor |
| 1545 | `pidfd_open(child_pid)` — opens a descriptor |
| 1554 | `libc::close(ready_r)` |
| 1558 / 1566 | function returns (`bail!` on init failure, or `Ok`) |
| end of scope | `_pipe_w_owner` and `_pipe_r_owner` drop → **`close()` on both numbers again** |

Both exits pass through the drop, so every container launch double-closes two descriptors. Note lines 1541 and 1545 open new descriptors *after* `ready_w` was freed and *before* the stale owner drops, so reuse of the freed number inside that window is structural, not hypothetical — and `spurd` is a multi-threaded tokio daemon whose other tasks open sockets and files concurrently.

The forked **child** escapes: it always ends in `execve` (1525), `libc::_exit` (1461), or `std::process::exit` (1478/1527), none of which run destructors. So the damage lands in the daemon, which is the worse of the two places.

### Symptoms

Which symptom appears depends on whether the freed descriptor number was reissued before the stale owner dropped. Both were reproduced deterministically with the standalone program under *Steps to Reproduce*:

| Situation | Build | Result |
|---|---|---|
| Number **not** reused | debug assertions on | `fatal runtime error: IO Safety violation: owned file descriptor already closed, aborting` — `SIGABRT`, exit 134 |
| Number **not** reused | release | no abort; the second `close()` executes silently on a stale number |
| Number **reused** | either | no abort; the stale owner closes an **unrelated live descriptor** — writes to that file then fail with `Bad file descriptor (os error 9)` |

The double close itself is unconditional; only which symptom surfaces depends on reuse. On a real single-node cluster the debug build aborted on **every** containerized job — five consecutive launches, five aborts, no run-to-run variation (Route B). A release build of the same tree survives but still closes both descriptors twice, visible in `strace` (Route C).

### Why it's a bug

Double-closing a descriptor that an `OwnedFd` still owns violates the ownership contract of `std::os::fd`, independent of which symptom a given build shows. The standard library treats it as a hard error — that is exactly what the `IO Safety violation` abort is reporting. Either ownership is transferred out (`into_raw_fd`) and the code closes by hand, or the owner is kept and the manual closes are removed; doing both is never correct.

### Why it matters

Both outcomes hit the node daemon rather than the job:

- **Debug builds abort `spurd`.** The node daemon dies mid-launch, so the node drops out of the cluster along with every other job on it. Worse, the container itself has already been forked and runs to completion and writes its output, but because the daemon died before confirming dispatch, the controller reports the job as never launched (`JobLaunchFailure`). The work ran, the scheduler says it didn't, and the node is gone. Plain `cargo build` produces debug assertions, including the `cargo build --all-targets --locked` in `.github/workflows/ci.yml`.
- **Release builds close a descriptor they no longer own.** Route C shows both closes firing, the second returning `EBADF` because that number happened to still be free. When it isn't free, the same close lands on whatever now holds the number — a descriptor another thread just opened. If that is the `pidfd` from line 1545 — returned as `RunningJob::_pidfd` and used specifically to track the child without PID recycling — the daemon is left holding a dangling tracking handle for a running job. Silent descriptor corruption in a daemon is considerably harder to diagnose than the abort.

### Expected

Each descriptor is closed exactly once. Ownership should either be transferred out of the `OwnedFd` (so the manual `close()` calls are the sole owner of the lifetime) or left with the `OwnedFd` (so the manual `close()` calls are removed) — never both.

### Operating System

Ubuntu 24.04 LTS (kernel 6.8.0-38-generic). Bug is OS/hardware-independent (Linux-only code path).

### CPU

N/A (hardware-independent)

### GPU

N/A (hardware-independent; no GPUs requested by the reproducing job)

### ROCm Version

N/A (hardware-independent)

### ROCm Component

_No response_

### Steps to Reproduce

Route A isolates the ownership mistake in a standalone program — seconds, no root, no cluster, no dependencies. Route B is the real cluster run. All output below is verbatim.

**Route A — minimal standalone repro.** Only `rustc` is needed. The program performs the exact sequence the parent branch performs (`pipe()` → `as_raw_fd()` → keep the owner alive → `libc::close()` → let the owner drop), and covers both the non-reused case (A) and the reused case (B).

1. Save as `repro.rs`:

```rust
// Minimal reproduction of the spurd container-launch sync-pipe double-close.
// Mirrors crates/spurd/src/executor.rs launch_container_job():
//   let (pipe_r, pipe_w) = nix::unistd::pipe()?;  // -> OwnedFd, OwnedFd
//   let ready_r = pipe_r.as_raw_fd();             // borrow only
//   let _pipe_r_owner = pipe_r;                   // owner stays alive
//   ... unsafe { libc::close(ready_r) } ...       // manual close #1
//   // owner drops at end of scope                -> close #2 on the same fd
use std::io::Write;
use std::os::fd::{AsRawFd, FromRawFd, OwnedFd};

extern "C" {
    fn pipe(fds: *mut i32) -> i32;
    fn close(fd: i32) -> i32;
}

fn make_pipe() -> (OwnedFd, OwnedFd) {
    let mut fds = [0i32; 2];
    assert_eq!(unsafe { pipe(fds.as_mut_ptr()) }, 0, "pipe() failed");
    // nix::unistd::pipe() hands back owning descriptors, exactly like this:
    unsafe { (OwnedFd::from_raw_fd(fds[0]), OwnedFd::from_raw_fd(fds[1])) }
}

/// Case A: spur's exact shape. Manual close, then the owner drops and closes again.
fn case_a_double_close() {
    let (pipe_r, pipe_w) = make_pipe();
    let ready_r = pipe_r.as_raw_fd(); // borrow, like spur
    let ready_w = pipe_w.as_raw_fd();
    let _pipe_r_owner = pipe_r; // owner kept alive, like spur
    let _pipe_w_owner = pipe_w;
    println!("case A: pipe fds = ({}, {})", ready_r, ready_w);
    unsafe {
        close(ready_w); // spur's manual close
        close(ready_r);
    }
    println!("case A: manual closes done; owners about to drop (second close)...");
    // _pipe_*_owner drop here -> close() again on an already-closed fd
}

/// Case B: same bug, but the fd number gets reused before the owner drops.
/// No abort even in debug -- instead an UNRELATED file is closed underneath us.
fn case_b_wrong_fd_closed() {
    let (pipe_r, pipe_w) = make_pipe();
    let ready_w = pipe_w.as_raw_fd();
    let _pipe_r_owner = pipe_r; // keep read end open so fd numbering is stable
    let stale_owner = pipe_w;
    unsafe { close(ready_w) }; // spur's manual close frees fd number `ready_w`

    // Anything opened now claims the lowest free fd -- i.e. the one just freed.
    let victim = std::fs::File::create("/tmp/fd_victim.txt").expect("create victim");
    println!(
        "case B: stale owner holds fd {}, unrelated victim file got fd {}",
        ready_w,
        victim.as_raw_fd()
    );
    assert_eq!(victim.as_raw_fd(), ready_w, "fd number was not reused; test inconclusive");

    drop(stale_owner); // closes the VICTIM's descriptor, not the pipe's

    match (&victim).write_all(b"payload") {
        Ok(()) => println!("case B: victim write OK (bug did not bite this run)"),
        Err(e) => println!("case B: victim write FAILED -> {} (its fd was closed by the stale owner)", e),
    }
}

fn main() {
    match std::env::args().nth(1).as_deref() {
        Some("a") => case_a_double_close(),
        Some("b") => case_b_wrong_fd_closed(),
        _ => eprintln!("usage: repro a|b"),
    }
    println!("reached end of main");
}
```

2. Build both profiles and run:

```bash
rustc -C debug-assertions=on     -o repro_debug   repro.rs   # as `cargo build`
rustc -O -C debug-assertions=off -o repro_release repro.rs   # as `cargo build --release`

./repro_debug   a; echo "exit=$?"
./repro_release a; echo "exit=$?"
./repro_debug   b; echo "exit=$?"
```

3. Observed (Ubuntu 24.04, kernel 6.8.0-38-generic, rustc 1.97.1):

```console
$ ./repro_debug a
case A: pipe fds = (3, 4)
case A: manual closes done; owners about to drop (second close)...
fatal runtime error: IO Safety violation: owned file descriptor already closed, aborting
exit=134

$ ./repro_release a
case A: pipe fds = (3, 4)
case A: manual closes done; owners about to drop (second close)...
reached end of main
exit=0

$ ./repro_debug b
case B: stale owner holds fd 4, unrelated victim file got fd 4
case B: victim write FAILED -> Bad file descriptor (os error 9) (its fd was closed by the stale owner)
```

Case A on the debug build is the daemon abort. Case A on release is the same double close with no guard. Case B shows the reused-number path, where there is no abort in either profile and an unrelated file loses its descriptor instead.

<details>
<summary>Route B — on a real single-node cluster (where the abort was first seen)</summary>

Any containerized job reaches `launch_container_job`, so one node is enough. Build unmodified `main` with `cargo build` (debug assertions on), then:

```bash
# minimal single-node config: one partition "gpu", one node, auth plugin "none",
# controller on 127.0.0.1:6817
spurctld -f spur.conf -D --state-dir ./state &
sudo spurd -f spur.conf -D --hostname node1 --controller http://127.0.0.1:6817 > spurd.log 2>&1 &

spur image import dockerd://busybox:latest      # any image; this used a 4 MB busybox

spur submit --partition gpu --container-image busybox:latest \
            --container-workdir / --output job.out \
            --controller http://127.0.0.1:6817 ./job.sh

pgrep -f spurd                                   # -> nothing
grep "IO Safety" spurd.log
spur queue --controller http://127.0.0.1:6817
```

Observed: the last line `spurd` writes is the launch success message, immediately followed by the abort as the function returns and the owners drop.

```
2026-08-17T18:44:53.635441Z  INFO spurd::executor: launching job job_id=1 work_dir="/home/amd/spur_fd_e2e"
2026-08-17T18:44:53.793724Z  INFO spurd::executor: containerized job launched (fork + pivot_root) job_id=1 pid=3106028 rootfs=/var/spool/spur/containers/job_1
fatal runtime error: IO Safety violation: owned file descriptor already closed, aborting
```

```
Aborted (core dumped)  spurd -f spur.conf -D --hostname node1 --controller http://127.0.0.1:6817
```

The container ran and `job.out` contained its output, yet the job sat pending forever:

```
JOBID PARTITION     NAME    USER ST  TIME NODES NODELIST(REASON)
    1       gpu  /home/a     amd PD  0:00     1 (JobLaunchFailure (dispatch confirmation failed: 0 of 1 nodes confirmed))
```

**Control, same live daemon:** the identical script submitted *without* `--container-image` runs to completion, `spurd` stays alive on its original PID, and `spurd.log` has zero `IO Safety` lines. Only the containerized path — the one that reaches `launch_container_job` — kills the daemon.

This route is deterministic: five consecutive container jobs, five aborts. One caveat if you reuse an existing `target/` directory — run `touch crates/spurd/src/executor.rs && cargo build` first. A binary left over from a patched build closes each descriptor exactly once and hides the defect completely, which is what produced an earlier run on this setup that appeared to survive.
</details>

<details>
<summary>Route C — release build, same cluster: no crash, but the double close is visible</summary>

Same unmodified tree built with `cargo build --release`, same single-node cluster, same container job. The daemon survives, the job reaches state `R`, and `spurd.log` has no `IO Safety` line — the guard is compiled out with debug assertions off. The double close is still there, and `strace` shows it directly. Start the agent under:

```bash
sudo strace -f -tt -y -o t.strace -e trace=pipe2,close,fcntl,read,write \
     spurd -f spur.conf -D --hostname node1 --controller http://127.0.0.1:6817 > spurd.log 2>&1 &
```

then submit the container job and follow the sync pipe. `-y` annotates each descriptor with the object it refers to. Verbatim, from the launching thread, `read_fd=11` / `write_fd=14`:

```
close(14<pipe:[2042784312]>) = 0                  # executor.rs:1532 (manual)
read(11<pipe:[2042784312]>,  <unfinished ...>     # executor.rs:1551
close(11<pipe:[2042784312]>) = 0                  # executor.rs:1554 (manual)
close(14) = -1 EBADF (Bad file descriptor)        # _pipe_w_owner drop -> second close
close(11) = -1 EBADF (Bad file descriptor)        # _pipe_r_owner drop -> second close
```

Each descriptor is closed twice by the same thread, the second close failing with `EBADF` because the number is already free. This run was harmless only because nothing claimed fd 11 or 14 during that ~200 µs window — timing, not correctness. Release builds carry the same defect and fail quietly; debug builds turn it into an immediate daemon abort.

When reading the trace, follow the sync-pipe fd numbers rather than grepping for any failed close: the forked child also produces a benign `close(N) = -1 EBADF` while `close_inherited_fds` sweeps `/proc/self/fd`.
</details>

<details>
<summary>rocminfo --support </summary>

```
N/A (hardware-independent — no GPU involvement)
```

</details>

### Affected code

- `crates/spurd/src/executor.rs:1418-1429` — `nix::unistd::pipe()` owners bound to `_pipe_r_owner`/`_pipe_w_owner` while raw copies are taken with `as_raw_fd()`.
- `crates/spurd/src/executor.rs:1448, 1485` — manual `libc::close()` in the child branch.
- `crates/spurd/src/executor.rs:1532, 1554` — manual `libc::close()` in the parent branch; both function exits then run the owners' destructors.
- No `into_raw_fd` / `mem::forget` / `ManuallyDrop` anywhere in the file.
- Line numbers are against `3f85b50`. Open PR #492 (`feat(spurd): survive agent restart without interrupting running jobs`) refactors `launch_container_job` and will shift them, but it does not change the ownership pattern — the `as_raw_fd()` borrows, the `_pipe_*_owner` bindings, and all four manual closes appear only as unmodified context in its diff.

Scope: `nix::unistd::pipe()` appears exactly once in the whole repository, so this is a single-site defect. The other `libc::close()` call sites are correct and should not be changed — `container.rs:1046` closes raw numbers scraped from `/proc/self/fd`, `landlock.rs:159/202` closes a syscall-returned ruleset fd that was never owned, and `pty.rs:70/73` plus `executor.rs:255-286` run post-fork in the child, where closing the child's own copies is intended.

### Suggested fix

**Recommended — let the `OwnedFd`s do the closing and delete the manual closes.** This removes the double close and is also leak-free if an early return happens before the closes (e.g. `fork()?` at line 1443):

```diff
     let ready_r = pipe_r.as_raw_fd();
     let ready_w = pipe_w.as_raw_fd();
-    // Keep OwnedFd alive so the fds aren't closed prematurely
-    let _pipe_r_owner = pipe_r;
-    let _pipe_w_owner = pipe_w;
+    // pipe_r / pipe_w stay the sole owners and close their fd exactly once on drop.

@@ nix::unistd::ForkResult::Child => {
-            unsafe {
-                libc::close(ready_r);
-            }
+            drop(pipe_r);

@@ // Signal parent: setup complete
             unsafe {
                 libc::write(ready_w, b"OK".as_ptr() as *const _, 2);
-                libc::close(ready_w);
             }
+            drop(pipe_w);

@@ nix::unistd::ForkResult::Parent { child } => {
-            unsafe {
-                libc::close(ready_w);
-            }
+            drop(pipe_w);

@@ let n = n.max(0) as usize;
-            unsafe {
-                libc::close(ready_r);
-            }
+            drop(pipe_r);
```

Both match arms may move the same binding because only one arm runs. `ready_r`/`ready_w` stay as `RawFd` copies for the `libc::read`/`write` and `close_inherited_fds` calls that need an `i32`, and every use precedes the corresponding drop. `FD_CLOEXEC` must still be set before any ownership transfer, since that `fcntl` borrows `&pipe_r`.

**Minimal alternative — one line, with a caveat.** Transfer ownership out so no destructor runs, and delete the two `_pipe_*_owner` bindings:

```rust
use std::os::fd::IntoRawFd;
let ready_r = pipe_r.into_raw_fd();   // consumes the OwnedFd
let ready_w = pipe_w.into_raw_fd();
```

Caveat: this leaves the descriptors raw, so if `fork()` at line 1443 fails and `?` returns early, both fds leak. The drop-based version has no such gap.

Either way, it would be worth exercising the container launch path in a root-gated integration test, since nothing in the suite reaches it today: `launch_container_job` (line 1407) has exactly one caller (line 667) and no test references it anywhere in the repository. CI compiles the buggy path but never executes it, which is why this has gone unnoticed since April.

### Related

- #649 (image `config.Env` discarded on import) lives in the same `launch_container_job` function but is an unrelated defect — that one concerns environment assembly, this one descriptor lifetime.
- #491 (PMI server task and listener fd leak) is also a different defect: an accumulating fd/task *leak* in `crates/spurd/src/pmi.rs` where `cleanup()` never stops the `run()` accept loop, not a double close in the launch path.

### Notes

- A fix is prepared and verified: with the drop-based patch applied and nothing else changed, the same container job that previously killed the daemon logged `job launched successfully` → `job finished exit_code=0`, reached state `R`, and `spurd` stayed alive on its original PID. Can open a PR referencing this issue.

SumonAMD:
## Summary

`launch_container_job` closes both ends of the fork sync pipe twice: once explicitly via
`libc::close`, and once again when the `OwnedFd` values that still own those descriptors are
dropped. This makes every containerized job launch violate Rust's I/O-safety contract.

Fixes #653.

## Root cause

On `main`, `launch_container_job` creates the pipe and then extracts both raw descriptors with
`as_raw_fd()`, which *borrows* rather than transfers ownership. The two `_pipe_*_owner` bindings
deliberately keep the `OwnedFd` values alive:

    let ready_r = pipe_r.as_raw_fd();   // borrow; pipe_r still owns this fd
    let ready_w = pipe_w.as_raw_fd();
    let _pipe_r_owner = pipe_r;
    let _pipe_w_owner = pipe_w;

Each end is then closed manually — the read end in the child arm right after the fork and in the
parent arm after the status read, the write end in the child after the `"OK"` write and in the parent
immediately after the fork — while the owners are still live and will close the same descriptors
again at end of scope.

Consequences:

- **Debug / any build with `debug_assertions`:** the second close trips Rust's I/O-safety guard and
  the process aborts with `IO Safety violation: owned file descriptor already closed`. On a
  single-node cluster this killed `spurd` on 5 out of 5 container job submissions, mid-launch, while
  a non-container job on the same daemon completed normally.
- **Release:** the guard is compiled out, so the extra close is silent. `strace -y` on a release
  build shows both closes of the same pipe fd, the second returning `-1 EBADF`. It is harmless only
  as long as the number has not been reissued; if it has, the stale close lands on an unrelated
  descriptor. The e2e container suite runs against release binaries, which is why it stayed green.

## The change

Make the `OwnedFd` values the sole owners and drop them at exactly the points where the manual
closes used to happen. The `RawFd` copies are kept, since `libc::read`, `libc::write`, and
`close_inherited_fds` need an `i32`.

- child: `libc::close(ready_r)` → `drop(pipe_r)`
- child: `libc::close(ready_w)` → `drop(pipe_w)` (after the `"OK"` write)
- parent: `libc::close(ready_w)` → `drop(pipe_w)`
- parent: `libc::close(ready_r)` → `drop(pipe_r)` (after the status read)

Both match arms move the same bindings, which is fine because only one arm ever runs in a given
process. Close ordering is unchanged, so the handshake still behaves identically: the child releases
the read end before wiring stdio, and the parent releases the write end before the blocking read, so
that read still sees EOF if the child dies without reporting.

## Verification

- `cargo fmt --all -- --check` and `cargo clippy --workspace --exclude spur-ffi --all-targets
  --locked` are clean, and `cargo test --locked` passes with `ulimit -n` raised. On hosts with a very
  high core count the suite hits `Too many open files` at the default 1024; that is pre-existing and
  reproduces identically on unpatched `main`.
- Debug `spurd` built from this branch, single-node cluster, containerized job: job reaches `R`,
  logs `job launched successfully` then `job finished exit_code=0`, the daemon survives on its
  original PID, and there are zero `IO Safety` lines across five consecutive container jobs. The same
  repro script run against unpatched `main` on the same node aborts the daemon every time.
- `strace -y` on the same run shows each pipe end closed exactly once per process, and every
  `fcntl(fd, F_GETFD)` drop-guard probe now succeeds, confirming no earlier close.

## Notes

- No behavior change beyond removing the second close; the fix is contained to the container launch
  path and does not touch non-container jobs.
- #492 moves this function but keeps the same ownership pattern, so this fix applies either way.

### DIFF
--- a/crates/spurd/src/executor.rs
+++ b/crates/spurd/src/executor.rs
@@ -1414,7 +1414,6 @@ async fn launch_container_job(
     let cgroup_path = setup_cgroup(job_id, cfg.cpus, cfg.memory_mb, &cfg.cpu_ids)?;
 
     // Sync pipe: child writes status, parent reads.
-    // Convert OwnedFd to raw fds for manual lifecycle management across fork.
     let (pipe_r, pipe_w) = nix::unistd::pipe().context("create sync pipe")?;
     // Prevent read end from leaking into exec'd process
     nix::fcntl::fcntl(
@@ -1424,9 +1423,6 @@ async fn launch_container_job(
     .ok();
     let ready_r = pipe_r.as_raw_fd();
     let ready_w = pipe_w.as_raw_fd();
-    // Keep OwnedFd alive so the fds aren't closed prematurely
-    let _pipe_r_owner = pipe_r;
-    let _pipe_w_owner = pipe_w;
 
     // Snapshot raw I/O fds before fork — the Copy JobIoRaw can be used
     // in the child without owning the fds (parent's OwnedFds keep them alive
@@ -1444,9 +1440,7 @@ async fn launch_container_job(
         nix::unistd::ForkResult::Child => {
             // === CHILD PROCESS ===
             // CRITICAL: synchronous code only. Tokio runtime is broken after fork.
-            unsafe {
-                libc::close(ready_r);
-            }
+            drop(pipe_r);
 
             // Reset signal handlers
             unsafe {
@@ -1482,8 +1476,8 @@ async fn launch_container_job(
             // Signal parent: setup complete
             unsafe {
                 libc::write(ready_w, b"OK".as_ptr() as *const _, 2);
-                libc::close(ready_w);
             }
+            drop(pipe_w);
 
             // Build final environment: base + container_env + hook environ.d
             let mut final_env = env_snapshot;
@@ -1528,9 +1522,7 @@ async fn launch_container_job(
         }
 
         nix::unistd::ForkResult::Parent { child } => {
-            unsafe {
-                libc::close(ready_w);
-            }
+            drop(pipe_w);
 
             // Drop the slave fd immediately so the master gets EOF when the child exits.
             let pty_master = job_io.into_master();
@@ -1550,9 +1542,7 @@ async fn launch_container_job(
             let mut buf = [0u8; 512];
             let n = unsafe { libc::read(ready_r, buf.as_mut_ptr() as *mut _, buf.len()) };
             let n = n.max(0) as usize;
-            unsafe {
-                libc::close(ready_r);
-            }
+            drop(pipe_r);
 
             if n < 2 || &buf[..2] != b"OK" {
                 let msg = String::from_utf8_lossy(&buf[..n]);


sajmera-pensando (approved):
**Correct fix — safe to merge.**

Verified the full function, including the two early-error write paths (`raw_io.wire()` failure at line 1460 and `container_init` failure at line 1476) and the `close_inherited_fds(ready_w)` call that intentionally preserves the write end through child setup. None of these interact badly with the fix.

The double-close was real: `_pipe_w_owner` / `_pipe_r_owner` were stashed to keep the `OwnedFd`s alive until end-of-function, but the manual `libc::close()` calls at each site meant every fd was closed once by hand and again on drop. In a Tokio process, fd numbers get recycled quickly, so the second close could silently close a completely unrelated fd opened in between.

The fix is the right shape: remove the stash variables, let `pipe_r` / `pipe_w` remain as the sole `OwnedFd` owners, and replace each `libc::close()` with a `drop()`. The `_exit(1)` early-exit path skips destructors so the fd closes with the process — no issue. The `std::process::exit(1)` path runs destructors, so after this fix `pipe_w` drops exactly once there too — also correct.

Net: simpler code, one ownership story, no double-close.

Issue #665, opened 2026-08-17
epic(tenancy): multi-tenancy isolation gaps beyond authorization


powderluv:
## Multi-tenancy: isolation gaps beyond authorization

This epic tracks the isolation work that remains **after** the authorization gaps are closed.
Authorization (identity binding, admin-gating the control plane, owner checks, fail-closed defaults —
tracked separately) makes Spur *access-correct*: a caller can only perform actions as itself. It does
not make Spur *isolation-correct*: it says nothing about physical co-residency, data remanence,
blast radius, noisy-neighbour, or cross-tenant information disclosure. Those are tracked here.

### The two-tier verdict

Assuming authorization is complete:

- **Dedicated, physically-disjoint per-tenant node pools** — achievable for mutually-distrusting
  tenants after the bounded fixes below **and** the asymmetric-key redesign (#658). If pools never
  share a host, the residue/noisy-neighbour gaps are contained within a tenant.
- **Shared nodes (two tenants on one host)** — **not achievable by configuration today.** The per-node
  tenant-exclusivity gate is a dead field (#657), and even after resurrecting it, safe sub-node
  sharing depends on a GPU VRAM-isolation guarantee (#661) that lives in the driver/hardware, not this
  codebase. On AMD Instinct without SR-IOV partition-level reset guarantees, two mutually-distrusting
  tenants cannot safely share a box.

### The gate

- [ ] #657 — tenancy: scheduler has no per-node tenant exclusivity — **highest leverage; every
  shared-node gap below depends on it**

### Design projects (new subsystem)

- [ ] #658 — asymmetric per-principal identity keys (required for mutual distrust at either tier)
- [ ] #659 — per-tenant network isolation

### Bounded changes / path-coverage

- [ ] #660 — GPU device cgroup on every launch path (finishes #148 / #154)
- [ ] #661 — GPU VRAM/KFD scrub between jobs (bounded insertion point; reliable guarantee is
  hardware/vendor)
- [ ] #662 — srun/salloc step cgroup + namespace confinement (applies #99's batch isolation to steps)
- [ ] #663 — restrict client-settable job priority (Slurm parity)
- [ ] #664 — owner-scope `get_job`/`get_job_steps`

### Already tracked elsewhere

- #439 — GPU is not tracked for fair-share / not enforced by QoS (the noisy-neighbour fairness half)
- #441 — dead config fields incl. `exclusive_user` (subsumed by #657 for the tenancy angle)
[...]

PR #666, opened 2026-08-17
fix(tenancy): clamp non-admin priority and redact cross-tenant job info (#663, #664)

### LINKED ISSUE #663: scheduler: restrict client-settable job priority (Slurm parity: --priority is admin-only)
## Summary

A job's base priority is taken verbatim from the client-supplied `spec.priority`, with no clamp and no
admin restriction. Because the compensating scheduling factors are capped, a single job submitted at
the maximum priority permanently outranks every normally-submitted job. In Slurm, `--priority` is
operator-restricted; Spur should match.

## Evidence (`6bc97d5`)

- `proto_to_job_spec` copies `spec.priority` verbatim, gated only by `> 0`
  (`crates/spurctld/src/server.rs:2968`); `Job::new` uses it as the base
  (`crates/spur-core/src/job.rs:791`, `spec.priority.unwrap_or(DEFAULT_PRIORITY)`).
- The pending queue sorts strictly by resulting priority (`crates/spurctld/src/cluster.rs:2916-2923`).
- The compensating factors are capped: fair-share ≤ 10.0 (`crates/spur-sched/src/priority.rs:15`),
  age ≤ 2.0 (`:13`) — so a job at `priority = u32::MAX` outranks every default-priority job forever.
- `submit_job` binds identity but never clamps or admin-gates priority, and it is not among the
  admin-gated RPCs.

## Consequence

Any user can front-run the entire queue with a single spec field — a fairness break, and at the limit
a scheduling denial for everyone else, without crossing an authorization boundary. This is a
Slurm-parity gap: Slurm restricts user-set priority to administrators.

## Proposed fix

Clamp non-admin `spec.priority` to a configured maximum (or ignore it for non-admins and require QOS/
nice for user-visible ordering), and admin-gate any request above that. Bounded change.

Related: #281 (priority formula multiplicative→additive), #285 (missing priority factors), #440
(default-priority collapse). Part of the multi-tenancy isolation roadmap.

_Surfaced during a residual-tenancy audit that assumed the authorization work is complete._

### LINKED ISSUE #664: tenancy: get_job/get_job_steps return any tenant's job metadata unscoped (parity with get_jobs)
## Summary

`get_job` and `get_job_steps` return any job's metadata to any caller with no owner scoping, while the
sibling list RPC `get_jobs` already scopes to the caller. The scoping mechanism exists; it was simply
never wired into the single-fetch and steps-fetch paths.

## Evidence (`6bc97d5`)

- `get_job` (`crates/spurctld/src/server.rs:614-637`) calls `get_job_for_display(job_id)` and returns
  `job_to_proto` with no `verified_identity()`, no `authoritative_user()`, and no owner check.
  `get_job_steps` (`:1598-1630`) is the same.
- The response serializes work_dir, the first command line, absolute stdout/stderr/stdin paths, the
  allocated nodelist, account, qos, comment, uid, and GPU detail (`server.rs:3128-3193`).
- The list RPC already does the right thing: `get_jobs` calls `authoritative_user`
  (`server.rs:559`) and turns the caller into a filter. `get_job` lacks that call.
- It does **not** leak job environment (`JobInfo` has no env field), so env-borne secrets are not
  exposed via this path.

## Consequence

Any authenticated user can `scontrol show job <id>` for any other tenant's job (iterating ids
enumerates the cluster) and read its working directory, command line, stdio paths, account/qos, and
allocated nodelist. This is a cross-tenant information-disclosure gap; the nodelist in particular is
useful for targeting the co-residency/residue gaps tracked separately.

Note on intent: Slurm exposes a good deal of cross-user job metadata by default, so part of resolving
this is deciding the intended visibility model (owner-only vs. cluster-visible with private fields
redacted). At minimum the single-fetch paths should match whatever policy `get_jobs` enforces.

## Proposed fix

Mirror `authoritative_user` and add `check_job_owner` to `get_job`/`get_job_steps`, or apply the
agreed visibility policy consistently across list and single-fetch. Bounded change.

Related: `area: controller`. Part of the multi-tenancy isolation roadmap.

_Surfaced during a residual-tenancy audit that assumed the authorization work is complete._

powderluv:
Fixes the two exploitable-today tenancy items from the multi-tenancy isolation roadmap (epic #665). Both survive the identity spine (#640) and both gate on the verified caller identity, so this targets `main` after #640.

## #663 — restrict client-settable job priority

An unprivileged caller could front-run the entire queue: `spec.priority` was taken verbatim (gated only `> 0`) and used as the base, while the compensating factors are capped (fair-share ≤ 10, age ≤ 2), so one job at `u32::MAX` outranks every default-priority job permanently.

- New `[scheduler] max_user_priority` (default `DEFAULT_PRIORITY` = 1000). A non-admin request above the ceiling is **clamped down and a warning returned**, not rejected — Slurm's `nice`-only model, where raising priority is operator-only.
- `clamp_priority()` is applied at **`submit_job`** and the **`update_job`** priority arm, so the ceiling can't be bypassed post-submit.
- Admin = the token's `admin` claim **or** the accounting `Admin` level.

## #664 — redact cross-tenant job info

`get_job`/`get_job_steps` returned any job's metadata to any caller with no owner scoping (unlike `get_jobs`), leaking work_dir, command, stdio paths, and — most usefully to an attacker — the **allocated nodelist**.

- New `[controller] job_info_visibility` = `redacted` (default) | `owner_only` | `full`.
- Owners and admins always see the full record. Under `redacted`, a non-owner gets identity/state/timing/account but work_dir, command, stdio paths, comment, resource detail, and the nodelist are blanked. `owner_only` returns `NOT_FOUND`; `full` is the legacy behaviour.
- Applied to `get_job` and `get_job_steps` (step names blanked under `redacted`). The disclosure decision is a pure function so the policy matrix is unit-tested.

## Notes

- **One PR, not two:** the fixes share `caller_is_admin` and touch the same two files, so they can't be cleanly independent PRs. Two commits' worth of concern, kept reviewable by putting each fix in distinct functions.
- Default behaviour changes: non-admins can no longer raise priority above 1000, and non-owners no longer see other users' nodelist/command/paths. Both are configurable (`max_user_priority`, `job_info_visibility = full`) for operators who want the old behaviour.
- `get_job_steps` under `owner_only` returns an empty step list rather than `NOT_FOUND` (the RPC already returns empty for unknown jobs, so ownership isn't a distinguisher).

## Testing

Verified on Linux (shark-a): `cargo fmt --check` clean, `clippy` clean, **spurctld 786/786**, **spur-core 518/518**. 5 new unit tests: priority clamp (above/at/below ceiling + admin-exempt), disclosure matrix (owner/admin × 3 policies), redaction field set.

Part of #665. Closes #663. Closes #664.

### DIFF
--- a/crates/spurctld/src/server.rs
+++ b/crates/spurctld/src/server.rs
@@ -405,21 +405,31 @@ impl ControllerService {
         })
     }
 
-    /// Clamp a non-admin caller's requested base priority to the configured ceiling.
+    /// Whether a caller is exempt from the non-admin restrictions (currently the priority ceiling):
+    /// an admin, or a caller with no verified identity. The latter preserves the pre-auth behaviour —
+    /// under `auth.mode = disabled`, or `permissive` with no credential presented, the cluster trusts
+    /// the client, so restricting it would break no-auth deployments (and internal callers, e.g. the
+    /// CLI raising priority in an unauthenticated cluster). `required` mode never reaches here without
+    /// an identity, so the restriction still applies to every real user there.
+    fn caller_is_privileged(&self, identity: Option<&spur_core::auth::Identity>) -> bool {
+        identity.is_none() || self.caller_is_admin(identity)
+    }
+
+    /// Clamp a non-privileged caller's requested base priority to the configured ceiling.
     ///
     /// Slurm lets an ordinary user only *lower* their priority (`nice >= 0`); raising it is
-    /// operator-only. We mirror that: a non-admin request above `[scheduler] max_user_priority`
+    /// operator-only. We mirror that: a request above `[scheduler] max_user_priority`
     /// (default [`DEFAULT_PRIORITY`]) is clamped down and a warning is returned — not rejected, so a
-    /// fat-fingered `--priority` still runs. Admins, and an unset priority, are untouched.
-    /// Returns the warning to surface to the caller, if any. Operates on the raw `Option<u32>` so
-    /// both the submit path (`JobSpec::priority`) and the `scontrol update` path (`req.priority`)
-    /// share one policy — a user cannot dodge the ceiling by boosting priority after submit.
+    /// fat-fingered `--priority` still runs. Privileged callers (admin or unauthenticated, see
+    /// [`Self::caller_is_privileged`]), and an unset priority, are untouched. Operates on the raw
+    /// `Option<u32>` so both the submit path (`JobSpec::priority`) and the `scontrol update` path
+    /// (`req.priority`) share one policy — a user cannot dodge the ceiling by boosting after submit.
     fn clamp_priority(
         priority: &mut Option<u32>,
-        caller_is_admin: bool,
+        caller_is_privileged: bool,
         max_user_priority: u32,
     ) -> Option<String> {
-        if caller_is_admin {
+        if caller_is_privileged {
             return None;
         }
         match *priority {
@@ -627,11 +637,11 @@ impl SlurmController for ControllerService {
         let mut core_spec = proto_to_job_spec(spec)?;
         Self::bind_spec_to_identity(&mut core_spec, identity.as_ref())?;
 
-        // Clamp a non-admin's base priority to the configured ceiling before it reaches the
-        // scheduler, so one submission cannot front-run the whole queue. Non-fatal: clamp + warn.
+        // Clamp a non-privileged caller's base priority to the configured ceiling before it reaches
+        // the scheduler, so one submission cannot front-run the whole queue. Non-fatal: clamp + warn.
         let priority_warning = Self::clamp_priority(
             &mut core_spec.priority,
-            self.caller_is_admin(identity.as_ref()),
+            self.caller_is_privileged(identity.as_ref()),
             self.cluster.config().scheduler.max_user_priority,
         );
 
@@ -973,7 +983,7 @@ impl SlurmController for ControllerService {
             }
         }
 
-        let caller_is_admin = self.caller_is_admin(Self::verified_identity(&request));
+        let caller_is_privileged = self.caller_is_privileged(Self::verified_identity(&request));
         let mut req = request.into_inner();
 
         // Handle hold/release via priority
@@ -990,10 +1000,11 @@ impl SlurmController for ControllerService {
             return Ok(Response::new(()));
         }
 
-        // Same ceiling as submit: a non-admin cannot raise priority above the cap post-submit.
+        // Same ceiling as submit: a non-privileged caller cannot raise priority above the cap
+        // post-submit.
         Self::clamp_priority(
             &mut req.priority,
-            caller_is_admin,
+            caller_is_privileged,
             self.cluster.config().scheduler.max_user_priority,
         );
 
@@ -3853,10 +3864,15 @@ mod tests {
     }
 
     #[test]
-    fn clamp_priority_never_touches_an_admin() {
+    fn clamp_priority_never_touches_a_privileged_caller() {
+        // Privileged = admin or unauthenticated (auth not enforced); see caller_is_privileged.
         let mut p = Some(u32::MAX);
         let warning = ControllerService::clamp_priority(&mut p, true, 1000);
-        assert_eq!(p, Some(u32::MAX), "an admin may set any priority");
+        assert_eq!(
+            p,
+            Some(u32::MAX),
+            "a privileged caller may set any priority"
+        );
         assert!(warning.is_none());
     }
 
--- a/crates/spurctld/src/server.rs
+++ b/crates/spurctld/src/server.rs
@@ -436,18 +436,21 @@ impl ControllerService {
 
     /// Apply the configured job-info visibility policy for `get_job`.
     ///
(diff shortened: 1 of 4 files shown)


crates/spurctld/src/server.rs
```diff
@@ -897,6 +990,13 @@ impl SlurmController for ControllerService {
             return Ok(Response::new(()));
         }
 
+        // Same ceiling as submit: a non-admin cannot raise priority above the cap post-submit.
```
yansun1996: Since `UpdateJob` returns `google.protobuf.Empty`, this return value has nowhere to go — the warning is silently dropped, not even logged. Could we at least `warn!()` here so a clamp on this path is visible somewhere? The docs (config.rs, configuration.rst, examples/spur.conf) currently say a warning is "returned" here, which isn't accurate for this path.

crates/spurctld/src/server.rs
```diff
@@ -1631,14 +1733,45 @@ impl SlurmController for ControllerService {
             return Ok(resp);
         }
 
-        let job_id = req.job_id;
+        // Scope step visibility to the same policy as get_job: a non-owner sees fewer step details.
+        // Step names can leak intent, so they are blanked under `redacted` and the list is empty
+        // under `owner_only`. Owner determined from the parent job; if it is gone from the display
+        // cache, only an admin is privileged.
+        // Owner from the parent job; if it is gone from the display cache, `""` makes only an admin
+        // (or an anonymous caller, per `viewer_is_privileged`) privileged.
```
yansun1996: These two comment blocks say almost the same thing back to back (looks like a leftover from an edit) — worth collapsing into one. Also, separately: defaulting `owner` to `""` when the job isn't in the display cache means the *true owner* of that job gets treated as unprivileged too (since their username != `""`), so they'd see their own step names blanked/hidden. `get_job` instead returns `NOT_FOUND` uniformly for the same "not in display cache" case — might be worth aligning `get_job_steps` to do the same before computing `owner`/`privileged`.

docs/admin-guide/configuration.rst
```diff
@@ -362,6 +377,16 @@ Scheduling loop cadence, per-cycle limits, and fairshare decay.
      - ``300``
      - Live
      - Maximum seconds a job may sit in COMPLETING before it is force-finished.
+   * - ``max_user_priority``
```
yansun1996, line 380: This says a warning is returned for both submit and `scontrol update`, but `UpdateJob` has no response field to carry one — only `sbatch`/submit actually surfaces it. Could the wording clarify that update-path clamping is currently silent to the caller?

examples/spur.conf
```diff
@@ -26,6 +35,12 @@ plugin = "backfill"
 interval_secs = 1
 max_jobs_per_cycle = 10000
 fairshare_halflife_days = 14
+# Highest base priority a non-admin may request (via --priority at submit or
+# scontrol update). Requests above this are clamped down, not rejected, with a
+# warning; admins are unaffected. Defaults to the base priority (1000), so a
+# non-admin can lower but not raise priority — matching Slurm, where boosting
+# priority is operator-only. Raise it to grant users a band above the baseline.
+# max_user_priority = 1000
```
yansun1996, line 47: Same as the configuration.rst note — this implies `scontrol update` returns a warning too, which isn't currently true.

yansun1996 (commented):
Nice, focused fix for two real cross-tenant leaks — the priority front-running and the unscoped `get_job`/`get_job_steps` disclosure. The policy design (pure `viewer_is_privileged`/`job_info_disclosure` functions, config-driven visibility levels) is clean and well unit-tested.

A few things worth addressing:

- **`update_job` has no ownership check at all** — nothing in that handler (or `ClusterManager::update_job`) verifies the caller owns the job before changing `time_limit`/`partition`/`comment`/`account`/`qos`, unlike `cancel_job`, which calls `check_job_owner` first. This predates this PR, but it sits in the very handler you're editing here for the priority ceiling, and it's a bigger tenancy hole than either bug this PR fixes (arbitrary cross-tenant mutation, not just read/priority). Could you file a follow-up issue for it so it doesn't get lost? Happy to see it addressed separately given this PR is already scoped tightly.
- The docs/comments say a priority clamp "returns a warning" — that's only true for `submit_job`. `UpdateJob` returns `google.protobuf.Empty`, so the `scontrol update` path's clamp is currently silent (not even logged). Worth fixing the wording in the three doc spots, and maybe adding a server-side `warn!` so it's at least observable.
- `get_job_steps` and `get_job` disagree on what happens when the job isn't in the display cache: `get_job` returns `NOT_FOUND` for everyone including the owner, but `get_job_steps` falls back to `owner = ""`, which makes even the true owner look unprivileged. See inline comment.
- A couple of the new doc comments run long (`clamp_priority`, `scoped_job_info`) — might be worth trimming per the repo's usual comment style.

Test-wise, the new unit tests nicely cover the pure policy functions, but nothing exercises the actual RPC handlers with an injected identity — a test like that would have caught the `get_job_steps` inconsistency above. Might be worth adding one owner-vs-non-owner integration test through `get_job`/`get_job_steps`.

sajmera-pensando (approved):
LGTM

PR #669, opened 2026-08-18
fix(spur-cli): accept Slurm's -n, -p and -P output flags on sacctmgr

nikhilsk:
## What this fixes

`sacctmgr` defines only `--controller` and `-i/--immediate` as global flags, so Slurm's header
and delimiter flags are rejected before the command runs:

```
$ sacctmgr -n -P show qos format=Name,Priority,MaxWall
error: unexpected argument '-n' found
```

`-n` is only the first flag clap trips over — fixing it alone moves the error to `-P`. The
underlying gap is that `sacctmgr` has no machine-readable output mode at all, so a script
migrated from Slurm has to parse the column-aligned human table, which is precisely what these
flags exist to avoid.

Semantics follow the `sacctmgr` manual:

| Flag | Behaviour |
| --- | --- |
| `-n`, `--noheader` | no header line |
| `-p`, `--parsable` | `\|` delimited, with a trailing `\|` |
| `-P`, `--parsable2` | `\|` delimited, without a trailing `\|` |

The trailing delimiter is the only difference between the two, and it is load-bearing: it
changes the field count that `cut`, `awk`, or `IFS` splitting sees. Both are added, since either
may appear in an existing script. Note the long name for `-P` is `--parsable2`.

## Approach

`format_engine` is shared by `sacctmgr`, `sacct`, `sinfo`, and `squeue`, so its existing
functions are left untouched and the new behaviour is additive: a `RowLayout` enum and an
`OutputStyle` that carries header visibility plus layout, resolved once from the flags and
passed to `show`. `OutputStyle::Aligned` delegates to the existing `format_row` /
`format_header`, so the default path runs the same code as before rather than a
reimplementation of it. The only structural change is that `print_header`'s body moved into
`OutputStyle::header_lines`, with the free function now delegating, so the dashed-rule logic
exists in one place.

Two details worth flagging for review:

- **Delimited rows are not trimmed.** `format_row` ends with `trim_end`, which is right for
  padded output. QOS and account resolvers return `""` for unset limits, so trimming delimited
  output would drop trailing separators and silently shrink the field count.
- **Delimited output ignores widths and truncation**, matching Slurm, and skips the literal
  spacing tokens that `parse_named_format` inserts between fields.

## Trade-offs

**`-p` and `-P` together.** Slurm's parser takes whichever came last. clap's derive cannot see
flag order, so `-P` wins when both are passed. Declaring them mutually exclusive was the
alternative, but that would reject input Slurm accepts.

**Entities without field specs.** `account`, `qos`, and `txn` resolve their columns through
`format_engine`, so delimited output works for them. `user`, `association`, and `tres` print
hardcoded fixed-width tables with no spec table behind them. Rather than accept `-p`/`-P` there
and emit padded text that no script can parse, those combinations fail with a message naming the
entity and the supported ones. Loud beats silently unparseable. `-n` is honoured for every
entity, since suppressing a header needs no column model. Giving the remaining entities proper
field-spec tables would unlock `format=` for them as well, and is better done deliberately than
folded in here.

**Field values are not escaped.** A free-text field carrying a literal `|` (account
`Descr`/`Org`, transaction `Info`) shifts every field after it, so `cut -d'|'` reads the wrong
column. Slurm behaves identically; this is documented rather than worked around.

**Flags may appear after `key=value` params.** `Show`/`List` no longer declare `params` as
`trailing_var_arg`. That attribute captured every token after the first param verbatim, so a
delimiter flag placed after a param was silently swallowed — `show qos format=Name -P` produced
aligned output, and `show qos name=gpu -P format=Name` dropped the `format=` filter and its
following token. With it removed, clap parses a global flag anywhere among the params. The
params are only ever `key=value`/`where`/`set` tokens, so nothing depended on the trailing
capture. A visible consequence: an unrecognised flag after the params now surfaces a clap
"unexpected argument" error instead of being captured as a param — which is the intended
correction.

`Add`/`Delete`/`Modify` carried the same attribute, and there the swallow was worse than
unparsable output: `parse_params` takes the token after a bare word as that word's value, so
`sacctmgr modify qos gpu set -i priority=10` bound `priority=10` to `-i`, applied nothing, and
reported success. They are now declared the same way as `Show`/`List`, with a test that fails
if the attribute comes back.

## Documentation

`docs/admin-guide/accounting.rst` gains a "Scripted output" section covering all three flags,
the trailing-delimiter difference and why it matters to a script's field count, that empty
fields keep their separators so the field count stays stable, that widths and truncation do not
apply, and that field values are not escaped. It also states the two things an operator cannot
infer: delimited output covers `account`, `qos`, and `txn`, and `-P` wins when both delimiter
flags are passed. The migration table carries the entity restriction as a limitation row, since
that is where someone migrating looks first.

## Testing

New unit tests across the two files, including:

- the reported invocation parses, and the flags compose with existing globals
- a delimiter flag placed after, and between, `key=value` params is parsed rather than swallowed
- `-P` has no trailing delimiter, `-p` has one, and a row whose last field is empty keeps its
  separators under both
- `format=` ordering is preserved in delimited output, and values containing commas (TRES
  strings) stay unambiguous
- delimited output drops padding and truncation
- `show txn` honours `-n` and renders delimited output through the resolved style
- **the default path is byte-identical** with no flag and with `-n`, asserted against the
  existing padded helpers, and the padded header block is pinned to exact expected strings
- the unsupported-entity refusal happens before any controller connection, and an unknown
  entity reports itself rather than the delimiter

Gates: `cargo fmt`, `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` (no
warnings), `cargo test --locked` (all passing), `cargo deny check`.

## Follow-up, not addressed here

A `format=` name that Spur does not recognise is silently dropped rather than reported — the
column simply vanishes (`format=Name,Flags` on `show qos` prints only `Name`). That predates
this change and is visible on `main`, but it matters more for delimited output, where a
dropped column shifts every field index after it. Fixing it means choosing between rejecting
unknown names, which changes behaviour for current `format=` callers, and implementing the
missing fields. Worth its own change.

### DIFF
--- a/crates/spur-cli/src/sacctmgr.rs
+++ b/crates/spur-cli/src/sacctmgr.rs
@@ -88,16 +88,11 @@ pub enum SacctmgrCommand {
     Show {
         /// Entity type: account, user, qos, association
         entity: String,
-        /// Optional filter
-        #[arg(trailing_var_arg = true)]
+        /// Optional `key=value` filters; a global flag may sit anywhere among them.
         params: Vec<String>,
     },
     /// List entities (alias for show)
-    List {
-        entity: String,
-        #[arg(trailing_var_arg = true)]
-        params: Vec<String>,
-    },
+    List { entity: String, params: Vec<String> },
 }
 
 pub async fn main() -> Result<()> {
@@ -866,23 +861,12 @@ async fn modify(entity: &str, params: &[String], addr: &str) -> Result<()> {
     }
 }
 
-/// Entities `show` prints as fixed-width tables, with no field-spec table behind them.
-/// A new entity rendered that way belongs here.
-const FIXED_WIDTH_ENTITIES: [&str; 8] = [
-    "user",
-    "users",
-    "association",
-    "associations",
-    "tres",
-    "txn",
-    "transaction",
-    "transactions",
-];
+/// Entities `show` prints as hardcoded fixed-width tables, with no field-spec table behind
+/// them. A new entity rendered that way belongs here.
+const FIXED_WIDTH_ENTITIES: [&str; 5] = ["user", "users", "association", "associations", "tres"];
 
-/// Accepting `-p`/`-P` for a fixed-width entity would hand a script padded text where it
-/// expects delimited fields, so refuse loudly instead of returning output that silently
-/// will not parse. Unrecognized entities are left to `show`, so they still report the
-/// unknown entity rather than a delimiter complaint.
+/// Delimiting a fixed-width table hands a script padded text it cannot parse, so refuse it.
+/// Unknown entities fall through to `show`, which reports them itself.
 fn reject_unsupported_delimiter(entity: &str, style: format_engine::OutputStyle) -> Result<()> {
     if !style.is_delimited() || !FIXED_WIDTH_ENTITIES.contains(&entity) {
         return Ok(());
@@ -890,7 +874,7 @@ fn reject_unsupported_delimiter(entity: &str, style: format_engine::OutputStyle)
 
     bail!(
         "sacctmgr: delimited output (-p/-P) is not supported for '{entity}'. \
-         Supported entities: account, qos"
+         Supported entities: account, qos, txn"
     )
 }
 
@@ -1029,12 +1013,9 @@ async fn show(
                 .into_inner()
                 .transactions;
 
-            format_engine::print_header(&fields);
+            style.print_header(&fields);
             for t in &txns {
-                println!(
-                    "{}",
-                    format_engine::format_row(&fields, &|spec| resolve_txn_field(t, spec))
-                );
+                println!("{}", style.row(&fields, &|spec| resolve_txn_field(t, spec)));
             }
             Ok(())
         }
@@ -2498,6 +2479,42 @@ mod tests {
         assert_eq!(params, vec!["format=Name,Priority".to_string()]);
     }
 
+    #[test]
+    fn a_delimiter_flag_after_a_param_is_still_parsed() {
+        let args =
+            SacctmgrArgs::try_parse_from(["sacctmgr", "show", "qos", "format=Name,Priority", "-P"])
+                .expect("a flag after a param must parse");
+
+        assert!(args.parsable2);
+        let SacctmgrCommand::Show { params, .. } = args.command else {
+            panic!("expected a show command");
+        };
+        assert_eq!(params, vec!["format=Name,Priority".to_string()]);
+    }
+
+    #[test]
+    fn a_delimiter_flag_between_params_does_not_swallow_the_next() {
+        let args = SacctmgrArgs::try_parse_from([
+            "sacctmgr",
+            "show",
+            "qos",
+            "name=gpu",
+            "-P",
+            "format=Name",
+        ])
+        .expect("a flag between params must parse");
+
+        assert!(args.parsable2);
+        let SacctmgrCommand::Show { params, .. } = args.command else {
+            panic!("expected a show command");
+        };
+        assert_eq!(
+            params,
+            vec!["name=gpu".to_string(), "format=Name".to_string()],
+            "the flag must not consume the following filter as its value"
+        );
+    }
+
     fn style_from(flags: &[&str]) -> format_engine::OutputStyle {
         let mut argv = vec!["sacctmgr"];
         argv.extend_from_slice(flags);
@@ -2559,28 +2576,71 @@ mod tests {
     }
 
     #[test]
-    fn header_suppression_applies_to_fixed_width_entities_too() {
+    fn noheader_flag_resolves_into_the_style() {
         assert!(style_from(&[]).shows_header());
         assert!(!style_from(&["-n"]).shows_header());
+    }
+
+    fn stub_txn() -> TransactionRecord {
+        TransactionRecord {
+            id: 7,
+            timestamp: None,
+            actor: "bob".into(),
+            actor_uid: 1000,
+            verified: true,
+            source: "api".into(),
+            action: "create".into(),
+            entity_type: "qos".into(),
+            entity_name: "gpu".into(),
+            outcome: "success".into(),
+            details: "{}".into(),
+        }
+    }
+
+    /// A `show txn` header block and row for the given flags, mirroring the arm's rendering.
+    fn txn_render(flags: &[&str], format: &str) -> (Vec<String>, String) {
+        let fields = txn_format_fields(Some(format)).expect("txn format must parse");
+        let style = SacctmgrArgs::try_parse_from(
+            ["sacctmgr"]
+                .iter()
+                .chain(flags)
+                .chain(&["show", "txn"])
+                .copied()
+                .collect::<Vec<_>>(),
+        )
+        .expect("flags must parse")
+        .output_style();
+        let row = style.row(&fields, &|spec| resolve_txn_field(&stub_txn(), spec));
+        (style.header_lines(&fields), row)
+    }
+
+    #[test]
+    fn txn_honours_noheader_and_delimited_output() {
+        let (header, row) = txn_render(&["-n", "-P"], "Action,Actor");
+        assert!(header.is_empty(), "-n must suppress the txn header");
+        assert_eq!(row, "create|bob");
+
+        let (header, _) = txn_render(&[], "Action,Actor");
+        assert!(!header.is_empty(), "the txn header must print by default");
     }
 
     #[test]
     fn delimited_output_is_refused_only_where_columns_are_not_modelled() {
         let delimited = style_from(&["-P"]);
-        for entity in ["account", "accounts", "qos"] {
+        for entity in [
+            "account",
+            "accounts",
+            "qos",
+            "txn",
+            "transaction",
+            "transactions",
+        ] {
             assert!(
                 reject_unsupported_delimiter(entity, delimited).is_ok(),
                 "{entity} should support delimited output"
             );
         }
-        for entity in [
-            "user",
-            "users",
-            "association",
-            "tres",
-            "txn",
-            "transactions",
-        ] {
+        for entity in FIXED_WIDTH_ENTITIES {
             assert!(
                 reject_unsupported_delimiter(entity, delimited).is_err(),
                 "{entity} should refuse delimited output"
@@ -2589,7 +2649,7 @@ mod tests {
 
         // Padded output stays available for every entity.
         let padded = style_from(&[]);
-        for entity in ["user", "association", "tres"] {
+        for entity in FIXED_WIDTH_ENTITIES {
             assert!(reject_unsupported_delimiter(entity, padded).is_ok());
         }
     }
--- a/crates/spur-cli/src/sacctmgr.rs
+++ b/crates/spur-cli/src/sacctmgr.rs
@@ -28,6 +28,34 @@ pub struct SacctmgrArgs {
     /// Immediate mode (no confirmation)
     #[arg(short = 'i', long, global = true)]
     pub immediate: bool,
+
+    /// Omit the header line
+    #[arg(short = 'n', long, global = true)]
+    pub noheader: bool,
+
+    /// Output '|' delimited with a trailing '|'
+    #[arg(short = 'p', long, global = true)]
+    pub parsable: bool,
+
+    /// Output '|' delimited without a trailing '|'
+    #[arg(short = 'P', long, global = true)]
+    pub parsable2: bool,
+}
+
+impl SacctmgrArgs {
+    /// Slurm takes whichever delimiter flag came last; clap's derive cannot see flag order,
+    /// so `-P` wins when both are given rather than rejecting input Slurm accepts.
+    fn output_style(&self) -> format_engine::OutputStyle {
+        let layout = if self.parsable2 {
+            format_engine::RowLayout::Delimited
+        } else if self.parsable {
+            format_engine::RowLayout::DelimitedTrailing
+        } else {
+            format_engine::RowLayout::Aligned
+        };
+
+        format_engine::OutputStyle::new(self.noheader, layout)
+    }
 }
 
 #[derive(Subcommand, Debug)]
@@ -79,13 +107,14 @@ pub async fn main() -> Result<()> {
 pub async fn main_with_args(args: Vec<String>) -> Result<()> {
     let args = SacctmgrArgs::try_parse_from(&args)?;
     let addr = args.controller.clone();
+    let style = args.output_style();
 
     match args.command {
         SacctmgrCommand::Add { entity, params } => add(&entity, &params, &addr).await,
         SacctmgrCommand::Delete { entity, params } => delete(&entity, &params, &addr).await,
         SacctmgrCommand::Modify { entity, params } => modify(&entity, &params, &addr).await,
         SacctmgrCommand::Show { entity, params } | SacctmgrCommand::List { entity, params } => {
-            show(&entity, &params, &addr).await
+            show(&entity, &params, &addr, style).await
         }
     }
 }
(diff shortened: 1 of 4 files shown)


crates/spur-cli/src/sacctmgr.rs
```diff
@@ -837,10 +866,45 @@ async fn modify(entity: &str, params: &[String], addr: &str) -> Result<()> {
...
+
+async fn show(
+    entity: &str,
+    params: &[String],
+    addr: &str,
+    style: format_engine::OutputStyle,
+) -> Result<()> {
     let p = parse_params(params);
+    let entity = entity.to_lowercase();
+    reject_unsupported_delimiter(&entity, style)?;
```
shiv-tyagi, line 886: `style` is threaded into `show()` here, but the `txn`/`transaction`/`transactions` match arm further down (untouched by this diff) still calls `format_engine::print_header`/`format_engine::format_row` directly instead of `style.print_header`/`style.row`. Every other branch (`account`, `qos`, `user`, `association`, `tres`) routes through `style`. Result: `-n` doesn't suppress the header for `sacctmgr show txn`, contradicting the PR description ("`-n` is honoured for every entity") and the new docs ("`-n` works for every entity").

Fix in the txn arm:
```rust
style.print_header(&fields);
for t in &txns {
    println!("{}", style.row(&fields, &|spec| resolve_txn_field(t, spec)));
}
```

No test catches this: `header_suppression_applies_to_fixed_width_entities_too` only asserts on `OutputStyle::shows_header()` in isolation, it never calls `show()` for `txn` and checks the printed output.

crates/spur-cli/src/sacctmgr.rs
```diff
@@ -837,10 +866,45 @@ async fn modify(entity: &str, params: &[String], addr: &str) -> Result<()> {
...
+fn reject_unsupported_delimiter(entity: &str, style: format_engine::OutputStyle) -> Result<()> {
+    if !style.is_delimited() || !FIXED_WIDTH_ENTITIES.contains(&entity) {
+        return Ok(());
+    }
+
+    bail!(
+        "sacctmgr: delimited output (-p/-P) is not supported for '{entity}'. \
+         Supported entities: account, qos"
+    )
+}
```
shiv-tyagi: Q: `FIXED_WIDTH_ENTITIES` groups `txn` with `user`/`association`/`tres` as entities format_engine "does not model," but `txn` already has a full `FormatToken` model (`txn_format_fields`, `txn_header`, `txn_field_spec`, `TXN_DEFAULT_FORMAT`/`TXN_ALL_FORMAT`), same as `qos`/`account`. The PR description's own rationale ("user, association, and tres print hardcoded fixed-width tables with no spec table behind them") doesn't mention txn, yet the code refuses it too, and the bail message ("Supported entities: account, qos") omits it.

Is `txn` deliberately excluded because `resolve_txn_field('i')` returns raw JSON `details` that could embed a literal `|` and corrupt the field count a script reads? If so, that reasoning should replace the "not modelled" framing. If not, `txn` could support `-p`/`-P` the same way `account`/`qos` do, once the fix above lands.

shiv-tyagi (commented):
Posted review PTAL

crates/spur-cli/src/sacctmgr.rs
yansun1996, line 886: `style` is threaded in here, but the `txn | transaction | transactions` arm below (`:1032`, `:1036`, outside this diff) still calls `format_engine::print_header(&fields)` and `format_engine::format_row(...)` directly. Those resolve to `OutputStyle::default()`, so `sacctmgr -n show txn` still prints the header and its dashed rule — while every other arm routes through `style`. `docs/admin-guide/accounting.rst:905` and `docs/migration-from-slurm.rst:73` both say "`-n` works for every entity", so I'd fix the code rather than soften the docs:

```rust
style.print_header(&fields);
for t in &txns {
    println!("{}", style.row(&fields, &|spec| resolve_txn_field(t, spec)));
}
```

crates/spur-cli/src/sacctmgr.rs
```diff
@@ -837,10 +866,45 @@ async fn modify(entity: &str, params: &[String], addr: &str) -> Result<()> {
     }
 }
 
-async fn show(entity: &str, params: &[String], addr: &str) -> Result<()> {
+/// Entities `show` prints as fixed-width tables, with no field-spec table behind them.
+/// A new entity rendered that way belongs here.
```
yansun1996: Small mismatch: this describes the listed entities as having "no field-spec table behind them", which is accurate for `user`/`association`/`tres` (they print hardcoded `{:<15}` literals), but `txn` has the full model — `TXN_DEFAULT_FORMAT`/`TXN_ALL_FORMAT`, `txn_header`, `txn_field_spec`, `txn_format_fields` — and renders through `FormatToken`s exactly like `qos` and `account`. So it's refused `-p`/`-P` for no structural reason. Once the arm routes through `style`, dropping the three `txn` aliases here would make delimited output work for it for free. If you'd rather keep it refused, could the comment give the actual reason? A `details` field carrying a literal `|` would be a good one.

docs/admin-guide/accounting.rst
```diff
@@ -862,6 +862,51 @@ in the cluster and silently exceeding the cap. ``MaxTRESPerJob``'s and
 a job's actual requested node count, since they bound a single job's or
 user's own footprint rather than group-wide capacity reuse.
 
+Scripted output
+---------------
+
+``sacctmgr show`` prints a column-aligned table for reading. For scripts, three
+Slurm flags change that rendering; they are global, so they may appear before or
+after the subcommand.
```
yansun1996: This contract doesn't quite hold. The flags work before the subcommand, and immediately after the entity, but `params` is declared `trailing_var_arg`, so once the first `key=value` is consumed every later token is captured verbatim — `sacctmgr show qos format=Name,Priority -P` silently produces aligned output. Worth either narrowing this sentence, or dropping `trailing_var_arg` on `Show`/`List` (the params are `key=value` pairs rather than a nested command line), which would let clap report the flag properly.

crates/spur-cli/src/sacctmgr.rs
```diff
@@ -2377,4 +2442,181 @@ mod tests {
...
+        );
+        assert_eq!(
+            style_from(&["-n"]).row(&fields, &|spec| resolve_qos_field(&q, spec)),
+            expected
+        );
+    }
+
+    #[test]
+    fn header_suppression_applies_to_fixed_width_entities_too() {
+        assert!(style_from(&[]).shows_header());
```
yansun1996, line 2601: This one doesn't render a fixed-width entity — it asserts on `style_from(...).shows_header()`, so it's testing the accessor rather than any `show` arm, and it passes even though `show txn` ignores `-n`. Could it either be renamed to something like `noheader_flag_resolves_into_the_style`, or extended to drive `show()` for one of those entities and assert on the printed output? The latter would have caught the `txn` gap.

crates/spur-cli/src/sacctmgr.rs
```diff
@@ -2377,4 +2442,181 @@ mod tests {
...
+                "{entity} should support delimited output"
+            );
+        }
+        for entity in [
+            "user",
+            "users",
+            "association",
+            "tres",
+            "txn",
+            "transactions",
```
yansun1996: This list is a hand-copy of `FIXED_WIDTH_ENTITIES` and drifts from it — `"associations"` and `"transaction"` are in the const but not here, so removing either alias wouldn't fail anything. Iterating `FIXED_WIDTH_ENTITIES` directly would keep the two in sync.

docs/admin-guide/accounting.rst
```diff
@@ -862,6 +862,51 @@ in the cluster and silently exceeding the cap. ``MaxTRESPerJob``'s and
...
+   sacctmgr -n -P show qos format=Name,Priority,MaxWall
+   sacctmgr -n -P show qos format=Name,Priority | cut -d'|' -f1
+
+The trailing delimiter is the only difference between ``-p`` and ``-P``, and it
+changes the field count that ``cut``, ``awk``, and ``IFS`` splitting see. Choose
+one deliberately. Delimited output keeps empty fields as empty, so a row whose
+last columns are unset still carries its separators and the field count stays
+stable across rows.
+
+Delimited output ignores column widths and truncation, so a long value is printed
```
yansun1996, line 901: Would it be worth a sentence here noting that field values aren't escaped? Account `Descr`/`Org` are free text, so a description containing a literal `|` shifts every field after it and a script's `cut -d'|'` reads the wrong column. Slurm behaves identically, so matching it is right — but it's the kind of thing worth stating, since the whole point of these flags is machine parsing.

crates/spur-cli/src/sacctmgr.rs
```diff
@@ -837,10 +866,45 @@ async fn modify(entity: &str, params: &[String], addr: &str) -> Result<()> {
...
+    "users",
+    "association",
+    "associations",
+    "tres",
+    "txn",
+    "transaction",
+    "transactions",
+];
+
+/// Accepting `-p`/`-P` for a fixed-width entity would hand a script padded text where it
```
yansun1996: Minor style point: this doc comment runs to four lines, and `join_fields` in `format_engine.rs:312` has a three-line one. We try to keep comment blocks to two lines and push longer reasoning into the commit message — both would collapse cleanly.

yansun1996 (changes requested):
Really nice piece of work — the `OutputStyle`/`RowLayout` split is clean, and having `Aligned` delegate to the existing `format_row`/`format_header` rather than reimplementing them is the right instinct. I checked the `print_header` → `header_lines` move against the pre-refactor body and it's byte-identical (same rule derivation, same trailing newlines, same empty-token behaviour), so the shared `sacct`/`sinfo`/`squeue` paths are genuinely unaffected — that was the thing most worth getting right here, and it is right. The trailing-delimiter distinction is correctly pinned at both layers, and the unknown-entity-before-delimiter ordering reads as resolved with a test behind it.

Two things I think need addressing before merge, details inline:

1. **`show txn` doesn't honour `-n`.** That arm still calls the free `format_engine::print_header`/`format_row` (`sacctmgr.rs:1032`/`:1036`) rather than going through `style`. Since the free function is `OutputStyle::default()`, the header and rule print regardless. Both docs state "`-n` works for every entity", so the code and the docs disagree today.
2. **A delimiter flag placed after a `key=value` param is silently swallowed, and eats the next token.** `params` is `trailing_var_arg`, so once the first param is consumed everything after it is captured verbatim. `sacctmgr show qos format=Name,Priority -P` quietly gives aligned output; and because `parse_params` treats a bare `-P` as a key and consumes the following token as its value, `sacctmgr show qos name=gpu -P format=Name` silently drops the `format=` filter and renders the default columns. Wrong output rather than an error, which is the part that worries me.

Smaller notes on test scoping and one docs addition are inline too. Nothing here touches the core design, which I think is sound.

crates/spur-cli/src/sacctmgr.rs
```diff
@@ -836,10 +865,34 @@ async fn modify(entity: &str, params: &[String], addr: &str) -> Result<()> {
...
+fn reject_unsupported_delimiter(entity: &str, style: format_engine::OutputStyle) -> Result<()> {
+    if !style.is_delimited() || DELIMITABLE_ENTITIES.contains(&entity) {
+        return Ok(());
+    }
+
+    bail!(
+        "sacctmgr: delimited output (-p/-P) is not supported for '{entity}'. \
+         Supported entities: account, qos"
+    )
+}
```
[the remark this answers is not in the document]
nikhilsk: Already addressed. The check now uses a denylist (`FIXED_WIDTH_ENTITIES`) rather
than the earlier `DELIMITABLE_ENTITIES` allowlist, so an unknown entity is not in
the list, `reject_unsupported_delimiter` returns `Ok`, and it falls through to the
`unknown entity` bail. `sacctmgr -P show <unknown>` reports the unknown entity, not
the delimiter error. This is pinned by `an_unknown_entity_reports_itself_rather_than_the_delimiter`.

---

crates/spur-cli/src/sacctmgr.rs
nikhilsk: Fixed in `c4e0335`. The `txn | transaction | transactions` arm now calls
`style.print_header(&fields)` and `style.row(...)` like every other arm, so `-n`
suppresses the header for `sacctmgr show txn`. Added `txn_honours_noheader_and_delimited_output`,
which renders the txn header/row through the resolved style and asserts `-n` yields
no header lines.

---

crates/spur-cli/src/sacctmgr.rs
nikhilsk: You're right — `txn` has the full `FormatToken` model, so grouping it with the
hardcoded fixed-width entities was wrong. Dropped `txn`/`transaction`/`transactions`
from `FIXED_WIDTH_ENTITIES` in `c4e0335`; delimited output now works for it the same
way as `qos`/`account`, and the bail message lists `account, qos, txn`. The literal
`|` in a `details`/`Info` field is the same non-escaping caveat that already applies
to account `Descr`/`Org`; it is now documented for both (see the `accounting.rst`
thread on escaping) rather than used as a reason to refuse the flag, which matches
Slurm.

---

crates/spur-cli/src/sacctmgr.rs
nikhilsk: Fixed in `c4e0335`, exactly as suggested:

```rust
style.print_header(&fields);
for t in &txns {
    println!("{}", style.row(&fields, &|spec| resolve_txn_field(t, spec)));
}
```

Covered by `txn_honours_noheader_and_delimited_output`.

---

crates/spur-cli/src/sacctmgr.rs
nikhilsk: Resolved by enabling `txn` for delimited output (removed from `FIXED_WIDTH_ENTITIES`),
so the misleading grouping is gone. The remaining const comment now describes only the
genuinely hardcoded entities (`user`/`association`/`tres`).

---

docs/admin-guide/accounting.rst
nikhilsk: Took the second option: dropped `trailing_var_arg` from `Show`/`List` so clap parses
a global flag anywhere among the `key=value` params. The docs sentence now states the
flags may appear anywhere on the command line, including after the entity and its
filters — which is now true.

Behaviour change: `sacctmgr show <entity> ... <unknown-flag>` now errors from clap
("unexpected argument") instead of silently capturing the token as a param. This is
the intended correction (the swallow was the bug), and no other code depended on the
trailing capture — params are only ever `key=value`/`where`/`set` tokens.

---

crates/spur-cli/src/sacctmgr.rs
nikhilsk: Renamed to `noheader_flag_resolves_into_the_style`, which is what it actually checks.
Added a separate `txn_honours_noheader_and_delimited_output` that drives the txn
rendering path (fields + resolved style + real resolver) and asserts on the produced
header block and row — this is the check that would have caught the `txn` gap. A full
end-to-end `show()` assertion needs a controller (the txn arm connects before printing),
so the regression is pinned at the rendering layer the arm now uses.

---

crates/spur-cli/src/sacctmgr.rs
nikhilsk: Fixed — the refused-entity and padded-entity loops now iterate `FIXED_WIDTH_ENTITIES`
directly, so removing or adding an alias can no longer drift from the const. The
supported-delimited loop lists `account, accounts, qos, txn, transaction, transactions`.

---

docs/admin-guide/accounting.rst
nikhilsk: Added a paragraph under the delimited-output section: values are not escaped, a
free-text field containing a literal `|` (account `Descr`/`Org`, transaction `Info`)
shifts later fields and breaks `cut -d'|'`, and Slurm behaves identically.

---

crates/spur-cli/src/sacctmgr.rs
nikhilsk: Both trimmed to two lines: `reject_unsupported_delimiter` in sacctmgr.rs and
`join_fields` in format_engine.rs.

---

shiv-tyagi (approved):
LGTM

yansun1996 (approved):
Thanks — I went back through each thread against the current code and they all check out. The `trailing_var_arg` removal went further than I asked (add/delete/modify too), and both entity loops now iterate `FIXED_WIDTH_ENTITIES` directly, so an alias can no longer drift out of sync. Approving.

PR #670, opened 2026-08-18
feat(spurctld)!: enforce QOS and association submit-time limits

shiv-tyagi:
Implements SPUR-191: validate QOS and association limits at job submit time, so unsatisfiable requests are rejected up front instead of pending forever (or being silently ignored).

## What this adds
- Submit-time limit gate (`enforce_submit_limits`), mirroring Slurm's `acct_policy_validate`:
  - Submit-count families (association `MaxSubmitJobs`/`GrpSubmitJobs`, QOS per-user/per-account/grp) deny unconditionally.
  - Association per-job wall denies unconditionally.
  - Stand-alone TRES/wall breaches deny only under the governing QOS's `DenyOnLimit`. A job with no QOS is treated as `DenyOnLimit`-on, so an unsatisfiable request is rejected rather than pending indefinitely.
- Human-readable denial messages that carry the exact Slurm reason code in parentheses, e.g. `you have reached the maximum submitted jobs for your association (AssocMaxSubmitJobLimit)`. `squeue`-facing reason codes are unchanged for scrapers.
- Read-back of association limits via `sacctmgr show user` (`MaxJobs`, `MaxSubmit`, `GrpSubmit`, `MaxWall`, `MaxTRES`, `GrpTRES`), closing the prior write-only gap.

## Breaking change
Limit values now follow Slurm sentinel semantics: unset = no limit, `-1` clears a limit, and `0` means **block all**. This flips the prior Spur meaning of `0` (was "no limit"). Scripts or configs that set `0` to mean unlimited must switch to `-1`.

Because binaries ship together, this is safe within a release. During a mixed-version window (old CLI against a new controller, or vice versa) unset limits can be misread, so upgrade the CLI and controller together.

## Reason codes
- Per-account submit uses Slurm's canonical `MaxSubmitJobsPerAccount` (previously the non-canonical `QOSMaxSubmitJobPerAccountLimit`).
- New codes `AssocGrpSubmitJobsLimit` and `QOSGrpSubmitJobsLimit`.

## Proto
Append-only, presence-aware (`optional`) fields on `AddUserRequest`, `UserInfo`, `CreateQosRequest`, and `QosInfo`. No tag renumbering, removal, or retyping; `package slurm;` unchanged.

## Docs
`docs/admin-guide/accounting.rst`: limit-value conventions (unset/`-1`/`0`), the new fields (`maxsubmitjobsperaccount`, `grpsubmit`, `DenyOnLimit`, `grpwall`), a "how limits are enforced at submit" section, corrected `modify` partial-patch semantics (omitted fields are preserved, not reset), and the new `show user` columns.

## Testing
- `cargo clippy --workspace --exclude spur-ffi --all-targets --locked`: no warnings.
- `cargo test --locked`: 2965 tests pass, 0 failures.
- Bare-metal 4-node cluster smoke matrix, 49/49: core scheduling, interactive (`srun`/`salloc`/`sattach`), and the full QOS/association enforcement matrix including sentinel semantics and every denial reason code. Verified twice on a freshly reset cluster.

## Follow-ups (not in scope)
- Submit-count TOCTOU under concurrent submits (bounded over-admission; the count read and the persist are not atomic).
- When an association and a QOS submit-count limit both trip, report the QOS reason (Slurm gives QOS precedence); today the association reason wins. Deny/allow outcome is identical.
- Render `MaxWall` as `d-hh:mm:ss` instead of raw minutes.
- Define per-account submit counting for account-less jobs.

### DIFF
--- a/crates/spur-cli/src/sacctmgr.rs
+++ b/crates/spur-cli/src/sacctmgr.rs
@@ -125,9 +125,15 @@ const QOS_KEYS: &[&str] = &[
     "maxwall",
     "maxtresperjob",
     "maxsubmitjobsperuser",
+    "maxsubmitjobsperaccount",
+    "maxsubmitpa",
+    "maxsubmitjobspa",
+    "grpsubmit",
+    "grpsubmitjobs",
     "maxtresperuser",
     "grptres",
     "grpwall",
+    "flags",
 ];
 
 /// Input keys the `add`/`modify user` handlers read (names + aliases). Gates
@@ -143,6 +149,8 @@ const USER_KEYS: &[&str] = &[
     "maxrunningjobs",
     "maxjobs",
     "maxsubmitjobs",
+    "grpsubmit",
+    "grpsubmitjobs",
     "maxtresperjob",
     "grptres",
     "maxwall",
@@ -187,17 +195,31 @@ async fn connect(addr: &str) -> Result<SlurmAccountingClient<crate::authclient::
     Ok(spur_proto::accounting_client(channel))
 }
 
-/// Parse a numeric limit value where `0` is a keyword meaning "no limit".
-/// Fails loudly instead of silently defaulting to `0` so a typo or
-/// out-of-range value never accidentally lifts a limit.
+/// Parse a numeric limit value. `-1` is Slurm's keyword for "no limit" and maps
+/// to the INFINITE sentinel (clears the stored limit); `0` is a literal value
+/// meaning "block all". Any other negative, or a value at/above INFINITE, is
+/// rejected. Fails loudly instead of silently defaulting so a typo never
+/// accidentally lifts or sets a limit.
 fn parse_limit(key: &str, val: &str) -> Result<u32> {
-    val.parse()
-        .map_err(|_| anyhow::anyhow!("invalid value for {key}=: '{val}'"))
+    let n: i64 = val
+        .parse()
+        .map_err(|_| anyhow::anyhow!("invalid value for {key}=: '{val}'"))?;
+    if n == -1 {
+        return Ok(spur_core::accounting::INFINITE);
+    }
+    if n < 0 || n >= spur_core::accounting::INFINITE as i64 {
+        bail!("invalid value for {key}=: '{val}'");
+    }
+    Ok(n as u32)
 }
 
 /// Same as `parse_limit`, but for wall-time fields that also accept Slurm's
-/// `d-hh:mm:ss`/`hh:mm:ss` duration syntax (see `parse_wall_time`).
+/// `d-hh:mm:ss`/`hh:mm:ss` duration syntax (see `parse_wall_time`). `-1` clears
+/// the limit (INFINITE sentinel).
 fn parse_wall_limit(key: &str, val: &str) -> Result<u32> {
+    if val == "-1" {
+        return Ok(spur_core::accounting::INFINITE);
+    }
     parse_wall_time(val).ok_or_else(|| anyhow::anyhow!("invalid value for {key}=: '{val}'"))
 }
 
@@ -254,6 +276,7 @@ struct UserUpsertFields {
     allowed_qos: String,
     max_running_jobs: u32,
     max_submit_jobs: u32,
+    grp_submit_jobs: u32,
     max_tres_per_job: String,
     grp_tres: String,
     max_wall_minutes: u32,
@@ -308,21 +331,28 @@ fn build_add_user_request(
     {
         bail!("defaultqos={default_qos} must be included in qos={allowed_qos}");
     }
+    // An unset numeric limit sends INFINITE (clear/no-limit) on `add`; a literal
+    // 0 would mean "block all" under the sentinel semantics.
+    let no_limit = spur_core::accounting::INFINITE;
     let max_running_jobs: u32 = find_alias(p, &["maxrunningjobs", "maxjobs"])
         .map(|(k, v)| parse_limit(k, v))
         .transpose()?
-        .unwrap_or(0);
+        .unwrap_or(no_limit);
     let max_submit_jobs: u32 = p
         .get("maxsubmitjobs")
         .map(|v| parse_limit("maxsubmitjobs", v))
         .transpose()?
-        .unwrap_or(0);
+        .unwrap_or(no_limit);
+    let grp_submit_jobs: u32 = find_alias(p, &["grpsubmit", "grpsubmitjobs"])
+        .map(|(k, v)| parse_limit(k, v))
+        .transpose()?
+        .unwrap_or(no_limit);
     let max_tres_per_job = p.get("maxtresperjob").cloned().unwrap_or_default();
     let grp_tres = p.get("grptres").cloned().unwrap_or_default();
     let max_wall_minutes: u32 = find_alias(p, &["maxwall", "maxwallduration"])
         .map(|(k, v)| parse_wall_limit(k, v))
         .transpose()?
-        .unwrap_or(0);
+        .unwrap_or(no_limit);
 
     Ok(UserUpsertFields {
         name,
@@ -332,6 +362,7 @@ fn build_add_user_request(
         allowed_qos,
         max_running_jobs,
         max_submit_jobs,
+        grp_submit_jobs,
         max_tres_per_job,
         grp_tres,
         max_wall_minutes,
@@ -368,7 +399,7 @@ async fn add(entity: &str, params: &[String], addr: &str) -> Result<()> {
             let max_jobs: u32 = find_alias(&p, &["maxrunningjobs", "maxjobs"])
                 .map(|(k, v)| parse_limit(k, v))
                 .transpose()?
-                .unwrap_or(0);
+                .unwrap_or(spur_core::accounting::INFINITE);
             let grp_tres = p.get("grptres").cloned().unwrap_or_default();
 
             let mut client = connect(addr).await?;
@@ -412,6 +443,7 @@ async fn add(entity: &str, params: &[String], addr: &str) -> Result<()> {
                     allowed_qos: Some(fields.allowed_qos.clone()),
                     max_running_jobs: Some(fields.max_running_jobs),
                     max_submit_jobs: Some(fields.max_submit_jobs),
+                    grp_submit_jobs: Some(fields.grp_submit_jobs),
                     max_tres_per_job: Some(fields.max_tres_per_job.clone()),
                     grp_tres: Some(fields.grp_tres.clone()),
                     max_wall_minutes: Some(fields.max_wall_minutes),
@@ -429,13 +461,17 @@ async fn add(entity: &str, params: &[String], addr: &str) -> Result<()> {
             if !fields.default_qos.is_empty() {
                 println!("  DefQOS     = {}", fields.default_qos);
             }
-            if fields.max_running_jobs > 0 {
+            let unset = spur_core::accounting::INFINITE;
+            if fields.max_running_jobs != unset {
                 println!("  MaxJobs    = {}", fields.max_running_jobs);
             }
-            if fields.max_submit_jobs > 0 {
+            if fields.max_submit_jobs != unset {
                 println!("  MaxSubmit  = {}", fields.max_submit_jobs);
             }
-            if fields.max_wall_minutes > 0 {
+            if fields.grp_submit_jobs != unset {
+                println!("  GrpSubmit  = {}", fields.grp_submit_jobs);
+            }
+            if fields.max_wall_minutes != unset {
                 println!("  MaxWall    = {} min", fields.max_wall_minutes);
             }
             if !fields.max_tres_per_job.is_empty() {
@@ -463,21 +499,22 @@ async fn add(entity: &str, params: &[String], addr: &str) -> Result<()> {
                 .get("usagefactor")
                 .and_then(|v| v.parse().ok())
                 .unwrap_or(1.0);
+            let no_limit = spur_core::accounting::INFINITE;
             let max_jobs: u32 = find_alias(&p, &["maxjobsperuser", "maxjobspu"])
                 .map(|(k, v)| parse_limit(k, v))
                 .transpose()?
-                .unwrap_or(0);
+                .unwrap_or(no_limit);
             let max_wall: u32 = p
                 .get("maxwall")
                 .map(|v| parse_wall_limit("maxwall", v))
                 .transpose()?
-                .unwrap_or(0);
+                .unwrap_or(no_limit);
             let max_tres = p.get("maxtresperjob").cloned().unwrap_or_default();
             let grp_wall: u32 = p
                 .get("grpwall")
                 .map(|v| parse_wall_limit("grpwall", v))
                 .transpose()?
-                .unwrap_or(0);
+                .unwrap_or(no_limit);
 
             let mut client = connect(addr).await?;
             client
@@ -495,7 +532,7 @@ async fn add(entity: &str, params: &[String], addr: &str) -> Result<()> {
                         p.get("maxsubmitjobsperuser")
                             .map(|v| parse_limit("maxsubmitjobsperuser", v))
                             .transpose()?
-                            .unwrap_or(0),
+                            .unwrap_or(no_limit),
                     ),
                     max_tres_per_user: Some(p.get("maxtresperuser").cloned().unwrap_or_default()),
                     grp_tres: Some(p.get("grptres").cloned().unwrap_or_default()),
@@ -505,6 +542,22 @@ async fn add(entity: &str, params: &[String], addr: &str) -> Result<()> {
                         .map(|v| parse_u32("preemptexempttime", v))
                         .transpose()?,
                     clear_preempt_exempt_time: false,
+                    max_submit_jobs_per_account: Some(
+                        find_alias(
+                            &p,
+                            &["maxsubmitjobsperaccount", "maxsubmitpa", "maxsubmitjobspa"],
+                        )
+                        .map(|(k, v)| parse_limit(k, v))
+                        .transpose()?
+                        .unwrap_or(no_limit),
+                    ),
+                    grp_submit_jobs: Some(
+                        find_alias(&p, &["grpsubmit", "grpsubmitjobs"])
+                            .map(|(k, v)| parse_limit(k, v))
+                            .transpose()?
+                            .unwrap_or(no_limit),
+                    ),
+                    flags: Some(p.get("flags").cloned().unwrap_or_default()),
                 })
                 .await
                 .context("CreateQos RPC failed")?;
@@ -513,10 +566,10 @@ async fn add(entity: &str, params: &[String], addr: &str) -> Result<()> {
                 " Adding QOS(s)\n  Name       = {}\n  Priority   = {}\n  Preempt    = {}",
                 name, priority, preempt
             );
-            if max_wall > 0 {
+            if max_wall != no_limit {
                 println!("  MaxWall    = {} min", max_wall);
             }
-            if max_jobs > 0 {
+            if max_jobs != no_limit {
                 println!("  MaxJobsPU  = {}", max_jobs);
             }
             println!(" QOS added.");
@@ -676,6 +729,16 @@ fn build_modify_qos_request(
             .get("clearpreemptexempttime")
             .map(|v| is_truthy(v))
             .unwrap_or(false),
+        max_submit_jobs_per_account: find_alias(
+            p,
+            &["maxsubmitjobsperaccount", "maxsubmitpa", "maxsubmitjobspa"],
+        )
+        .map(|(k, v)| parse_limit(k, v))
+        .transpose()?,
+        grp_submit_jobs: find_alias(p, &["grpsubmit", "grpsubmitjobs"])
+            .map(|(k, v)| parse_limit(k, v))
+            .transpose()?,
+        flags: p.get("flags").cloned(),
     })
 }
 
@@ -713,6 +776,9 @@ fn build_modify_user_request(
         max_submit_jobs: p
             .get("maxsubmitjobs")
             .map(|v| parse_limit("maxsubmitjobs", v))
+            .transpose()?,
+        grp_submit_jobs: find_alias(p, &["grpsubmit", "grpsubmitjobs"])
+            .map(|(k, v)| parse_limit(k, v))
             .transpose()?,
         max_tres_per_job: p.get("maxtresperjob").cloned(),
         grp_tres: p.get("grptres").cloned(),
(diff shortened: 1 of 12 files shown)


sajmera-pensando (changes requested):
Code review findings — please address before merge.

**[1] Race condition on submit-count check (TOCTOU)**
The gate in `enforce_submit_limits` reads the live submit count and then persists the new job in two non-atomic steps. Under concurrent submits at the limit boundary, two jobs can both pass the check and land in the Raft log. The PR acknowledges this as "bounded over-admission."

Please file a follow-up Jira ticket for this (suggested priority: **P3**) and link it in the PR description so it's tracked. No fix needed in this PR.

---

**[2] `ALTER TABLE` migrations missing `IF NOT EXISTS` guards**
The new `ALTER TABLE` statements in `db.rs` run unconditionally on startup. On a cluster where a previous partial upgrade already added one of the new columns, startup will crash. Please add `IF NOT EXISTS` column guards (e.g. `ALTER TABLE qos ADD COLUMN IF NOT EXISTS max_submit_per_account INTEGER`) to all new `ALTER TABLE` calls.

---

**[3] Missing test for `0` stored in DB in `accounting/mod.rs`**
The predicate flip `x > 0` → `x >= 0` in `opt_u32` (in `accounting/mod.rs`) is the key behavior change that makes `0` mean "block all" rather than NULL. There is no test covering a `grp_submit_jobs = 0` value coming out of the DB through this path. Please add one so this can't silently regress.

---

**[4] Inconsistent handling of unknown QOS flags**
`canonicalize_qos_flags()` in `grpc.rs` hard-errors on any unrecognized flag at write time. `parse_deny_on_limit()` in `limits_cache.rs` silently ignores unrecognized flags at read time. These should behave consistently. Suggest making the write path also silently ignore unknown flags (warn + skip) so that importing a Slurm sacctmgr dump with future flags like `NoDecay` doesn't break `create_qos` calls.

---

**[5] `format_user_row` not tested for mixed set/unset limits**
`format_user_row_renders_limits_and_blanks_unset` only covers the all-unset case. The realistic scenario — some limits set to concrete values, others `INFINITE` (blank) — is untested. Please add a test case with at least one set limit and one unset limit to guard the mixed rendering path.

crates/spurctld/src/accounting/grpc.rs
```diff
@@ -31,17 +31,45 @@ fn nullable_str(field: &Option<String>) -> Option<Option<&str>> {
 }
 
 /// Map an optional u32 limit proto field to a nullable-int patch: unset ->
-/// keep, 0 -> clear (no limit), n -> set. Errors if `n` overflows i32.
+/// keep, INFINITE (`u32::MAX`) -> clear (no limit / SQL NULL), n -> set the
+/// literal (including 0, which means "block all"). Errors if `n` overflows i32.
 fn nullable_limit(field: Option<u32>, what: &str) -> Result<Option<Option<i32>>, Status> {
```
yansun1996, line 36: This is the mixed-version edge I flagged in the summary. Before this change `Some(0)` was the clear sentinel; now `0` is stored literally and read back by the submit gate as "block all". An un-upgraded CLI, or any existing REST/FFI caller that hasn't been updated, will keep sending `0` to clear a limit and will instead silently block every submission for that association.

Would it be worth accepting both `0` and `INFINITE` as "clear" on the write path for one release, so the new meaning only arrives once clients have caught up? Failing that, a startup warning when any limit column holds a literal `0` would at least make the state discoverable.

yansun1996 (commented):
Nice piece of work — the submit gate is well-structured, the reason codes are faithful to Slurm, and closing the write-only read-back gap is a real improvement. The bare-metal matrix is reassuring. A few things I'd like to resolve before merge, plus some smaller notes.

**Main concern — the sentinel flip during a mixed-version window.** `nullable_limit` moved `Some(0) => clear` to `Some(INFINITE) => clear`. A login node still running the old CLI (or an existing REST/FFI caller) that sends `0` meaning "clear this limit" now writes a literal `0`, which the new gate reads as *block every submission* for that association — silently, with no error. The reverse direction is also visible: `limit_to_proto` now emits `4294967295` where an old `sacctmgr show` expects `0`. The PR body covers this as "upgrade the CLI and controller together", which is the right guidance, but would it be worth accepting `0` as "clear" on the write path for one release, or at least logging loudly at startup if any `0`-valued limit rows exist? A silent cluster-wide submit block is a rough failure mode to debug.

**Test coverage.** `nullable_limit`, `limit_to_proto`, and `canonicalize_qos_flags` in `accounting/grpc.rs` are the three pure functions this PR's semantics hinge on, and they currently have no unit tests — reverting `Some(INFINITE) => Ok(Some(None))` back to `Some(0)` would pass CI today. Same for `account_limits_from_record` in `accounting/mod.rs` (this is the point another reviewer raised about the `> 0` → `>= 0` flip; `limits_cache.rs` does cover its side, the association side doesn't). Beyond those: no test produces `AssocGrpSubmitJobsLimit` at all, and the cold-cache and unknown-QOS-name paths through `enforce_submit_limits` are untested.

**On the existing review comments:**

- The `ALTER TABLE` guards are already there — `db.rs` uses `ADD COLUMN IF NOT EXISTS` on all four new columns, so a partially-upgraded cluster should be fine.
- Agree with the note on `incoming_tasks` falling back to `1` on a parse failure — worth resolving the array spec once, before the gate, and surfacing the real error.
- On unknown QOS flags: the strict-write / lenient-read split is defensible (the write-path comment argues it well), but a warn-and-skip against an explicit allow-list would keep a Slurm dump containing e.g. `NoDecay` from failing the whole `create_qos`. Your call — worth a line in the docs either way.

**Smaller notes, no strong opinion:**

- `grp_submit_jobs` is read from the submitting user's own association row but compared against an account-wide count, so two users in one account with different `grpsubmit` values get different verdicts for the same aggregate. Probably wants a follow-up rather than a fix here.
- Both caches refresh on a timer with no invalidation on the write path, so a limit change takes up to one refresh interval to apply. That was invisible before; now it means a just-cleared limit keeps hard-rejecting. Worth a sentence in the docs.
- `submit_denial_message()`'s doc says "submit-time denial", but the `AssocMaxJobsLimit` and `QoSMaxJobsPerUser` arms aren't reachable from the gate (running-count limits still pend). Either the arms or the doc could go.
- `opt_u32` is now defined identically in `limits_cache.rs` and `accounting/mod.rs` — a shared fn next to `INFINITE` in `spur-core::accounting` would give both one place to test.
- Docs: the "Two families behave differently" section is missing a third — running-count limits (`maxjobs`, `maxjobsperuser`) never deny, they always pend. And "Where a QOS limit is defined, it takes precedence over the association's" isn't true of the submit gate, which checks association limits first and returns regardless of QOS. Since the reason codes are user-visible and slightly inconsistent (`AssocGrpSubmitJobsLimit`, `QOSGrpSubmitJobsLimit`, but a bare `MaxSubmitJobsPerAccount`), a small table of the exact strings would help scrapers.
- A handful of new comment blocks run to 3+ lines (`cluster.rs:635`, `4298`, `4351`; `qos.rs:83`, `199`; `account_limits.rs:41`, `129`; `job.rs:400`; `accounting.rs:12`; and the rewritten `AddUserRequest`/`CreateQosRequest` proto headers at 6 lines each). Repo convention keeps these to two — most would collapse fine.

crates/spurctld/src/accounting/grpc.rs
```diff
@@ -31,17 +31,45 @@ fn nullable_str(field: &Option<String>) -> Option<Option<&str>> {
...
+    match v {
+        Some(n) if n >= 0 => n as u32,
+        _ => spur_core::accounting::INFINITE,
+    }
+}
+
+/// Validate and canonicalize the QOS `flags` string. Only `DenyOnLimit` is
+/// recognized; an unknown token errors loudly rather than being silently
+/// dropped (a dropped flag reads as "set" but never takes effect).
+fn canonicalize_qos_flags(raw: &str) -> Result<String, Status> {
```
yansun1996, line 49: `nullable_limit`, `limit_to_proto`, and `canonicalize_qos_flags` are the three functions the sentinel change actually turns on, and they're pure — but the test module below only covers `validate_tres_*` and `fairshare_to_i32_*`. As it stands, reverting `Some(INFINITE) => Ok(Some(None))` back to `Some(0)` would still pass CI.

A few cheap cases would lock this down: `nullable_limit(None)` → keep, `Some(INFINITE)` → clear, `Some(0)` → literal `0`, and an over-`i32::MAX` value → error; `limit_to_proto(None)` and a stray negative → `INFINITE`, `Some(0)` → `0`; and for flags, unknown token → error, mixed case accepted, empty → empty.

crates/spurctld/src/accounting/mod.rs
```diff
@@ -124,7 +124,9 @@ pub async fn association_maps(
 }
 
 fn account_limits_from_record(r: db::AssociationRecord) -> AccountLimits {
-    let opt_u32 = |v: Option<i32>| v.filter(|&x| x > 0).map(|x| x as u32);
+    // A stored NULL (None) is "no limit"; a literal 0 is a real value (block
+    // all). A stray negative predates the sentinel flip; treat it as unset.
+    let opt_u32 = |v: Option<i32>| v.filter(|&x| x >= 0).map(|x| x as u32);
```
yansun1996: This `x > 0` → `x >= 0` flip is the association-side half of the new semantics, and it has no test — there are no cases for `account_limits_from_record` at all in this module. (`limits_cache.rs` does cover its own copy, which is why the gap is easy to miss.)

A record with `grp_submit_jobs: Some(0)` asserting it survives as `Some(0)`, plus `Some(-1)`/`None` → `None`, would cover it. If the duplicated closure moves to a shared fn in `spur-core::accounting`, one test could serve both call sites.

crates/spurctld/src/cluster.rs
```diff
@@ -4277,6 +4295,118 @@ impl ClusterManager {
...
+                if let Some(reason) = check_account_wall_limit(&job, &limits) {
+                    return Err(deny(reason));
+                }
+            }
+        }
+
+        // QOS submit-count + (DenyOnLimit-gated) stand-alone resource breach.
+        // A job with no QOS is treated as DenyOnLimit-on so an unsatisfiable
+        // association request is rejected rather than pending forever.
+        let mut deny_on_limit = true;
```
yansun1996, line 4523: Two things about this default that might be worth separating:

[1] It stays `true` not just when the job has no QOS (the documented case), but also when `qos_cache.is_loaded()` is false or the named QOS simply isn't in the cache yet. During the controller's startup window a job under a `DenyOnLimit=false` QOS would be hard-rejected on a stand-alone breach that should have pended — failing closed on a transient condition. Skipping the stand-alone checks entirely when the governing QOS can't be resolved would avoid that.

[2] The same flag then gates the *association* stand-alone check below. That means attaching a permissive QOS turns association enforcement off: the same job is rejected with no QOS, but admitted to pend forever under `qos=normal` with `DenyOnLimit` unset. Tracking association gating separately from the QOS flag would keep the two policies independent.

crates/spurctld/src/cluster.rs
```diff
@@ -4277,6 +4295,118 @@ impl ClusterManager {
...
+    /// (treated as on when the job has no QOS). `incoming` is the task count added.
+    fn enforce_submit_limits(&self, spec: &JobSpec, incoming: u32) -> Result<(), SubmitError> {
+        let jobs = self.jobs.read();
+        let user = spec.user.as_str();
+        let account = spec.account.as_deref().filter(|a| !a.is_empty());
+        let qos_name = spec.qos.as_deref().filter(|n| !n.is_empty());
+
+        // Representative job for the stand-alone/wall breach checks. The job_id
+        // is irrelevant here (these evaluate the request in isolation).
+        let job = Job::new(0, spec.clone());
```
yansun1996: Two allocations worth trimming on what is now a per-submission path:

[1] This clones the whole `JobSpec`, batch script included, purely to satisfy the `&Job` signature of `check_*_standalone_limits` / `check_account_wall_limit`. Those helpers only read `job.spec` — taking `&JobSpec` would drop the clone.

[2] Below this, the gate makes up to five separate `jobs.values().filter(..).count()` passes, all under one `self.jobs.read()`. They share the `is_submitted` predicate, so a single pass accumulating five counters would do, and would shorten the window during which the scheduler's `jobs.write()` is blocked. Skipping the counters whose limit isn't set would help further.

crates/spur-cli/src/sacctmgr.rs
```diff
@@ -184,17 +192,31 @@ async fn connect(addr: &str) -> Result<SlurmAccountingClient<crate::authclient::
...
+        bail!("invalid value for {key}=: '{val}'");
+    }
+    Ok(n as u32)
 }
 
 /// Same as `parse_limit`, but for wall-time fields that also accept Slurm's
-/// `d-hh:mm:ss`/`hh:mm:ss` duration syntax (see `parse_wall_time`).
+/// `d-hh:mm:ss`/`hh:mm:ss` duration syntax (see `parse_wall_time`). `-1` clears
+/// the limit (INFINITE sentinel).
 fn parse_wall_limit(key: &str, val: &str) -> Result<u32> {
```
yansun1996, line 219: `parse_limit` rejects anything `>= INFINITE`, but this path doesn't — after the `-1` check it falls through to `parse_wall_time`, whose bare-integer branch happily accepts `4294967295`. So `maxwall=4294967295` is taken as a valid minute count, equals the sentinel, and silently *clears* the limit instead of setting it.

Adding the same upper-bound check after `parse_wall_time` returns would make the two parsers symmetric.

crates/spur-cli/src/sacctmgr.rs
```diff
@@ -1613,17 +1733,88 @@ mod tests {
...
+        let row = format_user_row(&u);
+        let cols: Vec<&str> = row.split_whitespace().collect();
+        assert!(cols.contains(&"carol"));
+        assert!(cols.contains(&"4"));
+        assert!(cols.contains(&"8"));
+        assert!(cols.contains(&"1440"));
+        assert!(cols.contains(&"cpu=64"));
+        // grp_submit_jobs is the INFINITE sentinel: rendered as a blank cell, so
+        // no stray value shows up between MaxSubmit and MaxWall.
+        assert!(row.contains("8"));
```
yansun1996: This assertion is implied by `assert!(cols.contains(&"8"))` five lines up, so it can't fail independently — and the comment above it claims it proves the unset `GrpSubmit` cell is blank, which it doesn't check at all. `row.contains("8")` would still pass if a stray value appeared in that column.

Asserting on the fixed-width column slice (that the `GrpSubmit` field is all spaces) would make it real, and would also cover the mixed set/unset rendering case raised in the other review.

crates/spur-core/src/qos.rs
```diff
@@ -38,7 +38,126 @@ pub enum QosCheckResult {
     Blocked(PendingReason),
 }
 
-/// Check if a job would violate QOS limits.
+/// Whether the job's requested wall time exceeds `max_wall` minutes.
+fn wall_breach(job: &Job, max_wall: u32) -> bool {
```
yansun1996: `is_some_and` means a job submitted without `--time` never trips this, so `maxwall=0` blocks a job with `--time=1` but lets a time-less job straight through. The docs put wall time under the blanket "`0` = block all" convention, so that reads as a hole.

Treating a `0` cap as an unconditional breach (or `map_or(true, ..)` when the cap is `0`) would line the behaviour up with the documented rule. The same shape is in `account_limits.rs`.

docs/admin-guide/accounting.rst
```diff
@@ -45,6 +45,22 @@ each builds on the one before it:
...
+~~~~~~~~~~~~~
+
+Numeric limits (job counts, wall time) share one convention throughout this
+page:
+
+- **Unset / omitted** — no limit. This is the default for any limit you do not
+  name.
+- ``-1`` — clears a limit back to "no limit". Use it on ``modify`` to lift a cap.
+- ``0`` — a literal value meaning **block all**: a ``maxsubmitjobs=0`` rejects
+  every submission for that association. ``0`` does **not** mean "no limit".
```
yansun1996: Existing stored rows look safe — the old write path mapped `Some(0)` to SQL NULL, so a literal `0` wasn't reachable through gRPC — but *scripts* are the exposure. `sacctmgr modify user … set maxsubmitjobs=0` was the documented way to clear a limit before this change, and the same command now blocks every submission for that user. Any config-as-code or Slurm-migration script re-run after the upgrade will do exactly that.

Would it be worth a `.. warning::` here calling that out, with the `-1` remediation and a query admins can run to find rows currently holding a literal `0`?

shiv-tyagi: I don't think it is worth to create a mess with a warning there. The old behavior was anyways incorrect.

crates/spurctld/src/accounting/grpc.rs
shiv-tyagi: Clients are shipped with the controllers in a release. Not worth the hassle.

crates/spurctld/src/cluster.rs
```diff
@@ -627,6 +632,19 @@ impl ClusterManager {
...
+        // QOS) unsatisfiable stand-alone resource requests before the job ever
+        // enters the queue. The scheduler path stays a safety net for the
+        // limit-lowered-after-submit edge case. Each array task counts
+        // individually toward submit caps; the whole submission is rejected.
+        let incoming_tasks = match spec.array_spec.as_deref() {
+            Some(s) => spur_core::array::parse_array_spec(s)
+                .map(|a| a.task_ids.len() as u32)
+                .unwrap_or(1),
+            None => 1,
+        };
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed in 63e4be01. `submit_job` now parses the array spec up front and returns `SubmitError::invalid` on failure, so a malformed or oversized spec surfaces its real error before the submit-limit gate runs. The gate sees the parsed `task_ids.len()` (bounded by `MAX_ARRAY_SIZE`), not a silent `1`.

proto/slurm.proto
```diff
@@ -1258,6 +1258,8 @@ message CreateAccountRequest {
   optional string organization = 3;
   optional string parent_account = 4;
   optional double fairshare_weight = 5;
+  // Unset preserves the stored value; the INFINITE sentinel (4294967295) clears
+  // to no-limit; any other value (including a literal 0) is stored as-is.
   optional uint32 max_running_jobs = 6;
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed in 63e4be01. The message-level comment now matches the fields: empty string clears text fields, `INFINITE` clears a numeric limit, and a literal `0` is stored as-is (block all).

docs/admin-guide/accounting.rst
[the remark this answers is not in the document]
shiv-tyagi: Fixed in 63e4be01. The conventions section now states omission means "no limit" on `add` and "preserve existing" on `modify`.

crates/spurctld/src/accounting/grpc.rs
shiv-tyagi: Added in 63e4be01. The sentinel flip now lives in one place, `spur_core::accounting::limit_from_db` / `limit_to_wire`, with canonical tests `limit_from_db_maps_sentinels` and `limit_to_wire_maps_null_and_values` covering None-keep, `INFINITE`-clear, `Some(0)`-literal, negative-clear, and over-`i32::MAX`. The grpc layer keeps `nullable_limit_maps_sentinels`, `nullable_limit_rejects_over_i32_max`, and `canonicalize_qos_flags_normalizes_and_validates`. Reverting `Some(INFINITE) => Ok(Some(None))` to `Some(0)` now fails these.

crates/spurctld/src/accounting/mod.rs
shiv-tyagi: Done in 63e4be01, both parts. The closure moved to the shared `spur_core::accounting::limit_from_db`, and `accounting/mod.rs` covers the association path: `account_limits_preserve_zero_as_block_all` (`Some(0)` survives), `account_limits_map_null_and_negative_to_unset` (`None`/`-1` -> `None`), and `account_limits_pass_through_positive`.

crates/spurctld/src/cluster.rs
shiv-tyagi: [1] A named `--qos` that can't be resolved never reaches this default. `apply_default_qos` runs first (cluster.rs:561) and rejects an unknown QOS with `QOS '...' does not exist`, including when the cache is unloaded (test `apply_default_qos_explicit_invalid_hints_when_cache_unloaded`). So `deny_on_limit` only stays `true` here for genuinely no-QOS jobs, the documented case, not for a `DenyOnLimit=false` QOS during a startup window.

[2] Coupling association stand-alone gating to the governing QOS `DenyOnLimit` is intentional Slurm parity: in Slurm `DenyOnLimit` governs QOS and association limits alike, so a permissive QOS pends an association TRES breach instead of rejecting it, and a no-QOS job is treated as `DenyOnLimit`-on. This is documented in accounting.rst. Decoupling would diverge from Slurm, so I would keep them coupled.

crates/spurctld/src/cluster.rs
shiv-tyagi: Both allocation points addressed. The submit-count scans already skip any limit that isn't set, so the common no-limit path does zero passes. The `JobSpec` clone is dropped in 3ca38b54: `check_account_wall_limit`, `check_qos_standalone_limits`, `check_account_standalone_limits`, and the private `wall_breach` / `job_tres` / resource-breach helpers now take `&JobSpec`, so the gate passes `spec` directly with no clone. The running-count `check_*_limits` keep their `&Job` signatures and call inward with `&job.spec`.

Left the five scans as separate passes: with skip-unset they only run for set limits (usually one or two), and the predicates differ per counter (`user+account`, `account`, `user+qos`, `account+qos`, `qos`), so a single loop would still test every predicate per job for marginal gain.

crates/spur-cli/src/sacctmgr.rs
shiv-tyagi: Fixed in 63e4be01. After `parse_wall_time`, `parse_wall_limit` rejects a value that lands on `INFINITE`, so `maxwall=4294967295` errors instead of silently clearing. Symmetric with `parse_limit`.

crates/spur-cli/src/sacctmgr.rs
shiv-tyagi: Fixed in 63e4be01. `format_user_row_renders_limits_and_blanks_unset` now asserts the fixed-width `GrpSubmit` slice is all spaces, so it fails if a stray value lands in the unset column.

crates/spur-core/src/qos.rs
shiv-tyagi: Fixed in 63e4be01. `wall_breach` now treats a `0` cap as an unconditional breach, so `maxwall=0` blocks a time-less job too, matching the documented "0 = block all". Same change in `account_limits.rs`. Added `max_wall_zero_blocks_time_less_job` and `positive_max_wall_leaves_time_less_job_alone` in both.

biluriuday (approved):
LGTM

sajmera-pensando (approved):
LGTM

PR #674, opened 2026-08-18
[SPUR-183]: purge k0s join token on teardown even when reset fails

spraveenio:
## Problem

After `spur k8s down --reset` + `spur k8s up` on the same long-lived `spurctld`
leader, a worker failed to join: unit `active`, but absent from `kubectl get
nodes`, journal looping on `crypto/rsa: verification error ... kubernetes-ca`.

`down --reset` re-bootstraps the control plane with a **regenerated CA**. A join
token minted against the previous CA no longer validates.

## Root cause

Two carriers of a stale-CA token survived teardown:

1. **Node side (operative).** `ClusterSupervisor::stop` and
   `K0sAgent::stop_untracked` awaited `k0s reset` with `?` *before* `purge_unit()`.
   A failing `k0s reset` (broken/half-installed/swapped k0s binary) short-circuited
   the purge, leaving `/etc/k0s/token`. The `Restart=always` unit then restarted
   k0s with that stale token; the controller, seeing the unit `active`, never
   re-pushed a fresh one.
2. **Controller side.** The reconcile loop's `join_tokens` cache
   (`cluster_k8s.rs`) is process-scoped and was never cleared on `Down`. On the
   next `up`, roles are re-assigned before `converge_provisioning` runs, so its
   `assigned.is_empty()` clear never fires, and a cached old-CA token gets
   re-pushed.

## Fix

- **spurd:** purge the unit + token **unconditionally** in both teardown paths,
  surfacing the `k0s reset` error *after* the purge (extracted into
  `reset_and_purge`).
- **spurctld:** clear `join_tokens` in the `Down` reconcile arm (defense in depth;
  binds cache lifetime to the cluster incarnation).

No proto/config/CLI contract change. `--reset` semantics are unchanged — the purge
is simply made robust to a failing `k0s reset`.

## Testing

- Unit (CI): `reset_and_purge_removes_token_even_when_k0s_reset_fails` (spurd) —
  hermetic (`k0s reset` fails via a non-existent binary, no systemd/k0s host
  state), asserts token+unit removed and the reset error surfaced.
  `reconcile_phase_reports_no_error_for_down_and_degraded` (spurctld) extended to
  assert the `Down` tick clears a seeded token.
- Hardware A/B (2-node k0s testbed): reproduced the `kubernetes-ca` loop on the
  pre-fix build (worker k0s forced non-executable so `k0s reset` fails); on the
  fixed build the stale token is purged despite the failed reset and the worker
  rejoins to `kubectl ... Ready`, controller never restarted. Confirmed the CA
  regenerates across the cycle.
- `cargo fmt --check`, `cargo clippy --workspace --exclude spur-ffi --all-targets`,
  `cargo test -p spurd -p spurctld` all pass.


sajmera-pensando (approved):
Good fix. The root cause analysis is correct — both carriers of the stale token (the node-side `/etc/k0s/token` file and the controller's in-memory `join_tokens` cache) are identified and addressed.

**Design:** The `reset_and_purge` refactor correctly captures the ordering contract — run `k0s reset`, always purge regardless of outcome, then surface the error. Idiomatic Rust, easy to reason about. The controller-side `join_tokens.clear()` in the `Down` arm is the right place and a good defense-in-depth layer.

**Tests:** The unit test is hermetic (nonexistent binary makes `k0s reset` fail deterministically, no systemd/k0s host state) and verifies all three properties: error surfaced, token gone, unit file gone. The hardware A/B validation on a 2-node testbed is exactly the right level of confidence for this kind of fix.

**One minor nit:** `purge_unit` is best-effort and swallows errors internally, so the comment saying it "removes" the unit/token (lines 324 and 824) slightly overstates the guarantee. "Attempts to purge" or "best-effort purge" would be more accurate. Copilot flagged this too. Not a blocker.

No breaking changes to proto, config, CLI, or Raft-persisted state. Approve.

yansun1996 (approved):
Fix looks correct — the ordering change (purge unconditionally, surface the reset error after) closes both carriers of the stale-CA token cleanly, and the new test exercises the actual regression. LGTM.

PR #675, opened 2026-08-18
fix(spurctld): enforce QOS GrpWall at scheduling admission

nikhilsk:
## What this fixes

`GrpWall` on a QOS was accepted, persisted, cached and displayed — and never checked. An admin
who set a group wall-clock budget got no enforcement and no warning: `sacctmgr modify qos ... set
GrpWall=...` succeeded, `show qos` displayed the value, and nothing capped anything. An
unenforced limit is worse than a rejected one, because the operator believes a budget is in force.

The limit is now enforced at scheduling admission. Once a QOS reaches its budget, its jobs stop
being scheduled and wait with Slurm's `QOSGrpWallLimit` reason; they become eligible again as
consumption falls back below the cap.

## Approach

The enforcement decision stays a pure function. `check_qos_limits` gains a
`consumed_wall_minutes: Option<u64>` parameter, so the whole policy is unit-testable with no
database, and `None` — an unknown figure — leaves the limit unapplied.

Consumption needs no new accumulator and no persisted state: the `jobs` table already records
`start_time` and `end_time`, so a per-QOS figure is one aggregate query. Running jobs contribute the
time accrued so far, and a job that started before the window contributes only the part inside it.
The query sits behind a cache refreshed on the existing `accounting.fairshare_refresh_secs`
interval, mirroring `QosCache`, so nothing new lands in the scheduling hot loop.

Config gains one field, `accounting.grp_wall_window_days`, defaulting to 14. It is deliberately
independent of `scheduler.fairshare_halflife_days`: the half-life fades usage for priority scoring on
a curve that never reaches zero, while this window is a hard cutoff on a budget, so the two are tuned
for different reasons and a test pins the independence.

## Accounting never recorded a job's QOS

The aggregate groups by `jobs.qos`, and that column turned out to be dead: `record_job_start` took
no QOS argument, so every row ever written kept the `''` default. Grouping over an always-empty
column returns nothing, every budget would have read zero, and no job would ever have been blocked —
the same silent non-enforcement this PR is about, one layer down. It survived the unit tests because
they seed the cache or call the pure limit check directly, and it only surfaced when the whole chain
was run against a real controller and database.

So the QOS is now persisted, from all four writers: the scheduling path, reconciliation backfill,
the notifier, and the accounting gRPC service (`RecordJobStartRequest.qos`, appended as a new tag).
`record_job_start` takes a `JobStartRecord` instead of 13 positional arguments — five adjacent
`&str` parameters is what let a missing field hide in plain sight, and a transposed pair there would
compile silently and mis-attribute a job.

## Deliberate divergences from Slurm

These are documented in the admin guide, not just here:

- **Running jobs are never killed.** Slurm cancels running jobs when the limit is hit. Nothing in
  Spur's limit-checking path cancels a running job today, so parity would mean building a new and
  far more dangerous capability — one that destroys customer work if a budget or the accumulator
  is ever wrong. Admission-only enforcement matches how every other limit here behaves.
- **A rolling window replaces decayed usage.** Slurm decays usage via `PriorityDecayHalfLife` /
  `PriorityUsageResetPeriod`. Spur has no reset-period concept, and fair-share's decay is a soft
  priority weight that never reaches zero — the opposite of a budget. A trailing window is a hard
  budget with no decay curve.
- **QOS only, not associations.** `grp_wall_minutes` exists on `QosLimits` alone; `AccountLimits`
  has no group-wall field and the `associations` table has no column for it, so per-account
  budgets cannot be set today, let alone enforced. Adding them means a proto field, a migration
  and new CLI surface.
- **The budget is not applied until a consumption figure has been read.** With accounting disabled,
  or the database unreachable at controller start, the limit simply does not apply and scheduling
  continues; failing closed would turn an outage into a cluster-wide scheduling halt. After a first
  successful read, a failed refresh retains the last figure rather than reverting to unknown — the
  same as `QosCache` with its definitions — so a budget survives a brief outage with bounded
  staleness instead of letting spend run unbounded.

## Upgrade note

`PendingReason` rides in the Raft log, and there is no `#[serde(other)]` fallback, so a controller
still on an older binary cannot deserialize an entry naming the new `QosGrpWallLimit` variant.
Upgrade controllers before the new reason can be emitted. This matches every previous
reason-variant addition; the new config field is `#[serde(default)]`, so deployed configs keep
parsing.

Consumption also cannot be backfilled, since the QOS was never recorded on historical jobs. A budget
set immediately after upgrading therefore starts from zero and fills over the following
`grp_wall_window_days`. This is in the admin guide too.

## Testing

- `check_qos_limits` unit tests: blocks at the cap and over it, admits below it, admits when the
  figure is `None`, and ignores consumption entirely when no budget is set.
- The reason string is asserted byte-exact against Slurm's `QOSGrpWallLimit`, and round-trips
  through serde to cover the Raft-log contract.
- Three controller-level tests drive the real scheduling pass: a QOS at its budget is withheld and
  tagged, one below its budget is still scheduled, and an unseeded usage cache does not block.
- The blocking test was verified load-bearing — removing the enforcement turns it red.
- A `#[ignore]`d database test (`DATABASE_URL`, like its neighbours) covers the QOS write and the
  aggregate together: a finished job and a running one are attributed to their QOS, and a job older
  than the window contributes nothing. Also verified load-bearing — stubbing the `qos` bind back out
  fails it with `left: None`.
- `cargo fmt`, `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` (clean), and
  `cargo test --locked` (full suite, no failures).
- `cargo deny check` fails on RUSTSEC-2026-0258 (`h2`) on this branch, unrelated to this change
  and fixed separately.

End-to-end on a single-node cluster — controller, `spurd` and PostgreSQL 16, with
`grp_wall_window_days = 14` and a QOS capped at `GrpWall=1`:

| Job | QOS | Result |
| --- | --- | --- |
| 1 | `budget` (1 min) | ran to `COMPLETED` after 3m21s — not killed after exceeding the budget mid-run |
| 2 | `budget` | `PENDING`, reason `QOSGrpWallLimit`, at 3.36 minutes consumed |
| 3 | `free` (no budget) | scheduled and completed normally |

Job 3 is the control: an unbudgeted QOS keeps flowing while the exhausted one is held, so this is
per-QOS enforcement rather than a scheduling stall.


sajmera-pensando (changes requested):
Requesting changes on the  default coupling to  — see inline comment for details.

sajmera-pensando (changes requested):
Requesting changes on the grp_wall_window_days default coupling to fairshare_halflife_days — see inline comment for details.

crates/spur-core/src/config.rs
```diff
@@ -438,6 +438,12 @@ pub struct AccountingConfig {
     /// How often to refresh fairshare/QoS caches from the accounting database.
     #[serde(default = "default_fairshare_refresh_secs")]
     pub fairshare_refresh_secs: u32,
+    /// Trailing window over which a QOS's wall-clock consumption is measured for
+    /// `GrpWall`. Independent of `scheduler.fairshare_halflife_days`: that fades
+    /// old usage for priority scoring, while this is a hard cutoff on a budget, so
+    /// the two are tuned for different reasons and are expected to diverge.
+    #[serde(default = "default_grp_wall_window_days")]
+    pub grp_wall_window_days: u32,
```
yansun1996, line 483: **Blocking (must-fix).** This is separate from the settled question of where the default comes from — it's about the range of values accepted.

`SlurmConfig::validate` doesn't touch any `accounting.*` field, so `grp_wall_window_days = 0` is accepted. Downstream that becomes `make_interval(days => 0)`, and `COALESCE(end_time, now()) > now()` is false for every row including running ones — so the aggregate returns nothing, every QOS reports zero consumed, and every `grpwall` budget is silently unenforced. `sacctmgr show qos` still displays the cap, which is exactly the "operator believes a budget is in force" situation this PR set out to remove.

A value above `i32::MAX` reaches the same end by a different route: `.bind(window_days as i32)` wraps negative, the cutoff moves into the future, and the `WHERE` never matches.

Would you add a `validate()` arm rejecting `0` with a sane upper bound (`1..=3650`?), alongside the existing `max_batch_requeue == 0` / `max_launch_backoff_secs == 0` guards? That closes both paths in one place.

crates/spurctld/src/accounting/db.rs
```diff
@@ -505,6 +515,47 @@ pub async fn get_usage(
...
+    .fetch_all(pool)
+    .await?;
+
+    let consumed = rows
+        .iter()
+        .map(|row| {
+            let seconds = row
+                .get::<Option<i64>, _>("wall_seconds")
+                .unwrap_or(0)
+                .max(0);
```
yansun1996, line 567: **should-fix.** The `.max(0)` clamps the already-aggregated per-QOS sum, so it can't catch a single bad row — one row with `end_time < start_time` (clock skew, or a row written by a caller that got it wrong) contributes a negative term that cancels real consumption inside the same QOS, quietly under-reporting the budget.

Clamping per row inside the `SUM` instead would be robust to that:

```sql
SUM(GREATEST(EXTRACT(EPOCH FROM (
  LEAST(COALESCE(end_time, now()), now())
  - GREATEST(start_time, now() - make_interval(days => $1))
)), 0))::BIGINT as wall_seconds
```

Worth doing given that job-start rows can come from outside the controller.

crates/spurctld/src/accounting/db.rs
```diff
@@ -505,6 +515,47 @@ pub async fn get_usage(
...
+               SUM(
+                 EXTRACT(EPOCH FROM (
+                   LEAST(COALESCE(end_time, now()), now())
+                   - GREATEST(start_time, now() - make_interval(days => $1))
+                 ))
+               )::BIGINT as wall_seconds
+        FROM jobs
+        WHERE qos <> ''
+          AND start_time IS NOT NULL
+          AND COALESCE(end_time, now()) > now() - make_interval(days => $1)
```
yansun1996: **should-fix (query shape, arguably a follow-up).** `COALESCE(end_time, now()) > …` isn't index-usable, and there's no index on `jobs(qos)` or `jobs(end_time)` — so this scans the whole history table on every refresh. Paired with the 10s timeout on the caller side, a cluster with a long history reaches a point where the refresh always times out, the cache never loads, and the budget is unenforced with only a `warn!` to show for it.

An equivalent sargable form:

```sql
AND (end_time IS NULL OR end_time > now() - make_interval(days => $1))
```

A partial index on `(qos, start_time) WHERE qos <> ''` would help too, though that's fine as a separate change.

crates/spur-core/src/job.rs
```diff
@@ -287,6 +287,7 @@ pub enum PendingReason {
     QosGrpMemLimit,
     QosGrpNodeLimit,
     QosGrpGpuLimit,
+    QosGrpWallLimit,
```
yansun1996, line 300: **should-fix.** `QosGrpWallLimit` isn't in `REASON_VOCAB`, though every sibling (`QosGrpCpuLimit`, `QosGrpMemLimit`, `QosGrpNodeLimit`, `QosGrpGpuLimit`) is. That table is the canonical home for the display string and drives `reason_vocab_display_matches_slurm_25_11` / `reason_vocab_serde_roundtrips`; since it's a hand-maintained slice with no exhaustiveness guard, the omission is silent and the next person adding a reason inherits a table that's already incomplete.

The two assertions were re-implemented at `qos.rs:911` and `qos.rs:917` instead — adding `(PendingReason::QosGrpWallLimit, "QOSGrpWallLimit")` here and dropping both would keep the concept in one place.

On `qos.rs:917` specifically: serializing and deserializing within the same binary can't fail for a unit variant, so it doesn't cover the upgrade hazard its comment describes. The repo's pattern for that is the frozen wire-string fixture in `wal.rs` — `assert_eq!(json, "\"QosGrpWallLimit\"")` would at least catch a rename.

crates/spurctld/src/limits_cache.rs
```diff
@@ -105,6 +105,79 @@ impl Default for QosCache {
...
+
+        tokio::spawn(async move {
+            loop {
+                match tokio::time::timeout(
+                    Duration::from_secs(10),
+                    crate::accounting::db::consumed_wall_minutes_by_qos(&pool, window_days),
+                )
+                .await
+                {
+                    Ok(Ok(consumed)) => cache.replace(consumed),
```
yansun1996: **should-fix.** Not reopening the retain-vs-clear decision — I think keeping the last figure is right, and the reasoning in the thread above holds. This is about three ways the loop can leave enforcement off without saying so:

1. A successful query returning **zero rows** marks the cache loaded, after which every QOS reports `Some(0)` and the gate reads that as "budget untouched, admit everything." So a broken predicate or an empty window is indistinguishable from genuine zero usage — enforcement looks on, nothing is ever blocked, no warning.
2. If the first attempt fails or times out, the loop sleeps the full interval (default 300s) before retrying. `QosCache` deliberately does a separate 5s initial fetch for this case.
3. There's no log on first successful load, unlike `QosCache`'s `info!(count = …, "qos cache initialized")`, so an operator can't tell when the budget actually became enforceable.

The `info!` is the cheapest of the three and the one I'd most like to see here — it makes both other cases diagnosable. Borrowing `QosCache`'s initial-fetch shape would cover [2] as well.

crates/spurctld/src/cluster.rs
```diff
@@ -10222,6 +10241,93 @@ mod tests {
...
+            "a QOS at its GrpWall budget must not be scheduled"
+        );
+        assert_eq!(
+            cm.get_job(job_id).unwrap().pending_reason,
+            PendingReason::QosGrpWallLimit
+        );
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn grp_wall_budget_admits_while_consumption_is_below_the_cap() {
```
yansun1996, line 11239: **should-fix (test effectiveness).** This test and `grp_wall_budget_is_unenforced_while_usage_is_unavailable` below both assert only `pending.contains(&job_id)` — "the job was admitted." Delete the whole `grp_wall_minutes` block from `check_qos_limits` and both still pass, so neither can catch a regression in the thing they're named for. `grp_wall_budget_blocks_scheduling_and_tags_the_slurm_reason` above is the only load-bearing one of the three.

Asserting the reason as well would make them differential — that the job is pending *and not* tagged `QosGrpWallLimit`. For this one that also pins `>=` vs `>` at the boundary, which is currently only implied by the 599/600 pairing across two tests.

crates/spurctld/src/accounting/db.rs
```diff
@@ -1380,6 +1470,117 @@ mod job_history_tests {
         Ok(())
     }
 
+    /// GrpWall attributes consumption by the QOS recorded on each job, so a start
+    /// that fails to persist it leaves every budget reading zero and the limit
+    /// silently unenforced. Exercise the write and the aggregate together.
+    #[tokio::test]
+    #[ignore = "requires DATABASE_URL and PostgreSQL"]
+    async fn consumed_wall_minutes_attributes_a_job_to_the_qos_it_ran_under() -> anyhow::Result<()>
```
yansun1996, line 1546: **should-fix (missing branches).** Good that this covers finished, still-running, and fully-outside in one go. Two gaps, both a couple of extra rows in the same test:

- **Partial clipping** is untested — a job started 20 days ago that ended 2 days ago should contribute exactly `12 * 1440` minutes with a 14-day window. That's the only case that exercises `GREATEST(start_time, now() - make_interval(days => $1))`; break that clause today and this test still passes.
- **A `qos = ''` row** never appears, so the `WHERE qos <> ''` filter is unexercised — and that filter is the only thing keeping every un-QOS'd job in the cluster out of a single shared bucket. Worth asserting no `""` key comes back.

(Minor: the running-job assertion depends on how long the test body takes to reach the aggregate — about 60s of slack before it rounds to the next minute. Seeding at `now - 10m30s` would centre it.)

docs/admin-guide/accounting.rst
```diff
@@ -397,7 +405,56 @@ QOS keys
...
+* **The limit applies to a QOS only.** Slurm also accepts ``GrpWall`` on an
+  association; Spur does not store it there, so it cannot be set or enforced per
+  account.
+
+Enforcement needs a consumption figure, and until one has been read there is
+none: with accounting disabled, or with the database unreachable when the
+controller starts, the budget is not applied and scheduling continues. Once a
+figure has been read, a later refresh failure leaves the last one in place rather
+than discarding it, exactly as the QOS cache retains its definitions, so a budget
+keeps applying across a brief outage at up to one refresh interval of staleness.
```
yansun1996: **should-fix (docs).** The rewrite here resolved the earlier mismatch, but one clause didn't come with it: "at up to one refresh interval of staleness" isn't bounded. The loop only replaces the snapshot on success, so a three-day outage yields a three-day-stale figure — the retention is deliberate and fine, it's just the *bound* that isn't there. Something like "retained until the database returns" would match.

Related, and I think a genuine gap rather than wording: the refresh loop is only spawned inside the successful-pool-and-migration branch at startup. If the database is unreachable when the controller starts, the loop never starts at all, so the budget stays unenforced for the whole process lifetime — it does not begin applying once PostgreSQL recovers. The sentence above reads as though the gap closes on its own. Worth stating "…and stays unapplied until the controller is restarted" (or making startup connect retryable, which is the better fix but a bigger one).

docs/admin-guide/accounting.rst
```diff
@@ -397,7 +405,56 @@ QOS keys
      - Aggregate TRES cap across all jobs under this QOS.
    * - ``grpwall``
      - ``0`` (no limit)
-     - Aggregate wall-clock time across all jobs under this QOS.
+     - Aggregate wall-clock budget, in minutes, across all jobs under this QOS.
```
yansun1996: **should-fix (docs).** "in minutes" holds only for a bare integer — `grpwall` goes through the same wall-time parser as `maxwall`, which also accepts `hh:mm`, `hh:mm:ss`, `d-hh:mm` and `d-hh:mm:ss`. An admin reading this and wanting ten hours will write `grpwall=600`, but an admin who writes `grpwall=1:30` gets 90 minutes without this page telling them so.

Suggest: "Aggregate wall-clock budget across all jobs under this QOS. A bare integer is minutes; Slurm time formats (`hh:mm:ss`, `d-hh:mm:ss`) are also accepted." The same clarification would help the `maxwall` row above, though that's pre-existing.

docs/admin-guide/accounting.rst
```diff
@@ -397,7 +405,56 @@ QOS keys
...
+window contributes only the part inside it. The figure is refreshed on the same
+interval as the other accounting caches, ``fairshare_refresh_secs``.
+
+The window is deliberately independent of ``scheduler.fairshare_halflife_days``.
+The half-life fades old usage for priority scoring, a soft curve that never
+reaches zero; this window is a hard cutoff on a budget. Changing one must not
+move the other, so a cluster can pair, say, a seven-day half-life for responsive
+priority with a thirty-day window for monthly budget enforcement.
+
+Three deliberate differences from Slurm:
```
yansun1996, line 540: **should-fix (docs).** This list is the right place for a few more behaviours an operator would otherwise discover the hard way:

- **A requeue discards the earlier run.** The job-start upsert resets `start_time` and clears `end_time` on conflict, so a requeued job's first run drops out of the budget entirely.
- **Admission can overshoot between refreshes.** The figure is re-read on the refresh interval while the scheduler admits every pass, so with consumption sitting just under the cap an unbounded batch can start before the next read. Symmetrically, jobs held by the limit only become eligible again at the next refresh, not the moment consumption ages out — worth a clause on line 417 too.
- **A job whose end is never recorded accrues for the whole window.** If a row is left open (state wiped before reconciliation could repair it), it contributes the full window from then on and pins that QOS below any cap, with nothing surfacing the cause.
- **Spend is keyed on the QOS name and can't be reset.** Deleting and recreating a QOS with the same name inherits up to `grp_wall_window_days` of the old one's consumption, and there's no admin override — the only escapes are renaming or waiting out the window. For a control whose failure mode is "an entire QOS stops scheduling," that's worth saying out loud.

yansun1996 (changes requested):
Nice piece of work — the framing is right (an unenforced limit is worse than a rejected one), the pure-function split for `check_qos_limits` makes the policy properly testable, and catching that `jobs.qos` was dead is the kind of find that only comes from running the whole chain. The deliberate-divergences section and the checks that the new tests are load-bearing are appreciated.

I've tagged every comment with a severity. **There is exactly one I'd gate on**, and it's a small fix — the rest are suggestions.

### Blocking

**[1] `accounting.grp_wall_window_days` has no validation** — `0` collapses the window so every QOS reads zero consumed, and every budget is silently unenforced while `show qos` still shows the cap. That's the failure this PR exists to eliminate, re-created one layer up and reachable by a plausible typo. (Values above `i32::MAX` wrap negative through the `as i32` bind and land in the same place.) One arm in `validate()`, next to the existing `*_secs == 0` guards. Comment on `config.rs`.

### Should-fix, and each is a few lines — worth doing here

- `GREATEST(0, …)` inside the `SUM` rather than clamping the group total.
- One row in `REASON_VOCAB`, and drop the two re-implemented assertions in `qos.rs`.
- An `info!` on the first successful usage load, as `QosCache` does — without it there's no way to tell from the logs that enforcement ever came up.

### Should-fix, your call on scope

Test strengthening [2], docs accuracy [3], and one query-shape note. Details inline.

### Worth a line in the PR body rather than a code change

- `PendingReason` has no `#[serde(other)]`, and it reaches persisted state through both `WalOperation::JobStateChange` and `ClusterSnapshot.jobs`, with strict snapshot recovery. Once a job carries the new variant and a snapshot is taken, an older controller binary can't start from it — so this constrains rollback and mixed-version HA, not just replay of old entries. Earlier reason-variant additions shipped without a `!`, so I'm not asking for one, but the constraint is worth stating for release notes.
- `record_job_start` on the accounting service does no per-RPC authorization, and `qos` now flows from the wire into a scheduling-enforcement input. A caller can attribute backdated rows to another QOS and exhaust it, or rewrite a live job's `qos` to `""` (which the `qos <> ''` filter then excludes) to dodge its own budget. The unauthorized-write surface is pre-existing and not yours to fix here, but this change is what makes it load-bearing — worth a follow-up issue.

### Nits, no need to thread

`limits_cache.rs:108-115` is an 8-line doc block and `db.rs:171-175` a 5-line one whose tail narrates the refactor rationale — that reads more like commit-message material. `cluster.rs:2986` takes a read lock per candidate inside the retain loop, while the sibling `qos_by_job` lookup two lines above is hoisted once.

The two earlier threads on `job_id as i32` and the default's provenance both look settled to me — the casts are pre-existing and only moved by this PR, and the default is a flat `14` with a test guarding against re-coupling.

crates/spur-core/src/config.rs
nikhilsk: Fixed in 078ec9b. `validate()` now requires `1..=3650`, in one arm next to the `max_batch_requeue` and `max_launch_backoff_secs` guards, so both routes you described are closed in the same place.

Verified against a live controller rather than just the unit test:

```
Error: invalid value for accounting.grp_wall_window_days: 0 (must be between 1 and 3650)
```

Tests cover `0`, `3651` and `u32::MAX` rejected with the field named in the error, and `1` and `3650` accepted. The range is on the config reference row too, so a refused startup is explainable from the docs.

crates/spurctld/src/accounting/db.rs
nikhilsk: Done in d09b2ee, in the form you wrote.

Worth recording that this was not hypothetical. I put an end-before-start row alongside three good rows in PostgreSQL: the aggregate-level clamp reported 1530 minutes where the per-row clamp reports 1590, so one skewed row was cancelling an hour of real consumption inside the same QOS, exactly as you described.

The function's doc comment now says why the clamp sits inside the `SUM`, since otherwise the next reader is likely to simplify it back out.

crates/spurctld/src/accounting/db.rs
nikhilsk: Taken in d09b2ee. The predicate is now:

```sql
AND (end_time IS NULL OR end_time > now() - make_interval(days => $1))
```

Equivalent, since a running job has a NULL end and always qualified under the old form, and index-usable.

I have left the partial index out. Agreed it belongs in its own change, where it can be considered alongside whatever else the accounting tables need rather than bolted onto this one.

crates/spur-core/src/job.rs
nikhilsk: Fixed in d2638b3. The row is in `REASON_VOCAB` and both re-implemented assertions are gone, so the display string lives in one place and the existing vocabulary tests cover it.

You are right about `qos.rs:917`: it could not fail. Replaced with a frozen wire-string fixture beside the other WAL fixtures, asserting the serialized form is `"QosGrpWallLimit"`, so a rename has to be a deliberate decision.

On the wider point from your summary, agreed, and the PR body now states it. `ClusterSnapshot.jobs` is recovered strictly, so once a job carries the variant and a snapshot is taken, an older binary cannot start from that snapshot at all. That constrains rollback and mixed-version HA rather than only replay of old entries, and it belongs in the release notes.

crates/spurctld/src/limits_cache.rs
nikhilsk: dc38a67 covers [3] and [2]. The first successful load now logs, and the loop takes a short first attempt before settling into the interval, with both paths sharing one refresh function. From the live cluster:

```
INFO spurctld::limits_cache: grpwall usage cache initialized, budgets now enforced qos_count=5 window_days=14
```

[1] I have deliberately left as it is. An empty result is a legitimate state on a fresh cluster, so telling it apart from a broken predicate needs a signal the query itself cannot give, and the obvious heuristic (warn when the jobs table is non-empty but the aggregate is empty) is a guess I would rather not encode in the refresh loop. What the log line does buy is that the state is observable, and the predicate itself is now covered from the other side: the database test has a no-QOS row and a window-clipped row, both mutation-checked, so a broken filter or clamp fails a test rather than silently returning nothing.

crates/spurctld/src/cluster.rs
nikhilsk: Fixed in 70ee2d3. Both tests now assert the reason is not `QosGrpWallLimit` as well as that the job was admitted, so they fail on spurious tagging rather than only on enforcement being absent altogether.

That also pins the boundary where it is exercised: the 599-of-600 case asserts admitted and untagged, and the sibling test blocks at 600, so `>=` is now covered by an assertion rather than implied by the pairing.

crates/spurctld/src/accounting/db.rs
nikhilsk: Both added in d09b2ee, and both mutation-checked rather than assumed:

- Partial clipping: started 20 days ago, ended 2 days ago, asserted at `12 * 1440`. Dropping the `GREATEST(start_time, ...)` clause fails it with `left: Some(25920), right: Some(17280)`.
- A `qos = ''` row, asserting no `""` key comes back. Dropping the filter fails it with `left: Some(120), right: None`.

Running fixture reseeded at `now - 10m30s`, so the truncating division cannot tip either way while the body runs.

One item from your summary that stayed out of the code, since you flagged it as out of scope: the PR body now carries the `record_job_start` authorization gap as a follow-up. Agreed the unauthorized-write surface is pre-existing, and agreed this change is what makes it load-bearing, since `qos` now reaches a scheduling decision and the `qos <> ''` filter turns blanking it into an evasion.

docs/admin-guide/accounting.rst
nikhilsk: Both fixed in 923439d.

The unbounded claim is gone. The text now says the last figure is retained until the database returns, however old that figure is by then, and that consumption accrued during the outage is invisible until it does.

You were right that the second point is a gap rather than wording. The refresh loop only spawns inside the successful-pool-and-migration branch, so a database that is unreachable at controller start means the budget is never applied for that process's lifetime and does not begin applying when PostgreSQL recovers. The page now says exactly that, and tells the operator to restart the controller once the database is reachable.

I agree a retryable startup connect is the better fix. It applies to all four accounting caches rather than this one, so I have left it as a separate change rather than widening this PR.

docs/admin-guide/accounting.rst
nikhilsk: Fixed in 923439d, close to your wording. The accepted formats are spelled out, with a note that seconds are truncated (the parser reads them and drops them), and I kept your `600` against `10:00` example since that is the mistake an admin actually makes.

Left the `maxwall` row alone as pre-existing, though I agree it has the same gap and would be a one-line follow-up.

docs/admin-guide/accounting.rst
nikhilsk: All four are in 923439d, as their own list ("behaviours worth knowing before relying on a budget") rather than folded into the Slurm-divergence list above, since they are properties of this implementation rather than deliberate departures from Slurm.

I checked the requeue claim before writing it: the job-start upsert sets `start_time = EXCLUDED.start_time` and `end_time = NULL` on conflict, so a requeued job's first run does leave the budget entirely.

Delayed release is on the line above as you suggested. The never-closed row and the QOS-name-keyed spend each name the symptom an operator would actually see, since neither surfaces its own cause: `sacct` showing a job with no end time, and a recreated QOS inheriting the old one's consumption with no override and no way to reset it short of a new name.

sajmera-pensando (dismissed):
LGTM

PR #676, opened 2026-08-18
fix(spurd): start container jobs from the image's config.Env

### LINKED ISSUE #649: [Issue]: Container jobs launch with the host environment instead of the image's config.Env — torchrun from the rocm/pytorch image is not on PATH
### Problem Description

When a container image is deployed with `spur image import` + `sbatch --container-image`, the image's own environment (its OCI `config.Env`: `PATH`, `LD_LIBRARY_PATH`, `DEBIAN_FRONTEND`, …) is thrown away and **replaced** by the submitting host's environment.

Verified against `spur 0.8.0`, commit `9697ed6` ("fix(spur-cli): accept slurm-format reservation durations (#626)", 2026-08-14), pristine upstream `main` with a clean tree.

### Root cause (two stages)

**Stage 1 — import drops `config.Env`.**
- `crates/spur-cli/src/image.rs` → `extract_docker_save_tar()` extracts only `manifest[0].Layers` from the `docker save` tar and ignores `manifest[0].Config` (the config blob that holds `config.Env`).
- `crates/spur-net/src/oci.rs` → the registry puller's `Manifest` struct only deserializes `layers`; it never fetches the `config` blob.
- Result: the squashfs produced by both import paths contains no record of the image's environment.

**Stage 2 — launch never applies image env.**
- `crates/spurd/src/executor.rs` → `launch_container_job()` builds the child environment as `host/job env (env_snapshot)` + `--container-env` + hook `environ.d`. The image env is never read.
- The child then calls `execve(path, argv, envp)`, which **replaces the entire environment** with that set — so whatever the image baked in is gone.
- The base (`env_snapshot`) comes from `sbatch --export`, which defaults to `ALL` (`crates/spur-cli/src/sbatch.rs:206`, consumed at `sbatch.rs:909`), i.e. the full submitting-host environment.

**Scope — `Env` is not the only casualty.** The same discarded config blob also carries `Entrypoint`, `Cmd`, and `WorkingDir`. Grepping `Entrypoint` and `WorkingDir` across `crates/spur-cli/src` and `crates/spur-net/src` returns no hits, so none of them is recorded at import either.

### Symptoms

Example image: `rocm/pytorch:latest`, which installs `torchrun`, `python`, and `pip` under `/opt/venv/bin` and ROCm libraries under `/opt/rocm/lib` — all reachable *only* via the image's own `config.Env`. That image defines exactly three variables:

```console
$ docker inspect rocm/pytorch:latest --format '{{range .Config.Env}}{{println .}}{{end}}'
PATH=/opt/venv/bin:/opt/rocm/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
DEBIAN_FRONTEND=noninteractive
LD_LIBRARY_PATH=/opt/rocm/lib
```

| Field | `docker run rocm/pytorch:latest` (correct) | same image via spur container job (buggy) |
|---|---|---|
| `PATH` | `/opt/venv/bin:/opt/rocm/bin:/usr/local/sbin:…` (image's own) | submitting host's `PATH` — `/opt/venv/bin` absent |
| `LD_LIBRARY_PATH` | `/opt/rocm/lib` | host value / empty |
| image var (e.g. `DEBIAN_FRONTEND`) | `noninteractive` | unset |
| `command -v torchrun` | `/opt/venv/bin/torchrun` | **not found** |
| `torchrun --version` inside the job | runs normally | **`torchrun: command not found`** |

Concretely: submitting a job with `--container-image rocm/pytorch:latest` whose script calls `torchrun` fails with `command not found`, even though `/opt/venv/bin/torchrun` exists inside that image and `docker run` on the very same image executes it fine. The container's `PATH` is the submitting host's, so `/opt/venv/bin`, the directory the image installed its executables into, is never searched.

### Why it's a bug

`execve` does not inherit or merge the image's environment; the launcher is responsible for seeding it, and spur seeds it with the host's environment instead of the image's. This silently diverges from how every other container runtime (docker/podman/enroot/pyxis) treats an image: the image's `Env` is the base.

Three further reasons this is a gap rather than a deliberate design choice:

1. **spur's own documentation already promises behavior that the discarded config makes impossible.** `docs/user-guide/running-containers.rst` documents a flag as:

   ```rst
   * - ``--container-entrypoint``
     - Use the image's entrypoint (``sbatch``).
   ```

   Since import never parses `Entrypoint` out of the config blob, spur holds no record of the image's entrypoint, so there is nothing for that flag to use.

2. **A merged fix has already worked around a symptom of this same root cause.** #149 ("Container jobs submitted by non-root users fail with glibc version mismatch") was fixed by #151 ("fix(spurd): non-root container uses container's dynamic loader", merged 2026-04-30), which hand-wires the container's loader and library paths precisely because the image's own environment is not available at launch.

3. **The available workarounds do not cover the case.** `--container-env KEY=VAL` requires the submitter to already know each image's internal paths and re-type them in every job script, per image; the admin-side hook environment (`executor.rs:1488`) is site-wide and root-only. Neither lets a user simply run an image the way `docker run` does.

### Why it matters

This breaks the primary ROCm use case. `rocm/pytorch` ships `torchrun`, `python`, and ROCm libs on the image's own `PATH`/`LD_LIBRARY_PATH` (e.g. `/opt/venv/bin`, `/opt/rocm/lib`). Under spur those directories vanish, so:
- `torchrun`, `python`, `pip` become unresolvable (`command not found`),
- `LD_LIBRARY_PATH` for ROCm/RCCL is lost, breaking multi-GPU/multi-node launches,
- users see failures that look like a broken image even though `docker run` on the same image works.

### Expected

The image's `config.Env` should form the **base** environment of the container. Host/job vars, `--container-env`, and hooks may add to or override it, and `PATH`/`LD_LIBRARY_PATH` should be unioned (image first) rather than clobbered, so executables installed in the image stay resolvable while host additions still work.

### Operating System

Ubuntu 24.04 LTS (kernel 6.8.0-38-generic). Bug is OS/hardware-independent.

### CPU

N/A (hardware-independent)

### GPU

N/A (hardware-independent; observed with the rocm/pytorch image)

### ROCm Version

N/A (hardware-independent)

### ROCm Component

_No response_

### Steps to Reproduce

> **Note on the test image:** the repro below uses a small purpose-built busybox image that mimics the real-world layout (an executable installed under `/opt/venv/bin`, with `PATH` set only in the image's `config.Env`) rather than pulling `rocm/pytorch:latest`, which is tens of GB and slow/space-hungry to pull and import. spur never inspects image contents — both images take the identical import and launch code path — so the failure is the same. `rocm/pytorch:latest` was separately confirmed to place `torchrun`, `python`, and `pip` at `/opt/venv/bin` with `PATH`/`LD_LIBRARY_PATH` defined only in its image config (see the `docker inspect` output under *Symptoms*).

Reproduced end-to-end against a single-node spur cluster (controller + agent) on one Linux host — no Docker runtime used at launch, only spur.

All output below is verbatim from that run on commit `9697ed6`.

**Prerequisites:** a built spur checkout (`cargo build -p spurctld -p spurd -p spur-cli`), plus `docker`, `squashfs-tools` (`mksquashfs`/`unsquashfs`), cgroup v2, and root (for namespaces/pivot_root).

<details>
<summary>0. Bring up a single-node cluster (controller + agent)</summary>

```bash
B=target/debug; WORK=/tmp/spur_e2e; STATE=$WORK/state
mkdir -p "$WORK" "$STATE"

cat > "$WORK/spur.conf" <<EOF
cluster_name = "e2e"
[controller]
listen_addr = "127.0.0.1:6817"
state_dir = "$STATE"
[scheduler]
plugin = "backfill"
interval_secs = 1
[auth]
plugin = "none"
[[partitions]]
name = "gpu"
default = true
state = "UP"
nodes = "node1"
max_time = "1-00:00:00"
default_time = "1:00:00"
[[nodes]]
names = "node1"
cpus = 8
memory_mb = 16384
[network]
wg_enabled = false
agent_port = 6818
EOF

nohup $B/spurctld -f "$WORK/spur.conf" -D --state-dir "$STATE" > "$WORK/spurctld.log" 2>&1 &
sleep 3
sudo -b bash -c "cd $PWD && ./$B/spurd -f $WORK/spur.conf -D --hostname node1 --controller http://127.0.0.1:6817 > $WORK/spurd.log 2>&1"
sleep 4
$B/spur nodes --controller http://127.0.0.1:6817   # expect: node1 idle
```
</details>

1. Build an image whose executable lives **only** on the image's own `PATH`:

```dockerfile
FROM busybox:latest
RUN mkdir -p /opt/venv/bin \
 && printf '#!/bin/sh\necho TORCHRUN_RAN_FROM_IMAGE\n' > /opt/venv/bin/torchrun \
 && chmod +x /opt/venv/bin/torchrun
ENV PATH=/opt/venv/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
ENV IMG_MARKER=hello_from_image_env
```
```bash
docker build -t spur-envtest:latest .
```

2. Baseline — `docker run` sees the image env (correct):
```bash
$ docker run --rm spur-envtest:latest /bin/sh -c 'echo PATH=$PATH; echo IMG_MARKER=$IMG_MARKER; command -v torchrun && torchrun'
PATH=/opt/venv/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
IMG_MARKER=hello_from_image_env
/opt/venv/bin/torchrun
TORCHRUN_RAN_FROM_IMAGE
```
For comparison, the submitting shell's own `PATH` on the test host was:
```
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
```

3. Import + run the same image through spur:
```bash
$B/spur image import dockerd://spur-envtest:latest

cat > "$WORK/job.sh" <<'EOF'
#!/bin/sh
echo "PATH=$PATH"
echo "IMG_MARKER=${IMG_MARKER:-<EMPTY>}"
command -v torchrun && torchrun || echo "TORCHRUN=NOT_FOUND"
EOF

$B/spur submit --partition gpu --container-image spur-envtest:latest \
  --container-workdir / --chdir "$WORK" -o "$WORK/job.out" \
  --controller http://127.0.0.1:6817 "$WORK/job.sh"

sleep 6
cat "$WORK/job.out"
```

4. Actual output from the spur-launched container (buggy) — verbatim `job.out`:
```
PATH=/home/amd/spur_issue/spur_fresh/target/debug:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
IMG_MARKER=<EMPTY>
TORCHRUN=NOT_FOUND
```

The container's `PATH` is the host-side `PATH` (with spur's own bin directory prepended). Note that it contains `/home/amd/spur_issue/spur_fresh/target/debug` — a host-only build directory that does not exist inside the container — while the image's `/opt/venv/bin` is absent entirely. `IMG_MARKER`, set in the image's `config.Env`, arrives empty. The job itself ran to completion (`exit_code=0`); nothing failed except the environment.

**Expected (correct) behavior:** `PATH` should contain `/opt/venv/bin` (the image's own dir), `IMG_MARKER=hello_from_image_env`, and `torchrun` should resolve and run — matching the `docker run` baseline in step 2.

> Note: a **separate**, unrelated defect exists on this same commit — `spurd` double-closes the container launch sync-pipe file descriptors, which aborts debug builds with `fatal runtime error: IO Safety violation: owned file descriptor already closed`. It triggered on other runs of this build but did **not** trigger in the run transcribed above, so it is not a prerequisite for (nor a cause of) the environment problem shown here. Tracked separately.

<details>

```
N/A (hardware-independent — no GPU involvement)
```

</details>

### Affected code

- `crates/spur-cli/src/image.rs` — `extract_docker_save_tar()` reads only `manifest[0].Layers`; `Config` is ignored.
- `crates/spur-net/src/oci.rs` — `Manifest` deserializes only `layers`; config blob never fetched.
- `crates/spurd/src/executor.rs` — `launch_container_job()` builds `final_env = host env + container_env + hook_env` (no image env) before `execve`.

### Related

- #149 / #151 — merged fix that hand-wires the container's dynamic loader because the image's environment is unavailable at launch; same root cause treated at the symptom level.

### Notes

- A fix + regression tests are prepared and verified end-to-end (image env restored, `torchrun` resolves, host vars still pass through). Can open a PR referencing this issue.

SumonAMD:
Closes #649.

## What

Container jobs ran with the submitting host's environment instead of the
image's. `spur image import` discarded the OCI config blob on both the
`docker save` and registry paths, and `launch_container_job` built the child
environment from the job environment alone, so `execve` replaced everything the
image had set. Programs installed on the image's own `PATH` — `torchrun` under
`/opt/venv/bin` in `rocm/pytorch` — were unresolvable and `LD_LIBRARY_PATH` for
ROCm/RCCL was lost, while `docker run` on the same image works.

## Approach

Import records the image config into the rootfs at `etc/spur/image-config.json`
before mksquashfs, from the `docker save` archive and from the registry blob.
Launch reads it back and uses `config.Env` as the base environment.

Precedence, low to high: image `config.Env` → job environment (`--export`,
default `ALL`) → `--container-env` → `environ.d` and hook variables. `PATH` and
`LD_LIBRARY_PATH` are unioned with the image's entries first instead of being
replaced, so tools shipped in the image resolve while host additions still
apply; an explicit `--container-env PATH=...` still replaces outright.

The whole config blob is recorded, not just `Env`, so `Entrypoint`, `Cmd`, and
`WorkingDir` are available for a follow-up that makes `--container-entrypoint`
do what the docs already describe. Nothing reads them yet.

## Trade-offs

- Images imported by an earlier version record nothing and keep today's
  behaviour until re-imported. No flag day.
- After re-import, `PATH` ordering changes: a tool present both in the image's
  venv and in `/usr/bin` now resolves to the image's, matching `docker run`.
  That is the point of the fix, but it is a visible change.
- With `--export NONE`, or an `--export` list omitting `HOME`, an image that
  sets `HOME` in `config.Env` now supplies it; under the default `--export ALL`
  the submitter's value wins. This matches `docker run`, but I am happy to
  filter `HOME`/`USER` out of the image environment if reviewers prefer.
- The rootfs is content spur does not control, since `--container-image` accepts
  an arbitrary squashfs path. The recorded config is therefore read only through
  non-symlinked path components, only when it is a regular file, and only up to
  1 MiB; without those guards a crafted image could redirect the write out of
  the rootfs at import or stall a launch on a FIFO.
- #649 cites #149/#151 as a live workaround for this root cause. That code was
  superseded by #154, which moved both root and rootless launches onto
  `container_init` + `pivot_root`; the root cause itself still stands.

## Testing

- `cargo test` (2983 tests, 34 suites), `cargo fmt --check`, `clippy
  --all-targets` (0 warnings), SPDX headers, and the docs image (`sphinx -W`,
  pyspelling, markdownlint) all pass.
- Five unit tests in `spur-net::oci`: union and override, no recorded config,
  oversized config, symlinked path component, and a FIFO at the config path
  (guarded by a channel timeout so a regression fails rather than hangs).
- End to end on a single-node cluster: image imported by the previous binary
  (unchanged), by this one (`torchrun` runs, `PATH` image-first, image variables
  present, host variables still pass through), `--container-env` override,
  non-container job, registry import path, named containers, `--export NONE`,
  and a hostile squashfs with a FIFO at the sidecar path (job runs, node stays
  responsive).

### DIFF
--- a/crates/spur-cli/src/image.rs
+++ b/crates/spur-cli/src/image.rs
@@ -246,6 +246,21 @@ fn extract_docker_save_tar(tar_data: &[u8], rootfs: &str) -> Result<()> {
                 .with_context(|| format!("failed to extract layer {}", layer_file))?;
         }
 
+        // Record the image config so container jobs can start from its config.Env.
+        // Best effort: an archive without a readable config still imports.
+        if let Some(config_file) = manifest
+            .first()
+            .and_then(|entry| entry.get("Config"))
+            .and_then(|config| config.as_str())
+        {
+            let recorded = resolve_docker_save_file(&archive_root, config_file)
+                .and_then(|path| std::fs::read(path).with_context(|| format!("read {config_file}")))
+                .and_then(|config_json| spur_net::oci::record_image_config(dest, &config_json));
+            if let Err(e) = recorded {
+                eprintln!("warning: image environment not captured from {config_file}: {e:#}");
+            }
+        }
+
         Ok(())
     })();
 
--- a/crates/spur-net/src/oci.rs
+++ b/crates/spur-net/src/oci.rs
@@ -15,6 +15,7 @@
 //! 5. Extract layers in order to build rootfs
 //! 6. Pack rootfs into squashfs via mksquashfs
 
+use std::collections::HashMap;
 use std::ffi::OsStr;
 use std::io::Read;
 use std::path::{Path, PathBuf};
@@ -44,7 +45,15 @@ struct TokenResponse {
 struct Manifest {
     #[serde(default)]
     layers: Vec<LayerDescriptor>,
+    /// Image config blob, which carries `config.Env`.
+    #[serde(default)]
+    config: Option<ConfigDescriptor>,
     // v1 compat: some registries return "fsLayers" instead
+}
+
+#[derive(Deserialize)]
+struct ConfigDescriptor {
+    digest: String,
 }
 
 #[derive(Deserialize)]
@@ -354,6 +363,26 @@ async fn pull_and_extract(
         let media_type = &manifest.layers[i].media_type;
         extract_layer(data, Some(media_type), rootfs_dir)
             .with_context(|| format!("failed to extract layer {}", i + 1))?;
+    }
+
+    // Best effort: a registry that will not serve the config blob still yields
+    // a usable image, and jobs using it fall back to the host environment.
+    match manifest.config.as_ref() {
+        Some(config) => {
+            if let Err(e) = fetch_and_record_config(
+                &client,
+                image_ref,
+                &registry_url,
+                &config.digest,
+                token.as_deref(),
+                rootfs_dir,
+            )
+            .await
+            {
+                warn!(error = %e, digest = %config.digest, "image environment not captured");
+            }
+        }
+        None => debug!("manifest has no config descriptor; image environment not captured"),
     }
 
     Ok(())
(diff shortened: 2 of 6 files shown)


sajmera-pensando (changes requested):
The fix is correct and the unit tests are solid, but there are no automated e2e tests added to `tests/native_host/e2e/` or `tests/k8s/e2e/`. The manual cluster run described in the PR description is valuable, but if someone later breaks `executor.rs` or `oci.rs`, `cargo test` won't catch it.

Please add two e2e tests before merging:

**[1] Core fix** — import an image that has a known binary on its own `PATH` (not present on the host), submit a job that invokes that binary, assert it succeeds and that the container's `PATH` is image-first with host entries appended. This is the `torchrun` scenario and the one a unit test cannot catch because it requires a real import → real launch → real `execve`.

**[2] Backward compatibility** — import an image using the old binary (or manually remove `etc/spur/image-config.json` from the squashfs after import), submit a job, assert it runs normally using the job environment. Confirms that users who haven't re-imported their images don't get a regression.

Everything else (registry vs docker save, `--container-env` override, `--export NONE`, non-container jobs) can be follow-ups — these two cover the critical path and the no-regression case.

sajmera-pensando (approved):
Both requirements addressed. The three e2e tests in `tests/native_host/e2e/test_container.py` cover the core fix (image-only binary resolves, PATH is image-first), backward compatibility (no recorded config keeps job env unchanged), and the registry pull path. Implementation matches what was described. LGTM.

PR #679, opened 2026-08-18
fix(spur-core): cap hostlist expansion size to prevent controller OOM

### LINKED ISSUE #677: [Issue]: hostlist::expand has no size cap — a short nodelist can OOM-kill spurctld
### Problem Description

`hostlist::expand()` materializes one `String` per element of a bracket range whose bounds are parsed straight from the input as `u64`, with no limit on the resulting element count. A 28-byte pattern — `node[0-18446744073709551615]` — asks for 1.8 × 10^19 allocations. The controller expands it on a **read-only RPC that is unauthenticated under the default `auth.mode`**, so a single `sinfo -n` query drives `spurctld` into the kernel OOM killer and takes the whole control plane down.

This is the sibling of #645 in a different function. The `array.rs` fix (#646) added an early element-count bound to `parse_array_spec`; `hostlist.rs` has the same defect and was not covered.

### Operating System

Ubuntu 24.04 LTS (kernel 6.8.0-38-generic) — defect is OS-independent

### CPU

N/A — defect is hardware-independent (reproduced on a CPU-only loopback controller)

### GPU

N/A — defect is hardware-independent; no GPU is involved (control-plane RPC / scheduler code path)

### ROCm Version

N/A — defect is hardware-independent (pure spurctld control-plane bug, not ROCm-runtime dependent)

### ROCm Component

_No response_

### Steps to Reproduce

Route A shows the unbounded growth at library level in seconds — no root, no cluster. Route B is the end-to-end controller kill. All output is verbatim.

**Route A — library-level, shows no cap is ever enforced.**

1. Save as `crates/spur-core/examples/probe_hostlist.rs`:

```rust
// Growth of spur_core::hostlist::expand() with no cap on the element count.
use std::time::Instant;

/// Resident set size in MB, from /proc/self/statm (field 2 = resident pages).
fn rss_mb() -> u64 {
    let s = std::fs::read_to_string("/proc/self/statm").unwrap_or_default();
    let pages: u64 = s.split_whitespace().nth(1).and_then(|p| p.parse().ok()).unwrap_or(0);
    pages * 4096 / (1024 * 1024)
}

fn main() {
    for exp in 4..=7 {
        let pattern = format!("node[0-{}]", 10u64.pow(exp) - 1);
        let before = rss_mb();
        let t = Instant::now();
        let hosts = spur_core::hostlist::expand(&pattern).expect("expand");
        println!(
            "expand(\"{}\") -> {} hosts | RSS +{} MB | {:?}",
            pattern, hosts.len(), rss_mb().saturating_sub(before), t.elapsed()
        );
    }
    // The only validation in the range loop is `start > end`:
    println!("expand(\"node[5-1]\") -> {:?}",
        spur_core::hostlist::expand("node[5-1]").err().map(|e| e.to_string()));
}
```

2. Run it (the `ulimit` is a safety net so an accidental typo cannot disturb the host):

```bash
cargo build --release -q -p spur-core --example probe_hostlist
( ulimit -v 6000000; ./target/release/examples/probe_hostlist )
```

3. Observed:

```console
expand("node[0-9999]") -> 10000 hosts | RSS +0 MB | 370.454µs
expand("node[0-99999]") -> 100000 hosts | RSS +4 MB | 4.316946ms
expand("node[0-999999]") -> 1000000 hosts | RSS +48 MB | 43.753984ms
expand("node[0-9999999]") -> 10000000 hosts | RSS +502 MB | 433.708156ms
expand("node[5-1]") -> Some("invalid range: 5 > 1")
```

Linear in the requested count at roughly 52 bytes per host, with no limit at any magnitude — 10 million hosts in 0.43 s. Extrapolating, `node[0-18446744073709551615]` requests ~9.6 × 10^11 GB. The reversed-range line is the control: size is the only unvalidated property.

**Route B — end-to-end controller kill via the RPC.** Only the controller is needed; `get_nodes` does not involve `spurd`.

1. Bring up a controller under a memory cap. Use a cgroup (`MemoryMax`), **not** `ulimit -v` — tokio reserves a large virtual address space at startup, so a virtual-size cap kills the controller before it finishes booting.

```bash
B=target/release; WORK=/tmp/spur_hltest; STATE=$WORK/state
mkdir -p "$STATE"

cat > "$WORK/spur.conf" <<EOF
cluster_name = "hltest"

[controller]
listen_addr = "127.0.0.1:6817"
state_dir = "$STATE"

[scheduler]
plugin = "backfill"
interval_secs = 1

[auth]
plugin = "none"

[[partitions]]
name = "gpu"
default = true
state = "UP"
nodes = "node1"

[[nodes]]
names = "node1"
cpus = 8
memory_mb = 16384

[network]
wg_enabled = false
agent_port = 6818
EOF

sudo systemd-run --scope --unit=spur_hltest -p MemoryMax=1500M \
  "$PWD/$B/spurctld" -f "$WORK/spur.conf" -D --state-dir "$STATE" \
  > "$WORK/spurctld.log" 2>&1 &
sleep 3
$B/spur sinfo --controller http://127.0.0.1:6817     # sanity: node1 listed
```

2. Fire one read-only query with no credentials:

```bash
$B/spur sinfo -n 'node[0-18446744073709551615]' --controller http://127.0.0.1:6817
```

3. Observed — the kernel OOM killer takes the controller:

```
tokio-rt-worker invoked oom-killer: gfp_mask=0xcc0(GFP_KERNEL)
Memory cgroup out of memory: Killed process 3851418 (spurctld) anon-rss:1515520kB
spur_hltest.scope: A process of this unit has been killed by the OOM killer.
```

The controller is gone and every subsequent request fails. Its own log confirms the request needed no credentials:

```
unauthenticated request accepted (auth.mode = permissive) path=/slurm.SlurmController/GetNodes
```

4. Teardown:

```bash
sudo systemctl stop spur_hltest.scope 2>/dev/null; rm -rf /tmp/spur_hltest
```

**Variant — the persistent scheduler path.** Submit `sbatch -w 'node[0-18446744073709551615]'` instead. The job is accepted and persisted to the Raft log, and the controller then dies while scheduling it — and again on restart, because the pending spec is re-expanded every cycle.

<details>

```
Paste output here
```

</details>

## Root cause

`expand_single()` parses both range endpoints as `u64` and loops over the full inclusive range, pushing a formatted `String` per iteration. The only validation is `start > end`:

```rust
// crates/spur-core/src/hostlist.rs:306-324
let start: u64 = start_str
    .parse()
    .map_err(|_| HostlistError::InvalidRange(range_part.into()))?;
let end: u64 = end_str
    .parse()
    .map_err(|_| HostlistError::InvalidRange(range_part.into()))?;

if start > end {
    return Err(HostlistError::InvalidRange(format!("{} > {}", start, end)));
}

for i in start..=end {
    let name = format!("{}{:0>width$}{}", prefix, i, suffix, width = width);
    if suffix.contains('[') {
        expand_single(&name, results)?;
    } else {
        results.push(name);
    }
}
```

There is no cap before the loop and no incremental check inside it. `results` grows until the allocator or the kernel stops it. `HostlistError` (`hostlist.rs:8-13`) has only `InvalidPattern` and `InvalidRange`, no size variant, in contrast to `array.rs`, which has `MAX_ARRAY_SIZE` and `ArrayError::TooLarge`.

`count()` is not an escape hatch — it is implemented in terms of `expand()`:

```rust
// crates/spur-core/src/hostlist.rs:340-344
/// Count the number of hosts in a hostlist pattern without expanding.
pub fn count(pattern: &str) -> Result<usize, HostlistError> {
    // For now, just expand and count. Can optimize later.
    Ok(expand(pattern)?.len())
}
```

The doc comment says "without expanding"; the body expands.

## Reachability — three paths, the first unauthenticated by default

**1. `get_nodes` RPC — read-only, unauthenticated under the default `auth.mode`.** `spurctld` expands the caller-supplied `nodelist` filter directly, with no admin gate and no identity check:

```rust
// crates/spurctld/src/server.rs:1080-1086  (in `async fn get_nodes`, :1060)
let nodelist = req.nodelist.trim();
let allowed_names: Option<HashSet<String>> = (!nodelist.is_empty()).then(|| {
    spur_sched::node_match::expand_hostlist_or_split(nodelist)
        .into_iter()
        .collect()
});
```

Surfaces: `sinfo -n '<pattern>'` and `scontrol show node <pattern>`. `AuthMode::Permissive` is `#[default]` (`crates/spur-core/src/config.rs:619-629`), so this is the out-of-the-box posture, not a misconfiguration. Under `required` the same request still works for any authenticated user.

**2. Scheduler placement — persistent, survives restart.** Every scheduling cycle the backfill plugin builds a `NodePlacement` per pending job, which expands **both** user-controlled `--nodelist` and `--exclude`:

```rust
// crates/spur-sched/src/node_match.rs:34-47
let nodelist = job.spec.nodelist.as_deref()
    .filter(|s| !s.is_empty())
    .map(|s| HashSet::from_iter(expand_hostlist_or_split(s)));

let exclude = job.spec.exclude.as_deref()
    .filter(|s| !s.is_empty())
    .map(|s| HashSet::from_iter(expand_hostlist_or_split(s)))
    .unwrap_or_default();
```

`NodePlacement::new` is called from `find_suitable_nodes` (`spur-sched/src/backfill.rs:74`) and from the pending-reason classifiers in `spurctld/src/cluster.rs`. The job spec is persisted in the Raft log, so a single `sbatch -w 'node[0-18446744073709551615]'` re-triggers the expansion on **every scheduler tick and every controller restart** — the controller cannot stay up long enough to cancel the job. A narrower path expands pending `spec.nodelist` when any node is on dispatch cooldown (`cluster.rs:2287`).

**3. Admin RPC.** `remove_nodes` expands `req.nodes` (`spurctld/src/server.rs:2980`). Admin-gated, so it matters least. Client-side call sites (`squeue.rs:393`, `scontrol.rs:1268`, `node.rs:249`, `sinfo.rs:317`) kill only the user's own CLI process.

## Why it's a bug

Every other malformed-input class in this module is rejected with a `HostlistError` — unmatched `[`, reversed range (`node[5-1]`), `]`-before-`[` (#639), stray `]` before a comma (#643). Size is the one property never validated, even though the endpoints are `u64` and the caller is untrusted. The adjacent `array.rs` establishes the intended contract: a named maximum, enforced arithmetically *before* anything is materialized. Applying that bound here is consistency, not new policy.

## Why it matters

- **Remote DoS of the entire control plane, unauthenticated by default.** One 28-byte `sinfo -n` argument kills `spurctld`; every submit, query, and cancel then fails cluster-wide.
- **The scheduler variant is self-perpetuating.** A pending job's `--nodelist`/`--exclude` are re-expanded from the Raft-persisted spec every tick and every boot, so one submitted job can keep the controller from staying up — including long enough to cancel that job.
- **Cheap and accidental.** A typo in a generated nodelist (an extra digit in `node[0-99999999999]`) has the same effect.

## Expected

An out-of-range pattern is rejected with a `HostlistError` — the same treatment other malformed patterns already get — before any large allocation. Peak memory stays bounded by the cap regardless of input.

## Affected code (all line numbers against `main` @ `5ac18f01`)

- `crates/spur-core/src/hostlist.rs:306-324` — `expand_single()`: `u64` endpoints, `start > end` the only check, unbounded push loop.
- `crates/spur-core/src/hostlist.rs:8-13` — `HostlistError` has no size variant.
- `crates/spur-core/src/hostlist.rs:340-344` — `count()` delegates to `expand()`.
- `crates/spurctld/src/server.rs:1060`, `:1080-1086` — `get_nodes` RPC expands the caller's `nodelist`, unauthenticated by default.
- `crates/spur-sched/src/node_match.rs:34-47` — `NodePlacement::new` expands `spec.nodelist` and `spec.exclude`; reached from `spur-sched/src/backfill.rs:74` and `spurctld/src/cluster.rs`.
- `crates/spurctld/src/cluster.rs:2287` — pending `spec.nodelist` expanded on dispatch cooldown.
- `crates/spurctld/src/server.rs:2980` — `remove_nodes` expands `req.nodes` (admin-gated).

## Suggested fix

Mirror what #646 did for `parse_array_spec`: a named cap enforced arithmetically before materializing, plus an incremental backstop covering the recursive and comma-separated cases.

```rust
/// Maximum hosts a single pattern may expand to. Bounds peak memory on untrusted
/// input; well above any real cluster (~52 bytes/host, so ~52 MB at the cap).
pub const MAX_HOSTLIST_SIZE: usize = 1_000_000;
```

```rust
#[error("hostlist too large: {count} hosts exceeds maximum {max}")]
TooLarge { count: u64, max: usize },
```

In `expand_single`, before the loop:

```rust
if start > end {
    return Err(HostlistError::InvalidRange(format!("{} > {}", start, end)));
}
// Bound the element count BEFORE materializing. Saturating so the count
// arithmetic itself can't overflow on a u64::MAX range.
let range_count = (end - start).saturating_add(1);
if (results.len() as u64).saturating_add(range_count) > MAX_HOSTLIST_SIZE as u64 {
    return Err(HostlistError::TooLarge {
        count: (results.len() as u64).saturating_add(range_count),
        max: MAX_HOSTLIST_SIZE,
    });
}
```

The saturating arithmetic is load-bearing: a plain `end - start + 1` overflows on `node[0-18446744073709551615]`, the exact input this guard rejects. Keep an incremental guard inside the push loop too, so nested products (`rack[0-9999]-node[0-9999]`) are bounded by the cap overall.

Two follow-on notes:
- `expand_hostlist_or_split` (`spur-sched/src/node_match.rs:194-207`) falls back to a plain comma-split on `Err`, so an over-cap pattern becomes one literal name that matches nothing — safe, but a `--nodelist` job then silently never matches. Validating at `submit_job` would give a clean rejection and also close the persistent re-expansion variant.
- **`MAX_HOSTLIST_SIZE` is a policy choice for the maintainers.** 1,000,000 caps peak memory ~52 MB; matching `MAX_ARRAY_SIZE` (100,000) would be tighter. Either is defensible.

## Tests to add with the fix

- `expand("node[0-18446744073709551615]")` → `Err(TooLarge)`, returns promptly.
- `expand("node[0-999999]")` → `Ok`, 1,000,000 hosts (cap boundary unchanged).
- `expand("rack[0-9999]-node[0-9999]")` → `Err(TooLarge)` (nested product bounded overall).
- `expand("node[0-9],node[0-9]")` → `Ok`.
- `count("node[0-18446744073709551615]")` → `Err`.
- Existing valid patterns unchanged — `expand("node[001-003]")`, `expand("gpu[01-02],cpu[01-02]")`, `expand("node[5-1]")` → `InvalidRange`, `expand("node]1")` → `Ok` (#639), `expand("x],y")` → `["x]", "y"]` (#643).

## Related

- **#645 / PR #646** — `parse_array_spec` overflow + late size check. Same defect class in a different function; #646's early-bound approach is the template. This issue is the part that fix did not reach.
- **#642 / PR #643** (`split_top_level` depth underflow) and **#638 / #639** (`]`-before-`[` panic) — same file, but those are pattern *shape* validation; this is result *size*.

## Notes

- Verified on `main` @ `5ac18f01` (v0.8.0). The controller ran inside a 1.5 GB `systemd-run` scope, so only its own cgroup was OOM-killed and the host was never under memory pressure. Test cluster and scope torn down afterwards.
- A fix along the above lines is implemented and verified locally: full `spur-core` lib suite green (531 passed, 0 failed, incl. new regression tests) and `cargo build --workspace` clean. Happy to open a PR referencing this issue.

SumonAMD:
Closes #677

## Problem

`hostlist::expand()` parsed bracket-range endpoints as `u64` and materialized one
`String` per element with no limit on the resulting count. A 28-byte pattern —
`node[0-18446744073709551615]` — requests ~1.8e19 allocations, so `results` grows
until the kernel OOM-kills the process.

This is reachable from untrusted input on three paths, the first of which is a
read-only RPC that is unauthenticated under the default `auth.mode`:

- `get_nodes` (`spurctld/src/server.rs`) expands the caller-supplied `nodelist`
  filter — reached via `sinfo -n '<pattern>'` and `scontrol show node <pattern>`.
  One such query takes down the whole control plane.
- `NodePlacement::new` (`spur-sched/src/node_match.rs`) expands a pending job's
  `--nodelist` and `--exclude` every scheduling cycle. Because the spec is
  persisted in the Raft log, a single `sbatch -w '<pattern>'` re-triggers the
  expansion on every tick and every controller restart.
- `remove_nodes` (admin-gated, so lowest impact).

Size was the only property `hostlist.rs` never validated — unmatched `[`,
reversed ranges, `]`-before-`[` (#639), and stray `]` before a comma (#643) are
all already rejected.

## Fix

Mirrors the approach #646 took for `parse_array_spec` in the adjacent `array.rs`:

- Add `MAX_HOSTLIST_SIZE` (1,000,000) and a `HostlistError::TooLarge` variant.
- Bound the element count arithmetically in `expand_single` **before** anything is
  materialized. The bound uses `(end - start).saturating_add(1)` — this is
  load-bearing rather than stylistic, since a plain `end - start + 1` overflows on
  exactly the `u64::MAX` input the guard exists to reject.
- Keep an incremental check inside the push loop so nested products such as
  `rack[0-9999]-node[0-9999]` are bounded by the cap overall rather than per range.
  `results.len()` is the natural accumulator since it is already threaded through
  the recursion.

Peak memory is now bounded by the cap (~52 bytes/host, so ~52 MB) regardless of
input, and an over-cap pattern gets the same clean `HostlistError` treatment every
other malformed pattern already receives.

## Tests

Four regression tests added to `hostlist.rs`:

- `expand_rejects_oversized_range` — `node[0-18446744073709551615]` returns
  `Err(TooLarge)` promptly (guards the RPC DoS).
- `expand_allows_cap_boundary` — `node[0-999999]` still yields 1,000,000 hosts.
- `expand_rejects_nested_product_over_cap` — `rack[0-9999]-node[0-9999]` is bounded
  overall, not per range.
- `count_rejects_oversized_range` — `count()` is covered too, since it delegates to
  `expand()`.

Existing valid patterns are unchanged, including the `#639` and `#643` cases.

## Verification

Run locally on Ubuntu 24.04 with the pinned 1.95.0 toolchain and
`RUSTFLAGS="-D warnings"`, rebased onto `main` @ `54f3763b`:

- `cargo fmt --all --check` — clean
- `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` — zero warnings
- `cargo build --all-targets --locked` — clean, `Cargo.lock` unchanged
- `cargo test --locked` — all green (spur-core 541, spurctld 839, spurd 238), 0 failures

## Notes for review

- **`MAX_HOSTLIST_SIZE = 1_000_000` is a policy choice I'd like maintainers to
  confirm.** It is generous next to the largest real clusters while capping peak
  memory around 52 MB. Matching `MAX_ARRAY_SIZE` (100,000) would be tighter and
  more consistent — either is defensible, happy to change it.
- `TooLarge { count: u64, max: usize }` uses `u64` for `count` where `array.rs`
  uses `usize`. This is deliberate: the saturating range count is computed in `u64`
  before any narrowing, so the error can faithfully report a value above `usize`
  range on the pathological input.
- Out of scope but worth flagging: `expand_hostlist_or_split`
  (`spur-sched/src/node_match.rs`) falls back to a plain comma-split when `expand`
  returns `Err`, so an over-cap `--nodelist` becomes one literal name that matches
  nothing. That is safe and keeps the RPC responsive, but the job then silently
  never matches a node. Validating the nodelist at `submit_job` would give the user
  a clean rejection and close the persistent re-expansion variant at the door — happy
  to do that as a follow-up if you'd like.

### DIFF
--- a/crates/spur-core/src/hostlist.rs
+++ b/crates/spur-core/src/hostlist.rs
@@ -665,6 +665,16 @@ mod tests {
     }
 
     #[test]
+    fn expand_rejects_recursive_overshoot_past_cap() {
+        // A comma-list suffix pushes per element without consulting the pre-loop
+        // bound, so the running total steps past the cap instead of being
+        // rejected in advance. Exercises the incremental backstop, which the
+        // nested-range case above never reaches.
+        let err = expand("rack[0-999999]-node[1,2]").unwrap_err();
+        assert!(matches!(err, HostlistError::TooLarge { .. }));
+    }
+
+    #[test]
     fn count_rejects_oversized_range() {
         assert!(count("node[0-18446744073709551615]").is_err());
     }
--- a/crates/spur-core/src/hostlist.rs
+++ b/crates/spur-core/src/hostlist.rs
@@ -3,6 +3,8 @@
 
 use thiserror::Error;
 
+pub const MAX_HOSTLIST_SIZE: usize = 1_000_000;
+
 /// Errors from hostlist parsing.
 #[derive(Debug, Error)]
 pub enum HostlistError {
@@ -10,6 +12,8 @@ pub enum HostlistError {
     InvalidPattern(String),
     #[error("invalid range: {0}")]
     InvalidRange(String),
+    #[error("hostlist too large: {count} hosts exceeds maximum {max}")]
+    TooLarge { count: u64, max: usize },
 }
 
 /// Expand a Slurm hostlist pattern into individual hostnames.
@@ -314,12 +318,30 @@ fn expand_single(pattern: &str, results: &mut Vec<String>) -> Result<(), Hostlis
                     return Err(HostlistError::InvalidRange(format!("{} > {}", start, end)));
                 }
 
+                // Bound element count BEFORE materializing. Saturating so the count
+                // arithmetic itself can't overflow on a u64::MAX range.
+                let range_count = (end - start).saturating_add(1);
+                if (results.len() as u64).saturating_add(range_count) > MAX_HOSTLIST_SIZE as u64 {
+                    return Err(HostlistError::TooLarge {
+                        count: (results.len() as u64).saturating_add(range_count),
+                        max: MAX_HOSTLIST_SIZE,
+                    });
+                }
+
                 for i in start..=end {
                     let name = format!("{}{:0>width$}{}", prefix, i, suffix, width = width);
                     if suffix.contains('[') {
                         expand_single(&name, results)?;
                     } else {
                         results.push(name);
+                    }
+                    // Incremental backstop: nested products like rack[0-9999]-node[0-9999]
+                    // can exceed the cap even when each single range is under it.
+                    if results.len() > MAX_HOSTLIST_SIZE {
+                        return Err(HostlistError::TooLarge {
+                            count: results.len() as u64,
+                            max: MAX_HOSTLIST_SIZE,
+                        });
                     }
                 }
             } else {
@@ -622,5 +644,28 @@ mod tests {
     #[test]
     fn recursive_suffix_bracket_order_rejected() {
         assert!(expand("node[1-2]a]b[3-4]").is_err());
+    }
+
+    #[test]
+    fn expand_rejects_oversized_range() {
+        let err = expand("node[0-18446744073709551615]").unwrap_err();
+        assert!(matches!(err, HostlistError::TooLarge { .. }));
+    }
+
+    #[test]
+    fn expand_allows_cap_boundary() {
+        let hosts = expand("node[0-999999]").unwrap();
+        assert_eq!(hosts.len(), 1_000_000);
+    }
+
+    #[test]
+    fn expand_rejects_nested_product_over_cap() {
+        let err = expand("rack[0-9999]-node[0-9999]").unwrap_err();
+        assert!(matches!(err, HostlistError::TooLarge { .. }));
+    }
+
+    #[test]
+    fn count_rejects_oversized_range() {
+        assert!(count("node[0-18446744073709551615]").is_err());
     }
 }


sajmera-pensando (commented):
One minor issue to address in a follow-up PR:

**In-loop guard uses `>` instead of `>=` (line 340)**

The pre-loop check correctly fires when the total *would reach* `MAX_HOSTLIST_SIZE`. But the in-loop backstop uses `results.len() > MAX_HOSTLIST_SIZE`, which means it allows one element past the cap to be pushed before stopping. These two guards are inconsistent with each other.

For the comma-suffix path (e.g. `rack[0-999999]-node[1,2]`), `results` can reach `MAX_HOSTLIST_SIZE + 1` before the error fires. Practically harmless — one extra `String` — but it contradicts the stated guarantee and could confuse future readers. Changing `>` to `>=` on that line makes both guards consistent.

Could you raise a small follow-up PR with that one-character fix? Happy to approve it immediately.

Also worth tracking as a separate follow-up: the `--nodelist` / `--exclude` path at `submit_job`. As you noted in the description, an over-cap pattern currently silently becomes a literal name that matches nothing — the user gets no error, just a job that sits in queue forever. Slurm rejects at submission with a clear message. A follow-up that validates the nodelist at `submit_job` would close that gap and is the right fix.

sajmera-pensando (dismissed):
Fix is correct, well-reasoned, and addresses a real DoS vulnerability. All 542 spur-core tests pass with no regressions. The two items in my earlier comment (in-loop guard off-by-one, silent non-match at submit) are both follow-up nits — neither blocks this. Approving.

PR #680, opened 2026-08-18
fix(spur-core): use checked arithmetic in time parsers to prevent overflow

### LINKED ISSUE #678: [Issue]: Slurm time parser integer overflow — spurctld panics (debug) or sets wrong MaxTime (release)
### Problem Description

The Slurm-grammar half of the time parser in `crates/spur-core/src/config.rs` multiplies and adds user-supplied numbers with raw `*` and `+`. A partition `max_time` of `"2982617-00:00:00"` — about 8168 years — overflows `u32`. In a release build the value wraps to **1184 minutes** and `spurctld` starts up reporting `MaxTime=19:44:00`; in a debug build the same config file makes `spurctld` panic during config validation and fail to start.

Both outcomes contradict invariants the surrounding code documents explicitly. The immediately adjacent duration parser in the same file, `parse_suffix_duration_seconds`, uses `checked_mul` and `checked_add` throughout — which is what marks this as an oversight in one code path rather than a deliberate choice.

### Operating System

Ubuntu 24.04 LTS (kernel 6.8.0-38-generic) — defect is OS-independent

### CPU

N/A — defect is hardware-independent (reproduced on a CPU-only loopback controller)

### GPU

N/A — defect is hardware-independent; no GPU is involved (config/time-parser code path)

### ROCm Version

N/A — defect is hardware-independent (pure spurctld control-plane bug, not ROCm-runtime dependent)

### ROCm Component

_No response_

### Steps to Reproduce

Route A is the operator-facing failure and shows the release/debug split. Route B pins the defect to the parser in isolation. All output is verbatim.

**Route A — a partition `max_time` that overflows. No agent or GPU needed.**

1. Build both profiles and write a config whose only unusual field is `max_time` (2982617 days × 1440 min/day = 4294968480 = `u32::MAX` + 1185):

```bash
cargo build -q -p spurctld -p spur-cli              # debug assertions on
cargo build -q --release -p spurctld -p spur-cli

WORK=/tmp/spur_timetest; mkdir -p "$WORK/state"
cat > "$WORK/spur.conf" <<EOF
cluster_name = "timetest"

[controller]
listen_addr = "127.0.0.1:6899"
state_dir = "$WORK/state"

[scheduler]
plugin = "backfill"
interval_secs = 1

[auth]
plugin = "none"

[[partitions]]
name = "gpu"
default = true
state = "UP"
nodes = "node1"
max_time = "2982617-00:00:00"

[[nodes]]
names = "node1"
cpus = 8
memory_mb = 16384

[network]
wg_enabled = false
agent_port = 6899
EOF
```

2. Start the **debug** controller — it panics in config validation and never starts:

```console
$ target/debug/spurctld -f $WORK/spur.conf -D --state-dir $WORK/state
thread 'main' (4042790) panicked at crates/spur-core/src/config.rs:1628:21:
attempt to multiply with overflow
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Expected instead: `ConfigError::InvalidValue { field: "partitions.gpu.max_time", … }`, which is what `validate()` produces for every other bad value (e.g. `max_time = "1 hour"`).

3. Start the **release** controller — it accepts the value and reports a limit nobody configured:

```console
$ target/release/spurctld -f $WORK/spur.conf -D --state-dir $WORK/state &
$ target/release/spur scontrol show partition --controller http://127.0.0.1:6899
PartitionName=gpu Default=YES
   MaxTime=19:44:00 DefaultTime=UNLIMITED
```

`19:44:00` is 1184 minutes: `4294968480 mod 2^32`. The configured ~8168 years became 19h44m silently, with no warning in the log.

4. Teardown: `pkill -f spur_timetest; rm -rf /tmp/spur_timetest`

**Route B — the parser in isolation, both profiles.**

1. Save as `crates/spur-core/examples/probe_time.rs`:

```rust
// Unchecked arithmetic in the Slurm-grammar half of the time parser.
fn main() {
    for s in ["1-00:00:00", "2982616-00:00:00", "2982617-00:00:00", "71582789:00"] {
        println!("parse_time_minutes({:>18?}) => {:?}", s, spur_core::config::parse_time_minutes(s));
    }
    for s in ["1-00:00:00", "213503982334602-00:00:00"] {
        println!("parse_time_seconds({:>26?}) => {:?}", s, spur_core::config::parse_time_seconds(s));
    }
    // Control: the suffix path in the same file uses checked_mul/checked_add.
    let big = format!("{}d", "9".repeat(20));
    println!("control, suffix path => {:?}", spur_core::config::parse_time_minutes(&big));
}
```

2. Run both profiles:

```bash
cargo build -q --release -p spur-core --example probe_time && ./target/release/examples/probe_time
cargo build -q          -p spur-core --example probe_time && ./target/debug/examples/probe_time
```

3. Observed — release wraps:

```console
parse_time_minutes(      "1-00:00:00") => Some(1440)
parse_time_minutes("2982616-00:00:00") => Some(4294967040)
parse_time_minutes("2982617-00:00:00") => Some(1184)
parse_time_minutes(     "71582789:00") => Some(44)
parse_time_seconds(              "1-00:00:00") => Some(86400)
parse_time_seconds("213503982334602-00:00:00") => Some(61184)
control, suffix path => None
```

debug panics on the same input, exit 101:

```console
parse_time_minutes(      "1-00:00:00") => Some(1440)
parse_time_minutes("2982616-00:00:00") => Some(4294967040)
thread 'main' (4000483) panicked at crates/spur-core/src/config.rs:1628:21:
attempt to multiply with overflow
```

Three things to note. `2982617-00:00:00` (~8168 years) becomes 19h44m. `71582789:00` becomes 44 minutes. And `2982616-00:00:00` shows there is no sanity ceiling either — it does not overflow, so ~8168 years is accepted as a valid limit. The final line is the control: the hardened suffix path in the same file rejects an equally absurd value cleanly.

<details>

```
Paste output here
```

</details>

## Root cause

Four functions do unchecked arithmetic on parsed input.

```rust
// crates/spur-core/src/config.rs:1623-1643
fn parse_slurm_time_minutes(s: &str) -> Option<u32> {
    // days-hours:minutes:seconds
    if let Some((days, rest)) = s.split_once('-') {
        let days: u32 = days.parse().ok()?;
        let hms = parse_hms(rest)?;
        return Some(days * 24 * 60 + hms);   // <-- unchecked
    }

    // hours:minutes:seconds or hours:minutes or just minutes
    let parts: Vec<&str> = s.split(':').collect();
    match parts.len() {
        1 => parts[0].parse().ok(),
        2 => {
            let h: u32 = parts[0].parse().ok()?;
            let m: u32 = parts[1].parse().ok()?;
            Some(h * 60 + m)                 // <-- unchecked
        }
        3 => parse_hms(s),
        _ => None,
    }
}
```

`parse_hms` (`:1741-1754`) has the same shape (`h * 60 + m + …`), and the seconds variants `parse_slurm_time_seconds` (`:1658-1687`) and `parse_hms_seconds` (`:1723-1737`) repeat it in `u64` (`days * 86400 + …`, `h * 3600 + m * 60 + sec`).

The sibling function five lines below is careful about exactly this:

```rust
// crates/spur-core/src/config.rs:1692-1721 — parse_suffix_duration_seconds
num = Some(num.unwrap_or(0).checked_mul(10)?.checked_add(digit as u64)?);
...
total = total.checked_add(value.checked_mul(unit_secs)?)?;
```

So `"99999999999999999999d"` is correctly rejected as `None`, while the Slurm-grammar `"213503982334602-00:00:00"` wraps. Both are reachable through the same public function — `parse_time_minutes` (`:1612`) tries the Slurm grammar first, then falls back to the suffix parser.

There is no `[profile.release]` section and no `overflow-checks` setting anywhere in the workspace's TOML files, so the two profiles diverge on stock cargo defaults: debug panics, release wraps.

Overflow thresholds, for reference:

| Path | Overflows when |
|---|---|
| `parse_slurm_time_minutes`, `D-HH:MM:SS` | `days ≥ 2982617` (and at `days = 2982616` an `hms` above 255 tips it over) |
| `parse_slurm_time_minutes`, `HH:MM` | `hours ≥ 71582789` |
| `parse_slurm_time_seconds`, `D-HH:MM:SS` | `days ≥ 213503982334602` |

## Why it's a bug

Two invariants are stated in comments right next to the call sites, and overflow defeats both.

**1. `Config::validate()` promises a bad partition time fails loudly.**

```rust
// crates/spur-core/src/config.rs:1522-1529
// Reject malformed partition time strings at load rather than silently
// treating them as UNLIMITED (a typo like "1 hour" must fail, not
// remove the cap).
for pc in &self.partitions {
    let max = pc.max_time.as_ref()
        .map(|t| parse_partition_time(&format!("partitions.{}.max_time", pc.name), t))
```

`parse_partition_time` (`:1595-1607`) is documented as "Unlike `parse_time_minutes`, a typo like `"1 hour"` fails loudly instead of collapsing to UNLIMITED". It implements that by treating `None` as `ConfigError::InvalidValue`. Overflow returns `Some(wrapped)` rather than `None`, so a nonsense value slips through the one gate built to catch nonsense values — and instead of removing the cap, it silently *changes* the cap to an unrelated number.

**2. `build_partitions()` states its inputs were already validated.**

```rust
// crates/spur-core/src/config.rs:1552-1554
/// Convert partition configs to Partition structs. Time strings are assumed
/// valid here; [`Self::validate`] rejects malformed ones at load, so a stray
/// `None` below only ever means UNLIMITED, never a swallowed parse error.
```

That assumption is what makes `.and_then(|t| parse_time_minutes(t))` at `:1569-1570` safe. Since `validate` no longer rejects overflowing values, the invariant no longer holds.

The `default_time > max_time` consistency check at `:1540-1546` also compares post-wrap numbers, so it can pass or fail for reasons unrelated to what the operator wrote.

Independently of overflow, there is no sanity ceiling: `"2982616-00:00:00"` does not overflow and is accepted as `Some(4294967040)` — a time limit of roughly 8168 years.

## Why it matters

- **Debug builds will not start on such a config.** The panic happens inside config validation — a startup failure with an `attempt to multiply with overflow` backtrace into `spur-core`, instead of the actionable `ConfigError::InvalidValue` naming the offending field. Plain `cargo build` enables debug assertions, including the `cargo build --all-targets --locked` in `.github/workflows/ci.yml`.
- **Release builds silently enforce a limit nobody configured.** A partition set to ~8168 years reports and enforces `MaxTime=19:44:00`. Jobs inherit a wall clock orders of magnitude shorter than intended and get killed early, with nothing in the logs linking the kill to the config value.
- **The parser is reachable from unprivileged surfaces**, not just operator config (see below), so the debug-profile panic is reachable from an RPC argument too.

Being precise about severity: the triggering values are extreme, so a hand-typed `--time` is unlikely to hit it. The realistic routes are generated/templated configs, unit conversions producing a large day count, fuzzing, and deliberate input. The strongest arguments for fixing are the two defeated invariants and the profile-dependent behaviour, not a likely accident.

## Reachability (all line numbers against `main` @ `5ac18f01`)

| Surface | Call site |
|---|---|
| `spur.conf` partition `max_time` / `default_time` (validation) | `spur-core/src/config.rs:1529`, `:1535` via `parse_partition_time` |
| `spur.conf` partition build | `spur-core/src/config.rs:1569-1570` |
| Create/update partition RPC | `spurctld/src/server.rs:1932`, `:1942` |
| `scontrol update PartitionName=… MaxTime=…` handling | `spurctld/src/cluster.rs:3573`, `:3585` |
| REST job submit time limit | `spurctld/src/rest/handlers.rs:96` |
| `sbatch --time` | `spur-cli/src/sbatch.rs:894` (`parse_time_seconds`) |
| `srun --time` | `spur-cli/src/srun.rs:545` |
| `salloc --time` | `spur-cli/src/salloc.rs:280` |
| `scontrol` time fields | `spur-cli/src/scontrol.rs:1219`, `:1544` |

## Expected

Overflow returns `None`, the same as every other unparseable input, so:
- `parse_partition_time` turns it into `ConfigError::InvalidValue` naming the field — the loud failure `validate()` was written to produce.
- The RPC and REST paths return an invalid-argument error.
- Debug and release behave identically.

## Affected code (all line numbers against `main` @ `5ac18f01`)

- `crates/spur-core/src/config.rs:1628` — `days * 24 * 60 + hms` (the line both panics point at).
- `crates/spur-core/src/config.rs:1638` — `h * 60 + m` in the `HH:MM` arm.
- `crates/spur-core/src/config.rs:1741-1754` — `parse_hms`: `h * 60 + m + …`.
- `crates/spur-core/src/config.rs:1658-1687` — `parse_slurm_time_seconds`, same pattern in `u64` (`:1662`, `:1670`, `:1676`, `:1683`).
- `crates/spur-core/src/config.rs:1723-1737` — `parse_hms_seconds`, same (`:1729`, `:1735`).
- Correct by contrast: `crates/spur-core/src/config.rs:1692-1721` (`parse_suffix_duration_seconds`) uses `checked_mul`/`checked_add` throughout.

## Suggested fix

Use checked arithmetic and return `None` on overflow, mirroring the suffix path so the two halves of the same public function agree. `None` already means "not a valid duration" to every caller, so each one turns it into its own clean error with no call-site changes:

```diff
 fn parse_slurm_time_minutes(s: &str) -> Option<u32> {
     if let Some((days, rest)) = s.split_once('-') {
         let days: u32 = days.parse().ok()?;
         let hms = parse_hms(rest)?;
-        return Some(days * 24 * 60 + hms);
+        return days.checked_mul(24 * 60)?.checked_add(hms);
     }
 
     let parts: Vec<&str> = s.split(':').collect();
     match parts.len() {
         1 => parts[0].parse().ok(),
         2 => {
             let h: u32 = parts[0].parse().ok()?;
             let m: u32 = parts[1].parse().ok()?;
-            Some(h * 60 + m)
+            h.checked_mul(60)?.checked_add(m)
         }
         3 => parse_hms(s),
         _ => None,
     }
 }
```

The same mechanical change applies to `parse_hms`, `parse_slurm_time_seconds`, and `parse_hms_seconds`. Note `checked_add` returns the `Option` directly, so the arms lose their `Some(...)` wrapper rather than nesting inside it.

Worth deciding at review time: whether to also add a sanity ceiling (a `MAX_TIME_MINUTES`), since checked arithmetic alone still accepts `"2982616-00:00:00"` as a ~8168-year limit. Checked arithmetic is the correctness fix; a ceiling is a policy call and could be a follow-up.

## Tests to add with the fix

- `parse_time_minutes("2982617-00:00:00")` → `None` (u32 days overflow).
- `parse_time_minutes("2982616-05:00:00")` → `None` (overflow on the `+ hms`, not the multiply — 2982616×1440 leaves only 255 minutes of headroom and `05:00:00` is 300).
- `parse_time_minutes("71582789:00")` → `None` (`HH:MM` arm).
- `parse_time_seconds("213503982334602-00:00:00")` → `None` (u64 days overflow).
- `parse_partition_time("p.max_time", "2982617-00:00:00")` → `Err(InvalidValue)` (the invariant this issue is really about).
- Boundary and control cases unchanged: `parse_time_minutes("2982616-00:00:00")` → `Some(4294967040)`, plus the existing `"60"`, `"1:30"`, `"72:00:00"`, `"1-00:00:00"`, `"1h40m"`, `"INFINITE"` expectations.

## Related

- **#645 / PR #646** — `parse_array_spec` u32 overflow. Same class of defect (unchecked arithmetic on untrusted numeric input in `spur-core`), different function; that one's consequence was allocation, this one's is a silently wrong value.

## Notes

- Verified on `main` @ `5ac18f01` (v0.8.0). Route A ran a controller only, on loopback port 6899, with no agent and no GPUs; the temporary config and state were removed afterwards.
- A fix along the above lines is implemented and verified locally: full `spur-core` lib suite green (531 passed, 0 failed, incl. new regression tests) and `cargo build --workspace` clean. With the fix applied, the same config file makes the controller exit 1 with `Error: invalid value for partitions.gpu.max_time: 2982617-00:00:00` in both profiles. Happy to open a PR referencing this issue.

SumonAMD:
Closes #678

## Problem

Four time-parsing helpers in `config.rs` multiplied and added user-supplied
values with unchecked arithmetic. `parse_suffix_duration_seconds`, sitting
directly between them in the same file, already uses `checked_mul`/`checked_add`
throughout — the Slurm-grammar path just never got the same treatment.

The result is profile-dependent, and the release behavior is the worse of the two.
Starting `spurctld` with a partition whose `max_time = "2982617-00:00:00"`:

- **Debug:** `panicked at crates/spur-core/src/config.rs:1628:21: attempt to
  multiply with overflow`, exit 101.
- **Release:** the controller starts and becomes leader, and `spur show partition`
  reports `MaxTime=19:44:00`. 2982617 × 1440 wraps to 1184 minutes, so an operator
  who asked for ~8000 years silently gets a 19-hour cap, with nothing logged.

This defeats two invariants the code states in its own doc comments. The first is
on `parse_partition_time`: "a typo like `"1 hour"` fails loudly instead of
collapsing to UNLIMITED" — an overflowing value fails neither loudly nor to
UNLIMITED, it collapses to an arbitrary shorter limit. The second is the
round-up rule in `parse_hms`, which can round a wrapped value down instead.

The parsers are reachable from operator config, from the create/update partition
RPC and `scontrol update PartitionName=… MaxTime=…`, from the REST job-submit
handler, and from `--time` on `sbatch`/`srun`/`salloc`, so the input is not
exclusively trusted.

## Fix

Every affected arm already returns `Option`, so this is mechanical: each `*`
becomes `.checked_mul(..)?` and each `+` becomes `.checked_add(..)`, and overflow
falls out as `None` — the same signal a malformed string already produces. No new
error variant and no policy decision needed, since `parse_partition_time` already
converts `None` into `ConfigError::InvalidValue`.

Covered functions:

- `parse_slurm_time_minutes` — `D-HH:MM:SS` and `HH:MM` arms
- `parse_slurm_time_seconds` — days, bare-minutes, `HH:MM`, and `HH:MM:SS` arms
- `parse_hms_seconds` — both arms
- `parse_hms` — the seconds round-up expression

## Tests

- `parse_time_rejects_overflowing_slurm_durations` — covers overflow on the
  multiply (`2982617-00:00:00`), on the add (`2982616-05:00:00`, where 2982616
  days leaves 255 minutes of headroom and `05:00:00` needs 300), the `HH:MM` arm
  (`71582789:00`), and the u64 seconds path (`213503982334602-00:00:00`). It also
  asserts the largest non-overflowing value, `2982616-00:00:00`, still returns
  `Some(4_294_967_040)`, so nothing valid was narrowed.
- `parse_partition_time_rejects_overflowing_duration` — confirms the loud-failure
  contract now holds for overflow, not just for typos.

Verified end-to-end as well: after the fix, both debug and release print
`Error: invalid value for partitions.gpu.max_time: 2982617-00:00:00` and exit 1.

## Verification

Ubuntu 24.04, pinned 1.95.0 toolchain, `RUSTFLAGS="-D warnings"`, on `main`
@ `54f3763b`:

- `cargo fmt --all --check` — clean
- `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` — zero warnings
- `cargo build --all-targets --locked` — clean, `Cargo.lock` unchanged
- `cargo test --locked` — 33 test binaries green (spur-core 539, spurctld 839,
  spurd 238), 0 failures

## Notes for review

- I deliberately did **not** add a sanity ceiling (something like "no more than N
  years"). Rejecting only what cannot be represented is the smaller, more
  obviously-correct change, and a ceiling is a policy call that belongs to you. Say
  the word if you'd like one and I'll add it here.
- `parse_time_minutes`'s suffixed-duration fallback (`90m`, `2d12h`) was already
  overflow-safe and is untouched; this only brings the Slurm-grammar path up to the
  same standard.

### DIFF
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -1625,7 +1625,7 @@ fn parse_slurm_time_minutes(s: &str) -> Option<u32> {
     if let Some((days, rest)) = s.split_once('-') {
         let days: u32 = days.parse().ok()?;
         let hms = parse_hms(rest)?;
-        return Some(days * 24 * 60 + hms);
+        return days.checked_mul(24 * 60)?.checked_add(hms);
     }
 
     // hours:minutes:seconds or hours:minutes or just minutes
@@ -1635,7 +1635,7 @@ fn parse_slurm_time_minutes(s: &str) -> Option<u32> {
         2 => {
             let h: u32 = parts[0].parse().ok()?;
             let m: u32 = parts[1].parse().ok()?;
-            Some(h * 60 + m)
+            h.checked_mul(60)?.checked_add(m)
         }
         3 => parse_hms(s),
         _ => None,
@@ -1659,7 +1659,9 @@ fn parse_slurm_time_seconds(s: &str) -> Option<u64> {
     // days-hours:minutes:seconds
     if let Some((days, rest)) = s.split_once('-') {
         let days: u64 = days.parse().ok()?;
-        return Some(days * 86400 + parse_hms_seconds(rest)?);
+        return days
+            .checked_mul(86400)?
+            .checked_add(parse_hms_seconds(rest)?);
     }
 
     let parts: Vec<&str> = s.split(':').collect();
@@ -1667,20 +1669,22 @@ fn parse_slurm_time_seconds(s: &str) -> Option<u64> {
         1 => {
             // Just minutes → convert to seconds
             let mins: u64 = parts[0].parse().ok()?;
-            Some(mins * 60)
+            mins.checked_mul(60)
         }
         2 => {
             // HH:MM → hours and minutes (no seconds)
             let h: u64 = parts[0].parse().ok()?;
             let m: u64 = parts[1].parse().ok()?;
-            Some(h * 3600 + m * 60)
+            h.checked_mul(3600)?.checked_add(m.checked_mul(60)?)
         }
         3 => {
             // HH:MM:SS
             let h: u64 = parts[0].parse().ok()?;
             let m: u64 = parts[1].parse().ok()?;
             let sec: u64 = parts[2].parse().ok()?;
-            Some(h * 3600 + m * 60 + sec)
+            h.checked_mul(3600)?
+                .checked_add(m.checked_mul(60)?)?
+                .checked_add(sec)
         }
         _ => None,
     }
@@ -1726,13 +1730,15 @@ fn parse_hms_seconds(s: &str) -> Option<u64> {
         2 => {
             let h: u64 = parts[0].parse().ok()?;
             let m: u64 = parts[1].parse().ok()?;
-            Some(h * 3600 + m * 60)
+            h.checked_mul(3600)?.checked_add(m.checked_mul(60)?)
         }
         3 => {
             let h: u64 = parts[0].parse().ok()?;
             let m: u64 = parts[1].parse().ok()?;
             let sec: u64 = parts[2].parse().ok()?;
-            Some(h * 3600 + m * 60 + sec)
+            h.checked_mul(3600)?
+                .checked_add(m.checked_mul(60)?)?
+                .checked_add(sec)
         }
         _ => None,
     }
@@ -1750,7 +1756,10 @@ fn parse_hms(s: &str) -> Option<u32> {
     } else {
         0
     };
-    Some(h * 60 + m + if s > 0 { 1 } else { 0 }) // Round up seconds
+    // Round up seconds
+    h.checked_mul(60)?
+        .checked_add(m)?
+        .checked_add(if s > 0 { 1 } else { 0 })
 }
 
 /// Format minutes as D-HH:MM:SS or HH:MM:SS.
@@ -2279,6 +2288,31 @@ memory_mb = 1024000
         // Slurm grammar unchanged.
         assert_eq!(parse_time_seconds("60"), Some(3600));
         assert_eq!(parse_time_seconds("00:00:30"), Some(30));
+    }
+
+    #[test]
+    fn parse_time_rejects_overflowing_slurm_durations() {
+        // u32 minutes overflow: 2982617 days * 1440 is u32::MAX + 1185.
+        assert_eq!(parse_time_minutes("2982617-00:00:00"), None);
+        // Overflow on the `+ hms`, not the multiply: 2982616 days leaves only
+        // 255 minutes of headroom and "05:00:00" is 300.
+        assert_eq!(parse_time_minutes("2982616-05:00:00"), None);
+        // HH:MM arm.
+        assert_eq!(parse_time_minutes("71582789:00"), None);
+        // u64 seconds overflow.
+        assert_eq!(parse_time_seconds("213503982334602-00:00:00"), None);
+        // The largest non-overflowing value is still accepted unchanged.
+        assert_eq!(parse_time_minutes("2982616-00:00:00"), Some(4_294_967_040));
+    }
+
+    #[test]
+    fn parse_partition_time_rejects_overflowing_duration() {
+        // validate() promises a bad partition time fails loudly rather than
+        // silently enforcing a different cap.
+        assert!(matches!(
+            parse_partition_time("partitions.gpu.max_time", "2982617-00:00:00"),
+            Err(ConfigError::InvalidValue { .. })
+        ));
     }
 
     #[test]


sajmera-pensando (approved):
Well-constructed fix. All four parsers have real overflow bugs; the release behavior (silent wrap to a shorter limit with nothing logged) is worse than the debug behavior (panic), and the PR correctly fixes both. The mechanical approach — `checked_mul`/`checked_add` returning `None`, propagated by `?` — is exactly right because these functions already use `Option` for malformed input; no new error machinery needed.

The test for `2982616-05:00:00` is the most valuable one: it proves the add-overflow path is caught independently of the multiply, and it's a case that's easy to miss. The boundary assertion (`2982616-00:00:00` → `Some(4_294_967_040)`) confirms the fix doesn't narrow the valid range. Good.

One uncovered line (Codecov): almost certainly the `None` path of `mins.checked_mul(60)` in the bare-minutes arm of `parse_slurm_time_seconds`. You'd need 307+ trillion bare minutes to trigger it — not worth a test.

The decision to skip a sanity ceiling is right for this PR. Rejecting only what the type literally cannot represent is the smallest correct change; a policy max is a separate discussion.

Compared to Slurm: Slurm's C parser rejects overflow values explicitly. Old Spur was silently worse. New Spur matches Slurm's intent.

PR #683, opened 2026-08-18
feat(spurctld): QOS preemption hierarchy and minimum exempt time

sajmera-pensando:
## Context

Spur's preemption was missing two controls that Slurm provides, causing a practical problem: a team's QOS quota provides no runtime guarantee — a higher-priority job from any other pool can displace running work that is well within its allocated quota. This PR closes both gaps.

**What Spur already matched Slurm on (unchanged):**
- `preempt_mode` (`cancel`, `requeue`, `suspend`, `off`) — per partition, with QOS override
- Partition `priority_tier` — lower-tier partitions are protected from equal/lower-tier preemption
- Named reservation protection — jobs inside an active time-windowed reservation are shielded

### [1] QOS preemption hierarchy (`PreemptType=preempt/qos`)

Previously, any job with 2× higher effective priority could preempt any other job regardless of which QOS or pool it belonged to. There was no way to say "tier A may preempt tier B but not tier C."

This PR adds:
- `scheduler.preempt_type = "qos_priority"` in `spur.conf` — enables the hierarchy
- `preempt = <comma-separated QOS names>` field on each QOS — the allow-list of tiers this QOS may preempt
- When enabled, `try_preempt` enforces: a pending job may only preempt a running job when the pending job's QOS explicitly lists the running job's QOS in its `preempt` allow-list. An empty allow-list means the QOS may not preempt anything.

```bash
# Example two-tier setup
sacctmgr add qos name=burst  priority=100   preemptmode=cancel
sacctmgr add qos name=priority priority=10000 preempt=burst
# "priority" can preempt "burst"; "burst" cannot preempt anyone;
# "reserved" (no preempt list) cannot be preempted by either.
```

### [2] Minimum guaranteed running time (`PreemptExemptTime`)

Previously, a newly-started job was immediately eligible for preemption — no grace window.

This PR adds `preempt_exempt_time` at three levels (most specific wins: QOS > partition > global):
- Global: `scheduler.preempt_exempt_time = <seconds>` in `spur.conf`
- Per partition: `preempt_exempt_time` in `[[partitions]]`, or via `scontrol update PartitionName=<name> PreemptExemptTime=<secs>`
- Per QOS: `preemptexempttime=<secs>` via `sacctmgr`; use `clearpreemptexempttime` to revert to inheriting

## Approach

Both controls are implemented as additional filters in `try_preempt` in `scheduler_loop.rs`, layered on top of the existing priority-gap, node-overlap, and reservation-tier checks. Both default to the existing behavior (`preempt_type = "none"`, `preempt_exempt_time = 0`), so this is fully backward-compatible for existing clusters.

Key design choices:
- `preempt_type` and the global `preempt_exempt_time` live in `[scheduler]` (TOML, live-reloadable)
- Per-partition `preempt_exempt_time` lives in `[[partitions]]` and is also settable at runtime via `scontrol update`. It survives `scontrol reconfigure` when not specified in TOML (the WAL entry is skipped, leaving the runtime value intact).
- `preempt_exempt_time = 0` is a valid explicit value meaning "immediately preemptable"; to revert a QOS/partition to inheriting the global default, use `clearpreemptexempttime` / `--clear-preempt-exempt-time`.
- `start_time = None` on a running job (should not happen, but defensively handled) is treated as immediately preemptable rather than infinitely protected.
- For multi-partition jobs, the maximum partition-level `preempt_exempt_time` is used (most protective, consistent with how `preempt_mode` picks the most aggressive mode).

The implementation is wired end-to-end: DB schema (with `ALTER TABLE ... ADD COLUMN IF NOT EXISTS` migrations) → limits cache → gRPC → proto → `scontrol` / `sacctmgr`.

## Testing

`cargo clippy --workspace --exclude spur-ffi --all-targets --locked` — clean  
`cargo test --locked` — all pass

Unit tests added in `scheduler_loop.rs`:
- `effective_exempt_secs` resolution (QOS > partition > global, multi-partition max)
- QOS allow-list permit/block/empty cases

Integration tests added in `cluster.rs` (full `ClusterManager` + `try_preempt`):
- `qos_preempt_hierarchy_blocks_preemption_when_not_in_allow_list`
- `qos_preempt_hierarchy_allows_preemption_when_in_allow_list`
- `preempt_exempt_time_protects_recently_started_job`

E2e tests added in `tests/native_host/e2e/test_preemption_qos_hierarchy.py`:
- `TestQosPreemptHierarchyBlocked` — with `preempt_type=qos_priority` and empty allow-list, preemption is blocked even with 1000× priority gap
- `TestQosPreemptHierarchyAllowed` — with `"low-allow"` in the allow-list, preemption fires and the low job is cancelled
- `TestPreemptExemptTime` — 20s exempt window: job survives inside it, preemption fires after it expires
- `TestReconfigurePreservesExemptTime` — runtime `scontrol update` exempt time survives `scontrol reconfigure` (regression for the reconfigure-wipe bug) — **passes** on `shrey-dev`

Note: `TestPreemptExemptTime` and the two accounting tests require a host where `spurd` can dispatch jobs (cgroup/namespace permissions); they could not be run on the available test host due to a pre-existing environment issue unrelated to this PR (same `JobLaunchFailure` on unmodified `test_preemption.py`).

## Docs

- `docs/admin-guide/configuration.rst`: `preempt_type` and `preempt_exempt_time` added to `[scheduler]` and `[[partitions]]` tables
- `docs/admin-guide/accounting.rst`: `preempt` and `preemptexempttime` added to QOS keys table; new "QOS preemption hierarchy" subsection with two-tier worked example

## Proto

New fields use new tag numbers (no renumbering). New WAL field uses `#[serde(default)]`. All DB columns added with `IF NOT EXISTS`. Backward-compatible.

### DIFF
--- a/crates/spur-cli/src/sacctmgr.rs
+++ b/crates/spur-cli/src/sacctmgr.rs
@@ -216,6 +216,20 @@ fn parse_i32(key: &str, val: &str) -> Result<i32> {
         .map_err(|_| anyhow::anyhow!("invalid value for {key}=: '{val}'"))
 }
 
+/// Parse an unsigned integer field, failing loudly rather than silently
+/// defaulting so a typo never quietly changes a limit on a partial `modify`.
+fn parse_u32(key: &str, val: &str) -> Result<u32> {
+    val.parse()
+        .map_err(|_| anyhow::anyhow!("invalid value for {key}=: '{val}'"))
+}
+
+/// Return true when a key=value pair explicitly opts in to a boolean action
+/// (value is "1", "yes", or "true", case-insensitive). Bare keys (no `=value`)
+/// never match because `parse_params` maps them to an empty string.
+fn is_truthy(val: &str) -> bool {
+    matches!(val.to_lowercase().as_str(), "1" | "yes" | "true")
+}
+
 /// Look up a field by its accepted aliases in priority order, returning the key
 /// the caller actually wrote alongside its value. Parsers use the returned key
 /// so an invalid-value error names the exact parameter the user passed (e.g.
@@ -488,7 +502,8 @@ async fn add(entity: &str, params: &[String], addr: &str) -> Result<()> {
                     grp_wall_minutes: Some(grp_wall),
                     preempt_exempt_time: p
                         .get("preemptexempttime")
-                        .and_then(|v| v.parse::<u32>().ok()),
+                        .map(|v| parse_u32("preemptexempttime", v))
+                        .transpose()?,
                     clear_preempt_exempt_time: false,
                 })
                 .await
@@ -655,8 +670,12 @@ fn build_modify_qos_request(
             .transpose()?,
         preempt_exempt_time: p
             .get("preemptexempttime")
-            .and_then(|v| v.parse::<u32>().ok()),
-        clear_preempt_exempt_time: p.contains_key("clearpreemptexempttime"),
+            .map(|v| parse_u32("preemptexempttime", v))
+            .transpose()?,
+        clear_preempt_exempt_time: p
+            .get("clearpreemptexempttime")
+            .map(|v| is_truthy(v))
+            .unwrap_or(false),
     })
 }
 
@@ -1004,7 +1023,7 @@ fn resolve_qos_field(q: &QosInfo, spec: char) -> String {
         'p' => q.priority.to_string(),
         'P' => q.preempt_mode.clone(),
         'Q' => q.preempt.clone(),
-        'E' => blank_if_zero(q.preempt_exempt_time),
+        'E' => q.preempt_exempt_time.map(blank_if_zero).unwrap_or_default(),
         'U' => format!("{}", q.usage_factor),
         'G' => q.grp_tres.clone(),
         'T' => q.max_tres_per_job.clone(),
@@ -1835,6 +1854,87 @@ mod tests {
         assert!(
             !header.contains("MaxJobs"),
             "default should omit MaxJobs: {header}"
+        );
+    }
+
+    // --- parse_u32 / is_truthy ---
+
+    #[test]
+    fn parse_u32_accepts_valid_integer() {
+        assert_eq!(parse_u32("preemptexempttime", "120").unwrap(), 120);
+        assert_eq!(parse_u32("preemptexempttime", "0").unwrap(), 0);
+    }
+
+    #[test]
+    fn parse_u32_rejects_non_integer() {
+        let err = parse_u32("preemptexempttime", "abc").unwrap_err();
+        assert!(
+            err.to_string().contains("preemptexempttime"),
+            "error names the field: {err}"
+        );
+        assert!(
+            err.to_string().contains("abc"),
+            "error names the bad value: {err}"
+        );
+    }
+
+    #[test]
+    fn parse_u32_rejects_negative() {
+        assert!(parse_u32("preemptexempttime", "-1").is_err());
+    }
+
+    #[test]
+    fn is_truthy_accepts_canonical_values() {
+        for v in &["1", "yes", "YES", "true", "True", "TRUE"] {
+            assert!(is_truthy(v), "{v} should be truthy");
+        }
+    }
+
+    #[test]
+    fn is_truthy_rejects_other_values() {
+        for v in &["0", "no", "false", "", "maybe"] {
+            assert!(!is_truthy(v), "{v} should not be truthy");
+        }
+    }
+
+    // --- clearpreemptexempttime must require explicit value ---
+
+    #[test]
+    fn clear_preempt_exempt_time_requires_truthy_value() {
+        // A bare key (parse_params maps it to "") must NOT trigger the clear.
+        let p: std::collections::HashMap<String, String> = [
+            ("name".into(), "q".into()),
+            ("clearpreemptexempttime".into(), "".into()),
+        ]
+        .into_iter()
+        .collect();
+        let req = build_modify_qos_request(&p).unwrap();
+        assert!(!req.clear_preempt_exempt_time, "bare key must not clear");
+    }
+
+    #[test]
+    fn clear_preempt_exempt_time_fires_on_explicit_one() {
+        let p: std::collections::HashMap<String, String> = [
+            ("name".into(), "q".into()),
+            ("clearpreemptexempttime".into(), "1".into()),
+        ]
+        .into_iter()
+        .collect();
+        let req = build_modify_qos_request(&p).unwrap();
+        assert!(req.clear_preempt_exempt_time);
+    }
+
+    #[test]
+    fn preempt_exempt_time_bad_value_errors() {
+        let p: std::collections::HashMap<String, String> = [
+            ("name".into(), "q".into()),
+            ("preemptexempttime".into(), "not-a-number".into()),
+        ]
+        .into_iter()
+        .collect();
+        assert!(
+            build_modify_qos_request(&p).is_err(),
+            "malformed preemptexempttime should error"
         );
     }
 }
--- a/crates/spur-cli/src/sacctmgr.rs
+++ b/crates/spur-cli/src/sacctmgr.rs
@@ -118,6 +118,7 @@ const QOS_KEYS: &[&str] = &[
     "preemptmode",
     "preempt",
     "preemptexempttime",
+    "clearpreemptexempttime",
     "usagefactor",
     "maxjobsperuser",
     "maxjobspu",
@@ -488,6 +489,7 @@ async fn add(entity: &str, params: &[String], addr: &str) -> Result<()> {
                     preempt_exempt_time: p
                         .get("preemptexempttime")
                         .and_then(|v| v.parse::<u32>().ok()),
+                    clear_preempt_exempt_time: false,
                 })
                 .await
                 .context("CreateQos RPC failed")?;
@@ -654,6 +656,7 @@ fn build_modify_qos_request(
         preempt_exempt_time: p
             .get("preemptexempttime")
             .and_then(|v| v.parse::<u32>().ok()),
+        clear_preempt_exempt_time: p.contains_key("clearpreemptexempttime"),
     })
 }
 
--- a/crates/spur-cli/src/sacctmgr.rs
+++ b/crates/spur-cli/src/sacctmgr.rs
@@ -116,6 +116,8 @@ const QOS_KEYS: &[&str] = &[
     "description",
     "priority",
     "preemptmode",
+    "preempt",
+    "preemptexempttime",
     "usagefactor",
     "maxjobsperuser",
     "maxjobspu",
@@ -469,6 +471,7 @@ async fn add(entity: &str, params: &[String], addr: &str) -> Result<()> {
                     description: Some(desc),
                     priority: Some(priority),
                     preempt_mode: Some(preempt.clone()),
+                    preempt: p.get("preempt").cloned(),
                     usage_factor: Some(usage_factor),
                     max_jobs_per_user: Some(max_jobs),
                     max_wall_minutes: Some(max_wall),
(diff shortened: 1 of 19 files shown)


yansun1996 (commented):
Nice work closing the QOS-hierarchy and exempt-time gaps — the resolution logic (`Option<u32>` precedence, multi-partition max, requeue/redispatch resetting the exempt window, reconfigure preserving runtime overrides) all checks out correctly against the code. A few issues below, mostly in the CLI/proto/docs layer around the two new fields, that I think are worth fixing before merge — the core scheduler logic in `try_preempt` looks solid.

Two small test-coverage notes:
- `crates/spur-core/src/wal.rs`: the new `PartitionUpdate.preempt_exempt_time` field has `#[serde(default)]` but no frozen old-payload deserialize test (unlike the existing `EvictTerminalJobs`/`K0sSetPhase` tests) — worth adding one so a future regression that drops the default is caught before it reaches a running cluster.
- The three `qos_preempt_allow_list_*` unit tests in `scheduler_loop.rs` just re-assert `Vec::contains` inline rather than calling `try_preempt`; real coverage comes from the `cluster.rs` integration tests, so these could be dropped without losing coverage (or left as-is, just flagging they're redundant).

crates/spur-cli/src/sacctmgr.rs
```diff
@@ -644,6 +653,10 @@ fn build_modify_qos_request(
             .get("grpwall")
             .map(|v| parse_wall_limit("grpwall", v))
             .transpose()?,
+        preempt_exempt_time: p
+            .get("preemptexempttime")
+            .and_then(|v| v.parse::<u32>().ok()),
+        clear_preempt_exempt_time: p.contains_key("clearpreemptexempttime"),
```
yansun1996: Two issues here:

1. `clear_preempt_exempt_time` only checks key *presence*, not its value — `clearpreemptexempttime=false` would still clear it. Worse, `parse_params` treats a bare trailing token (no `=value`) as a key expecting the *next* CLI arg as its value, so `sacctmgr modify qos name=x set clearpreemptexempttime` (bare, as the last token — which is how every doc example writes it) never gets inserted into the map at all, and the clear silently no-ops.
2. `preempt_exempt_time`'s `.and_then(|v| v.parse::<u32>().ok())` silently drops malformed input (e.g. `preemptexempttime=abc`) instead of erroring. This file already has a documented convention for exactly this situation — `parse_i32`/`parse_f64` return `Result` so "a typo never quietly changes a weight" — could this field use the same pattern?

Suggest requiring an explicit value for the clear flag (`clearpreemptexempttime=1`/`=yes`, matching how `scontrol`'s `--clear-preempt-exempt-time` equivalent is checked) and switching to a `parse_u32`-with-context helper for the exempt-time value.

crates/spurctld/src/accounting/grpc.rs
```diff
@@ -548,6 +548,7 @@ impl SlurmAccounting for AccountingService {
             description: req.description.as_deref(),
             priority: req.priority,
             preempt_mode: req.preempt_mode.as_deref(),
+            preempt: req.preempt.as_deref(),
```
yansun1996: The `preempt` allow-list is written through with no validation that the listed QOS names exist — unlike `allowed_qos` in `add_user` just above, which validates via `db::missing_qos` before writing. A typo here becomes silent dead config, and if a QOS with that exact name is created later, it would retroactively inherit preempt rights nobody intended to grant. Worth reusing `db::missing_qos` the same way?

proto/slurm.proto
```diff
@@ -1388,6 +1400,9 @@ message QosInfo {
   string max_tres_per_user = 10;
   string grp_tres = 11;
   uint32 grp_wall_minutes = 12;
+  // Comma-separated QOS names this QOS may preempt (empty = none allowed).
+  string preempt = 13;
+  uint32 preempt_exempt_time = 14;  // 0 = defer to partition/global default
```
yansun1996: This field (and `PartitionInfo.preempt_exempt_time`) is a plain `uint32`, while the corresponding request messages correctly use `optional uint32`. That means the server's `.unwrap_or(0)` collapses "explicitly set to 0" and "never set" into the same wire value, and `sacctmgr`/`scontrol`'s display logic (`blank_if_zero`, `if ... > 0`) hides zero either way. An admin who explicitly sets `preemptexempttime=0` to force "immediately preemptable" can't distinguish that from "inherited" via the CLI — which is the exact ambiguity this PR is meant to resolve. Since there's still room on both messages, would it be worth making these `optional uint32` too, so presence round-trips end to end?

docs/admin-guide/accounting.rst
```diff
@@ -374,6 +374,22 @@ QOS keys
...
        lower-``priority`` QOS when its effective priority exceeds twice the
        victim's.
+   * - ``preempt``
+     - ``""`` (no restriction)
+     - Comma-separated list of QOS names that jobs in this QOS are allowed to
+       preempt. Only enforced when ``scheduler.preempt_type = "qos_priority"``
+       (see :doc:`configuration`). An empty value means this QOS may not preempt
+       any other QOS under that mode. Example: ``preempt=low,batch`` allows jobs
+       in this QOS to preempt ``low`` and ``batch`` jobs. To clear the list at
+       runtime: ``sacctmgr modify qos name=<name> set clearpreemptexempttime``.
```
yansun1996: Two things about this example:
- It's used to describe clearing the `preempt` allow-list here, but `clearpreemptexempttime` only affects `preempt_exempt_time` — clearing `preempt` is `sacctmgr modify qos name=<name> set preempt=` (empty value).
- As written (bare, no `=value`), this hits the parser gap noted on `sacctmgr.rs` and silently does nothing — could you add `=1` (or whatever value ends up being required after that fix)?

Same issue recurs a few lines down at 392, and again in the "Minimum exempt time" example below.

docs/admin-guide/accounting.rst
```diff
@@ -374,6 +374,22 @@ QOS keys
...
+   * - ``preemptexempttime``
+     - ``0`` (inherit partition / global)
```
yansun1996: The underlying field is `Option<u32>` defaulting to `None`, and `0` is itself a distinct, meaningful, settable value ("immediately preemptable, no exemption") — not the same as "unset." Could this say "unset (inherits partition/global)" instead, to match how the partition-level row in `configuration.rst` correctly describes its default as `null`?

docs/admin-guide/accounting.rst
```diff
@@ -399,6 +415,70 @@ QOS keys
...
+.. code-block:: bash
+
+   # Global: no job can be preempted in its first 5 minutes.
+   # (Set in spur.conf: scheduler.preempt_exempt_time = 300)
+
+   # Per-QOS: burst jobs are protected for 2 minutes regardless of global.
+   sacctmgr modify qos name=burst set preemptexempttime=120
+
+   # Per-partition via scontrol (runtime, no restart needed):
+   scontrol update-partition PartitionName=gpu PreemptExemptTime=600
```
yansun1996: `update-partition` only accepts explicit `--name=... --preempt-exempt-time=...` flags — it doesn't parse inline `Key=Value` pairs. That syntax belongs to the separate generic `scontrol update <EntityName>=... key=value` command (used elsewhere in this doc, e.g. `scontrol update nodename=gpu001 state=drain`). Should this be `scontrol update PartitionName=gpu PreemptExemptTime=600` instead? Same fix would apply to the equivalent line in `configuration.rst`.

tests/native_host/e2e/test_preemption_qos_hierarchy.py
```diff
@@ -0,0 +1,306 @@
...
+            f"preempt_exempt_time not set after scontrol update: {out}"
+        )
+
+        # Trigger a reconfigure.
+        c.scontrol("reconfigure")
+        time.sleep(3)  # let the controller apply the reload
+
+        # The override must survive: it was set at runtime, not from TOML.
+        # In Spur, a live scontrol update writes a WAL entry; reconfigure
+        # sends its own WAL update from the TOML values. After this fix,
```
yansun1996: Nit: this narrates the fix/review history rather than the invariant itself — could this collapse to something like `# reconfigure must not clobber a runtime-set override when TOML doesn't specify the field`?

crates/spur-cli/src/sacctmgr.rs
sajmera-pensando: Addressed — added  helper (errors on malformed input) and  (requires explicit //). Bare key now silently no-ops as documented. Both paths (add and modify) updated.

crates/spurctld/src/accounting/grpc.rs
sajmera-pensando: Addressed — added validation via `db::missing_qos` before writing, matching the pattern used for `allowed_qos` on `add_user`. A typo in the allow-list now returns a 404 instead of becoming silent dead config.

proto/slurm.proto
sajmera-pensando: Addressed — both `QosInfo.preempt_exempt_time` and `PartitionInfo.preempt_exempt_time` changed to `optional uint32`. Absence now means "unset/inherited" and 0 means "explicitly no exemption". Updated server-side population and CLI display accordingly.

docs/admin-guide/accounting.rst
sajmera-pensando: Addressed — fixed both instances. The clear command for the preempt allow-list is now `preempt=` (empty value); `clearpreemptexempttime=1` is correct only for clearing the exempt time. Also added `=1` to the bare key example.

docs/admin-guide/accounting.rst
sajmera-pensando: Addressed — default changed to "unset (inherits partition/global)" to match how the partition-level row in `configuration.rst` describes it as `null`.

docs/admin-guide/accounting.rst
sajmera-pensando: Addressed — corrected to `scontrol update PartitionName=gpu PreemptExemptTime=600` (inline Slurm key=value syntax). Same fix applied in both `accounting.rst` and `configuration.rst`.

tests/native_host/e2e/test_preemption_qos_hierarchy.py
sajmera-pensando: Addressed — collapsed to: `# reconfigure must not clobber a runtime-set override when the TOML partition entry does not specify preempt_exempt_time.`

crates/spur-cli/src/sacctmgr.rs
sajmera-pensando: Addressed — added `parse_u32()` helper (errors on malformed input) and `is_truthy()` (requires explicit `=1`/`=yes`/`=true`). Bare key now silently no-ops as documented. Both paths (add and modify) updated.

crates/spurctld/src/accounting/grpc.rs
sajmera-pensando: Addressed — added validation via `db::missing_qos` before writing, matching the pattern used for `allowed_qos` on `add_user`. A typo in the allow-list now returns a 404 instead of becoming silent dead config.

docs/admin-guide/accounting.rst
sajmera-pensando: Addressed — fixed both instances. The clear command for the preempt allow-list is now `preempt=` (empty value); `clearpreemptexempttime=1` is only for clearing the exempt time. Also added `=1` to the bare key example.

docs/admin-guide/accounting.rst
sajmera-pensando: Addressed — corrected to `scontrol update PartitionName=gpu PreemptExemptTime=600` (inline Slurm key=value syntax). Same fix applied in both `accounting.rst` and `configuration.rst`.

tests/native_host/e2e/test_preemption_qos_hierarchy.py
sajmera-pensando: Addressed — collapsed to a single line stating the invariant: reconfigure must not clobber a runtime-set override when the TOML partition entry does not specify the field.

proto/slurm.proto
sajmera-pensando: Addressed — both `QosInfo.preempt_exempt_time` and `PartitionInfo.preempt_exempt_time` changed to `optional uint32`. Absence now means "unset/inherited" and 0 means "explicitly no exemption". Updated server-side population and CLI display accordingly.

docs/admin-guide/accounting.rst
sajmera-pensando: Addressed — default changed to "unset (inherits partition/global)" to match how the partition-level row in `configuration.rst` describes it as `null`.

yansun1996 (approved):
LGTM

PR #685, opened 2026-08-19
fix(security): gate control-plane mutation RPCs on verified admin

powderluv:
Hardening: the RPCs that define cluster tenancy — partitions, node config, reservations, admission tokens, and the accounting service — performed no authorization. `Identity::require_admin` existed but had zero callers, so any reachable caller could rewrite the partition/account tables or self-promote via the accounting API.

## Change

- New `ControllerService::require_admin(&request, op)`: a deny-by-default admin prologue built on the existing identity spine (`verified_identity` from request extensions + `caller_is_admin`, i.e. token admin claim OR accounting Admin level). Returns `PermissionDenied` when there is no verified admin identity.
- Gates 11 controller mutation RPCs (after leader-forward, before `into_inner`): `create/update/delete_partition`, `reconfigure`, `update_node`, `create_token`, `list_tokens`, `revoke_token`, `create/update/delete_reservation`.
- New `accounting::grpc::require_admin` — the first production caller of `Identity::require_admin` — gating `create_account`, `delete_account`, `add_user`, `remove_user`, `create_qos`, `delete_qos`. Read-only `list_*`/`get_*` and the internal `record_job_*` telemetry stay open.
- Fixes `create_token`/`revoke_token` to forward to the leader via `forward_request` (preserves the caller's credential under HA; their previous manual forward dropped it).

## Tests

Deny-by-default table over all 11 gated controller RPCs (verified non-admin AND anonymous both rejected), an admin-passes test that actually creates the partition, and the accounting-mutation gate (admin allowed, non-admin/anonymous rejected — closes the `add_user` self-promotion path).

## Behaviour change / deferred

- This is a deliberate behaviour change: a caller with no verified admin identity (permissive-with-no-credential, or disabled mode) is now refused where it previously succeeded.
- `drain_node`/`deregister_node` left ungated pending a check of the node-self-service path (they are DoS, not tenancy-repooling — `update_node`'s label path, the actual re-pooling vector, IS gated).
- The accounting gate uses the token admin claim only (the accounting service holds no association-cache handle); the controller gate uses the richer `caller_is_admin`.

Verified on Linux CI. Part of the multi-tenancy authorization hardening (epic #665).

### DIFF
--- a/crates/spurctld/src/server.rs
+++ b/crates/spurctld/src/server.rs
@@ -6976,6 +6976,16 @@ mod tests {
                 .any(|p| p.name == "team-a"),
             "the admin's partition must exist after the gated call succeeds"
         );
+
+        // A structurally different RPC (token issuance) also passes the admin gate.
+        let token_resp = svc
+            .create_token(admin_request(CreateTokenRequest::default()))
+            .await
+            .expect("an admin caller may create a token");
+        assert!(
+            !token_resp.into_inner().token_id.is_empty(),
+            "the admin's token must be issued"
+        );
     }
 
     // --- authoritative_user ---
--- a/crates/spurctld/src/server.rs
+++ b/crates/spurctld/src/server.rs
@@ -1559,14 +1559,9 @@ impl SlurmController for ControllerService {
             })
             .unwrap_or_default();
 
-        // A node re-registering with different labels reaches the same NodeLabelsUpdate WAL op as
-        // the gated `update_node`; require the same admin bar there, but let first registration and
-        // unchanged re-registration through, so a node agent can self-register without one.
-        if let Some(existing) = self.cluster.get_node(&request.get_ref().hostname) {
-            if existing.labels != request.get_ref().labels {
-                self.require_admin(&request, "relabel an existing node on re-registration")?;
-            }
-        }
+        // Decided here, enforced atomically inside `register_node` against the label diff — a
+        // pre-check here instead would race a concurrent registration for the same node.
+        let caller_privileged = self.caller_is_privileged(Self::verified_identity(&request));
 
         let req = request.into_inner();
         let resources = req.resources.map(proto_to_resource_set).unwrap_or_default();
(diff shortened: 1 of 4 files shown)


crates/spurctld/src/server.rs
```diff
@@ -6687,6 +6741,131 @@ mod tests {
...
+            delete_reservation,
+            DeleteReservationRequest {
+                name: "r".into(),
+                ..Default::default()
+            }
+        );
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn control_plane_mutation_allows_admin() {
```
yansun1996, line 6967: Consider extending this to cover an admin-success case for at least one more RPC (a token or reservation call would be structurally different from partitions) — right now only `create_partition`'s admin-passes path is positively tested; the rest infer it from the anonymous branch.

yansun1996 (commented):
Nice, focused hardening — the gate placement (post-leader-forward, pre-`into_inner`) is consistent across all 17 sites, and the credential-preserving forward fix for `create_token`/`revoke_token` closes a real gap. Two suggestions before merge:

1. `control_plane_mutation_allows_admin` only exercises the admin-success path for `create_partition`. Since the other 9 gated RPCs currently rely only on the anonymous-not-rejected assertion as an implicit proxy, worth adding at least one more admin-success case (e.g. a token or reservation RPC) so a regression that broke only the "admin passes" branch would get caught.
2. This changes previously-succeeding behavior (an authenticated non-admin calling `create_partition`/`create_reservation`/`create_token`/etc. now gets `PermissionDenied`). Would it be worth a short note in `docs/admin-guide/configuration.rst`? There's already a precedent section for this shape of admin-exemption writeup (the `max_user_priority` ceiling).

Nits (non-blocking): the two `require_admin` doc comments run 10+ lines — could trim to the essential "why" and let the rest live in the PR description. Also `self.pool()?` runs before `require_admin` in the six accounting handlers; harmless today (pool-unavailable is already visible via the ungated `list_accounts`), but worth flipping the order for consistency with the controller side, which always authorizes before touching state.

sajmera-pensando (approved):
LGTM

PR #697, opened 2026-08-19
fix(spur-cli): stop sprio and sshare panicking on every invocation

### LINKED ISSUE #696: sprio and sshare panic on every invocation: -h is declared twice
## What happens

`sprio` and `sshare` panic before doing any work, on every invocation, including with no arguments at all:

```
$ sprio
thread 'main' panicked at clap_builder-4.6.6/src/builder/debug_asserts.rs:125:17:
Command sprio: Short option names must be unique for each argument, but '-h' is in
use by both 'noheader' and 'help' (call `cmd.disable_help_flag(true)` to remove the
auto-generated `--help`)

$ sshare
thread 'main' panicked at clap_builder-4.6.6/src/builder/debug_asserts.rs:125:17:
Command sshare: Short option names must be unique for each argument, but '-h' is in
use by both 'noheader' and 'help' ...
```

`sprio -h` and `sshare -h` panic the same way. Neither command can be run, and because it aborts rather than exiting with a usage error, a calling script sees a crash rather than a message it can act on.

## What I expected

Both commands print their table, and `-h` suppresses the header as the flag is declared to do.

## Where it comes from

Both declare `-h`/`--noheader` while leaving clap's generated `--help` in place, which also claims `-h`. `squeue` and `sinfo` declare the same flag but turn the generated help off, so they are unaffected.

Two things kept this hidden. The conflict is an argument-definition check that fires when the parser is built, before any argument is looked at, so it does not depend on what the user typed. And neither command has a test that constructs its parser, so nothing in the test suite ever builds these two.

The check itself only runs in debug builds. A release binary will not panic, but `-h` is still declared twice, so it cannot be relied on to mean `--noheader` there.

## Impact

`sprio` and `sshare` are unusable in any debug build. Every other CLI command in the same binary works, so the binary looks healthy until one of these two is called.

## Environment

Spur 0.8.0, debug build, single-node local cluster.

nikhilsk:
Fixes #696.

`sprio` and `sshare` could not be run at all. Both declared Slurm's `-h`/`--noheader` while leaving clap's generated `--help` in place, which also claims `-h`. clap rejects a duplicate short option when the parser is *constructed*, so both aborted before looking at a single argument, with or without one:

```
$ sprio
thread 'main' panicked at clap_builder-4.6.6/src/builder/debug_asserts.rs:125:17:
Command sprio: Short option names must be unique for each argument, but '-h' is in
use by both 'noheader' and 'help'
```

`spur priority` and `spur share` hit the same two parsers, so the panic was reachable four ways. Because it aborts rather than exiting with a usage error, a calling script saw a crash rather than a message it could act on.

## Approach

`squeue` and `sinfo` already declare the same Slurm flag and already solve this: turn the generated help off and re-add `--help` as long-only. This does the same on both commands, so the four spellings behave like the rest of the CLI. `-h` suppresses the header as declared, `--help` still prints help.

The more interesting question was why it shipped. clap's check fires at parser construction, and nothing in the tree ever constructed these two parsers: neither module had a test, and every command owns its own parser, so there was no single place where a malformed definition would be caught. Two commands were unusable in every debug build without a test noticing.

So the second commit walks every parser in the binary in one test. That is a class fix rather than a fix for these two: any command that claims a short option twice now fails in CI rather than at a user's terminal. The list is maintained by hand because a Slurm-named command's parser is not reachable from a single root type, and a comment says a new command must be added to it.

## Trade-offs

The exhaustive test's list can go stale for a *newly added* command, which is the same weakness as the hand-maintained tables elsewhere in the tree. It is still strictly better than the status quo, where nothing checks any of them. Deriving the list would mean restructuring dispatch, which is not worth doing for this.

Nothing else changed. `--help` on every command in this binary prints its help prefixed with `Error:` and exits 1, `squeue` and `sacct` included. That is pre-existing and repo-wide, so it is not touched here.

Docs need no change: both commands are already listed as supported, which is precisely what the panic contradicted.

## Testing

- Verified all four invocations against a live single-node cluster, before and after. `sprio`, `sshare`, `spur priority` and `spur share` now print their tables; `sshare -h` prints no header; `sprio --help` prints help.
- Mutation-checked the tests rather than trusting them: removing `disable_help_flag` from `sprio` turns all three red (`short_h_means_noheader`, `help_is_still_reachable_by_its_long_name`, `every_command_parser_is_well_formed`), each on the duplicate-short-option assertion.
- `cargo fmt --check`, `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` clean, and `cargo test --locked` green.

### DIFF
--- a/crates/spur-cli/src/main.rs
+++ b/crates/spur-cli/src/main.rs
@@ -330,3 +330,40 @@ fn print_usage() {
     eprintln!("  salloc sbatch srun squeue scancel sinfo sacct sacctmgr scontrol");
     eprintln!("  sprio sshare sstat sdiag sreport strigger sattach scrontab smd");
 }
+
+#[cfg(test)]
+mod tests {
+    use clap::CommandFactory;
+
+    // Every command owns its own parser, so nothing builds them all until a user runs
+    // one. A malformed definition (a short option claimed twice, say) is caught only
+    // when its parser is constructed, so each one is listed here by hand and a new
+    // command must be added alongside it.
+    #[test]
+    fn every_command_parser_is_well_formed() {
+        crate::exec::ExecArgs::command().debug_assert();
+        crate::image::ImageArgs::command().debug_assert();
+        crate::k8s::K8sArgs::command().debug_assert();
+        crate::net::NetArgs::command().debug_assert();
+        crate::node::NodeArgs::command().debug_assert();
+        crate::sacct::SacctArgs::command().debug_assert();
+        crate::sacctmgr::SacctmgrArgs::command().debug_assert();
+        crate::salloc::SallocArgs::command().debug_assert();
+        crate::sattach::SattachArgs::command().debug_assert();
+        crate::sbatch::SbatchArgs::command().debug_assert();
+        crate::scancel::ScancelArgs::command().debug_assert();
+        crate::scontrol::ScontrolArgs::command().debug_assert();
+        crate::scrontab::ScrontabArgs::command().debug_assert();
+        crate::sdiag::SdiagArgs::command().debug_assert();
+        crate::sinfo::SinfoArgs::command().debug_assert();
+        crate::smd::SmdArgs::command().debug_assert();
+        crate::sprio::SprioArgs::command().debug_assert();
+        crate::squeue::SqueueArgs::command().debug_assert();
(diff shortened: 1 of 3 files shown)


biluriuday (approved):
LGTM

shiv-tyagi (approved):
LGTM

PR #698, opened 2026-08-19
feat(spurctld): record reservation admin actions in accounting txn log

biluriuday:
### Summary
`scontrol` reservation create/update/delete previously left **no record of who ran the command** — the accounting DB only tracked jobs. This adds an audit trail: a general-purpose `txn` (transaction) table records reservation admin actions — including **denied and failed attempts** — with the acting user, outcome, and requested parameters, and exposes them via a new `GetTransactions` gRPC RPC and a Slurm-compatible `sacctmgr show txn`.

This closes a gap that stock Slurm also has: its `txn_table` doesn't cover `scontrol` reservations, and its reservation records carry no actor.

### What's included
- **Audit `txn` table** in the accounting schema — entity-agnostic (reservations today; extensible to other admin ops), storing actor, actor_uid, `verified`, `source`, action, entity, outcome, and a JSON `details` payload.
- **Write path:** best-effort, leader-side audit fired from the reservation handlers and the expired-reservation purge (`source=system`), mirroring the existing job-accounting notifier (async + bounded retry).
- **Read path:** `GetTransactions` RPC + `sacctmgr show txn|transaction` with Slurm-aligned filters (`Actor=`/`Action=`/`Start=`/`End=`, plus `Entity=`/`Outcome=`) and columns (default `Time,Action,Actor,Where,Info`; extras `Outcome,Verified,Source,ID,ActorUID`).
- **Shared CLI time-arg util** — de-duplicates `parse_time_arg`/`datetime_to_proto` across `sacct`, `sreport`, and the new command.
- **Optional retention:** `accounting.txn_retention_days` (leader-only periodic purge; unset = keep forever, matching Slurm).
- Docs (`accounting.rst`, `configuration.rst`) and unit + e2e tests.

### Design choices / trade-offs
- **Correct-by-construction attribution:** the audit fires strictly *after* the operation resolves and records the *authoritative* actor (JWT subject when verified, else the client-asserted name with `verified=no`). A `success` row can only follow a committed Raft op; `denied`/`error` imply no state change.
- **Best-effort, non-transactional:** a DB outage never blocks or fails a reservation command; the only possible skew is a committed reservation with a missing audit row (logged as an error), never a false success.
- **Non-breaking / safe rolling upgrade + rollback:** no persisted Raft/WAL type changes (WAL replay writes zero audit rows); additive proto (new tags only), additive schema (`CREATE ... IF NOT EXISTS`), one optional config field, additive CLI.
- **Read path is ungated**, consistent with `get_job_history` and the rest of the accounting service; audit-read confidentiality therefore requires `auth.mode = required` (documented).

### Sample output
Produced by (on a 3-node bare-metal cluster, accounting enabled):
```bash
sudo scontrol create-reservation --name=maint-2026 --start-time=now --duration=120 --nodes=spur-node1 --flags=ignore_jobs --users=alice,bob
sudo scontrol create-reservation --name=maint-2026 ...        # duplicate -> error
sudo scontrol update-reservation --name=maint-2026 --duration=240
scontrol delete-reservation maint-2026                        # run as 'vm' (non-owner) -> denied
sudo scontrol delete-reservation maint-2026
```

Default (Slurm-compatible columns):
```
$ sacctmgr show txn
Time                 Action   Actor          Where                    Info
----                 ------   -----          -----                    ----
2026-08-19T11:04:39  delete   root           reservation:maint-2026   {}
2026-08-19T11:04:39  delete   vm             reservation:maint-2026   {"error":"user 'vm' cannot delete reservation 'maint-2026' owned by 'root'"}
2026-08-19T11:04:39  update   root           reservation:maint-2026   {"add_accounts":[],"add_nodes":[],"add_users":[],"duration_minutes":240,"remove_accounts":[],"remove_nodes":[],"remove_users":[]}
2026-08-19T11:04:39  create   root           reservation:maint-2026   {"accounts":[],"duration_minutes":120,"error":"reservation 'maint-2026' already exists","flags":["ignore_jobs"],"nodes":["spur-node1"],"start_time":"now","users":[]}
2026-08-19T11:04:38  create   root           reservation:maint-2026   {"accounts":[],"duration_minutes":120,"flags":["ignore_jobs"],"nodes":["spur-node1"],"start_time":"now","users":["alice","bob"]}
```

Full columns (Spur extensions `Outcome`, `Verified`, `Source`, `ID`):
```
$ sacctmgr show txn format=ID,Time,Actor,Action,Where,Outcome,Verified,Source,Info
ID  Time                 Actor    Action   Where                    Outcome  Verified  Source  Info
--  ----                 -----    ------   -----                    -------  --------  ------  ----
5   2026-08-19T11:04:39  root     delete   reservation:maint-2026   success  no        api     {}
4   2026-08-19T11:04:39  vm       delete   reservation:maint-2026   denied   no        api     {"error":"user 'vm' cannot delete reservation 'maint-2026' owned by 'root'"}
3   2026-08-19T11:04:39  root     update   reservation:maint-2026   success  no        api     {...deltas...}
2   2026-08-19T11:04:39  root     create   reservation:maint-2026   error    no        api     {"error":"reservation 'maint-2026' already exists",...}
1   2026-08-19T11:04:38  root     create   reservation:maint-2026   success  no        api     {"nodes":["spur-node1"],"users":["alice","bob"],...}
```

Filtering (`Actor=`, `Action=`, `Entity=`, `Outcome=`, `Start=`, `End=`):
```
$ sacctmgr show txn Actor=root Action=create Outcome=success
```

(`Verified=no` here because the cluster runs the default `auth.mode = permissive` with no tokens — actors are trust-on-wire; under `required` with JWTs it reads `yes`.)

### Testing
- `cargo test --locked` — green. Unit tests cover the attribution mapping (`verified`/`actor_uid`, incl. a UID > 2³¹), outcome/details serialization, timestamp validation, the query-limit cap, time-arg parsing (incl. `now--7days` rejection), and CLI filter parsing. DB round-trip tests are `#[ignore]`-gated (no external services in the default run).
- Verified end-to-end on the 3-node bare-metal cluster (Postgres-backed): create / update / delete plus duplicate-create (`error`) and non-owner delete (`denied`) all produce the correct rows (output above). e2e assertions parse rows into columns and match exact values.

### Follow-ups (out of scope)
- **Security (pre-existing):** the whole `SlurmAccounting` gRPC service is unauthenticated — including `add_user`, which sets `admin_level`. Under default `permissive`/`disabled` mode an anonymous caller can self-grant accounting Admin. The coherent fix is service-wide authz; until then, use `auth.mode = required`. This also bounds who can read the audit log.
- Keyset (`ts,id`) **pagination** for large audit history.
- Live-reloadable retention; document that purge is a hard delete (no archive).
- Extend auditing to `sacctmgr` account/user/QoS CRUD and node ops (the table is already entity-agnostic).

### DIFF
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -490,6 +490,10 @@ pub struct AccountingConfig {
     /// chain. Mirrors Slurm's `AccountingStorageEnforce=qos`. Default false.
     #[serde(default)]
     pub require_qos: bool,
+    /// Delete `txn` audit-log rows older than this many days. `None` (default)
+    /// keeps them forever, matching Slurm's default-off purge behavior.
+    #[serde(default)]
+    pub txn_retention_days: Option<u32>,
 }
 
 fn default_fairshare_refresh_secs() -> u32 {
@@ -508,6 +512,7 @@ impl Default for AccountingConfig {
             grp_wall_window_days: default_grp_wall_window_days(),
             default_qos: String::new(),
             require_qos: false,
+            txn_retention_days: None,
         }
     }
 }
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -491,7 +491,8 @@ pub struct AccountingConfig {
     #[serde(default)]
     pub require_qos: bool,
     /// Delete `txn` audit-log rows older than this many days. `None` (default)
-    /// keeps them forever, matching Slurm's default-off purge behavior.
+    /// or `0` disables purging (rows kept forever, matching Slurm's default-off
+    /// behavior); a positive value enables the periodic purge.
     #[serde(default)]
     pub txn_retention_days: Option<u32>,
 }
--- a/.wordlist.txt
+++ b/.wordlist.txt
@@ -8,7 +8,9 @@ Cargo
 config
 cron
 crontab
+cryptographically
 daemonize
+datetime
 dbd
 deprovision
 dev
--- a/crates/spur-cli/src/main.rs
+++ b/crates/spur-cli/src/main.rs
@@ -34,6 +34,7 @@ mod srun;
 mod sshare;
 mod sstat;
 mod strigger;
+mod timearg;
 mod token;
 
 use std::path::Path;
--- a/crates/spur-cli/src/sacct.rs
+++ b/crates/spur-cli/src/sacct.rs
@@ -7,6 +7,7 @@ use spur_proto::proto::GetJobHistoryRequest;
 
 use crate::exit_fmt::format_exit;
 use crate::format_engine;
+use crate::timearg::{datetime_to_proto, parse_time_arg};
 
 /// Display accounting data for jobs.
 #[derive(Parser, Debug)]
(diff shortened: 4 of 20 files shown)


crates/spur-core/src/config.rs
```diff
@@ -490,6 +490,10 @@ pub struct AccountingConfig {
     /// chain. Mirrors Slurm's `AccountingStorageEnforce=qos`. Default false.
     #[serde(default)]
     pub require_qos: bool,
+    /// Delete `txn` audit-log rows older than this many days. `None` (default)
+    /// keeps them forever, matching Slurm's default-off purge behavior.
+    #[serde(default)]
+    pub txn_retention_days: Option<u32>,
```
shiv-tyagi, line 497: Q: what happens when `txn_retention_days` is set to `0`? `main.rs` treats it the same as unset (`.filter(|d| *d > 0)`), so the purge loop never starts. That's a reasonable choice (matches Slurm's off-by-default), but worth a one-line note here since `0` reads as "purge everything older than 0 days" to most admins, not "disabled".

shiv-tyagi (dismissed):
Nice audit trail addition, closes a real gap versus stock Slurm (its txn_table doesn't cover scontrol reservations, and reservation records carry no actor). Clippy clean, tests pass, no proto/WAL compat issues. One small doc suggestion inline, non-blocking.

crates/spur-core/src/config.rs
biluriuday: when it is set to 0, it is treated as unset and the purging does not start. I have re-worded the docs to clarify the behavior.

shiv-tyagi (approved):

PR #699, opened 2026-08-19
fix(spurctld): adopt each node's real WireGuard address as its k0s mesh IP

spraveenio:
## What this fixes

`spur k8s up` over a live WireGuard mesh could silently corrupt the mesh whenever the chosen control-plane set was not an alphabetical prefix of the order in which nodes physically joined.

`provision_assignments` allocated each node's `k0s_mesh_ip` from a fresh, independent address pool and never consulted the real `spur0` address the node already advertises via `Node.address`. When the CP set was a non-prefix subset (or join order was scrambled), the assigned mesh IPs drifted off the real addresses. `ApplyMesh` then pushed `AllowedIPs` built from that wrong table, and because a WireGuard `AllowedIPs` entry is exclusive per peer per interface, this reassigned real IPs away from their rightful peers — breaking every tunnel, including nodes that were never being reconfigured. No error was surfaced; `spur k8s status` just showed a generic `degraded`.

## Approach

Adopt each node's real mesh address instead of re-deriving it:

- `Node.address` **within** `mesh_cidr` → it *is* the mesh IP → use it as `k0s_mesh_ip`.
- `Node.address` **outside** `mesh_cidr` (e.g. a control plane advertising its routable underlay address) → reports no mesh address → falls through to pool allocation (bootstrap gets `.1`).
- Two nodes claiming the same **in-mesh** address → genuine conflict → fail loud with a per-node "network mismatch" reason.

The bootstrap is treated the same as any other node — `.1` results naturally when the operator gave the controller `.1` via `spur net init`, and etcd seeding / CP ordering key off the node **name**, not the IP, so a divergent join order now works instead of corrupting the mesh.

This aligns Spur with how both Slurm (static `NodeAddr`) and Kubernetes (kubelet adopts the host's `--node-ip`) treat a node's own transport address: adopted from a single source of truth, never re-derived by a second component. The pod-CIDR plane already follows this model correctly.

## Compatibility

Reads an existing field (`Node.address`) and writes the same `k0s_mesh_ip` field already persisted — **no proto, config, or persisted-state change**, no Raft/WAL break.

## Testing

- Unit tests in `cluster_k8s.rs` (exercise the real `provision_assignments` via a `ClusterManager` fixture): adversarial (non-prefix) CP selection, scrambled join order, underlay-advertising node pools (not refused), conflicting in-mesh addresses fail loud, idempotency preserved, pool fallback for unmeshed nodes. All green.
- Full `spurctld` suite passes; `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` and `cargo fmt --check` clean.
- **Hardware validated before/after** on a 2-node mesh testbed with two binaries built from the same tree differing only by this patch. The worker joined at a real address the pool would not pick (real `.3`; pool would give `.2`): pre-fix assigned `.2` (worker stuck, assigned IP unreachable); post-fix adopted `.3` (both nodes active, cluster `ready`, mesh IP reachable). The full-fidelity adversarial 3+-CP-over-mesh case (the original incident) needs ≥3 mesh nodes and is covered by the unit tests.

fixes SPUR-196

### DIFF
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -15656,33 +15656,77 @@ mod tests {
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn provision_refuses_conflicting_real_addresses() {
-        // Two meshed nodes claiming the same real address is a genuine operator mistake: assigning
-        // both would push an exclusive AllowedIPs entry for two peers. Refuse with a reason.
+        // Two unassigned meshed nodes advertising the same real address is a genuine operator mistake:
+        // assigning either would push an exclusive AllowedIPs entry for an address two peers claim. Both
+        // stay unprovisioned and both carry the reason, but the call still succeeds so any other node in
+        // the same tick provisions normally.
         let dir = TempDir::new().unwrap();
         let cm = test_cluster(&dir).await;
         register_meshed_node(&cm, "n005", "10.44.0.7");
         register_meshed_node(&cm, "n006", "10.44.0.7"); // duplicate
+        register_meshed_node(&cm, "n007", "10.44.0.9"); // healthy bystander
         record_cp_set(&cm, "n005", &["n005"]);
 
-        let err = crate::cluster_k8s::provision_assignments(
-            &cm,
-            &provision_net_calico(),
-            &cm.k0s_state(),
-        )
-        .unwrap_err();
-        assert!(
-            format!("{err:#}").to_lowercase().contains("mismatch")
-                || format!("{err:#}").to_lowercase().contains("conflict"),
-            "duplicate real address must fail loud: {err:#}"
+        crate::cluster_k8s::provision_assignments(&cm, &provision_net_calico(), &cm.k0s_state())
+            .unwrap();
+
+        // The healthy node adopts its real address and provisions despite the conflicting pair.
+        wait_for("healthy bystander provisions", || {
+            cm.get_node("n007").and_then(|n| n.k0s_role).is_some()
+        });
+        assert_eq!(
+            cm.get_node("n007").unwrap().k0s_mesh_ip.as_deref(),
+            Some("10.44.0.9")
         );
-        // The reason is surfaced on BOTH sides of the conflict so `spur k8s status` points at each.
-        wait_for("both nodes carry the conflict reason", || {
-            let a = cm.get_node("n005").and_then(|n| n.k0s_last_error);
-            let b = cm.get_node("n006").and_then(|n| n.k0s_last_error);
-            a.map(|e| e.to_lowercase().contains("mismatch"))
-                .unwrap_or(false)
-                && b.map(|e| e.to_lowercase().contains("mismatch"))
-                    .unwrap_or(false)
+
+        // Neither conflicting node is assigned, and both name each other in an exact-shape reason so
+        // `spur k8s status` points at the whole conflict.
+        assert!(cm.get_node("n005").and_then(|n| n.k0s_role).is_none());
+        assert!(cm.get_node("n006").and_then(|n| n.k0s_role).is_none());
+        let want = "network mismatch: mesh address 10.44.0.7 claimed by n005, n006";
+        wait_for("both conflicting nodes carry the exact reason", || {
+            cm.get_node("n005")
+                .and_then(|n| n.k0s_last_error)
+                .as_deref()
+                == Some(want)
+                && cm
+                    .get_node("n006")
+                    .and_then(|n| n.k0s_last_error)
+                    .as_deref()
+                    == Some(want)
+        });
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn provision_refuses_newcomer_colliding_with_assigned_node() {
+        // An already-assigned member owns its mesh IP. A newcomer advertising that same address is the
+        // one refused — the incumbent keeps its assignment, the newcomer stays out with a reason.
+        let dir = TempDir::new().unwrap();
+        let cm = test_cluster(&dir).await;
+        register_meshed_node(&cm, "n005", "10.44.0.7");
+        record_cp_set(&cm, "n005", &["n005"]);
+        crate::cluster_k8s::provision_assignments(&cm, &provision_net_calico(), &cm.k0s_state())
+            .unwrap();
+        wait_for("incumbent assigned", || {
+            cm.get_node("n005").and_then(|n| n.k0s_role).is_some()
+        });
+
+        register_meshed_node(&cm, "n006", "10.44.0.7"); // collides with the incumbent's owned address
+        crate::cluster_k8s::provision_assignments(&cm, &provision_net_calico(), &cm.k0s_state())
+            .unwrap();
+
+        // Incumbent untouched; newcomer refused and told who owns the address.
+        assert_eq!(
+            cm.get_node("n005").unwrap().k0s_mesh_ip.as_deref(),
+            Some("10.44.0.7")
+        );
+        assert!(cm.get_node("n006").and_then(|n| n.k0s_role).is_none());
+        let want = "network mismatch: mesh address 10.44.0.7 already owned by n005";
+        wait_for("newcomer carries the ownership reason", || {
+            cm.get_node("n006")
+                .and_then(|n| n.k0s_last_error)
+                .as_deref()
+                == Some(want)
         });
     }
 
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -7390,6 +7390,35 @@ mod tests {
         });
     }
 
+    /// Register a node already on the WireGuard mesh: it advertises its real `spur0` address (what
+    /// `detect_node_address` reports when WireGuard is up) and a wg pubkey, so `provision_assignments`
+    /// adopts that real address as its `k0s_mesh_ip` instead of allocating from a pool.
+    fn register_meshed_node(cm: &ClusterManager, name: &str, mesh_addr: &str) {
+        cm.register_node(
+            name.into(),
+            name.into(),
+            ResourceSet {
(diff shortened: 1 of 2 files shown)


crates/spurctld/src/cluster_k8s.rs
```diff
@@ -225,6 +225,32 @@ pub(crate) fn provision_assignments(
...
+            continue; // already assigned; its persisted mesh IP is authoritative
+        }
+        let Some(addr) = real_mesh_address(n, &net.mesh_cidr)? else {
+            continue; // no in-mesh address reported — pool allocation handles it below
+        };
+        if let Some(other) = claimed.insert(addr, n.name.clone()) {
+            let reason = format!("network mismatch: mesh address {addr} also claimed by {other}");
+            let _ = cluster.set_node_k0s_error(&n.name, Some(reason.clone()));
+            anyhow::bail!("{} {reason}", n.name);
+        }
```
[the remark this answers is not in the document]
spraveenio: Good catch on both points — addressed in 77d2434.

1. **Both sides now surfaced.** On a duplicate in-mesh address, the reason is recorded on the originally-claiming node (`other`) as well as the second-seen node, so `spur k8s status` points at each side of the conflict. The message now names both: `network mismatch: mesh address <ip> claimed by both <a> and <b>`.
2. **No longer swallowing the write error.** Extracted a `record_node_k0s_error` helper (mirroring the existing `clear_node_error`) that `warn!`s on a failed `set_node_k0s_error` instead of `let _ =`, so the reason isn't silently dropped on a Raft/WAL write error.

The unit test `provision_refuses_conflicting_real_addresses` was strengthened to assert both nodes carry the reason.

crates/spurctld/src/cluster_k8s.rs
```diff
@@ -225,6 +225,38 @@ pub(crate) fn provision_assignments(
...
+    // mesh when the control-plane set isn't an alphabetical prefix of join order. A node whose comm
+    // address is OUTSIDE `mesh_cidr` (e.g. a control plane advertising its routable underlay address)
+    // is simply not reporting a mesh address — it falls through to pool allocation, not an error.
+    // Reserve every adopted address up front so pooled nodes never collide, and fail loud only on a
+    // genuine conflict: two nodes claiming the same in-mesh address.
+    let mut real_ip: HashMap<String, Ipv4Addr> = HashMap::new();
+    let mut claimed: HashMap<Ipv4Addr, String> = HashMap::new();
+    for n in &nodes {
+        if n.k0s_role.is_some() {
+            continue; // already assigned; its persisted mesh IP is authoritative
```
yansun1996, line 249: Since this skip only runs for not-yet-assigned nodes, once a node has a `k0s_role` its `k0s_mesh_ip` is never revisited — even if its real WireGuard address later changes (restart, re-key, etc.). Since `build_mesh_membership` re-pushes `ApplyMesh` from `k0s_mesh_ip` on every reconcile tick, a stale value here would keep pushing a wrong `AllowedIPs` entry, which is a milder version of the exact bug this PR fixes. Might be worth a follow-up to detect/reconcile drift on already-assigned nodes, or at least a comment noting it as a known limitation.

crates/spurctld/src/cluster_k8s.rs
```diff
@@ -225,6 +225,38 @@ pub(crate) fn provision_assignments(
...
+        if let Some(other) = claimed.insert(addr, n.name.clone()) {
+            // Surface the conflict on BOTH sides so `spur k8s status` points at each node sharing the
+            // address, not just the second one seen.
+            let reason = format!(
+                "network mismatch: mesh address {addr} claimed by both {} and {other}",
+                n.name
+            );
+            record_node_k0s_error(cluster, &n.name, &reason);
+            record_node_k0s_error(cluster, &other, &reason);
+            anyhow::bail!("{reason}");
```
yansun1996: Since the conflict-detection loop runs to completion before the role-assignment loop, this `bail!` aborts the whole call — meaning every node goes unprovisioned on this tick (and every tick after, since the caller just logs and retries), not just the conflicting pair. Was the full-cluster blast radius intentional, or would it be worth excluding just the conflicting pair and letting the rest of the nodes provision normally?

crates/spurctld/src/cluster_k8s.rs
```diff
@@ -225,6 +225,38 @@ pub(crate) fn provision_assignments(
         }
     }
 
+    // Adopt each node's REAL mesh address: a node on the mesh advertises its `spur0` IP as
```
yansun1996: This comment block (and a couple of others added nearby, e.g. around the `mesh_ip` selection a bit further down and the `real_mesh_address` doc comment) run 4-8 lines. Might be worth trimming these down to a line or two capturing the key *why*, with the rest of the design rationale left to the PR description / commit message.

yansun1996 (commented):
Nice fix — adopting each node's real advertised address as the source of truth for `k0s_mesh_ip` instead of re-deriving it from a pool is the right call, and it mirrors how the pod-CIDR plane already works. The unit tests do a good job exercising the adversarial CP-selection, scrambled-join-order, underlay-fallback, and conflict cases through the real `provision_assignments` path.

A few things worth a look, none of which I think need to block this PR:

- **`Node.address` is self-reported** (a node's advertised address wins over the gRPC peer address at registration), and this PR is the first place that value becomes network-security-relevant — it now feeds WireGuard `AllowedIPs` via `k0s_mesh_ip`. A misbehaving or misconfigured node claiming another node's real address could trigger the "network mismatch" error against a legitimate peer. This is a pre-existing trust assumption, not something introduced here, but might be worth a follow-up look now that the blast radius includes mesh availability.
- In `provision_refuses_conflicting_real_addresses`, the error-message assertion is `contains("mismatch") || contains("conflict")` — fairly loose. Might be worth tightening to an exact match (or asserting both node names appear) so a future message-wording change doesn't silently weaken the test.

See inline notes for two other spots.

crates/spurctld/src/cluster_k8s.rs
spraveenio: Good point — this is the drift case: once a node has a `k0s_role`, its `k0s_mesh_ip` is never revisited, so if its real WireGuard address later changes, `build_mesh_membership` keeps pushing a stale `AllowedIPs`.

Before scoping a fix: is this a real lifecycle for our mesh? Does a node's `spur0` address actually change across a spurd restart / re-registration / re-key in practice, or is it allocated once and stable for the node's lifetime once it joins? If it's stable, this stays a documented known-limitation; if it can churn, drift detection/reconcile on already-assigned nodes is worth a follow-up. Leaning toward filing a follow-up issue either way and keeping this PR scoped to the join-order bug — does that match your read?

crates/spurctld/src/cluster_k8s.rs
spraveenio: Addressed in 4f8c28a — the full-cluster blast radius was not intentional, and the conflict is now contained.

`provision_assignments` no longer `bail!`s. It groups unassigned claimants by advertised in-mesh address and resolves per address:
- an already-assigned node owns its `k0s_mesh_ip`, so a newcomer colliding with it is the one refused;
- two unassigned newcomers on the same address both stay out (no winner picked — a real duplicate is an operator mistake);
- every non-conflicting node provisions normally that tick.

Refused nodes are skipped via a `refused` set and each carries a per-node reason (`network mismatch: mesh address <ip> already owned by <owner>` / `... claimed by <a>, <b>`) for `spur k8s status`. Tests updated: `provision_refuses_conflicting_real_addresses` now asserts the healthy bystander still provisions while both conflicting nodes stay out, and `provision_refuses_newcomer_colliding_with_assigned_node` covers the incumbent-wins case.

crates/spurctld/src/cluster_k8s.rs
spraveenio: Trimmed in 4f8c28a. The `mesh_ip` selection comment and the `real_mesh_address` doc comment are down to two lines each; the adopt-block comment kept a short why plus the conflict-handling contract (which the code now enforces rather than the old `bail!`), with the rest of the rationale in the commit message / PR description.

crates/spurctld/src/cluster.rs
```diff
@@ -15487,6 +15516,192 @@ mod tests {
         assert!(cm.get_node("node-a").and_then(|n| n.k0s_role).is_some());
     }
 
+    // ---- SPUR-196: k0s adopts each node's real WireGuard address as its mesh IP ----
```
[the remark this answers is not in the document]
spraveenio: Already addressed in 77d2434 (before this review landed) — the `SPUR-196` references in these comments were reworded to neutral technical descriptions. Current source has no ticket IDs in the changed blocks.

crates/spurctld/src/cluster_k8s.rs
```diff
@@ -225,6 +225,32 @@ pub(crate) fn provision_assignments(
         }
     }
 
+    // Adopt each node's REAL mesh address (SPUR-196): a node on the mesh advertises its `spur0` IP as
+    // `Node.address`; when that address falls within `mesh_cidr` it IS the node's mesh IP, and is the
+    // single source of truth for it. Re-deriving from an independent pool is what silently corrupts the
+    // mesh when the control-plane set isn't an alphabetical prefix of join order. A node whose comm
+    // address is OUTSIDE `mesh_cidr` (e.g. a control plane advertising its routable underlay address)
+    // is simply not reporting a mesh address — it falls through to pool allocation, not an error.
```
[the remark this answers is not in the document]
spraveenio: Same as above — this block was reworded in 77d2434 and no longer embeds a ticket ID. 4f8c28a further reworks this comment for the contained-conflict behavior.

crates/spurctld/src/cluster_k8s.rs
yansun1996: we can handle this case as a follow-up since it requires the self-healing and reconcile, thanks

yansun1996 (approved):
LGTM

PR #700, opened 2026-08-19
feat(spurctld): record preemption provenance on job state

sajmera-pensando:
## Why

When a job is preempted, users have no way to see why it happened or which job caused it. `squeue` shows `Reason=BeginTime` for requeued jobs (opaque), and cancelled-by-preemption jobs show nothing beyond `State=CANCELLED` in `sacct`. The preempting job ID, the mode (cancel/requeue/suspend), and the QOS authorization chain are only in log traces — never persisted.

## What this changes

Three new `#[serde(default)]` fields on `Job` (WAL-safe — old entries replay with `None`):
- `preempted_by: Option<JobId>` — which job triggered the preemption
- `preempt_mode: Option<String>` — `"Requeue"`, `"Cancel"`, or `"Suspend"`
- `preempt_qos: Option<String>` — the QOS name that authorized it under `preempt_type = QosPriority`; empty for plain priority-based preemption

These are threaded from `try_preempt` through `preempt_job()` into all three WAL ops (`JobPreemptRequeue`, `JobPreemptCancel`, `JobSuspend`) → WAL apply → `notify_job_end` → accounting DB (three new nullable columns).

`PendingReason::BeginTime` is replaced by a new `Preempted` variant for requeue-preempted jobs, so `squeue -o "%R"` shows `Preempted` (with a desc naming the preempting job) instead of the opaque `BeginTime`.

Also fixes a pre-existing gap in `sacct`: `PREEMPTED` was missing from both the state filter and the row mapper in `get_job_history`, so `sacct -s PREEMPTED` returned nothing and preempted jobs displayed as `COMPLETED`. Added `"PR" | "PREEMPTED"` to `parse_acct_state` and the corresponding entries to both the filter and the mapper in `grpc.rs`.

## Querying preempted jobs

One command works for any preempted job regardless of mode:

```
scontrol show job <id>          # while job is live (running, pending, suspended, or recently terminal)
sacct -j <id> --format=...      # permanent record in accounting DB
```

### Requeue mode — job pending after preemption

```
JobId=1 JobName=low-job
   UserId=root Account=
   Partition=default QOS=
   JobState=PENDING Reason=preempted by job 2
   NumNodes=1 NumTasks=1 CPUs/Task=1
   SubmitTime=2026-08-19T23:04:04 StartTime=N/A EndTime=N/A
   WorkDir=/root
   StdOut=/root/spur-1.out StdErr=/root/spur-1.out
   ExitCode=0:0 DerivedExitCode=0:0 Priority=1000
   PreemptedBy=2 PreemptMode=Requeue PreemptQOS=N/A
```

### Requeue mode — same job re-running (provenance cleared on re-dispatch)

```
JobId=1 JobName=low-job
   UserId=root Account=
   Partition=default QOS=
   JobState=RUNNING Reason=None
   NumNodes=1 NumTasks=1 CPUs/Task=1
   NodeList=SCSSAJMERA01
   SubmitTime=2026-08-19T23:04:04 StartTime=2026-08-19T23:04:13 EndTime=N/A
   WorkDir=/root
   StdOut=/root/spur-1.out StdErr=/root/spur-1.out
   ExitCode=0:0 DerivedExitCode=0:0 Priority=1000
```

### Cancel mode — scontrol show job (terminal)

```
JobId=1 JobName=low-cancel
   UserId=root Account=
   Partition=default QOS=
   JobState=CANCELLED Reason=None
   NumNodes=1 NumTasks=1 CPUs/Task=1
   NodeList=SCSSAJMERA01
   SubmitTime=2026-08-19T23:18:37 StartTime=2026-08-19T23:18:37 EndTime=2026-08-19T23:18:38
   WorkDir=/root
   StdOut=/root/spur-1.out StdErr=/root/spur-1.out
   ExitCode=-1:0 DerivedExitCode=0:0 Priority=1000
   PreemptedBy=2 PreemptMode=Cancel PreemptQOS=N/A
```

### Cancel mode — sacct (permanent accounting record)

```
$ sacct -j 1 --format=JobID,JobName,State,ExitCode,PreemptedBy,PreemptMode,PreemptQOS --noheader
1   low-cancel   PREEMPTED   -1:0   2   Cancel   N/A

$ sacct -s PREEMPTED --noheader
1   low-cancel   root   default   PREEMPTED   00:00:00   1   -1:0

$ sacct -s PR --noheader
1   low-cancel   root   default   PREEMPTED   00:00:00   1   -1:0
```

Note: `scontrol show job` shows `CANCELLED` (the live job state after cancel-preemption). `sacct` correctly shows `PREEMPTED` — the accounting convention, matching Slurm, is that the finalization event is always recorded as PREEMPTED regardless of the terminal live state.

## Approach / trade-offs

- **WAL-safe**: all new fields use `#[serde(default)]`, so old log entries replay without error. Verified with a backward-compat unit test that applies a legacy WAL entry (no provenance fields) and asserts the fields default to `None`.
- **Accounting DB**: `ADD COLUMN IF NOT EXISTS` with nullable/empty defaults — no migration failures on existing rows.
- **Suspend path**: provenance fields are carried inside the `JobSuspend` WAL op (not written after the fact), so they survive Raft replication and leader failover.
- **Stale provenance**: cleared in `JobStart` WAL apply, so a re-dispatched run's accounting row never inherits the prior preemption's data. Verified by the `preempt_requeue_wal_apply_with_provenance_clears_on_restart` unit test.
- **Proto**: three fields appended to `JobInfo` at tags 57–59 (append-only, backward compatible).
- **`preempt_job()` signature change**: internal only — no proto/REST breaking change.

## Testing

- Unit tests cover WAL apply for all three modes (requeue/cancel/suspend), field propagation, backward-compat replay of old WAL entries, and the `PendingReason::Preempted` display/behavior.
- E2E tests assert `PreemptedBy=`, `PreemptMode=`, `PreemptQOS=` in `scontrol show job` output for requeue, cancel, and QOS-driven preemption against a real running cluster.
- All outputs above captured from live single-node clusters running this branch.

### DIFF
--- a/crates/spur-cli/src/sacct.rs
+++ b/crates/spur-cli/src/sacct.rs
@@ -67,7 +67,7 @@ pub struct SacctArgs {
 
 const SACCT_DEFAULT_FORMAT: &str = "%.8i %.15j %.10u %.10a %.10P %.8T %10M %.8D %6x";
 const SACCT_LONG_FORMAT: &str =
-    "%.8i %.15j %.10u %.10a %.10P %.8T %10M %.8D %6x %.10X %.19S %.19E %.10l";
+    "%.8i %.15j %.10u %.10a %.10P %.8T %10M %.8D %6x %.10X %.19S %.19E %.10l %8y %8m %12q";
 const SACCT_BRIEF_FORMAT: &str = "%.8i %.8T %6x";
 
 pub fn sacct_header(spec: char) -> &'static str {
@@ -91,6 +91,9 @@ pub fn sacct_header(spec: char) -> &'static str {
         'C' => "NCPUS",
         'R' => "ReqMem",
         'Q' => "QOS",
+        'y' => "PreemptedBy",
+        'm' => "PreemptMode",
+        'q' => "PreemptQOS",
         _ => "?",
     }
 }
@@ -115,6 +118,9 @@ fn sacct_field_spec(name: &str) -> Option<char> {
         "ncpus" => Some('C'),
         "reqmem" => Some('R'),
         "qos" => Some('Q'),
+        "preemptedby" => Some('y'),
+        "preemptmode" => Some('m'),
+        "preemptqos" => Some('q'),
         _ => None,
     }
 }
@@ -216,6 +222,27 @@ fn resolve_sacct_field(job: &spur_proto::proto::JobInfo, spec: char) -> String {
         'n' => job.nodelist.clone(),
         'C' => (job.num_tasks * job.cpus_per_task.max(1)).to_string(),
         'Q' => job.qos.clone(),
+        'y' => {
+            if job.preempted_by == 0 {
+                "N/A".into()
+            } else {
+                job.preempted_by.to_string()
+            }
+        }
+        'm' => {
+            if job.preempt_mode.is_empty() {
+                "N/A".into()
+            } else {
+                job.preempt_mode.clone()
+            }
+        }
+        'q' => {
+            if job.preempt_qos.is_empty() {
+                "N/A".into()
+            } else {
+                job.preempt_qos.clone()
+            }
+        }
         _ => "?".into(),
     }
 }
@@ -233,6 +260,7 @@ fn parse_acct_state(s: &str) -> Option<i32> {
         "CA" | "CANCELLED" => Some(5),
         "TO" | "TIMEOUT" => Some(6),
         "NF" | "NODE_FAIL" => Some(7),
+        "PR" | "PREEMPTED" => Some(8),
         "DL" | "DEADLINE" => Some(10),
         "R" | "RUNNING" => Some(1),
         "PD" | "PENDING" => Some(0),
@@ -377,6 +405,7 @@ mod tests {
             ("CANCELLED", JobState::JobCancelled),
             ("TIMEOUT", JobState::JobTimeout),
             ("NODE_FAIL", JobState::JobNodeFail),
+            ("PREEMPTED", JobState::JobPreempted),
             ("DEADLINE", JobState::JobDeadline),
         ];
         for (s, expected) in cases {
--- a/crates/spurctld/src/accounting/grpc.rs
+++ b/crates/spurctld/src/accounting/grpc.rs
@@ -185,6 +185,9 @@ impl SlurmAccounting for AccountingService {
             end_time,
             req.exit_signal,
             req.derived_exit_code,
+            None,
+            "",
+            "",
         )
         .await
         .map_err(|e| Status::internal(e.to_string()))?;
@@ -214,6 +217,7 @@ impl SlurmAccounting for AccountingService {
                 4 => Some("FAILED".into()),
                 5 => Some("CANCELLED".into()),
                 6 => Some("TIMEOUT".into()),
+                8 => Some("PREEMPTED".into()),
                 10 => Some("DEADLINE".into()),
                 _ => None,
             })
@@ -256,6 +260,7 @@ impl SlurmAccounting for AccountingService {
                     "FAILED" => JobState::JobFailed as i32,
                     "CANCELLED" => JobState::JobCancelled as i32,
                     "TIMEOUT" => JobState::JobTimeout as i32,
+                    "PREEMPTED" => JobState::JobPreempted as i32,
                     "DEADLINE" => JobState::JobDeadline as i32,
                     "RUNNING" => JobState::JobRunning as i32,
                     "PENDING" => JobState::JobPending as i32,
@@ -295,6 +300,9 @@ impl SlurmAccounting for AccountingService {
                 srun_step_dispatch: false,
                 req_gpus: 0,
                 req_gpus_detail: String::new(),
+                preempted_by: r.preempted_by.unwrap_or(0) as u32,
+                preempt_mode: r.preempt_mode.clone(),
+                preempt_qos: r.preempt_qos.clone(),
             })
             .collect();
 
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -1428,6 +1428,8 @@ impl ClusterManager {
         self.propose(WalOperation::JobSuspend {
             job_id,
             at: chrono::Utc::now(),
+            preempted_by: None,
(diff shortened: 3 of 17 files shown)


crates/spur-cli/src/sacct.rs
```diff
@@ -233,6 +260,7 @@ fn parse_acct_state(s: &str) -> Option<i32> {
         "CA" | "CANCELLED" => Some(5),
         "TO" | "TIMEOUT" => Some(6),
         "NF" | "NODE_FAIL" => Some(7),
+        "PR" | "PREEMPTED" => Some(9),
```
yansun1996: `"PR" | "PREEMPTED"` maps to state code `9` here, but `9` is `JOB_SUSPENDED` in `proto/slurm.proto` (`JOB_PREEMPTED = 8`). The row-mapper side elsewhere already uses `JobState::JobPreempted as i32` (8) correctly, so this filter looks like it's using the wrong constant. Should this be `8`?

crates/spurctld/src/accounting/grpc.rs
```diff
@@ -214,6 +217,7 @@ impl SlurmAccounting for AccountingService {
                 4 => Some("FAILED".into()),
                 5 => Some("CANCELLED".into()),
                 6 => Some("TIMEOUT".into()),
+                9 => Some("PREEMPTED".into()),
```
yansun1996: Same mismatch — `9 => Some("PREEMPTED")` should be `8` to match `JOB_PREEMPTED` in the proto. As written, filtering `GetJobHistoryRequest` on `JOB_SUSPENDED` (9) would actually return preempted jobs.

docs/user-guide/monitoring-jobs.rst
```diff
@@ -333,6 +344,34 @@ an entity: ``job``, ``node``, ``partition``, ``reservation``, or ``step``.
    spur show node node01
    scontrol show partition gpu
```
yansun1996, line 347: A couple of small mismatches with the actual CLI output: `PreemptedBy` prints `N/A` (not `0`) when unset, and the whole `PreemptedBy=.../PreemptMode=.../PreemptQOS=...` line is only shown at all when the job was preempted — not "0 if not preempted." Likewise `PreemptQOS` prints `N/A`, not empty, for plain priority-based preemption (the `sacct` section earlier in this same doc already says `N/A` — just needs to match here too).

crates/spurctld/src/cluster.rs
```diff
@@ -1707,12 +1719,34 @@ impl ClusterManager {
         match mode {
             PreemptMode::Off => anyhow::bail!("preemption disabled for job {}", job_id),
             PreemptMode::Suspend => {
-                self.suspend_job(job_id, "")?;
+                // Propose JobSuspend directly (rather than suspend_job) so
```
yansun1996, line 1722: The `PreemptMode::Suspend` branch re-checks job existence/state, which was already validated in the same synchronous call just above. Doesn't look reachable to diverge — could probably drop the inner check.

yansun1996 (commented):
Nice feature, and I like that the WAL fields are all `#[serde(default)]` with a legacy-replay test — that's exactly the right pattern for this kind of addition.

Two things worth fixing before merge:

1. **Suspend→Resume doesn't clear preemption provenance.** `JobSuspend`'s WAL-apply arm sets `preempted_by`/`preempt_mode`/`preempt_qos`, but the `JobResume` apply arm (`crates/spurctld/src/cluster.rs`, around line 5210) never clears them — unlike `JobStart`, which explicitly clears all three fields on redispatch with the comment "a new run supersedes any prior preemption provenance." The same reasoning applies to resume: a job that's suspended via preemption, resumed, and then runs to a normal completion will have its accounting record — and live `scontrol show job`/`sacct` output — permanently show it as "preempted via Suspend." `notify_job_finished` reads these fields unconditionally on every finalize, so this isn't just a display quirk, it corrupts the terminal accounting row. Worth clearing the three fields in the `JobResume` arm too, and adding a test alongside the existing requeue/`JobStart` provenance-clearing test for this resume path.
2. **`sacct -s PREEMPTED` / `-s PR` filters on the wrong state code.** Looks like it's using `9` (which is `JOB_SUSPENDED` in the proto) instead of `8` (`JOB_PREEMPTED`). The row *display* side already uses the correct code, so this looks like a copy-paste of the wrong constant in the filter path. See inline comments.

A few smaller doc/style notes below. Also, a couple of the doc's claims about `PreemptedBy=0`/`PreemptQOS=Empty` don't match what the CLI actually prints (`N/A` in both cases) — easy fix.

One more doc note that doesn't land on a changed line: since this PR adds a new `Preempted` pending reason (`squeue -o "%R"` for requeue-preempted jobs), could we add it to the "Common reasons" list in the squeue section? It currently lists `Priority`/`Resources`/`Dependency`/`Reservation`/QOS reasons but not the new one.

crates/spur-cli/src/sacct.rs
sajmera-pensando: Fixed — changed to  (). Also added  to the  unit test to catch this class of regression.

crates/spurctld/src/accounting/grpc.rs
sajmera-pensando: Fixed — changed to `8` (`JOB_PREEMPTED`). The row-mapper side was already correct; this was indeed a copy-paste error in the filter path only.

docs/user-guide/monitoring-jobs.rst
sajmera-pensando: Fixed both — `PreemptedBy` now says "only shown when the job has been preempted" (the line is omitted entirely for non-preempted jobs), and `PreemptQOS` consistently says `N/A` for plain priority-based preemption, matching the sacct section and the actual CLI output.

crates/spurctld/src/cluster.rs
sajmera-pensando: Removed — the inner check is unreachable since the outer guard at the top of `preempt_job` already validates state and returns an error if the job is missing or not Running. Added a comment explaining why.

sgopinath1 (approved):

yansun1996 (approved):
LGTM

PR #703, opened 2026-08-20
fix(spur-cli): honor sacct -j/--jobs job-ID filter

### LINKED ISSUE #689: sacct: -j/--jobs job-ID filter is ignored and returns all jobs
## What happened

`sacct -j <id>` ignores the job-ID filter entirely. It returns the full job history (up to `--limit`) regardless of the value passed, including for IDs that do not exist.

```
$ sacct -n -j 30 -o JobID,State | head
31 CANCELLED
30 CANCELLED
29 CANCELLED
28 CANCELLED
...

$ sacct -n -j 999999 -o JobID,State | head
31 CANCELLED
30 CANCELLED
29 CANCELLED
28 CANCELLED
...
```

Both invocations print the identical full list. `-j 30` and `-j 999999` behave the same as running `sacct` with no filter.

## What I expected

`sacct -j 30` returns only job 30 (and its steps). `sacct -j 999999` returns nothing. This matches Slurm's `sacct -j` behavior.

## Where it comes from

The flag is declared but never used. In `crates/spur-cli/src/sacct.rs` the `--jobs` value is parsed:

```rust
/// Show only these job IDs
#[arg(short = 'j', long)]
pub jobs: Option<String>,
```

but `args.jobs` is never read when building the query. `GetJobHistoryRequest` is constructed with `user`, `account`, `start_after`, `start_before`, `states`, and `limit` only, and the returned jobs are printed without any client-side ID filtering. So the id is silently dropped.

## Impact

Any workflow or script that relies on `sacct -j <id>` to inspect a single job gets the whole history instead. It is a silent wrong-result (no error), which is worse than rejecting an unsupported flag, and it breaks the common `sacct -j` idiom for scripting.

## Environment

- Spur 0.8.0, bare-metal cluster (4 nodes), accounting backed by PostgreSQL.
- Reproduced consistently against a live controller.

Note: `scontrol show job <id>` filters by id correctly, so the gap is specific to `sacct`.

maybeharshit:
## Motivation

`sacct -j <id>` accepted the `-j/--jobs` flag but silently ignored it. The value was parsed and then dropped, and `GetJobHistory` carried no job-ID filter at any layer, so every invocation returned the full history (up to `--limit`) regardless of the ID passed — including IDs that do not exist. Being a silent wrong-result (no error), it broke the common `sacct -j <id>` scripting idiom.

Fixes #689.

## Technical Details

The gap spanned the entire accounting read path, so the fix does too:

- **CLI** (`sacct.rs`): parse the comma-separated `--jobs` value and put the IDs in the request. Slurm step suffixes (`30.0`, `30.batch`) collapse to the base job ID — Spur accounting records whole jobs, not steps — and unparseable tokens are dropped, mirroring `squeue`'s handling.
- **Proto** (`slurm.proto`): add `repeated uint32 job_ids = 7` to `GetJobHistoryRequest`. Append-only (new tag), so FFI/REST/wire compatibility is preserved.
- **Controller** (`accounting/grpc.rs`): forward the IDs to the DB layer.
- **DB** (`accounting/db.rs`): filter with `job_id IN (...)` in the query builder. Filtering is **server-side** on purpose — a client-side filter over the default 100-row window would wrongly return nothing for a job older than the last 100 records.

`squeue -j` already worked this way (`GetJobsRequest.job_ids`); this brings `sacct` to parity. `scontrol show job <id>` was unaffected since it uses a fetch-by-id RPC.

## Test Plan

- New `parse_job_ids` unit test: comma-separated lists, whitespace, step suffixes, junk-dropping, empty input.
- Extended the PG-gated `get_job_history` integration test: single ID, multi-ID subset, non-existent ID → empty, and job-ID ∩ user composition (runs in the "DB tests (accounting)" CI job).
- `cargo test --locked` — all pass.
- `cargo fmt --check` — clean.

### DIFF
--- a/crates/spur-cli/src/sacct.rs
+++ b/crates/spur-cli/src/sacct.rs
@@ -160,7 +160,11 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
         })
         .unwrap_or_default();
 
-    let job_ids = args.jobs.as_deref().map(parse_job_ids).unwrap_or_default();
+    let job_ids = args
+        .jobs
+        .as_deref()
+        .map(crate::job_id_arg::parse_job_ids)
+        .unwrap_or_default();
 
     let channel = crate::authclient::connect(&args.controller)
         .await
@@ -271,21 +275,6 @@ fn parse_acct_state(s: &str) -> Option<i32> {
         "PD" | "PENDING" => Some(0),
         _ => None,
     }
-}
-
-/// Parse a Slurm-style `-j/--jobs` value into numeric job IDs.
-///
-/// Accepts a comma-separated list and tolerates step suffixes (`30.0`,
-/// `30.batch`) by matching the base job ID, since Spur accounting records whole
-/// jobs rather than individual steps. Unparseable entries are dropped, mirroring
-/// squeue's job-ID handling.
-fn parse_job_ids(s: &str) -> Vec<u32> {
-    s.split(',')
-        .filter_map(|tok| {
-            let base = tok.trim().split('.').next().unwrap_or_default();
-            base.parse::<u32>().ok()
-        })
-        .collect()
 }
 
 fn format_elapsed(job: &spur_proto::proto::JobInfo) -> String {
@@ -416,20 +405,5 @@ mod tests {
         .await
         .unwrap_err();
         assert!(err.to_string().contains("no recognized fields"));
-    }
-
-    #[test]
-    fn parse_job_ids_handles_lists_steps_and_junk() {
-        assert_eq!(parse_job_ids("30"), vec![30]);
-        assert_eq!(parse_job_ids("30,31,32"), vec![30, 31, 32]);
-        // Surrounding whitespace is tolerated.
-        assert_eq!(parse_job_ids(" 30 , 31 "), vec![30, 31]);
-        // Step suffixes collapse to the base job ID (Spur records jobs, not steps).
-        assert_eq!(parse_job_ids("30.0"), vec![30]);
-        assert_eq!(parse_job_ids("30.batch,31.0"), vec![30, 31]);
-        // Unparseable tokens are dropped, not turned into a bogus filter.
-        assert_eq!(parse_job_ids("30,abc,31"), vec![30, 31]);
-        assert!(parse_job_ids("").is_empty());
-        assert!(parse_job_ids("abc").is_empty());
     }
 }
--- a/crates/spur-cli/src/sacct.rs
+++ b/crates/spur-cli/src/sacct.rs
@@ -160,6 +160,8 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
         })
         .unwrap_or_default();
 
+    let job_ids = args.jobs.as_deref().map(parse_job_ids).unwrap_or_default();
+
     let channel = crate::authclient::connect(&args.controller)
         .await
         .context("failed to connect to controller")?;
@@ -183,6 +185,7 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
             start_after,
             start_before,
             states,
+            job_ids,
             limit: args.limit,
         })
         .await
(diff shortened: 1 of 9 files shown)


shiv-tyagi (commented):
Reviewed the fix and tested it end-to-end on a bare-metal cluster (5 jobs, accounting on PostgreSQL): single ID, multi-ID, `--jobs=`, step suffix (`3.batch` -> job 3), non-existent ID (empty), and junk-token dropping all return exactly the expected rows. The wiring CLI -> proto -> controller -> SQL is correct, server-side filtering is the right call, and the proto field is appended so wire compat holds.

A few things to consider, inline. One that has no diff line to anchor to:

**docs: `docs/user-guide/monitoring-jobs.rst` sacct flags table omits `--jobs`/`-j`.** This PR makes `-j` functional (it was silently ignored), which is a user-facing change that per AGENTS.md ships with docs. Add a row:

```
   * - ``--jobs``
     - ``-j``
     - Filter by job ID (comma list; step suffixes like ``30.0`` match the base job).
```

crates/spurctld/src/accounting/db.rs
```diff
@@ -398,13 +398,15 @@ pub struct JobRecord {
 }
 
 /// Query job history.
+#[allow(clippy::too_many_arguments)]
```
shiv-tyagi: `get_job_history` is now 8 positional args, silenced with `#[allow(clippy::too_many_arguments)]`. That lint is a signal here: the call sites pass `..., &[], &[id1], 100` where `states` and `job_ids` are both `&[]` and easy to transpose without the compiler noticing. Prefer folding the filters into a struct:

```rust
#[derive(Default)]
pub struct JobHistoryQuery<'a> {
    pub user: Option<&'a str>,
    pub account: Option<&'a str>,
    pub start_after: Option<DateTime<Utc>>,
    pub start_before: Option<DateTime<Utc>>,
    pub states: &'a [String],
    pub job_ids: &'a [i32],
    pub limit: u32,
}
```

Drops the `#[allow]` and makes call sites self-documenting.

crates/spur-cli/src/sacct.rs
```diff
@@ -240,6 +243,21 @@ fn parse_acct_state(s: &str) -> Option<i32> {
     }
 }
 
+/// Parse a Slurm-style `-j/--jobs` value into numeric job IDs.
+///
+/// Accepts a comma-separated list and tolerates step suffixes (`30.0`,
+/// `30.batch`) by matching the base job ID, since Spur accounting records whole
+/// jobs rather than individual steps. Unparseable entries are dropped, mirroring
+/// squeue's job-ID handling.
```
shiv-tyagi: This says step-suffix handling mirrors squeue, but `squeue` (`squeue.rs:114`) and `sprio` (`sprio.rs:49`) both parse with a plain `j.trim().parse::<u32>()` and do not strip `.step` suffixes. So `squeue -j 30.0` drops the token while `sacct -j 30.0` matches job 30. The sacct behavior is better, the parity claim is just wrong. See the duplication comment below for a fix that resolves both.

crates/spur-cli/src/sacct.rs
```diff
@@ -240,6 +243,21 @@ fn parse_acct_state(s: &str) -> Option<i32> {
...
+fn parse_job_ids(s: &str) -> Vec<u32> {
```
shiv-tyagi: Third copy of comma-list job-ID parsing (`squeue.rs:109`, `sprio.rs:45`, now here). Extract one shared helper and have all three call it. Bonus: it unifies the step-suffix handling so `squeue`/`sprio` gain the same leniency instead of silently diverging.

crates/spur-cli/src/sacct.rs
```diff
@@ -426,4 +444,19 @@ mod tests {
...
+    #[test]
+    fn parse_job_ids_handles_lists_steps_and_junk() {
+        assert_eq!(parse_job_ids("30"), vec![30]);
+        assert_eq!(parse_job_ids("30,31,32"), vec![30, 31, 32]);
+        // Surrounding whitespace is tolerated.
+        assert_eq!(parse_job_ids(" 30 , 31 "), vec![30, 31]);
+        // Step suffixes collapse to the base job ID (Spur records jobs, not steps).
+        assert_eq!(parse_job_ids("30.0"), vec![30]);
+        assert_eq!(parse_job_ids("30.batch,31.0"), vec![30, 31]);
+        // Unparseable tokens are dropped, not turned into a bogus filter.
```
shiv-tyagi: Nit: these comments narrate what the asserts already state plainly. Drop them if you touch the file.

shiv-tyagi (dismissed):
LGTM

shiv-tyagi (approved):
LGTM. Re-verified after the rebase: all review comments addressed (JobHistoryQuery struct, shared job-ID parsing helper, docs row), proto field remains append-only (job_ids = 7, no collision with the new base), clippy clean, full test suite green.

PR #704, opened 2026-08-20
docs(admin-guide): say where GrpWall is enforced and what a zero budget does

nikhilsk:
## What this adds

Two facts about `GrpWall` that an operator could previously only learn by experiment, plus the entry a migrating admin looks for first.

**Where the limit is enforced.** `grp_wall_minutes` is read in `check_qos_limits` and nowhere else — neither `check_qos_submit_limits` nor `check_qos_standalone_limits` consults it. A breach therefore never rejects a submission, not even under `DenyOnLimit`: the job is accepted and pends. An admin who sets `DenyOnLimit` expecting an over-budget QOS to bounce submissions gets a growing pending queue instead, with nothing in the docs to explain why.

**What a zero budget does.** Limit values carry Slurm's sentinel meanings, so `grpwall=0` is a literal budget of zero that blocks every job in the QOS rather than lifting the cap. But because enforcement needs a consumption figure before it applies, `grpwall=0` blocks nothing until one has been read — the same fail-open gap that governs the budget generally also governs its most destructive value, in both directions: set as a hard stop it lets jobs through, and set in the belief that `0` means "unlimited" it freezes the QOS once the first refresh lands.

**The migration table.** The three deliberate divergences from Slurm — admission-only enforcement with running jobs never killed, a trailing window instead of decayed usage, and no association-level cap — were documented only in the admin guide. `docs/migration-from-slurm.rst` never mentioned `GrpWall` at all, and its "Behavioral Differences and Current Limitations" table is where someone porting a Slurm setup reads first. The row cross-references the admin guide rather than restating it.

## Why separately

The GrpWall enforcement change is already merged. These gaps were found while reviewing that work against the QOS submit-time limits that landed alongside it: the submit gate and the sentinel semantics are what make the enforcement point and the meaning of `0` worth stating, and neither existed when the original docs were written.

## Testing

Docs-only; no code paths touched.

- Every claim verified against `crates/spur-core/src/qos.rs` on `main`: `grp_wall_minutes` appears only in `check_qos_limits`, guarded by `if let (Some(cap), Some(consumed))` with a `consumed >= cap` comparison, which is what makes both the zero-budget and unknown-consumption statements true.
- reST structure checked with docutils on both files.
- `pyspelling` run differentially against `main`'s copy of each file: no newly flagged words, so no wordlist change was needed.
- Both `:ref:` targets used (`grpwall-budgets`, `deny-on-limit`) resolve to real labels; `grpwall-budgets` is added here.

### DIFF
--- a/docs/admin-guide/accounting.rst
+++ b/docs/admin-guide/accounting.rst
@@ -516,6 +516,8 @@ Set ``DenyOnLimit`` through the QOS ``flags`` key:
 
    sacctmgr modify qos name=highprio set flags=DenyOnLimit
 
+.. _grpwall-budgets:
+
 Group wall-clock budgets (GrpWall)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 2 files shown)


sgopinath1 (approved):
LGTM

PR #708, opened 2026-08-20
fix(spur-core): emit node-local load balancing for HA mesh k0s clusters

spraveenio:
## What this fixes

Fixes SPUR-198

HA k0s (`spur k8s up --replicas 3/5`) over the WireGuard mesh had no cluster-wide konnectivity endpoint. Each controller's generated k0s config set `spec.api.address` to its own mesh IP and nothing provided a cluster-scoped endpoint, so the leader-gated `konnectivity-agent` DaemonSet pinned to whichever single controller last held the lease. The agent needs `serverCount` distinct server IDs but only ever reached one, re-dialing that controller ~1/s forever. The other controllers received zero agents and logged `"No agent available"` at ~1900 lines/s — enough to fill the disk and cost etcd its quorum on a long-running cluster. Reads kept serving from the apiserver watch cache, so the cluster looked healthy (`phase: ready`) until it failed.

## Approach

A VIP cannot float over WireGuard's cryptokey routing, so k0s node-local load balancing (NLLB) is the fix rather than an `externalAddress`/VIP. On the calico (mesh-native) path, when the cluster has more than one control plane, the generated config now emits:

```yaml
spec:
  network:
    nodeLocalLoadBalancing:
      enabled: true
      type: EnvoyProxy
```

Each node then runs a local Envoy that round-robins konnectivity across all controllers on `:8132`. Single-CP clusters do not emit it (an Envoy per node is needless there).

The control-plane count is threaded from `converge_provisioning` through `controller_k0s_config` into `k0s_controller_config_yaml`. No `ClusterConfig` schema change and no proto change — NLLB is always-on for multi-CP mesh clusters, so there is no config surface that could re-enable the broken state.

## Caveat (documented)

k0s does not hot-reload NLLB. Flipping an *existing* cluster onto this fix requires restarting workers before their Envoy starts; a fresh `spur k8s up` is unaffected. This is documented in the managed-kubernetes admin page.

## Testing

- Unit (`k0s_config_tests`): multi-CP calico config contains the NLLB stanza; single-CP calico config omits it; existing calico assertions (api address, SANs, bird mode, CIDRs, MTU) and the kube-router `None` path unchanged.
- `cargo test --locked` — spur-core 576 passed / 0 failed, spurctld 871 passed / 0 failed.
- `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` — no warnings.
- `cargo fmt --all -- --check` — clean.

## Out of scope

- The image logrotate defect (`su root syslog` missing) that turned the log noise into a disk-full outage — separate provisioning ticket.
- Extending the Ready gate / `spur k8s status` to detect konnectivity imbalance — optional follow-up.

### DIFF
--- a/crates/spur-core/src/k0s.rs
+++ b/crates/spur-core/src/k0s.rs
@@ -65,6 +65,12 @@ mod local_path_tests {
 /// the underlay to leave room for WireGuard's ~50-byte overhead, avoiding fragmentation). Returns
 /// `None` for any `cni` other than `"calico"` (the k0s default, kube-router, needs no config file).
 /// `sans` are extra API-server certificate SANs (e.g. the control-plane's mesh + underlay IPs).
+///
+/// `cp_count` is the number of control planes. For a multi-CP (`cp_count > 1`) cluster there is no
+/// VIP that can float over WireGuard cryptokey routing, so each node needs a local balanced endpoint
+/// across all controllers: enabling node-local load balancing (EnvoyProxy) gives konnectivity a
+/// cluster-wide `:8132` endpoint instead of pinning every agent to one controller. Single-CP clusters
+/// do not need it.
 pub fn k0s_controller_config_yaml(
     cni: &str,
     pod_cidr: &str,
@@ -72,6 +78,7 @@ pub fn k0s_controller_config_yaml(
     cni_mtu: u16,
     api_address: &str,
     sans: &[String],
+    cp_count: usize,
 ) -> Option<String> {
     if cni != "calico" {
         return None;
@@ -97,6 +104,11 @@ pub fn k0s_controller_config_yaml(
     y.push_str("    calico:\n");
     y.push_str("      mode: bird\n");
     y.push_str(&format!("      mtu: {cni_mtu}\n"));
+    if cp_count > 1 {
+        y.push_str("    nodeLocalLoadBalancing:\n");
+        y.push_str("      enabled: true\n");
+        y.push_str("      type: EnvoyProxy\n");
+    }
     Some(y)
 }
 
@@ -178,6 +190,7 @@ mod k0s_config_tests {
             1450,
             "192.0.2.1",
             &["192.0.2.1".to_string(), "203.0.113.9".to_string()],
+            1,
         )
         .unwrap();
         assert!(y.contains("address: 192.0.2.1"));
@@ -197,9 +210,44 @@ mod k0s_config_tests {
             "198.51.100.0/24",
             1450,
             "192.0.2.1",
-            &[]
+            &[],
+            3,
         )
         .is_none());
+    }
+
+    #[test]
+    fn multi_cp_calico_enables_node_local_load_balancing() {
+        // Documentation-range addresses only (RFC 5737 TEST-NET); no real infrastructure IPs.
+        let y = k0s_controller_config_yaml(
+            "calico",
+            "192.0.2.0/24",
+            "198.51.100.0/24",
+            1450,
+            "192.0.2.1",
+            &["192.0.2.1".to_string()],
+            3,
+        )
+        .unwrap();
+        assert!(y.contains("nodeLocalLoadBalancing:"));
+        assert!(y.contains("enabled: true"));
+        assert!(y.contains("type: EnvoyProxy"));
+    }
+
+    #[test]
+    fn single_cp_calico_omits_node_local_load_balancing() {
+        // Documentation-range addresses only (RFC 5737 TEST-NET); no real infrastructure IPs.
+        let y = k0s_controller_config_yaml(
+            "calico",
+            "192.0.2.0/24",
+            "198.51.100.0/24",
+            1450,
+            "192.0.2.1",
+            &["192.0.2.1".to_string()],
+            1,
+        )
+        .unwrap();
+        assert!(!y.contains("nodeLocalLoadBalancing"));
     }
 }
 
--- a/crates/spur-core/src/k0s.rs
+++ b/crates/spur-core/src/k0s.rs
@@ -66,11 +66,9 @@ mod local_path_tests {
 /// `None` for any `cni` other than `"calico"` (the k0s default, kube-router, needs no config file).
 /// `sans` are extra API-server certificate SANs (e.g. the control-plane's mesh + underlay IPs).
 ///
-/// `cp_count` is the number of control planes. For a multi-CP (`cp_count > 1`) cluster there is no
-/// VIP that can float over WireGuard cryptokey routing, so each node needs a local balanced endpoint
-/// across all controllers: enabling node-local load balancing (EnvoyProxy) gives konnectivity a
-/// cluster-wide `:8132` endpoint instead of pinning every agent to one controller. Single-CP clusters
-/// do not need it.
+/// For a multi-CP cluster (`cp_count > 1`) no VIP can float over WireGuard cryptokey routing, so
+/// node-local load balancing (EnvoyProxy) is enabled to give konnectivity a cluster-wide balanced
+/// endpoint instead of pinning every agent to one controller.
 pub fn k0s_controller_config_yaml(
     cni: &str,
     pod_cidr: &str,
--- a/crates/spurctld/src/cluster_k8s.rs
+++ b/crates/spurctld/src/cluster_k8s.rs
@@ -776,9 +776,8 @@ async fn fetch_component_state(cluster: &ClusterManager, node: &str) -> Option<S
 }
 
 /// The mesh-native k0s controller config for `node` (api on its mesh IP + Calico bird), or None for
-/// the default kube-router mode (`cni != "calico"`) / a node without a mesh IP. `cp_count` is the
-/// number of control planes; a multi-CP cluster additionally enables node-local load balancing so
-/// konnectivity gets a cluster-wide endpoint over the mesh.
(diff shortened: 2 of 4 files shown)


crates/spurctld/src/cluster_k8s.rs
```diff
@@ -841,7 +849,7 @@ async fn converge_provisioning(
         }
         // Mesh-native cluster: generate the k0s config (api on the mesh IP + Calico bird) when
         // cni=calico; None keeps the default kube-router. The bootstrap seeds etcd — no join token.
-        let k0s_config = controller_k0s_config(net, node);
+        let k0s_config = controller_k0s_config(net, node, cp_count);
         spawn_start_component(cluster, &node.name, role, None, k0s_config, None);
```
yansun1996, line 852: A few lines up, this whole branch is only reached when the node's `fetch_component_state` is NOT `"active"` — the same is true of the secondary-CP loop below (~line 892) for `K0sRole::Controller`. On an already-`Ready` HA cluster every controller is already `active`, so after this fix ships, upgrading never regenerates their `k0s.yaml` — and workers never receive `k0s_config` at all (only `K0sRole::Controller` gets it), so restarting workers alone can't apply NLLB to an existing cluster either. Would it be worth adding a config-drift check here (or forcing already-active controllers through one restart on upgrade) so existing HA clusters actually pick this up? Otherwise the note in the docs should say to reprovision the control plane rather than "restart the workers."

docs/deployment/managed-kubernetes.rst
```diff
@@ -140,6 +140,21 @@ over ``--replicas``. The first control-plane node is the etcd bootstrap.
...
+cannot follow WireGuard cryptokey routing — so ``spur k8s up`` enables k0s
+node-local load balancing (``nodeLocalLoadBalancing`` with ``EnvoyProxy``) on
+every control plane. This gives each node a local Envoy that round-robins across
+all controllers, so konnectivity has a cluster-wide ``:8132`` endpoint instead of
+pinning every agent to a single controller. It is on automatically for 3/5
+control planes and off for a single control plane (no balancing needed).
+
+.. note::
+
+   k0s does not hot-reload node-local load balancing. A fresh
```
yansun1996: Given the gating in `converge_provisioning`, restarting workers won't actually apply NLLB to an existing cluster — the controllers themselves never regenerate `k0s.yaml` while already active, and workers don't carry the config to begin with. Could this note instead say the control plane needs to be reprovisioned (e.g. `spur k8s down --reset` + `spur k8s up`) until there's a way to refresh an already-active controller's config in place?

crates/spur-core/src/k0s.rs
```diff
@@ -65,13 +65,20 @@ mod local_path_tests {
 /// the underlay to leave room for WireGuard's ~50-byte overhead, avoiding fragmentation). Returns
 /// `None` for any `cni` other than `"calico"` (the k0s default, kube-router, needs no config file).
 /// `sans` are extra API-server certificate SANs (e.g. the control-plane's mesh + underlay IPs).
+///
```
yansun1996, line 68: This doc comment block is now 5 lines — could trim to 2-3 and keep just the essential "why" (no floating VIP over WireGuard cryptokey routing → NLLB needed), with any extra background left to the commit message.

crates/spurctld/src/cluster_k8s.rs
```diff
@@ -776,8 +776,14 @@ async fn fetch_component_state(cluster: &ClusterManager, node: &str) -> Option<S
 }
 
 /// The mesh-native k0s controller config for `node` (api on its mesh IP + Calico bird), or None for
-/// the default kube-router mode (`cni != "calico"`) / a node without a mesh IP.
-fn controller_k0s_config(net: &ClusterNetworking, node: &spur_core::node::Node) -> Option<String> {
+/// the default kube-router mode (`cni != "calico"`) / a node without a mesh IP. `cp_count` is the
```
yansun1996: Similarly, this grew to 4 lines — a one-line addition ("`cp_count > 1` also enables node-local load balancing for konnectivity") would keep it consistent with the surrounding style.

yansun1996 (commented):
Nice root-cause writeup and clean fix for the calico path. One thing I want to flag before this merges: the migration note for existing clusters doesn't look right given how `converge_provisioning` is gated — see the inline comment. The rest is minor (comment length, a docs precision nit).

docs/deployment/managed-kubernetes.rst
```diff
@@ -140,6 +140,21 @@ over ``--replicas``. The first control-plane node is the etcd bootstrap.
    scope the cluster to the whole inventory, which trivially satisfies this —
    any registered node is then a valid control-plane choice.
 
+On a multi-control-plane mesh cluster there is no floating VIP — a VRRP address
+cannot follow WireGuard cryptokey routing — so ``spur k8s up`` enables k0s
+node-local load balancing (``nodeLocalLoadBalancing`` with ``EnvoyProxy``) on
+every control plane. This gives each node a local Envoy that round-robins across
+all controllers, so konnectivity has a cluster-wide ``:8132`` endpoint instead of
+pinning every agent to a single controller. It is on automatically for 3/5
```
yansun1996: "on automatically for 3/5 control planes" — the code's actual condition is `cp_count > 1`; it's only 3/5 in practice because `--replicas` is restricted elsewhere. Minor, but "any multi-control-plane count (today 3 or 5)" would be more precise and wouldn't need updating if that validator ever changes.

crates/spur-core/src/k0s.rs
```diff
@@ -197,10 +210,45 @@ mod k0s_config_tests {
...
             "192.0.2.1",
-            &[]
+            &[],
+            3,
         )
         .is_none());
     }
+
+    #[test]
+    fn multi_cp_calico_enables_node_local_load_balancing() {
```
yansun1996, line 218: Both new tests here exercise `k0s_controller_config_yaml` directly (pure YAML generation) — neither exercises `converge_provisioning`'s active-node gating, so neither would have caught the issue noted on `cluster_k8s.rs`. Worth adding a regression test that simulates an already-active controller and asserts whether its config gets regenerated on a later tick, if the gating logic changes to address that.

crates/spurctld/src/cluster_k8s.rs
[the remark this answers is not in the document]
spraveenio: Confirmed and agreed — the active-node gating means an already-`active` controller `continue`s before `controller_k0s_config` is ever called, and workers never receive `k0s_config` at all, so restarting workers cannot apply NLLB to an existing HA cluster.

Resolving this PR docs-only: the migration note now states that an already-active control plane is never rewritten in place and that an existing HA cluster must reprovision the control plane (`spur k8s down --reset` + `spur k8s up`) rather than restart workers. The fix itself is correct for a fresh `spur k8s up` (validated on hardware).

An in-place config-drift refresh for already-active controllers (regenerate + rolling-restart on upgrade) is a real feature with its own testing surface, so I'm deferring it to a separate ticket rather than expanding this PR.

spraveenio: Good catch — you're right on both counts: the secondary-CP loop is also gated on `!= "active"`, and only `K0sRole::Controller` gets `k0s_config` so workers never carry it. Restarting workers alone can't apply NLLB to an already-`Ready` cluster.

For this PR I've taken the docs-only route you suggested: the note now says to reprovision the control plane (`spur k8s down --reset` + `spur k8s up`) instead of restarting workers. The config-drift check / forced one-restart-on-upgrade for already-active controllers is worth doing but has its own testing surface, so I'll file it as a follow-up rather than fold it in here.

docs/deployment/managed-kubernetes.rst
spraveenio: Done — reworded exactly along these lines. The note now explains that Spur only writes a controller's `k0s.yaml` while bringing it up (an already-active control plane is never rewritten in place), and directs existing HA clusters to reprovision the control plane with `spur k8s down --reset` followed by `spur k8s up`.

crates/spur-core/src/k0s.rs
spraveenio: Trimmed to 3 lines, keeping just the "why" (no floating VIP over WireGuard cryptokey routing → NLLB for a cluster-wide konnectivity endpoint).

crates/spurctld/src/cluster_k8s.rs
spraveenio: Trimmed to a one-line addition: "`cp_count > 1` also enables node-local load balancing for konnectivity."

docs/deployment/managed-kubernetes.rst
spraveenio: Reworded to "on automatically for any multi-control-plane count (today 3 or 5)" so it matches the actual `cp_count > 1` condition and won't need updating if the `--replicas` validator changes.

crates/spur-core/src/k0s.rs
spraveenio: Agreed the existing two tests only cover pure YAML generation, not the active-node gating. Since this PR keeps the gating as-is and resolves the upgrade concern in docs (reprovision the control plane), there's no new gating behavior to assert here yet. I'll bundle the gating-level regression test with the follow-up ticket that adds in-place config refresh for already-active controllers, where the behavior it would guard actually changes.

yansun1996 (dismissed):
LGTM

PR #712, opened 2026-08-21
test: cover resource accounting and untested CLI helpers

SumonAMD:
## Summary

Coverage measurement on `main` put `spur-core/src/resource.rs` at 57% of functions, and left `sstat.rs` with no tests at all. The release path of the accounting types and sstat's field parsing had nothing holding their behaviour in place, so a regression in either would have been silent.

This adds 16 tests across those two files. **No production code changes** — every diff is an appended `#[cfg(test)]` module, so no existing code moves.

## What's covered

**`spur-core/src/resource.rs`** (10 tests) — takes the file to 100% of functions

- `subtract` saturating at zero on over-release instead of wrapping to a huge number, and pruning a device drained to zero so `has_devices` goes false
- `subtract` treating a release of never-held gres as a no-op
- `allocated_count` filtering GPUs by type (`mi300x`, `mi210`, `any`, `None`, unknown) and summing countable non-GPU gres by amount rather than by device tally
- The `is_empty` / `has_devices` split on a leftover empty device vector, so the two aren't mistaken for each other
- `build_exclusive_allocation` claiming every CPU and GPU, taking memory from the caller, and creating no entry for absent hardware or zero counts
- `aggregate_allocations` summing scalars, merging entries that share a `device_id`, and returning empty for an empty iterator
- `total_gpus` counting devices while `gpu_counts` groups by type

**`spur-cli/src/sstat.rs`** (6 tests) — field-list aliases (`ncpus`/`cpus`, `reqmem`, `gres`) with case-insensitive matching and trimming, unknown fields skipped without dropping valid neighbours, `format_duration` rolling over to `D-HH:MM:SS` exactly at the 86400s boundary, negative durations (clock skew on a signed proto duration) formatting as a magnitude, header widths matching rendered headers, and out-of-range states falling back to `UNKNOWN`.

## Coverage

| File | Functions before | Functions after |
| --- | --- | --- |
| `spur-core/src/resource.rs` | 57.45% | 100% |
| `spur-cli/src/sstat.rs` | 0% | 53.85% |

## Relation to other PRs

- `net.rs` and `token.rs` were originally part of this PR. They now sit in #716 instead, alongside the arithmetic fix their boundary cases document, so the two PRs touch disjoint files and can merge in either order.
- The `sprio`/`sshare` `-h` collision fix and the clap parser-validation test are deliberately **not** here — #697 already covers both.

## Test plan

- [x] `cargo test --locked` (full workspace) — 3151 passed, 0 failed
- [x] `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` — 0 warnings, 0 errors
- [x] `cargo build --all-targets --locked`
- [x] `scripts/check-license-headers.sh` on all changed files
- [x] `.githooks/pre-commit` passed

### DIFF
--- a/crates/spur-cli/src/sstat.rs
+++ b/crates/spur-cli/src/sstat.rs
@@ -258,3 +258,59 @@ fn format_duration(total_seconds: i64) -> String {
         format!("{:02}:{:02}:{:02}", hours, minutes, seconds)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn parse_field_list_accepts_aliases_case_and_padding() {
+        assert_eq!(parse_field_list("jobid"), [StatField::JobId]);
+        assert_eq!(parse_field_list("NCPUS"), [StatField::Cpus]);
+        assert_eq!(parse_field_list("cpus"), [StatField::Cpus]);
+        assert_eq!(parse_field_list("reqmem"), [StatField::MemAlloc]);
+        assert_eq!(parse_field_list("gres"), [StatField::GpuAlloc]);
+        assert_eq!(
+            parse_field_list("  JobID , Elapsed  "),
+            [StatField::JobId, StatField::Elapsed]
+        );
+    }
+
+    #[test]
+    fn parse_field_list_skips_unknown_fields_but_keeps_the_rest() {
+        assert_eq!(
+            parse_field_list("jobid,bogus,elapsed"),
+            [StatField::JobId, StatField::Elapsed]
+        );
+        assert!(parse_field_list("bogus").is_empty());
+    }
+
+    #[test]
+    fn format_duration_rolls_over_into_days() {
+        assert_eq!(format_duration(0), "00:00:00");
+        assert_eq!(format_duration(59), "00:00:59");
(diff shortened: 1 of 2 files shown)


sajmera-pensando (approved):
Test-only PR — all 248 added lines are inside `#[cfg(test)]` blocks, so there are zero production code changes and no risk of regression.

**What's here:** 16 unit tests across two files:
- `spur-core/src/resource.rs` (10 tests): brings function coverage from 57% to 100%. Locks in saturating subtraction (over-release clamps to zero rather than wrapping), device pruning, GPU type filtering, exclusive allocation, and multi-node aggregation.
- `spur-cli/src/sstat.rs` (6 tests): locks in field aliases (`ncpus`/`cpus`), case-insensitive parsing, silent unknown-field skipping, `D-HH:MM:SS` rollover at 86400s, and negative duration handling.

**One question for a follow-up:** `is_empty_and_has_devices_disagree_on_empty_device_vec` documents a state where a device key exists but holds an empty vector. Is that state actually reachable in production, or is it hypothetical? If it's reachable, it may indicate a gap in the pruning logic in `subtract` — worth a quick check, not a blocker for this PR.

LGTM otherwise.

Issue #439, opened 2026-07-14
Scheduler: GPU resources are not enforced by QoS limits and not tracked for fair-share


powderluv:
### Summary
GPU resources are never enforced by QoS limits and never recorded for fair-share. On a GPU cluster this means there is effectively **no GPU quota and no GPU-aware fairness** — the resource that matters most is unmanaged.

### Evidence
- **QoS limits ignore GPU.** `sum_running_tres` (`crates/spurctld/src/cluster.rs:3920`) totals only CPU / Node / Memory:
  ```rust
  let (mut cpu, mut node, mut mem) = (0u64, 0u64, 0u64);
  ... tres.set(TresType::Cpu, cpu); tres.set(TresType::Node, node); tres.set(TresType::Memory, mem);
  ```
  `TresType::Gpu` is parseable in `grp_tres` / `max_tres_per_job` / `max_tres_per_user` strings but is never summed or checked, so a `gpu=N` cap is silently a no-op. (`max_tres_per_user` additionally checks CPU only.)
- **GPU usage is never written.** `update_usage` (`crates/spurctld/src/accounting/db.rs:250`) computes `cpu_seconds = duration * num_tasks * cpus_per_task` and the `INSERT` (line 262) only writes `cpu_seconds`; `gpu_seconds` (schema `db.rs:65`) stays `0`.
- **Fair-share is CPU-only.** `compute_fairshare` (`crates/spurctld/src/accounting/fairshare.rs:31`) decays `record.cpu_seconds` only.

### Impact
A user can request GPUs well beyond any `gpu:N` QoS cap with no effect, and GPU consumption never influences fair-share priority. For an AMD GPU scheduler this is the highest-value quota dimension and it is unmanaged.

### Suggested fix
- Sum `TresType::Gpu` in `sum_running_tres` and check it in `check_qos_limits` (CPU/Node/Memory + GPU across `max_tres_per_job`, `max_tres_per_user`, `grp_tres`).
- Populate `gpu_seconds` in `update_usage` and include a GPU term (optionally TRES-billing-weighted) in `compute_fairshare`.

_Surfaced during the SPUR-managed-k0s quota-enforcement design analysis._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #714, opened 2026-08-21
fix(spurctld): count distinct nodes for the QoS group node limit (#709)

### LINKED ISSUE #709: [Issue]: Resources underutilized because of incorrect QosGrpNodeLimit calculation.
### Problem Description

Node has 8 GPUs. One job is use 1 GPU. But cant run another job on the node using other GPUs.
Wastage of GPUs.

=================
Evidence:

```
snalluru@crs-m2m-cpu-spur-009:~$ spur version
spur 0.8.0 (9ea4beec)

# my qos has node=4 limit
snalluru@crs-m2m-cpu-spur-009:~$ spur sacctmgr show qos | grep cvs
amd-cvs-qos     10000    off        1                                                                                                  node=4

# some other user is using 2 nodes already.
snalluru@crs-m2m-cpu-spur-009:~$ squeue -A amd-cvs
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
             31015  amd-spur     wrap speriasw  R 1-02:33:42      2 crsuse2-m2m-[228,277]
			 
# I started a job on 2 nodes (but only 1 gpu). The job starts running. The other GPUs on the node are free.
snalluru@crs-m2m-cpu-spur-009:~$ srun -A amd-cvs -n 2 -N 2 sleep 60
srun: Pending job allocation 37255...
srun: job 37255 running on crsuse2-m2m-[044,250]


# When I start another job (all I need is 1 gpu), it does not run. It waits..
snalluru@crs-m2m-cpu-spur-009:~$ srun -A amd-cvs -n 2 -N 2 sleep 60
srun: Pending job allocation 37257...

# Waits because it reaches QOSGrpNodeLimit
snalluru@crs-m2m-cpu-spur-009:~$ squeue -A amd-cvs
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
             37257  amd-spur    sleep snalluru PD       0:00      2 (QOSGrpNodeLimit)
             31015  amd-spur     wrap speriasw  R 1-02:34:28      2 crsuse2-m2m-[228,277]
             37255  amd-spur    sleep snalluru  R       0:43      2 crsuse2-m2m-[044,250]
```

Not fully sure if this is related to  #439 

### Operating System

Ubuntu "24.04.3 LTS (Noble Numbat)"

### CPU

 AMD EPYC 9654 96-Core Processor

### GPU

 AMD Instinct MI355X

### ROCm Version

7.0.1

### ROCm Component

_No response_

### Steps to Reproduce

_No response_

<details>

```
Paste output here
```

</details>

### Additional Information

_No response_

powderluv:
Fixes #709.

## Problem

An account with a QoS `node=N` limit hits `QOSGrpNodeLimit` and stops scheduling even though its nodes have free GPUs. Root cause: the running-side node usage is the **sum of each running job's requested `num_nodes`**, not the count of **distinct physical nodes** the account/QoS actually occupies.

- `job_tres` sets `TresType::Node = job.spec.num_nodes` (`cluster.rs`).
- `sum_running_tres` accumulates `job_tres` over the account's/QoS's running jobs, so two jobs that share a physical node double-count it.
- That total is enforced in `qos.rs` (`qos_resource_breach`) and `account_limits.rs`.

`Job.allocated_nodes` (the real placement, set at `JobStart`) was available but unused by the accounting. Slurm counts the association/QoS node usage from a per-assoc node bitmap — a shared node counts once.

## Fix

In `sum_running_tres`, compute `Node` as the size of the **union of `allocated_nodes`** across the matching running jobs (a `HashSet`), overriding the summed value. CPU/Memory/GPU stay additive (they are consumed per job even on a shared node). A running job with no recorded placement falls back to its requested `num_nodes` so the limit is never under-counted. Single choke point — one function, no signature change — so all three running-side consumers (QoS grp, QoS per-user, account grp) are corrected at once. The pending side and `PassReservations` deliberately keep `num_nodes` (an unplaced job has no placement to union against).

## Tests

- `sum_running_tres_counts_distinct_nodes_not_summed_num_nodes` — two running jobs `{n1,n2}`+`{n2,n3}` → Node = 3, not 4.
- `sum_running_tres_falls_back_to_num_nodes_when_placement_missing` — unplaced running job counts its requested nodes.
- `qos_grp_node_uses_distinct_allocated_nodes_not_summed_num_nodes` — end-to-end #709 repro through `pending_jobs()`: grp `node=4`, running jobs share a node (distinct=3), a new 1-node job is admitted (3+1=4 ≤ 4) instead of blocked (phantom 4+1=5).
- `qos_grp_node_still_blocks_when_distinct_cap_reached` — guards against over-admitting.

The two repro tests were confirmed to fail before the change and pass after.

## Scope / follow-up

Resolves the running-side over-count (the reporter's stated scenario). It does **not** cover the pure-packing case where the account already sits at the full distinct-node cap and a new job would land *only* on already-counted nodes (needing 0 new distinct nodes) — that requires making the pending side overlap-aware, i.e. selecting candidate nodes before the limit re-check, an inversion of the current gate-before-select ordering. Correctly a follow-up. Distinct from #439 (GPU-per-node not tracked for QoS/fairshare), though both surface as "nodes look full while GPUs are free."

### DIFF
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -6683,12 +6683,32 @@ fn partition_limit_block(job: &Job, part: &Partition) -> Option<spur_core::job::
 
 fn sum_running_tres(jobs: &HashMap<JobId, Job>, pred: impl Fn(&Job) -> bool) -> TresRecord {
     let mut tres = TresRecord::new();
+    // Node TRES is the count of DISTINCT physical nodes the matching running
+    // jobs occupy (the union of their `allocated_nodes`), not the sum of each
+    // job's requested `num_nodes`. Slurm derives the association/QOS node usage
+    // from a per-assoc/QOS node bitmap, so a node shared by two of an
+    // account's/QOS's jobs counts once. Summing `num_nodes` double-counts a
+    // packed node and blocks work the node still has capacity for (ROCm/spur#709).
+    // CPU/Memory/GPU stay additive — those are consumed per job even on a
+    // shared node — so only Node is overridden below.
+    let mut distinct_nodes: HashSet<&str> = HashSet::new();
+    let mut unplaced_nodes: u64 = 0;
     for j in jobs.values() {
         if j.state != JobState::Running || !pred(j) {
             continue;
         }
         tres.add(&job_tres(j));
+        if j.allocated_nodes.is_empty() {
+            // Running job without recorded placement (transient/edge): fall back
+            // to its requested count so the limit is never under-counted.
+            unplaced_nodes += j.spec.num_nodes as u64;
+        } else {
+            distinct_nodes.extend(j.allocated_nodes.iter().map(String::as_str));
+        }
     }
+    // `job_tres` set Node to the summed `num_nodes` above; replace it with the
+    // real distinct-node occupancy.
+    tres.set(TresType::Node, distinct_nodes.len() as u64 + unplaced_nodes);
     tres
 }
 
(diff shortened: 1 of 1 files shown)


shreyajmera (approved):
lgtm

sajmera-pensando (approved):
LGTM

PR #716, opened 2026-08-21
fix: use checked arithmetic for CIDR prefixes and token TTLs

### LINKED ISSUE #715: [Issue]: unchecked integer arithmetic on CIDR prefixes and token TTLs aborts in debug, wraps in release
### Problem Description

Three helpers in `spur-cli` and `spur-net` do arithmetic on CLI- and config-supplied
values without checking it, so out-of-range input aborts the process in a debug build and,
in a release build, wraps into either a silently wrong value or a misleading error. All
three inputs are plain long flags or config fields with no validation in front of them.

## 1. `address_network` overflows on a /0 or above /32

`crates/spur-cli/src/net.rs`:

    let mask = !((1u32 << (32 - prefix_len)) - 1);

`prefix_len` comes straight from `#[arg(long, default_value_t = 16)]`, so any value is
accepted. A `/0` shifts a `u32` by a full 32 bits, and anything above `/32` underflows
the `32 - prefix_len` subtraction on `u8` before the shift is even reached.

Reproduced by calling the helper directly:

    address_network("10.42.7.9", 0)
    thread panicked at crates/spur-cli/src/net.rs:456:18:
    attempt to shift left with overflow

    address_network("10.42.7.9", 33)
    thread panicked at crates/spur-cli/src/net.rs:456:27

Reachable in production from `spur net join --prefix-len 0`, where the result becomes a
WireGuard peer's `allowed_ips` at `net.rs:251`. In release builds the shift wraps rather
than aborting, so the peer is configured with a nonsense mask instead of failing.

**Expected:** `/0` yields the empty mask `0.0.0.0`; a prefix above `/32` is rejected with
a clear error.

## 2. `parse_ttl` overflows on a large TTL

`crates/spur-cli/src/token.rs`:

    Ok(days.parse::<u32>()? * 86400)

The `h` and `m` branches have the same shape. `50000d` is 4,320,000,000 seconds, past
`u32::MAX`:

    parse_ttl("50000d")
    thread panicked at crates/spur-cli/src/token.rs:121:12:
    attempt to multiply with overflow

In a release build it wraps to 25,032,704 seconds, about 290 days, and that value flows
into `generate_token` as the token's lifetime. The caller asked for roughly 137 years and
silently gets 290 days. The wrap always shortens rather than extends the lifetime, so this
is a correctness bug rather than a security hole.

**Expected:** a TTL that cannot be represented is rejected rather than wrapped.

## 3. `AddressPool::new` overflows on a /0 CIDR

`crates/spur-net/src/address.rs`:

    if prefix_len > 30 {
        return Err(AddressError::InvalidCidr("prefix length must be <= 30".into()));
    }
    let mask = !((1u32 << (32 - prefix_len)) - 1);

The guard bounds only the top of the range, so a `/0` passes it and reaches the same
unchecked shift as defect 1. `allocate` would do the same to its `host_bits` at
`address.rs:94`:

    AddressPool::new("10.0.0.0/0")
    thread panicked at crates/spur-net/src/address.rs:72:22:
    attempt to shift left with overflow

Reachable from `spur net init --cidr 10.0.0.0/0`, which passes the string straight to
`AddressPool::new` at `net.rs:192`. It is also reachable through config: `wg_cidr` is
validated by `is_valid_cidr`, which only checks `prefix <= 32`, and then flows to
`AddressPool::new` at `cluster_k8s.rs:218`. The existing tests cover `/33` but never `/0`.

**Expected:** a `/0` is rejected with a clear error, since a pool needs host bits to hand
out — which is what the `<= 30` bound already exists for.

## Suggested fix

Validate the prefix length up front and handle `/0` explicitly in `address_network`;
replace the TTL multiplication with `checked_mul`; widen the existing `AddressPool` prefix
guard to a range so one check covers both of its overflow sites. All three are small, local
changes. Note that `/0` should stay *valid* for `address_network`, where `0.0.0.0/0` is a
legitimate WireGuard `allowed_ips` default route, but should be *rejected* by
`AddressPool::new` — the two call sites want different answers.

Found on `main` at d6001b47.

### Operating System

Ubuntu 24.04 LTS (Noble Numbat)

### CPU

AMD EPYC 9575F 64-Core Processor

### GPU

8 x AMD Instinct MI350X (gfx950)

### ROCm Version

ROCm 7.14.0

### ROCm Component

_No response_

### Steps to Reproduce

All three defects are in pure helper functions, so they reproduce from a checkout with no
cluster, no GPU and no daemon running.

```bash
git clone https://github.com/ROCm/spur.git && cd spur
git checkout d6001b47

# 1. address_network: /0 shift overflow, and >/32 subtraction underflow
cat >> crates/spur-cli/src/net.rs <<'EOF'

#[cfg(test)]
mod repro_tests {
    use super::*;

    #[test]
    fn zero_prefix_should_give_empty_mask() {
        assert_eq!(address_network("10.42.7.9", 0).unwrap(), "0.0.0.0");
    }

    #[test]
    fn prefix_above_32_should_be_rejected() {
        assert!(address_network("10.42.7.9", 33).is_err());
    }
}
EOF

# 2. parse_ttl: multiply overflow
cat >> crates/spur-cli/src/token.rs <<'EOF'

#[cfg(test)]
mod repro_tests {
    use super::*;

    #[test]
    fn oversized_ttl_should_be_rejected() {
        assert!(parse_ttl("50000d").is_err());
    }
}
EOF

# 3. AddressPool::new: /0 slips past the >30 guard into the same shift
cat >> crates/spur-net/src/address.rs <<'EOF'

#[cfg(test)]
mod repro_tests {
    use super::*;

    #[test]
    fn zero_prefix_cidr_should_be_rejected() {
        assert!(AddressPool::new("10.0.0.0/0").is_err());
    }
}
EOF

cargo test -p spur-cli -p spur-net repro_tests
```
**Output:**
```
test net::repro_tests::zero_prefix_should_give_empty_mask ... FAILED
  panicked at crates/spur-cli/src/net.rs:456:18: attempt to shift left with overflow
test net::repro_tests::prefix_above_32_should_be_rejected ... FAILED
  panicked at crates/spur-cli/src/net.rs:456:27
test token::repro_tests::oversized_ttl_should_be_rejected ... FAILED
  panicked at crates/spur-cli/src/token.rs:121:12: attempt to multiply with overflow
test address::repro_tests::zero_prefix_cidr_should_be_rejected ... FAILED
  panicked at crates/spur-net/src/address.rs:72:22: attempt to shift left with overflow
```

rocminfo --support output

```
Paste output here
```

### Additional Information

**Hardware-independent.** All three defects are integer arithmetic in `spur-cli` and
`spur-net` helper functions on values taken from command-line flags or config fields.
Nothing in any of these paths touches the GPU, HSA or the driver, so the CPU/GPU/ROCm
fields above describe only the machine that happened to reproduce it.

**Debug versus release matters.** Rust aborts on overflow in debug builds, which is what
the panics above show. A release build wraps silently instead, which is the worse outcome:

- `address_network("10.42.7.9", 0)` returns a nonsense mask instead of failing, and that
  value is written into a WireGuard peer's `allowed_ips` at `net.rs:251`.
- `parse_ttl("50000d")` wraps 4,320,000,000 to 25,032,704 seconds, so a token requested
  for roughly 137 years is issued for about 290 days. The wrap always shortens the
  lifetime rather than extending it, so this is a correctness bug and not a privilege
  issue.
- `AddressPool::new("10.0.0.0/0")` computes an all-ones mask instead of aborting, leaving
  the base unmasked, and `allocate` then derives `max_host = 0` and reports
  `PoolExhausted`. So `spur net init --cidr 10.0.0.0/0` fails with a misleading "address
  pool exhausted" rather than a clear complaint about the prefix.

**Environment**
- spur 0.8.0 at `main` commit d6001b47
- rustc 1.97.1, cargo 1.97.1
- Kernel 6.8.0-38-generic

A fix is ready and will follow as a PR: validate the prefix length up front and handle
`/0` explicitly, replace the TTL multiplication with `checked_mul`, and widen the
`AddressPool` prefix guard to a range. It includes regression tests for `/0`, above `/32`,
the overflowing TTL, and the `/0` CIDR, all four of which fail against the current code.

SumonAMD:
## Summary

Nine new tests covering `/0`, `/33` and `u8::MAX` prefixes alongside the valid `/8`, `/16`, `/24`, `/25` and `/32` masks and malformed addresses including a truncated IP; TTL suffix expansion with surrounding padding, overflow rejection, and rejection of non-numeric and bare-suffix input; and `/0` CIDR rejection plus the `/1` boundary that must still allocate. The last commit folds in four assertions that previously lived in #712, so that PR could drop these two files and stop overlapping with this one.

- `address_network` (`spur-cli/src/net.rs`) built its WireGuard netmask with `!((1u32 << (32 - prefix_len)) - 1)`. A `/0`, which `spur net join --prefix-len 0` accepts, shifted a u32 by a full 32 bits, and any prefix above `/32` underflowed the `32 - prefix_len` subtraction before the shift was even reached. The result feeds a peer's `allowed_ips`.
- `parse_ttl` (`spur-cli/src/token.rs`) multiplied without checking, so `spur token create --ttl 50000d` overflowed u32 and wrapped to roughly 290 days. The caller asked for ~137 years and got a much shorter-lived token with no error.
- `AddressPool::new` (`spur-net/src/address.rs`) bounded only the top of the prefix range, so a `/0` passed the guard and reached the same unchecked shift; `allocate` would have done the same to its `host_bits`. It is reachable from `spur net init --cidr 10.0.0.0/0`, and from config as well, since `is_valid_cidr` only checks `prefix <= 32`.

Closes #715

## Approach

- `net.rs`: validate the prefix length up front and mask with an explicit `/0` case. This is the same guard shape already used by `cidr_contains` in `spurctld/src/cluster_k8s.rs` (`prefix > 32` bail plus an explicit `prefix == 0`). `u32::MAX << (32 - n)` is arithmetically identical to the old expression for every valid `n`, including `/32`, so behavior is unchanged for input that already worked.
- `token.rs`: fold the four multiply branches into a single `checked_mul` that reports the overflow. `parse_ttl("4294967295")` still returns `u32::MAX`, so the accepted range is not narrowed — only the silent wrap is removed.
- `address.rs`: widen the existing prefix guard into a range instead of teaching the mask about an empty prefix. A pool needs host bits to hand out, which is what the upper bound already exists for, and one guard covers both overflow sites (`new`'s mask and `allocate`'s `host_bits`). Note that `/0` stays valid in `spur-cli`, where `0.0.0.0/0` is a legitimate `allowed_ips` default route — the two call sites genuinely want different answers.

## Testing

Nine new tests covering `/0`, `/33` and `u8::MAX` prefixes alongside the valid `/16`, `/24`, `/25` and `/32` masks and malformed addresses; TTL suffix expansion, overflow rejection and non-numeric rejection; and `/0` CIDR rejection plus the `/1` boundary that must still allocate.

Each new boundary test was verified to fail against the previous code, by reverting the production change and keeping the tests:

| Test | Failure without the fix |
|---|---|
| `address_network_handles_zero_prefix` | `net.rs:456` attempt to shift left with overflow |
| `address_network_rejects_prefix_above_32` | `net.rs:456` attempt to subtract with overflow |
| `parse_ttl_rejects_values_that_overflow` | `token.rs:121` attempt to multiply with overflow |
| `test_zero_prefix_rejected` | `address.rs:72` attempt to shift left with overflow |

- `cargo fmt --all --check` — clean.
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — 0 warnings.
- `cargo build --all-targets --locked` — clean.
- `cargo test --locked` — 3144 passed, 0 failed.

### DIFF
--- a/crates/spur-cli/src/net.rs
+++ b/crates/spur-cli/src/net.rs
@@ -453,7 +453,49 @@ fn cmd_mesh(
 /// Given an IP and prefix length, compute the network address.
 fn address_network(ip: &str, prefix_len: u8) -> Result<String> {
     let addr: Ipv4Addr = ip.parse().context("invalid IP address")?;
-    let mask = !((1u32 << (32 - prefix_len)) - 1);
+    if prefix_len > 32 {
+        bail!("invalid prefix length /{}, expected 0-32", prefix_len);
+    }
+    // A /0 has no network bits, and shifting a u32 by a full 32 would overflow.
+    let mask = if prefix_len == 0 {
+        0
+    } else {
+        u32::MAX << (32 - prefix_len)
+    };
     let network = u32::from(addr) & mask;
     Ok(Ipv4Addr::from(network).to_string())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn address_network_masks_host_bits() {
+        assert_eq!(address_network("10.42.7.9", 24).unwrap(), "10.42.7.0");
+        assert_eq!(address_network("10.42.7.9", 16).unwrap(), "10.42.0.0");
+        assert_eq!(address_network("10.42.7.130", 25).unwrap(), "10.42.7.128");
+        assert_eq!(address_network("10.42.7.9", 32).unwrap(), "10.42.7.9");
+    }
+
+    /// A /0 shifts the mask by 32 bits, which is where the unchecked shift
+    /// overflowed instead of yielding the empty mask.
+    #[test]
+    fn address_network_handles_zero_prefix() {
+        assert_eq!(address_network("10.42.7.9", 0).unwrap(), "0.0.0.0");
+    }
+
+    /// Above /32 the `32 - prefix_len` subtraction underflowed before the shift
+    /// was ever reached, so the length is now rejected up front.
+    #[test]
+    fn address_network_rejects_prefix_above_32() {
+        assert!(address_network("10.42.7.9", 33).is_err());
+        assert!(address_network("10.42.7.9", u8::MAX).is_err());
+    }
+
+    #[test]
+    fn address_network_rejects_malformed_addresses() {
+        assert!(address_network("not-an-ip", 24).is_err());
+        assert!(address_network("::1", 24).is_err());
+    }
 }
--- a/crates/spur-cli/src/net.rs
+++ b/crates/spur-cli/src/net.rs
@@ -474,6 +474,7 @@ mod tests {
     fn address_network_masks_host_bits() {
         assert_eq!(address_network("10.42.7.9", 24).unwrap(), "10.42.7.0");
         assert_eq!(address_network("10.42.7.9", 16).unwrap(), "10.42.0.0");
+        assert_eq!(address_network("10.42.7.9", 8).unwrap(), "10.0.0.0");
         assert_eq!(address_network("10.42.7.130", 25).unwrap(), "10.42.7.128");
         assert_eq!(address_network("10.42.7.9", 32).unwrap(), "10.42.7.9");
     }
@@ -496,6 +497,7 @@ mod tests {
     #[test]
     fn address_network_rejects_malformed_addresses() {
         assert!(address_network("not-an-ip", 24).is_err());
+        assert!(address_network("10.42.7", 24).is_err());
         assert!(address_network("::1", 24).is_err());
     }
 }
--- a/crates/spur-cli/src/token.rs
+++ b/crates/spur-cli/src/token.rs
@@ -3,7 +3,7 @@
 
 //! `spur token` subcommands for admission token management.
 
-use anyhow::Result;
+use anyhow::{Context, Result};
 use clap::{Parser, Subcommand};
 
 use spur_proto::proto::{CreateTokenRequest, ListTokensRequest, RevokeTokenRequest};
@@ -117,17 +117,21 @@ fn cmd_user_token(user: &str, admin: bool, ttl: Option<String>, config_path: &st
 
 fn parse_ttl(s: &str) -> Result<u32> {
     let s = s.trim();
-    if let Some(days) = s.strip_suffix('d') {
-        Ok(days.parse::<u32>()? * 86400)
+    let (value, unit_secs) = if let Some(days) = s.strip_suffix('d') {
+        (days, 86400)
     } else if let Some(hours) = s.strip_suffix('h') {
-        Ok(hours.parse::<u32>()? * 3600)
+        (hours, 3600)
     } else if let Some(mins) = s.strip_suffix('m') {
-        Ok(mins.parse::<u32>()? * 60)
+        (mins, 60)
     } else if let Some(secs) = s.strip_suffix('s') {
-        Ok(secs.parse::<u32>()?)
+        (secs, 1)
     } else {
-        Ok(s.parse::<u32>()?)
-    }
+        (s, 1)
+    };
+    value
+        .parse::<u32>()?
+        .checked_mul(unit_secs)
+        .with_context(|| format!("TTL {} exceeds {} seconds", s, u32::MAX))
 }
 
 async fn cmd_create(controller: &str, ttl: Option<String>) -> Result<()> {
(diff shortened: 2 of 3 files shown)


sajmera-pensando (approved):
Three real bugs, all correctly fixed, well tested.

**net.rs** — the `/0` case and the above-32 rejection are both right. Accepting `/0` in `address_network` (WireGuard default route) while rejecting it in `AddressPool::new` (needs host bits to allocate) is the correct distinction. The new formula `u32::MAX << (32 - n)` is equivalent to the old one for all valid non-zero inputs, so no behavior change on the happy path.

**token.rs** — refactoring the four branches into a single `checked_mul` is cleaner and harder to get wrong when adding a future suffix. Two minor nits: the error message `"TTL 50000d exceeds 4294967295 seconds"` mixes the days-denominated input with a seconds-denominated limit, which reads a little oddly; and a zero-value TTL (e.g. `"0d"`) currently succeeds and produces a 0-second token — probably worth a follow-up guard or at least a test.

**address.rs** — minimal, correct. One line, right fix.

Tests cover the specific bugs and boundary cases well. Good to merge.

PR #717, opened 2026-08-21
fix(spurctld): categorize dispatch confirmation failure reasons

yansun1996:
## Summary

A batch job whose dispatch was rejected by a node agent surfaced the same generic pending reason no matter the cause: `dispatch confirmation failed: N of M nodes confirmed`. Operators had no way to distinguish a GPU allocation mismatch from an unreachable agent from a prolog failure without digging into controller logs.

This classifies each per-node dispatch outcome into a specific category and builds the job-level pending reason from the category breakdown across all nodes, e.g.:

```
dispatch confirmation failed (0/1 confirmed): 1 gpu/resource allocation mismatch
```

Categories: `prolog failure`, `gpu/resource allocation mismatch`, `agent unreachable`, `agent rejected launch`, `no agent address`, `dispatch task panicked`, and a `dispatch error` catch-all. Multiple failing nodes with different causes show all of them, e.g. `1 gpu/resource allocation mismatch, 1 prolog failure`.

Tracked as SPUR-190.

## Approach

Extends the existing per-node `DispatchError` enum (which already separated `PrologFailed` and `ResourcesUnavailable`) by splitting its former `Other` catch-all into `Unreachable` (connect failure or an RPC that timed out without a response) and `AgentRejected` (an explicit non-prolog rejection message from the agent), and adds a `category()` method used to tally failures into a summary instead of a blind count. This is purely controller-side and additive — no proto/wire changes, so no compatibility risk.

**Known limitation:** the `gpu/resource allocation mismatch` category is inferred from the agent's gRPC status code (`ResourceExhausted`) rather than an explicit field on `LaunchJobResponse`, unlike `prolog failure` which already has a dedicated `LaunchFailureKind` proto value. A natural follow-up would move that category onto the same explicit, versioned mechanism.

## Testing

- Unit tests covering every category individually plus a mixed multi-node, multi-category case.
- `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` and `cargo fmt --all --check` clean.
- Validated end-to-end on an isolated single-node deployment for the prolog-failure path, confirming the categorized reason shows up correctly through the CLI. The GPU-mismatch path needs real GPU hardware state to trigger faithfully and is covered by unit tests only.

### DIFF
--- a/crates/spurctld/src/scheduler_loop.rs
+++ b/crates/spurctld/src/scheduler_loop.rs
@@ -936,7 +936,27 @@ enum DispatchError {
     /// The agent rejected the dispatch because the controller-allocated resources
     /// are already in use locally — the controller's view of this node is stale.
     ResourcesUnavailable,
+    /// Connect failed, or the RPC timed out without a response, as opposed to
+    /// the agent responding with an explicit rejection.
+    Unreachable(anyhow::Error),
+    /// The agent explicitly rejected the launch for a reason it does not have
+    /// a `LaunchFailureKind` for yet.
+    AgentRejected(String),
     Other(anyhow::Error),
+}
+
+impl DispatchError {
+    /// A short, stable label for grouping failures across nodes into one
+    /// job-level reason instead of a raw per-node error dump.
+    fn category(&self) -> &'static str {
+        match self {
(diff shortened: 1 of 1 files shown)


sgopinath1 (approved):
LGTM

PR #718, opened 2026-08-22
test(spurctld): reproduce issue #709's actual QOS grp node scenario

snalluru-amd:
## Motivation
 reproduce issue #709's actual QOS grp node scenario

## Technical Details
fix(spurctld)#714 stops running jobs that SHARE a node from being double-counted against a QOS/account grp node= limit, but issue #709's own reproduction has two running jobs on disjoint node pairs (no shared node between them), so summed num_nodes and the distinct-node union are numerically identical there and the fix changes nothing for it.

This adds a failing regression test, qos_grp_node_blocks_job_packable_onto_already_used_nodes, that reproduces the reported scenario directly: a QOS grp node=4 limit fully occupied by two disjoint-node running jobs, then a new job needing only 1 more node is blocked even though the account's existing nodes have plenty of idle capacity it could pack onto instead of requiring a 5th distinct node.

Confirmed failing both on main and on this branch's fix - the actual fix needs to make the QOS/account admission gate (classify_pending_jobs -> qos_block_with/account_block_with) aware that a pending job could reuse already-allocated nodes of the same account/QOS, rather than assuming it always needs brand-new distinct nodes. That gate-before-select redesign is left for a follow-up commit on this branch.

First commit is to document the existing behavior. Second commit is the actual fix and updated tests.

### DIFF
--- a/crates/spur-core/src/account_limits.rs
+++ b/crates/spur-core/src/account_limits.rs
@@ -46,10 +46,18 @@ fn job_tres(spec: &JobSpec) -> (u64, u64, u64, u64) {
 /// `account_running_tres` (empty for the submit-time standalone evaluation),
 /// exceed the per-job or account-group cap? Wall is handled separately because
 /// the association wall cap denies unconditionally at submission.
+///
+/// `grp_node_charge` is the node count charged against `grp_tres` specifically
+/// — distinct from `job_node` below because a caller that knows the job could
+/// pack onto nodes the account already occupies may pass a smaller number than
+/// `spec.num_nodes`. The per-job node cap always uses the job's real requested
+/// node count, since it bounds a single job's own footprint rather than
+/// account-wide reuse.
 fn account_resource_breach(
     spec: &JobSpec,
     limits: &AccountLimits,
     account_running_tres: &TresRecord,
+    grp_node_charge: u64,
 ) -> Option<PendingReason> {
     let (job_cpu, job_node, job_mem, job_gpu) = job_tres(spec);
 
@@ -75,7 +83,7 @@ fn account_resource_breach(
             return Some(PendingReason::AssocGrpCpuLimit);
         }
         if grp.get(TresType::Node) > 0
-            && account_running_tres.get(TresType::Node) + job_node > grp.get(TresType::Node)
+            && account_running_tres.get(TresType::Node) + grp_node_charge > grp.get(TresType::Node)
         {
             return Some(PendingReason::AssocGrpNodeLimit);
         }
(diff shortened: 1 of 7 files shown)


sajmera-pensando (changes requested):
**Overall:** The design is sound and the tests are thorough. One correctness bug introduced by this PR needs to be fixed before merge — it can cause `grp_node` limits to be silently exceeded.

---

**Bug: same-pass double-crediting of occupied nodes**

`occupied_nodes()` only looks at `JobState::Running` jobs. It has no visibility into jobs admitted earlier in the same scheduling pass. Two pending jobs evaluated in the same pass can independently find the same occupied node, both compute `grp_node_charge=0`, and both get admitted — even if that node only has capacity for one of them. Whichever job loses the placement race lands on a new distinct node that was never charged against the limit. The `grp_node` cap is exceeded.

This is new to this PR. Before, admission always charged `num_nodes` and `PassReservations` mirrored that exactly. Now `grp_node_charge` can be 0, but `PassReservations.reserve()` still records `job.spec.num_nodes` via `job_tres()` — so the aggregate headroom seen by later jobs in the pass is more conservative than what was actually charged, but the specific nodes that were credited are invisible to later `new_distinct_nodes_needed` calls.

**Suggested fix:** add `qos_claimed_nodes: HashMap<String, HashSet<String>>` and `account_claimed_nodes: HashMap<String, HashSet<String>>` to `PassReservations`. In `reserve()`, insert the job's `preferred_nodes` into the relevant set. Pass those sets into `new_distinct_nodes_needed` and exclude already-claimed nodes from the reusable candidates. Also update `reserve()` to record `grp_node_charge` (not `num_nodes`) for the node dimension of the TRES aggregate, so the headroom check is consistent with what admission actually charged.

---

**Minor: concern re partial reuse + node goes unavailable between admission and placement.** Admission charges the shortfall but backfill may land on a completely different node if preferred nodes become busy between the pass and actual placement. This is a known TOCTOU limitation and acceptable as a known soft cap. Worth a brief code comment noting it so future readers understand `grp_node` is best-effort in this path, not a hard guarantee.

sajmera-pensando (approved):
All four follow-up commits look correct — approving.

**Same-pass double-credit fix (847a72c):** Implemented exactly as suggested. PassReservations gains qos_claimed_nodes/account_claimed_nodes, reserve() now takes the actual grp_node_charge instead of raw num_nodes for the TRES node dimension, and both block functions return Result<u64, PendingReason> so the charge flows back cleanly. The same-pass gap is closed.

**Two self-found bugs (9164d10 + 77040d3):** Both are real and well-caught.

- Cross-dimension pollution: account_block_with runs before qos_block_with on the same &mut Job, so account-credited nodes already in preferred_nodes were making NodePlacement restrictive before the QOS gate ran its own reuse computation — wrongly blocking any job with both an account and a QOS whose only slack was packing onto its own occupied node. Fixed cleanly with NodePlacement::new_ignoring_preferred_nodes().
- Unschedulable nodes credited as reusable: the missing is_schedulable()/is_k0s_reserved() checks in new_distinct_nodes_needed would have starved jobs by biasing placement toward Down/Draining/k0s nodes they can never actually land on. Fixed by matching the same gates real backfill uses.

The TOCTOU note on the cross-pass soft-cap is also in place. Good work finding and fixing these proactively.

Issue #23, opened 2026-03-23
spur image import fails with "Permission denied" when run as non-root user


amd-vpenumal:
## Summary
Running `spur image import <image>` as a non-root user immediately fails
with a permission error
 
## Steps to Reproduce
1. Build spur as a regular user
2. Run: ./target/release/spur image import ubuntu:22.04
 
## Output (actual)
  Importing library/ubuntu:22.04 from docker.io (arch: amd64)
  Error: Permission denied (os error 13)

PR #719, opened 2026-08-22
feat(spur-core): WireGuard mesh + k0s lifecycle CLI for node add/remove

spraveenio:
## Summary

Adds the WireGuard-mesh and k0s-lifecycle primitives needed to run and maintain a
Spur cluster over an encrypted mesh, and to add/remove nodes without a full teardown:

- **`spur net remove-peer --key [--interface]`** — the missing counterpart to
  `net add-peer`, for cleanly dropping a node's WireGuard peer when it leaves the
  mesh (idempotent; removing an absent peer succeeds).
- **`net add-peer --endpoint`** — carry each peer's underlay endpoint into the
  membership so worker↔worker (and controller↔controller) tunnels actually form,
  instead of a hub-and-spoke where only node↔controller works.
- **k0s online worker add/remove + mesh membership** plumbing in `spurctld`/`spurd`
  (`cluster_k8s.rs`, `agent_server.rs`), and honoring `[network] wg_interface` from
  `spur.conf` in `spurd` instead of hardcoding `spur0`.
- Docs: `docs/deployment/ansible.rst` updated for the mesh/k0s node-membership flow.

## Approach / design notes

- Membership is built from mesh facts (each node's `wg_pubkey` + mesh IP + underlay
  endpoint) so the reconcile can program a full mesh, not just AllowedIPs.
- `spur net mesh --config <file> --self <ip>` applies a full-mesh peering from a
  shared membership doc — the supported route to all-to-all when a caller (e.g. the
  ansible toolkit) drives it on every node.

## Test summary (validated live on a 4-node hardware testbed)

WireGuard mesh `10.44.0.0/16`, pods `10.42.0.0/16`, services `10.43.0.0/16`.

- **Mesh bring-up:** 1 controller + 3 workers join the mesh; all-to-all reachable
  over the mesh IPs (controller↔worker and worker↔worker).
- **k0s over mesh:** `spur k8s up` with a control-plane node + worker; cluster
  reaches `ready`, k8s node InternalIPs are the mesh IPs, controller stays meshed.
- **Cross-node pods on the mesh:** a pod on each of two different workers reach each
  other over the pod CIDR, and the traffic rides the WireGuard tunnel — verified via
  per-peer routes (`10.42.x via <peer-mesh-ip>`), pod-CIDR folded into peer
  AllowedIPs, and rising `wg` transfer counters on the exact peer during the ping.
- **Online node add/remove:** worker added via `k8s add-nodes` and removed cleanly;
  `net remove-peer` drops the stale peer so no ghost peer remains.
- **HA over the mesh (3 controllers):** Raft quorum forms over mesh IPs; stopping the
  leader triggers a re-election among the survivors while the cluster keeps serving,
  and the restarted controller rejoins as a follower.

Unit tests added for `net remove-peer` arg parsing; `cargo test` / `cargo clippy`
clean.

## Dependency

**ROCm/spur-toolkit#23** (ansible: multi-controller WireGuard mesh for HA) **depends
on this PR** — its playbooks call `spur net mesh`, `net add-peer --endpoint`, and
`net remove-peer` introduced here. Merge this PR first, then spur-toolkit#23.

### DIFF
--- a/crates/spurctld/src/cluster_k8s.rs
+++ b/crates/spurctld/src/cluster_k8s.rs
@@ -38,6 +38,10 @@ const RECONCILE_INTERVAL: Duration = Duration::from_secs(30);
 /// Network CIDRs the reconcile loop needs (from `[network]` + `[cluster]` config).
 #[derive(Clone, Debug)]
 pub struct ClusterNetworking {
+    /// Whether the WireGuard mesh is enabled (network.wg_enabled). Independent of `[cluster].enabled`:
+    /// a k0s cluster can run without the mesh, in which case the reconcile skips reading peer
+    /// endpoints from `wg` (nothing to advertise, and `wg show` would just error every tick).
+    pub wg_enabled: bool,
     /// WireGuard mesh CIDR (network.wg_cidr) — node mesh IPs are allocated from here.
     pub mesh_cidr: String,
     /// WireGuard interface name (network.wg_interface) — the controller reads its peer endpoints
@@ -92,7 +96,27 @@ pub async fn run(cluster: Arc<ClusterManager>, raft: Arc<RaftHandle>, net: Clust
         // reconcile_mesh is idempotent + prunes) so node-local drift (reboot, wg restart), a failed
         // push, and controller failover all self-heal. Only meaningful with ≥2 meshed nodes; the
         // membership diff gates only the log line, not the push.
-        let mesh = build_mesh_membership(&cluster, &net.mesh_cidr, &net.mesh_interface);
+        // build_mesh_membership shells out to `wg` (peer_endpoints); keep that off the async
+        // runtime with spawn_blocking, matching the agent side's apply_mesh handler.
+        let mesh = {
+            let cluster = cluster.clone();
+            let (wg_enabled, mesh_cidr, mesh_interface) = (
+                net.wg_enabled,
+                net.mesh_cidr.clone(),
+                net.mesh_interface.clone(),
+            );
+            match tokio::task::spawn_blocking(move || {
+                build_mesh_membership(&cluster, wg_enabled, &mesh_cidr, &mesh_interface)
+            })
+            .await
+            {
+                Ok(m) => m,
+                Err(e) => {
+                    warn!(error = %e, "build_mesh_membership task panicked; skipping mesh push this tick");
+                    MeshMembership::default()
+                }
+            }
+        };
         if mesh.nodes.len() >= 2 {
             if mesh.nodes != last_mesh {
                 info!(
@@ -631,30 +655,34 @@ pub fn node_statuses(cluster: &ClusterManager) -> Vec<ClusterNodeStatus> {
 /// mesh (no pubkey) are skipped rather than fabricated, so an incomplete membership is never emitted.
 pub fn build_mesh_membership(
     cluster: &ClusterManager,
+    wg_enabled: bool,
     mesh_cidr: &str,
     mesh_interface: &str,
 ) -> MeshMembership {
-    // Snapshot the controller's own peer→endpoint table so worker↔worker peers get an endpoint
-    // (net join only wires worker→controller; without this the full mesh has no worker path).
-    // Log on failure instead of swallowing it: an empty table silently regresses the mesh to
-    // hub-and-spoke, so operators need the reason (wrong interface name, `wg` unavailable, etc.).
-    let endpoints = spur_net::wireguard::peer_endpoints(mesh_interface).unwrap_or_else(|e| {
-        warn!(
-            interface = %mesh_interface, error = %e,
-            "could not read WireGuard peer endpoints; mesh membership will omit them, so \
-             worker↔worker tunnels may fall back to hub-and-spoke"
-        );
+    // Peer endpoints only matter under WireGuard; when the mesh is off, skip the `wg` shell-out
+    // entirely (it would just error every reconcile tick). Otherwise snapshot the controller's
+    // peer→endpoint table so worker↔worker peers get an endpoint (net join only wires
+    // worker→controller). Log on failure — an empty table silently regresses to hub-and-spoke.
+    let endpoints = if wg_enabled {
+        spur_net::wireguard::peer_endpoints(mesh_interface).unwrap_or_else(|e| {
+            warn!(
+                interface = %mesh_interface, error = %e,
+                "could not read WireGuard peer endpoints; mesh membership will omit them, so \
+                 worker↔worker tunnels may fall back to hub-and-spoke"
+            );
+            std::collections::HashMap::new()
+        })
+    } else {
         std::collections::HashMap::new()
-    });
+    };
     mesh_from_nodes(cluster.get_nodes(), mesh_cidr, &endpoints)
 }
 
 /// Pure core of [`build_mesh_membership`] (testable without a `ClusterManager`).
 ///
-/// Includes any meshed node (non-empty `wg_pubkey`) with a known mesh IP — `k0s_mesh_ip` if it has
-/// a role, else its advertised mesh-interface address (the configured `network.wg_interface`,
-/// `spur0` by default) — so the controller/login nodes stay in membership and aren't pruned.
-/// `endpoints` supplies each peer's underlay endpoint for worker↔worker tunnels.
+/// Includes any meshed node (non-empty `wg_pubkey`) with a known mesh IP — its `k0s_mesh_ip` if it
+/// has a role, else its advertised address when that falls inside `mesh_cidr` — so controller/login
+/// nodes stay in membership. `endpoints` supplies each peer's underlay endpoint for worker↔worker.
 fn mesh_from_nodes(
     nodes: Vec<spur_core::node::Node>,
     mesh_cidr: &str,
--- a/crates/spurctld/src/cluster_k8s.rs
+++ b/crates/spurctld/src/cluster_k8s.rs
@@ -40,6 +40,9 @@ const RECONCILE_INTERVAL: Duration = Duration::from_secs(30);
 pub struct ClusterNetworking {
     /// WireGuard mesh CIDR (network.wg_cidr) — node mesh IPs are allocated from here.
     pub mesh_cidr: String,
+    /// WireGuard interface name (network.wg_interface) — the controller reads its peer endpoints
+    /// from this to re-advertise worker↔worker underlay endpoints in the mesh membership.
+    pub mesh_interface: String,
     /// Pod CIDR (cluster.pod_cidr) — per-node /24s are carved from here.
     pub pod_cidr: String,
     /// Service CIDR (cluster.service_cidr) — for the generated k0s config.
@@ -89,7 +92,7 @@ pub async fn run(cluster: Arc<ClusterManager>, raft: Arc<RaftHandle>, net: Clust
         // reconcile_mesh is idempotent + prunes) so node-local drift (reboot, wg restart), a failed
         // push, and controller failover all self-heal. Only meaningful with ≥2 meshed nodes; the
         // membership diff gates only the log line, not the push.
-        let mesh = build_mesh_membership(&cluster);
+        let mesh = build_mesh_membership(&cluster, &net.mesh_cidr, &net.mesh_interface);
         if mesh.nodes.len() >= 2 {
             if mesh.nodes != last_mesh {
                 info!(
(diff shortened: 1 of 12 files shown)


crates/spurctld/src/cluster_k8s.rs
```diff
@@ -626,27 +629,41 @@ pub fn node_statuses(cluster: &ClusterManager) -> Vec<ClusterNodeStatus> {
...
-    mesh_from_nodes(cluster.get_nodes())
+pub fn build_mesh_membership(
+    cluster: &ClusterManager,
+    mesh_cidr: &str,
+    mesh_interface: &str,
+) -> MeshMembership {
+    // Snapshot the controller's own peer→endpoint table so worker↔worker peers get an endpoint
+    // (net join only wires worker→controller; without this the full mesh has no worker path).
+    let endpoints = spur_net::wireguard::peer_endpoints(mesh_interface).unwrap_or_default();
+    mesh_from_nodes(cluster.get_nodes(), mesh_cidr, &endpoints)
```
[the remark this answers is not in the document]
spraveenio: Good catch. Fixed in 386fcff: `build_mesh_membership` now logs a `warn!` (interface + error) when `peer_endpoints` fails instead of `unwrap_or_default()`, so a silent regression to hub-and-spoke is diagnosable.

crates/spurctld/src/cluster_k8s.rs
```diff
@@ -626,27 +629,41 @@ pub fn node_statuses(cluster: &ClusterManager) -> Vec<ClusterNodeStatus> {
...
 }
 
 /// Pure core of [`build_mesh_membership`] (testable without a `ClusterManager`).
-fn mesh_from_nodes(nodes: Vec<spur_core::node::Node>) -> MeshMembership {
+///
+/// Includes any meshed node (non-empty `wg_pubkey`) with a known mesh IP — `k0s_mesh_ip` if it has
+/// a role, else its advertised `spur0` address — so the controller/login nodes stay in membership
+/// and aren't pruned. `endpoints` supplies each peer's underlay endpoint for worker↔worker tunnels.
```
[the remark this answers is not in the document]
spraveenio: Fixed in 386fcff: the doc now says "its advertised mesh-interface address (the configured `network.wg_interface`, `spur0` by default)" instead of hardcoding `spur0`.

docs/deployment/ansible.rst
```diff
@@ -292,10 +292,63 @@ The mesh defaults are ``spur_wg_cidr=10.44.0.0/16``, ``spur_wg_port=51820``, and
...
 
+The mesh interface is enabled as a ``wg-quick@<iface>`` systemd unit
+(``spur_wg_persist=true``, the default) so it is recreated on boot from
+``/etc/wireguard/<iface>.conf``; the controller's reconcile then re-pushes peer
+membership. Set ``spur_wg_persist=false`` to skip boot enablement (the interface then
+only lasts until reboot).
+
 .. note::
 
    WireGuard is **single-controller only**; HA requires the ``direct`` transport.
```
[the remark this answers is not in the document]
spraveenio: Fixed in 386fcff: the note now states HA over WireGuard is supported (bootstrap `net init` + full-mesh `net mesh` pass wiring controller↔controller tunnels for Raft), replacing the contradictory "single-controller only" wording.

sajmera-pensando (changes requested):
Requesting changes for the three items in my comment: unit test for `peer_endpoints()` parsing, e2e follow-up issue/PR, and consolidating WireGuard docs into a single bringup guide.

yansun1996 (commented):
The hub-and-spoke fix (real endpoint propagation) and `remove-peer` are both solid, and the unit tests for `parse_peer_endpoints` are well-targeted. This already went through one review round (unit test, e2e ticket, docs consolidation) — those three are genuinely addressed. A few things surfaced in a fresh pass, mostly in the new docs, plus one real code-level race worth closing before merge.

**Docs — a couple of things outside this diff's hunks, so noting here instead of inline:**
- `ansible.rst` now contradicts itself: the topology table (`... (single controller only)`) and a `.. note::` a bit further down (`WireGuard is **single-controller only**...`) still say WireGuard is single-controller-only, while a new note added by this PR (to address earlier review feedback) says HA over WireGuard *is* supported. Neither of the two older claims was touched by this PR, so they're easy to miss, but they now read as directly contradicting the new note a few paragraphs later — could both be updated to match?
- Similarly, `spur net add-peer --program-routes` installs a kernel route for a peer's pod CIDR, but `remove-peer` has no counterpart to remove it when the peer is dropped. Low priority — `--program-routes` is explicitly scoped to the no-CNI bare-mesh test path — but worth a quick note either way.

**Docs — likely worth fixing before merge (see inline):**
- The `spur_wg_persist` / SPUR-managed-k0s section of `ansible.rst`, and the `remove_nodes.yml` / "inventory-driven HA" references in the new `wireguard.rst`, describe Ansible variables and playbooks (`spur_wg_persist`, `spur_k8s_*`, `k8s_up.yml`, `k8s_add_nodes.yml`, WireGuard cleanup in `remove_nodes.yml`) that don't exist in `spur-toolkit` yet — they ship only in the still-open `spur-toolkit#23`. A reader following these docs today will hit a missing playbook, an unconsumed variable, or find that `remove_nodes.yml` doesn't actually clean up the WireGuard peer it claims to. Worth caveating these sections until `#23` merges, or holding them until then.

**Code — worth fixing:**
- Confirmed a real race: `spur net remove-peer` is a purely local operation (direct `wg` mutation, no RPC to the controller). If the departed node hasn't *also* been deregistered from the cluster, the controller's reconcile loop (every 30s) rebuilds mesh membership from live node inventory — which still includes the departed node — and re-pushes it to survivors, silently recreating the peer that was just removed. The new "Removing a node from the mesh" section doesn't mention this ordering requirement.
- `build_mesh_membership`'s `peer_endpoints()` call shells out to `wg` synchronously inside the async reconcile loop, unlike the equivalent call on the agent side which wraps it in `spawn_blocking` — worth the same treatment for consistency (this PR already established the pattern one file over).

**Minor / optional:** a doc comment on `mesh_from_nodes` references a `network.wg_interface` parameter the function doesn't take (should just describe the CIDR-containment fallback); a couple of comment blocks run 3-4 lines where the convention here is 1-3; and `peer_endpoints()` still gets called every 30s even for k0s clusters that don't use WireGuard at all (`cluster.enabled` and `network.wg_enabled` are independent config flags), producing a permanent `warn!` every tick with no way to quiet it short of disabling k0s — low severity since membership just ends up empty, but worth a `wg_enabled` gate or a `debug!` downgrade.

docs/deployment/ansible.rst
```diff
@@ -292,9 +292,64 @@ The mesh defaults are ``spur_wg_cidr=10.44.0.0/16``, ``spur_wg_port=51820``, and
 ``spur_wg_interface=spur0``. The control node needs the ``ansible.utils`` collection and
 ``netaddr`` (see Prerequisites).
 
+The mesh interface is enabled as a ``wg-quick@<iface>`` systemd unit
```
yansun1996, line 299: This paragraph (and the SPUR-managed k0s section that follows it) describes `spur_wg_persist` and the `spur_k8s_*` variables / `k8s_up.yml` / `k8s_add_nodes.yml` playbooks. As of today none of these exist in `spur-toolkit` main — they're only in the still-open `spur-toolkit#23`. A reader following this doc right now would hit an unconsumed variable or a missing playbook. Would it be worth caveating this section as depending on `spur-toolkit#23` (not yet merged), or holding it until that PR lands?

docs/deployment/wireguard.rst
```diff
@@ -0,0 +1,163 @@
...
+
+   # On the controller — drop the departed node's peer (idempotent):
+   sudo spur net remove-peer --key <departed-node-pubkey> --interface spur0
+
+   # On the departed node — stop and de-persist the interface:
+   sudo systemctl disable --now wg-quick@spur0
+   sudo rm -f /etc/wireguard/spur0.conf
+
+In an HA mesh, remove the peer on **every** controller, not just the bootstrap —
+otherwise the others keep a stale peer. The Ansible ``remove_nodes.yml`` playbook
```
yansun1996: "The Ansible `remove_nodes.yml` playbook handles both the controller-side `remove-peer`... and the node-side teardown" — checked the current `remove_nodes.yml` on `spur-toolkit` main and it has no WireGuard logic at all yet; that's added by the still-open `spur-toolkit#23`. Today, removing a node via that playbook would leave a ghost peer on every controller — exactly what this paragraph is warning against. Could this be caveated the same way as the `ansible.rst` section, until `#23` merges?

docs/deployment/wireguard.rst
```diff
@@ -0,0 +1,163 @@
...
+   [controller]
+   hosts = ["10.44.0.1", "10.44.0.3", "10.44.0.4"]
+   peers = [
+     "10.44.0.1:6821",
+     "10.44.0.3:6821",
+     "10.44.0.4:6821",
+   ]
+
+Raft elects a leader automatically; clients and agents may target any controller
+and are redirected to the current leader. See :doc:`native-host` for the full
```
yansun1996, line 171: "...and :doc:`ansible` for the inventory-driven equivalent" — same issue: there's no inventory-driven multi-controller WireGuard flow in `ansible.rst`/the toolkit yet (the Ansible role still hard-asserts a single controller until `spur-toolkit#23` merges). This cross-reference currently points to something that doesn't exist.

docs/deployment/wireguard.rst
```diff
@@ -0,0 +1,163 @@
...
+=============================
+
+When a node leaves, drop its peer entry so it does not linger as a "ghost" peer
+(and, on the node itself, tear the interface down so it does not rejoin on
+reboot):
+
+.. code-block:: bash
+
+   # On the controller — drop the departed node's peer (idempotent):
+   sudo spur net remove-peer --key <departed-node-pubkey> --interface spur0
```
yansun1996, line 122: Worth a note here on sequencing: `spur net remove-peer` only mutates the local `wg` interface — it doesn't touch cluster state. If the node hasn't also been deregistered from the cluster (`spur node remove` / equivalent), the controller's reconcile loop rebuilds mesh membership from live node inventory on its next tick (≤30s) and will still include this node, silently re-adding the peer it was just told to drop. Might be worth stating explicitly that the node should be removed from the cluster first (or atomically), not just here on the mesh.

crates/spurctld/src/cluster_k8s.rs
```diff
@@ -626,27 +629,51 @@ pub fn node_statuses(cluster: &ClusterManager) -> Vec<ClusterNodeStatus> {
...
+pub fn build_mesh_membership(
+    cluster: &ClusterManager,
+    mesh_cidr: &str,
+    mesh_interface: &str,
+) -> MeshMembership {
+    // Snapshot the controller's own peer→endpoint table so worker↔worker peers get an endpoint
+    // (net join only wires worker→controller; without this the full mesh has no worker path).
+    // Log on failure instead of swallowing it: an empty table silently regresses the mesh to
+    // hub-and-spoke, so operators need the reason (wrong interface name, `wg` unavailable, etc.).
+    let endpoints = spur_net::wireguard::peer_endpoints(mesh_interface).unwrap_or_else(|e| {
```
yansun1996: This shells out to `wg` (via `peer_endpoints`) synchronously inside the async reconcile loop. `agent_server.rs`'s `apply_mesh` handler wraps its equivalent `wg` calls in `spawn_blocking` specifically to keep this off the async runtime — would it be worth doing the same here for consistency?

crates/spurctld/src/cluster_k8s.rs
```diff
@@ -626,27 +629,51 @@ pub fn node_statuses(cluster: &ClusterManager) -> Vec<ClusterNodeStatus> {
...
+        std::collections::HashMap::new()
+    });
+    mesh_from_nodes(cluster.get_nodes(), mesh_cidr, &endpoints)
 }
 
 /// Pure core of [`build_mesh_membership`] (testable without a `ClusterManager`).
-fn mesh_from_nodes(nodes: Vec<spur_core::node::Node>) -> MeshMembership {
+///
+/// Includes any meshed node (non-empty `wg_pubkey`) with a known mesh IP — `k0s_mesh_ip` if it has
+/// a role, else its advertised mesh-interface address (the configured `network.wg_interface`,
```
yansun1996: This doc comment says the fallback uses "its advertised mesh-interface address (the configured `network.wg_interface`, `spur0` by default)", but `mesh_from_nodes` doesn't take an interface parameter — the fallback (`real_mesh_address(&n, mesh_cidr)`, line 669) only checks whether the node's advertised address falls inside `mesh_cidr`, regardless of interface name. Could this be reworded to describe the CIDR-containment check instead? While touching it, this block (and the one at line 637-640) run 4-5 lines — might be worth trimming toward the project's usual 1-3 line comments.

sajmera-pensando (approved):
LGTM

PR #720, opened 2026-08-23
fix(spurctld): report array task throttle reason

### LINKED ISSUE #618: Array-throttled tasks show pending Reason=None

joshkmartinez:
Array tasks held behind an `--array=...%N` concurrency limit currently stay pending with `Reason=None`. This adds Slurm's `JobArrayTaskLimit` reason and records it when the per-array cap is full.

Pending-reason changes are applied in order, so a task gets the throttle reason only while it is actually held. When a slot reopens, that stale reason is cleared without overwriting a newer blocker from the same scheduling pass. The existing task ordering and slot accounting are unchanged.

Fixes #618

Tested with:

- `cargo test -p spurctld array_throttle --locked`
- 563 portable `spur-core` tests and 916 `spurctld` tests
- warnings-as-errors Clippy for `spur-core` and `spurctld`
- a local one-controller gRPC smoke with this branch's controller and a Spur 0.7 client; both `queue %R` and `show job` reported `JobArrayTaskLimit`, and cancelling the admitted child immediately cleared the reason on the next task
- formatting, diff, and SPDX checks

The full workspace commands still need Linux CI. On macOS, PMIx fails to link, `spur-cli` imports a `nix::getgroups` API that is unavailable on Apple, and the shell-hook tests execute through `/proc/self/fd`.

### DIFF
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -442,7 +442,7 @@ pub struct ClusterManager {
 
 struct PendingJobClassification {
     jobs: Vec<Job>,
-    blocked: Vec<(JobId, PendingReason)>,
+    reason_updates: Vec<(JobId, PendingReason)>,
     bb_stage_candidates: Vec<JobId>,
 }
 
@@ -3172,11 +3172,11 @@ impl ClusterManager {
         self.classify_pending_jobs().jobs
     }
 
-    /// Classify pending jobs once, apply displayed block reasons, advance burst-buffer
+    /// Classify pending jobs once, apply pending-reason updates, advance burst-buffer
     /// stage-in for selected candidates, and return the jobs eligible for scheduling.
     pub fn pending_jobs_and_tag_reasons(&self) -> Vec<Job> {
         let classification = self.classify_pending_jobs();
-        self.apply_blocked_pending_reasons(classification.blocked);
+        self.apply_pending_reason_updates(classification.reason_updates);
         self.advance_bb_staging_for(&classification.bb_stage_candidates);
         classification.jobs
     }
@@ -3213,13 +3213,13 @@ impl ClusterManager {
                 })
             })
             .collect();
-        let mut blocked = Vec::new();
+        let mut reason_updates = Vec::new();
 
         // Structural blockers retain their precedence before begin-time eligibility;
         // unlike consumables, they do not reserve capacity while the job waits.
         {
             let partitions = self.partitions.read();
-            retain_unblocked(&mut candidates, &mut blocked, |job| {
+            retain_unblocked(&mut candidates, &mut reason_updates, |job| {
                 partition_block(job, &partitions)
             });
         }
@@ -3238,7 +3238,7 @@ impl ClusterManager {
                 .collect()
         };
 
-        retain_unblocked(&mut candidates, &mut blocked, |job| {
+        retain_unblocked(&mut candidates, &mut reason_updates, |job| {
             if job.spec.dependency.is_empty() {
                 return None;
             }
@@ -3257,7 +3257,7 @@ impl ClusterManager {
             .collect();
 
         let reservations = self.get_reservations();
-        retain_unblocked(&mut candidates, &mut blocked, |job| {
+        retain_unblocked(&mut candidates, &mut reason_updates, |job| {
             reservation_block(job, &reservations, now)
         });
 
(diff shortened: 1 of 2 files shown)


crates/spurctld/src/cluster.rs
```diff
@@ -3297,8 +3297,21 @@ impl ClusterManager {
             };
             let count = array_counts.entry(array_id).or_insert(0);
```
yansun1996, line 3298: Since this is inside a `for candidate in &mut candidates` loop, would it be cleaner to `continue` on the over-cap branch instead of nesting the recovery check inside the `else`?

```rust
if *count >= max {
    record_pending_reason_update(&mut reason_updates, candidate, PendingReason::JobArrayTaskLimit);
    candidate.scheduling_eligible = false;
    continue;
}
if candidate.job.pending_reason == PendingReason::JobArrayTaskLimit {
    record_pending_reason_update(&mut reason_updates, candidate, PendingReason::None);
    candidate.job.set_pending_reason(PendingReason::None);
}
*count += 1;
```

crates/spurctld/src/cluster.rs
```diff
@@ -3297,8 +3297,21 @@ impl ClusterManager {
...
+                    PendingReason::JobArrayTaskLimit,
+                );
                 candidate.scheduling_eligible = false;
             } else {
+                if candidate.job.pending_reason == PendingReason::JobArrayTaskLimit {
+                    record_pending_reason_update(
+                        &mut reason_updates,
+                        candidate,
+                        PendingReason::None,
+                    );
```
yansun1996: `candidate.job` here is a clone taken from the read-lock snapshot earlier in this function, not a reference into the real `self.jobs` map — so this direct `set_pending_reason(None)` call doesn't actually reach shared state; only the queued `reason_updates` entry (applied later under the write lock) does. Since it looks load-bearing but isn't, might be worth dropping it to avoid confusing a future reader.

crates/spurctld/src/cluster.rs
```diff
@@ -16707,9 +16770,9 @@ mod tests {
             pending.contains(&solo),
             "the eligible job must still receive the remaining license slot"
         );
-        assert_ne!(
+        assert_eq!(
```
yansun1996, line 16768: `array_throttle_does_not_consume_scarce_license_pool` — since the array-throttled task's `scheduling_eligible` gets set to `false` before the license gate runs, `retain_eligible` skips it outright, so this test doesn't actually exercise a task that's independently over both the array cap and out of licenses in the same pass. Might be worth a small test confirming `JobArrayTaskLimit` still wins (and no stale reason lingers) when a candidate would fail two gates at once.

yansun1996 (commented):
Nice fix — routing the array-throttle check through the existing gate mechanism instead of the raw `scheduling_eligible` flip is the right call, and the write-lock re-check guard correctly prevents a stale clear from clobbering a newer reason. A couple of small suggestions below; nothing blocking.

Also, +1 to the bot's suggestion to rename `tag_blocked_pending_reasons` — it now applies clears too, not just blocks, so the name doesn't quite match what it does anymore.

crates/spurctld/src/cluster.rs
```diff
@@ -3671,23 +3688,23 @@ impl ClusterManager {
         cancelled
     }
 
-    /// Reclassify pending jobs and apply their displayed block reasons.
+    /// Reclassify pending jobs and apply their pending-reason updates.
     /// Leader-only; enforced by the scheduler-loop caller, not this function
     /// itself (mirrors `cancel_unsatisfiable_dependency_jobs()`).
     #[cfg(test)]
     fn tag_blocked_pending_reasons(&self) {
```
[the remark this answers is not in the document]
joshkmartinez: Good catch. I renamed the helper to `refresh_pending_reasons` and updated the related docs so they cover both setting and clearing reasons.

crates/spurctld/src/cluster.rs
joshkmartinez: Good call. I flattened this with an early `continue` for candidates over the array cap.

crates/spurctld/src/cluster.rs
joshkmartinez: Good catch. I removed the cloned-job mutation. The queued update is now the only path that updates shared state.

crates/spurctld/src/cluster.rs
joshkmartinez: Good catch. I added a regression test for a task that is both array-throttled and short on licenses. It verifies that `JobArrayTaskLimit` wins while the cap is full, then hands off to `Licenses` once the slot opens so the stale array reason does not linger.

yansun1996 (approved):
LGTM

PR #721, opened 2026-08-24
test(spur-net): add native-host WireGuard mesh e2e suite

spraveenio:
## Summary

Adds `tests/native_wg_host/e2e/` — a real-WireGuard-mesh e2e suite that stands up kernel `wg` interfaces (`spur net init`/`join`) across the SSH test nodes and verifies mesh + k0s-over-mesh behaviours that previously had no automated coverage. Follow-up to the WireGuard mesh + k0s lifecycle work in #719 (SPUR-212).

The suite reuses the base native-host harness (`tests/native_host/e2e/cluster.py`) via `pythonpath` — only the mesh-specific mechanics and fixtures are new. Rebased onto the sharded e2e model from #724.

## Scenarios (validated on a 3-node hardware bed)

- **Mesh bring-up** — 1 controller + N workers, all-to-all reachability over the mesh IPs including worker↔worker (endpoint propagation), with a handshake-settle retry (WireGuard is lazy about the first packet).
- **k0s-over-mesh** — cluster reaches ready, node InternalIPs are the mesh IPs, and the controller stays meshed (not pruned).
- **Cross-node pod CIDR** folded into peer AllowedIPs.
- **Graceful k8s remove/add** — cordon + drain a worker, `spur k8s remove-nodes` it, and confirm the removal is **k0s-only**: the node stays a registered Spur worker in the mesh with the **same WireGuard key**; `spur k8s add-nodes` restores it as a k8s worker with no re-key.
- **Online add/remove** leaves no ghost wg peer; `net remove-peer` idempotency.
- **Login-node reachability** (meshed but outside k0s scope).

Scaffolded but **skipped** (do not run in CI): the cross-node pod-datapath proof (D3 — needs a kubectl pod-placement helper; the rising-`wg`-counter assertion is implemented and reusable) and the 3-controller HA re-election (D5 — needs a multi-controller fixture the base harness does not yet provide).

## Design

- **Opt-in + skip-gated.** Gated three ways: `SPUR_TEST_NODES`, `SPUR_TEST_WG=1`, and a per-node preflight that skips cleanly when a real mesh can't be stood up (`wg` tool / kernel module absent). Where WireGuard is unavailable the tests **skip**, not fail.
- **Per-test k0s reset.** `spur k8s down` resets k0s asynchronously via the controller; the fixture kills the daemons right after, which would leak etcd state across tests. The fixture therefore force-resets k0s directly on every node (stop services, `k0s reset`, wipe dirs) at both setup and teardown, so each test starts clean. This was the root cause of initial cross-test flakiness, found and fixed on the hardware bed.
- **CI: folded into the sharded native-host job** — both testpaths in one pytest invocation with `SPUR_TEST_WG=1`, plus a wg-install step on the nodes. WG tests fall under the `not gpu` shards and distribute via the existing md5 sharding.

## ⚠️ Reviewer note

Whether the self-hosted runner's base VM image (`spur-ci-base.qcow2`) ships the **WireGuard kernel module** is not verifiable from this repo. If it is absent and `wireguard-go` is unavailable, the mesh tests skip rather than exercise a real data plane. If kmod support is missing, prefer baking `wireguard`/`wireguard-tools` into the image build over the apt-install step added here.

This is a **test-infra** change (`test` type — excluded from the release changelog); no `crates/` code was modified.

## Test plan

- [x] `python -m py_compile` + `pytest --collect-only`: 13 tests collect, no marker warnings
- [x] Gating verified: without `SPUR_TEST_WG` the suite is ignored; with it, collected and shardable
- [x] Hardware validation on a real 3-node WireGuard bed (mesh bring-up, k0s-over-mesh, graceful remove/add with wg-key persistence, login reachability all pass on hardware)
- [ ] Observe the folded WG tests in the `E2E / native-host` shards on the self-hosted runner (kmod availability)


tests/native_wg_host/e2e/conftest.py
```diff
@@ -0,0 +1,224 @@
...
+    Yields a :class:`WgMesh` with the interface already provisioned via
+    ``net init`` on node 0 and ``net join`` on the rest, promoted to a full
+    mesh. Torn down (interfaces + confs removed) after the test.
+    """
+    if len(ssh_nodes) < 2:
+        pytest.skip(f"WG mesh needs >= 2 nodes (got {len(ssh_nodes)})")
+    c = SpurCluster(ssh_nodes, make_remote_dir(), remote_bin_dir)
+    c.provision()  # resolves node_names; no daemons started
+    mesh = WgMesh(ssh_nodes, c.node_names, remote_bin_dir, _sudo_prefix(), iface=WG_IFACE)
+    indices = list(range(len(ssh_nodes)))
```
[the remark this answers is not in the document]
spraveenio: Fixed in d1a97b7 — the `wg_mesh` fixture now calls `c.root_agent_preflight()` after `provision()`, so on a node without NOPASSWD sudo (and no `SPUR_TEST_SSH_PASSWORD`) the mesh-only tests skip cleanly, matching the k0s fixtures.

tests/README.md
```diff
@@ -5,4 +5,5 @@ Test suites organized by deployment target. See [docs/developer/building.rst](..
 | Path | Description |
 |------|-------------|
 | `native_host/e2e/` | Deploys Spur on bare-metal nodes via SSH |
+| `native_wg_host/e2e/` | Stands up a real WireGuard mesh over SSH nodes; verifies mesh, k0s-over-mesh, and HA. Opt-in via `SPUR_TEST_WG=1` |
```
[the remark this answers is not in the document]
spraveenio: Fixed in d1a97b7 — the README entry now says the suite verifies mesh bring-up and k0s-over-mesh, and notes the cross-node pod-datapath and HA scenarios are scaffolded but currently skipped.

docs/developer/building.rst
```diff
@@ -211,6 +211,45 @@ Running the Tests
 
 Tests that require more nodes than provided, or missing GPU/container prerequisites, are automatically skipped.
 
+End-to-End Tests (WireGuard mesh)
+---------------------------------
+
+The WireGuard mesh E2E suite lives in ``tests/native_wg_host/e2e/`` and stands up a **real** WireGuard mesh (kernel ``wg`` interfaces via ``spur net init``/``join``) across the SSH test nodes, then verifies mesh connectivity, k0s-over-mesh, cross-node pod datapath, online node add/remove, HA, and login-node reachability. It reuses the base harness (``cluster.py``) from ``tests/native_host/e2e/``.
```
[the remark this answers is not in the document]
spraveenio: Fixed in d1a97b7 — building.rst now splits 'Automated today' from 'Scaffolded but currently skipped' (D3 pod datapath pending a pod-placement helper; D5 HA pending a multi-controller fixture), so readers don't assume those run in CI.

tests/native_wg_host/e2e/conftest.py
```diff
@@ -0,0 +1,310 @@
...
+    try:
+        c.k8s_down(reset=True)
+        c.wait_k8s_phase("down", timeout=120)
+    except Exception:
+        pass
+    _reset_k0s_all_nodes(c)
+    mesh = getattr(c, "wg_mesh", None)
+    if mesh is not None:
+        mesh.teardown(getattr(c, "wg_mesh_indices", []))
+    c.teardown()
```
yansun1996: In `_teardown_wg_k0s`, `c.teardown()` (which stops `spurd`/`spurctld`) runs *after* `_reset_k0s_all_nodes(c)`. Since the daemons are still alive while the manual k0s reset runs, their own async reconcile (triggered by the `k8s_down(reset=True)` call just above) can race the manual `rm -rf`/`k0s reset` — this looks like the same class of race the setup-side reset is designed to avoid, just reintroduced here. Would it be worth swapping the order — call `c.teardown()` before `_reset_k0s_all_nodes(c)` — so nothing is contending with the manual cleanup?

yansun1996 (commented):
Nice suite — real WireGuard mesh + k0s-over-mesh coverage that was genuinely missing, and the skip-gating (opt-in env vars + per-node preflight) is well thought out so it degrades cleanly on environments that can't run it. A few things worth addressing before merge, plus some smaller notes.

**Worth fixing:**
- A teardown-ordering issue in the k0s cluster fixture that can reintroduce the exact race the fixture is designed to avoid (see inline comment).
- Two doc/README lines that don't quite match what the code does (inline comments).
- A stray reviewer-facing note left in a CI comment block that should be resolved or moved out before merge.

**Smaller suggestions**, called out inline: a redundant SSH round-trip in mesh bring-up, a couple of assertions/helpers that could be tightened or wired in, and a docstring whose stated reasoning doesn't quite match how the mesh IP is actually resolved downstream.

Also, since this suite shares nodes/CI shards with the existing `native_host/e2e` suite, it might be worth adding a lightweight guard (or at least a comment) noting that a WireGuard interface should not be left behind for a non-WG test to trip over — not blocking, just a thought for a follow-up.

docs/developer/building.rst
```diff
@@ -211,6 +211,49 @@ Running the Tests
...
+   export SPUR_TEST_NODES=10.0.1.10,10.0.1.11,10.0.1.12
+   export SPUR_TEST_WG=1
+
+   # Full mesh suite
+   pytest tests/native_wg_host/e2e/ -v
+
+   # Only the mesh-bring-up scenarios (no k0s convergence needed)
+   pytest tests/native_wg_host/e2e/ -v -m "wireguard and not k0s"
+
+In CI this suite runs as the ``E2E / native-host-wg`` status check. See ``.github/workflows/e2e.yml`` — note the WireGuard kernel-module availability on the runner image is a prerequisite; where it is absent the tests skip rather than fail.
```
yansun1996: This says the suite reports as an `E2E / native-host-wg` status check, but the workflow only defines `E2E / native-host` and `E2E / k8s` — this suite is folded into the existing native-host job/shards (matching the PR description). Could this line just say it reports under the existing `E2E / native-host` status?

tests/README.md
```diff
@@ -5,4 +5,5 @@ Test suites organized by deployment target. See [docs/developer/building.rst](..
 | Path | Description |
 |------|-------------|
 | `native_host/e2e/` | Deploys Spur on bare-metal nodes via SSH |
+| `native_wg_host/e2e/` | Stands up a real WireGuard mesh over SSH nodes; verifies mesh bring-up and k0s-over-mesh (cross-node pod-datapath and HA scenarios are scaffolded but currently skipped). Opt-in via `SPUR_TEST_WG=1` |
```
yansun1996: This groups "cross-node pod-datapath and HA" together as scaffolded/skipped, but the pod-datapath test (`test_cross_node_pod_ping_rides_the_tunnel`) is actually implemented and asserts a real transfer-counter rise — only the HA scenario is a `pytest.skip` stub. Worth narrowing this to HA only so it doesn't undersell the coverage that's already there (building.rst already gets this distinction right).

.github/workflows/e2e.yml
```diff
@@ -222,6 +222,37 @@ jobs:
             ssh $SSH_OPTS -i "$KEY" ci@"$node" "echo '${PROF}' | sudo apparmor_parser -r 2>/dev/null || true"
           done
 
+      # WireGuard mesh suite (tests/native_wg_host/e2e/) is folded into the
+      # native-host shards, so every node needs `wg`/`wg-quick` and a working
+      # data plane. The suite's own preflight (_wg_preflight) skips cleanly when
+      # a usable mesh can't be provided, so a shard that lands no WG tests — or a
+      # runner image without the kmod — yields SKIPS, not failures.
+      #
+      # UNVERIFIED FROM THE REPO: whether the runner base image
```
yansun1996: This reads like a note-to-reviewer ("UNVERIFIED FROM THE REPO... confirm at review time") rather than a comment that should live in the file long-term. Once the kmod-availability question is resolved, could this be trimmed down or moved out (e.g. into the PR description or a tracked follow-up) rather than staying as a multi-line aside in the workflow?

tests/native_wg_host/e2e/wg_cluster.py
```diff
@@ -0,0 +1,337 @@
...
+            self.net_join(i, head_endpoint, head_key)
+            # Head learns the worker (its underlay endpoint) so the tunnel is bidirectional.
+            self.net_add_peer(
+                on_index=head,
+                peer_key=self.pubkeys[i],
+                allowed_ip=f"{self.mesh_ip_for(i)}/32",
+                endpoint=f"{self.nodes[i].host}:{WG_PORT}",
+            )
+        # Promote hub-and-spoke to full mesh so worker↔worker also connects.
+        self.full_mesh_by_add_peer(indices)
```
yansun1996: The join loop above already calls `net_add_peer` for each worker against the head, so `full_mesh_by_add_peer(indices)` re-issues those same head↔worker pairs again here. Since it's only the worker↔worker pairs that are actually new, would it be worth scoping this call to just those (e.g. `indices[1:]`) to save the extra round-trips per test?

tests/native_host/wg_e2e/conftest.py
```diff
@@ -0,0 +1,310 @@
...
+    try:
+        c.start(config_overrides=overrides, agent_as_root=True)
+    except Exception:
+        mesh.teardown(indices)
+        c.teardown()
+        raise
+    return c
+
+
+def _reset_k0s_all_nodes(c):
```
yansun1996: This resets every node via `exec_allow_fail(... || true)` with no check that the stop/reset actually completed before returning. Since this runs at the top of the *next* test's setup too, a node that fails to clean up here would likely surface as a confusing failure in a different, later test. Might be worth polling briefly for k0s to actually be down before proceeding, or at least logging (rather than swallowing) a failed step.

tests/native_wg_host/e2e/test_wg_mesh.py
```diff
@@ -0,0 +1,740 @@
...
+        `spur0` correctly errors ("No such device"). The fixture guarantees
+        `spur0` is up, so this exercises the real absent-peer path — no `k8s up`
+        needed (which would add a 600s convergence wait unrelated to this check).
+        """
+        c = wg_k0s_cluster
+        # A syntactically valid but not-present WireGuard key (32 bytes base64).
+        absent_key = "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA="
+        out = c.cli_as_user("root", ["spur", "net", "remove-peer",
+                                     "--key", absent_key, "--interface", WG_IFACE])
+        assert "Peer removed" in out or "removed" in out.lower(), (
```
yansun1996: This only checks the CLI's output string (`"removed" in out`) rather than confirming the peer table itself is unchanged. The wg-key-persistence and ghost-peer checks elsewhere in this file do real state comparisons — would it be worth doing the same here (e.g. a `wg show ... dump`/peer-list check before and after) so this test would actually catch a no-op removal?

tests/native_wg_host/e2e/wg_cluster.py
```diff
@@ -0,0 +1,337 @@
...
+            if p.public_key == peer_key:
+                return p
+        return None
+
+    def wg_transfer(self, index: int, peer_key: str) -> tuple[int, int]:
+        """(rx_bytes, tx_bytes) for a specific peer, or (0, 0) if absent."""
+        p = self.wg_peer(index, peer_key)
+        return (p.rx_bytes, p.tx_bytes) if p else (0, 0)
+
+    def has_route(self, index: int, cidr: str) -> bool:
```
yansun1996: `has_route` doesn't appear to be called anywhere in the test suite, though the pod-datapath scenario's docstring mentions per-peer routes as part of what it proves. Either wiring this into that assertion or removing it (if it's no longer needed) would avoid the dead helper.

tests/native_host/e2e/wg_cluster.py
```diff
@@ -0,0 +1,337 @@
...
+        self.bin_dir = bin_dir
+        self.iface = iface
+        self._sudo = sudo_prefix
+        # Filled by init_mesh(): index -> public key.
+        self.pubkeys: dict[int, str] = {}
+        # index -> mesh IP. Caller-supplied (mirrors spur-toolkit's per-node
+        # `spur_wg_address` inventory var) or the k0s-matching default.
+        self.mesh_ips: dict[int, str] = wg_addresses or self._default_mesh_ips()
+
+    def _default_mesh_ips(self) -> dict[int, str]:
```
yansun1996, line 63: This docstring's rationale (sorting by name to match "k0s's own pool order") doesn't quite match how mesh IPs are actually resolved on the controller side — from what I can tell, a meshed node's real advertised `spur0` address is adopted as authoritative rather than re-derived from a sorted pool. This works today because node 0 is always pinned to `.1`, but the stated "why" could mislead a future change to this ordering. Might be worth rewording to describe the actual invariant (avoid colliding with the reserved bootstrap address) rather than a k0s-pool-order assumption.

tests/native_wg_host/e2e/wg_cluster.py
```diff
@@ -0,0 +1,337 @@
...
+                    continue
+                target = self.mesh_ip_for(dst)
+                while True:
+                    if self.ping(src, target):
+                        break
+                    assert time.time() < deadline, (
+                        f"node {self.node_names[src]} cannot reach {target} "
+                        f"({self.node_names[dst]}) over the mesh after {settle_s}s"
+                    )
+                    time.sleep(3)
```
shiv-tyagi: `deadline` is computed once before the double loop and shared across every (src, dst) pair. The docstring promises "each pair is retried for up to *settle_s* seconds", but on an N-node mesh with N(N-1) pairs, an early pair that's slow to settle eats into the budget every later pair gets. That's flaky-test bait, and the failure message won't hint at the real cause. It also reimplements the exact "poll until predicate or timeout" pattern `wait_until` (same file) already provides correctly.

```suggestion
    def assert_all_to_all(self, indices: list[int], settle_s: int = 30) -> None:
        """Every node in *indices* must ping every other over the mesh IP.

        WireGuard is lazy: a freshly added worker↔worker peer has no live tunnel
        until the first packet triggers a handshake, so an immediate ping can
        drop even though the peer/endpoint config is correct. Each pair is
        therefore retried for up to *settle_s* seconds (the ping itself sends the
        handshake-triggering traffic) before it is considered unreachable.
        """
        for src in indices:
            for dst in indices:
                if src == dst:
                    continue
                target = self.mesh_ip_for(dst)
                wait_until(
                    lambda s=src, t=target: self.ping(s, t),
                    timeout_s=settle_s, interval_s=3,
                    desc=f"node {self.node_names[src]} reaching {target} ({self.node_names[dst]})",
                )
```

tests/native_wg_host/e2e/wg_cluster.py
```diff
@@ -0,0 +1,337 @@
...
+
+def mesh_ip(index: int) -> str:
+    """Deprecated positional mesh address (0 -> .1, 1 -> .2, ...).
+
+    Kept for the raw-mesh (no-k0s) tests where the fixture owns the whole mesh
+    and any consistent mapping works. The k0s tests must instead use
+    ``WgMesh.mesh_ip_for`` / ``mesh_ips`` so the addresses match k0s's own pool
+    order (see :meth:`WgMesh._assign_mesh_ips`).
+    """
+    return f"10.44.0.{index + 1}"
```
shiv-tyagi: This is imported in conftest.py but never called anywhere. Its own docstring already says "Deprecated ... Kept for the raw-mesh tests" but no raw-mesh test uses it, `WgMesh._default_mesh_ips` does the equivalent internally. Shipping new code that's already deprecated on arrival is the kind of unnecessary compat cruft this repo avoids pre-1.0. Drop it.

tests/native_wg_host/e2e/wg_cluster.py
```diff
@@ -0,0 +1,337 @@
...
+
+    def net_remove_peer(self, on_index: int, peer_key: str) -> str:
+        return self._spur(self.nodes[on_index], [
+            "net", "remove-peer", "--key", peer_key, "--interface", self.iface,
+        ])
+
+    def net_remove_peer_allow_fail(self, on_index: int, peer_key: str) -> str:
+        return self._spur_allow_fail(self.nodes[on_index], [
+            "net", "remove-peer", "--key", peer_key, "--interface", self.iface,
+        ])
```
shiv-tyagi: No caller anywhere in the suite. Either wire it into a test or drop it.

tests/native_wg_host/e2e/test_wg_mesh.py
```diff
@@ -0,0 +1,740 @@
...
+            f"remove-peer on an absent key should succeed idempotently:\n{out}"
+        )
```
shiv-tyagi: `"Peer removed" in out` is logically subsumed by `"removed" in out.lower()` since any string containing "Peer removed" also contains "removed" once lowercased. That first branch can never be the deciding one.

```suggestion
        assert "removed" in out.lower(), (
            f"remove-peer on an absent key should succeed idempotently:\n{out}"
        )
```

tests/native_wg_host/e2e/test_wg_mesh.py
```diff
@@ -0,0 +1,740 @@
...
+    # through the SSH channel does not survive as real stdin, so the apply would
+    # otherwise get empty input and create nothing.
+    remote_path = f"/tmp/{name}.yaml"
+    c.nodes[cp_index].write_file(remote_path, manifest)
+    out = c.nodes[cp_index].exec_allow_fail(
+        f"{c._sudo_prefix()}k0s kubectl apply -f {remote_path} 2>&1"
+    )
+    assert "created" in out or "configured" in out or "unchanged" in out, (
+        f"kubectl apply for pod {name} did not succeed:\n{out}"
+    )
```
shiv-tyagi: This exact check (`"created" in out or "configured" in out or "unchanged" in out`) is duplicated verbatim at L155 and L178. Worth pulling into one helper, e.g. `_assert_kubectl_applied(out, what)`, so a future change to kubectl's output strings is a one-line fix instead of three.

tests/native_wg_host/e2e/test_wg_mesh.py
```diff
@@ -0,0 +1,740 @@
...
+                   timeout_s=300, desc="both worker kubelets Ready")
+        # …and calico must be up, or a pod scheduled onto a worker whose
+        # calico-node is still initializing never gets a pod IP.
+        wait_until(lambda: _calico_ready(c, cp_index),
+                   timeout_s=480, desc="calico-node DaemonSet Running cluster-wide")
+
+        # Pre-pull the pod image on both workers so pod start doesn't race the
+        # IP-wait on a cold cache; skip cleanly if the registry is unreachable.
+        if not _prepull_busybox(c, [a_index, b_index]):
+            pytest.skip(f"{BUSYBOX_IMAGE} unavailable on the workers (offline registry)")
```
shiv-tyagi: This "wait workers Ready + wait calico Ready + prepull busybox (or skip)" block is repeated verbatim in `test_clusterip_service_reachable_cross_node` (L457-462), just with different variable names. Worth factoring into a shared helper given a third pod-datapath test is likely to land here later.

tests/native_host/wg_e2e/test_wg_mesh.py
```diff
@@ -0,0 +1,740 @@
...
+  leader is stopped, survivors re-elect, and the restarted controller rejoins
+  as a follower.
+* D6 — login-node reachability: spur-controller + k8s-controller + login node,
+  where the login node is meshed but outside k0s scope, still reaches every
+  mesh node.
+
+Every test is marked ``wireguard``; those that need k0s convergence are also
+marked ``k0s`` (hardware-only — CI VMs without the kernel bits skip via the
+suite preflight).
+"""
```
shiv-tyagi: This scenario list documents D1-D6 only, but the file also implements D7 (`TestGracefulK8sRemoveAdd`, L579) and D9 (`TestServiceCidrOverMesh`, L428), and D8 doesn't exist anywhere. Please update the list (and account for the gap) so it matches what's actually implemented.

tests/native_wg_host/e2e/test_wg_mesh.py
```diff
@@ -0,0 +1,740 @@
...
+        cluster to only the k8s controller (node1)."""
+        c = wg_login_topology
+        spur_ctrl, k8s_ctrl, login = c.node_names[0], c.node_names[1], c.node_names[2]
+
+        # Scope k0s to the k8s controller only; spur-ctrl (node0) and login
+        # (node2) are meshed but not k0s members.
+        c.k8s_up(["--nodes", k8s_ctrl, "--control-plane-node", k8s_ctrl])
+        c.wait_k8s_phase("ready", timeout=600)
+
+        login_index = 2
```
shiv-tyagi: `spur_ctrl` and `login` are unpacked but never referenced again, `login_index = 2` is hardcoded separately instead of `c.node_names.index(login)`, which is the pattern used elsewhere in this file (see `third_index` in D4).

tests/native_wg_host/e2e/test_wg_mesh.py
```diff
@@ -0,0 +1,740 @@
...
+
+Every test is marked ``wireguard``; those that need k0s convergence are also
+marked ``k0s`` (hardware-only — CI VMs without the kernel bits skip via the
+suite preflight).
+"""
+
+from __future__ import annotations
+
+import re
+import time
```
shiv-tyagi: Nit: unused import, nothing in this file calls `time.*` directly (the polling helpers all go through `wait_until`, imported separately).

tests/native_wg_host/e2e/conftest.py
```diff
@@ -0,0 +1,310 @@
...
+    config.addinivalue_line(
+        "markers",
+        "wireguard: WireGuard mesh E2E tests (need the wg kernel module + "
+        "wireguard-tools + rootful sudo; slow)",
+    )
+    config.addinivalue_line(
+        "markers",
+        "k0s: native spur-managed k0s cluster tests (rootful spurd + systemd + "
+        "etcd; slow)",
+    )
```
shiv-tyagi: Nit: `k0s` is already registered with the identical description in `native_host/e2e/conftest.py`. pytest dedupes `addinivalue_line` calls so this is harmless, but it's a copy that'll drift if one gets edited and not the other. Consider declaring only `wireguard` here.

.gitignore
```diff
@@ -19,6 +19,9 @@ Thumbs.db
 *.pem
 *.key
 
+# codie (per-engineer, not shared)
+/codie-config.yaml
+
```
shiv-tyagi: Q: `codie-config.yaml` and `local-testbed.yaml` (also added below at L47-48) aren't referenced anywhere in the new suite or docs. Rebase artifact from #724, or an unrelated local-tooling ignore that snuck into this diff? If unrelated, this belongs in its own PR, not one titled "add WireGuard mesh e2e suite".

tests/README.md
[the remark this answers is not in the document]
spraveenio: Done (re-confirming post-rebase, since the earlier fix SHA was squashed away): README now states the suite verifies mesh bring-up, k0s-over-mesh, and pod/service datapath, with only the HA scenario called out as scaffolded/skipped.

tests/native_wg_host/e2e/conftest.py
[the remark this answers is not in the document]
spraveenio: Done (re-confirming post-rebase): the `wg_mesh` fixture calls `c.root_agent_preflight()` after `provision()`, so a node without NOPASSWD sudo (and no `SPUR_TEST_SSH_PASSWORD`) skips the mesh-only tests cleanly, matching the k0s fixtures.

docs/developer/building.rst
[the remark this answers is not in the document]
spraveenio: Done (re-confirming post-rebase): building.rst separates what is automated today from the scaffolded-but-skipped HA scenario, so readers don't assume the skipped path runs in CI.

tests/native_wg_host/e2e/conftest.py
spraveenio: Reordered in `_teardown_wg_k0s`: `c.teardown()` now runs *before* `_reset_k0s_all_nodes(c)`, so spurctld/spurd are stopped first and the async reconcile kicked off by `k8s_down(reset=True)` can no longer race the manual `k0s reset`/`rm -rf`. Comment on the method explains the ordering.

docs/developer/building.rst
spraveenio: Reworded — building.rst now says the suite is folded into the existing `E2E / native-host` job (shares its sharded VM cluster) rather than reporting as a separate `E2E / native-host-wg` check.

tests/README.md
spraveenio: Narrowed — README now lists mesh bring-up, k0s-over-mesh, and cross-node pod/service datapath as covered, and calls out only the 3-controller HA scenario as scaffolded/skipped.

.github/workflows/e2e.yml
spraveenio: Trimmed — the multi-line UNVERIFIED note is gone. The step keeps a short comment explaining it installs wg best-effort and the suite preflight skips (not fails) when a usable data plane is absent.

tests/native_wg_host/e2e/wg_cluster.py
spraveenio: Scoped — `bring_up` now calls `full_mesh_by_add_peer(indices[1:])`. The head↔worker pairs are already wired by the join loop (net_join + the head's add-peer), so the full-mesh pass only needs the worker↔worker pairs.

tests/native_host/wg_e2e/conftest.py
spraveenio: Added a post-reset check per node: after the best-effort stop/reset, it tests `/var/lib/k0s` and logs a warning naming the node if the datadir survived, so a failed cleanup is attributed to the offending node instead of surfacing as a confusing failure in the next test.

tests/native_wg_host/e2e/test_wg_mesh.py
spraveenio: Strengthened — the idempotent-remove test now snapshots the wg peer table before and after and asserts it is unchanged, so a no-op removal (or an accidental disturbance of an existing peer) would actually be caught, not just the CLI success string.

tests/native_wg_host/e2e/wg_cluster.py
spraveenio: Removed `has_route` — it had no callers. The pod-datapath proof relies on the rising per-peer transfer counter, not a route check.

tests/native_host/e2e/wg_cluster.py
spraveenio: Reworded the `_default_mesh_ips` docstring to state the real invariant: the controller adopts each meshed node's advertised `spur0` address as authoritative rather than re-deriving from a sorted pool, so the only thing this map must preserve is not colliding with node 0's reserved `.1`. The name-sort is just for stable assignment, not tied to any k0s pool order.

tests/native_wg_host/e2e/wg_cluster.py
spraveenio: Rewrote `assert_all_to_all` to use the existing `wait_until` per (src,dst) pair, so each pair gets its own full `settle_s` budget instead of sharing one deadline computed before the loop. Removed the hand-rolled poll loop.

tests/native_wg_host/e2e/wg_cluster.py
spraveenio: Dropped `mesh_ip` — it was deprecated-on-arrival with no callers; `WgMesh._default_mesh_ips` covers the raw-mesh path.

tests/native_wg_host/e2e/wg_cluster.py
spraveenio: Removed `net_remove_peer_allow_fail` (and its now-orphaned `_spur_allow_fail` helper) — no callers.

tests/native_wg_host/e2e/test_wg_mesh.py
spraveenio: Simplified to `assert "removed" in out.lower()`. Also added a before/after peer-table comparison so the test proves a real no-op (per yansun's related comment).

tests/native_wg_host/e2e/test_wg_mesh.py
spraveenio: Extracted `_assert_kubectl_applied(out, what)`; the three apply sites (pod, httpd pod, service) now call it.

tests/native_wg_host/e2e/test_wg_mesh.py
spraveenio: Factored the Ready+calico+prepull block into `_ready_two_workers_with_image(...)`, shared by the D3 and D9 cross-node datapath tests.

tests/native_host/wg_e2e/test_wg_mesh.py
spraveenio: Updated the scenario list to the real set — D1, D2, D3, D4, D5 (skipped), D6, D7, D9 — each annotated with its test class, ordered by scenario id, and noting D8 has no mesh-specific behavior to assert.

tests/native_wg_host/e2e/test_wg_mesh.py
spraveenio: Fixed — D6 no longer unpacks the unused `spur_ctrl`, and `login_index` is now derived via `c.node_names.index(login)` to match the `third_index` pattern in D4.

tests/native_wg_host/e2e/test_wg_mesh.py
spraveenio: Removed the unused `import time`.

tests/native_wg_host/e2e/conftest.py
spraveenio: Consolidated — both `wireguard` and `k0s` markers are now registered once in `tests/pytest.ini`, and the WG conftest's `pytest_configure` is gone. This keeps standalone WG runs valid (a sibling conftest isn't loaded then) while removing the duplication drift.

.gitignore
spraveenio: Removed both `/codie-config.yaml` and `/local-testbed.yaml` from this PR's `.gitignore` — they were a rebase artifact unrelated to the WG suite. `.gitignore` now matches main.

shiv-tyagi (changes requested):
One thing which I missed in earlier review, why is the new suite under `tests/native_wg_hosts` and not `tests/native_host`?

It should be somewhere under `tests/native_host`. Lets keep things organised.

.github/workflows/e2e.yml
```diff
@@ -292,10 +318,14 @@ jobs:
...
           export SPUR_TEST_GPU_VENV=/opt/spur-ci/gpu-venv
+          # Opt in to the WireGuard mesh suite; folded into these shards. The
+          # suite's preflight skips it where a real mesh can't be stood up.
+          export SPUR_TEST_WG=1
           chmod +x ${{ env.RUN_DIR }}/release-binaries/*
           source /opt/spur-ci/test-venv/bin/activate
           rc=0
-          pytest ${{ env.RUN_DIR }}/e2e-assets/tests/native_host/e2e/ -v \
+          pytest ${{ env.RUN_DIR }}/e2e-assets/tests/native_host/e2e/ \
+                 ${{ env.RUN_DIR }}/e2e-assets/tests/native_host/wg_e2e/ -v \
```
shiv-tyagi: This does not look right. We should not touch this. Wireguard tests are eventually e2e tests only and should live under tests/native_host/e2e. There is already a lot of existing scaffolding there, if needed we can rafactor and improve but this would pollute the suite.

.github/workflows/e2e.yml
```diff
@@ -278,6 +278,32 @@ jobs:
...
+            sudo apt-get update -qq && sudo DEBIAN_FRONTEND=noninteractive \
+              apt-get install -y -qq wireguard-tools || true
+          fi
+          # Prefer the kernel module; fall back to userspace wireguard-go.
+          sudo modprobe wireguard 2>/dev/null \
+            || command -v wireguard-go >/dev/null 2>&1 \
+            || sudo DEBIAN_FRONTEND=noninteractive apt-get install -y -qq wireguard-go || true
+          echo "wg: $(command -v wg || echo MISSING); kmod: $(lsmod | grep -q wireguard && echo yes || echo no)"
+          REMOTE
+          done
```
shiv-tyagi: I will bake this into VM images so that we can save time here.

shiv-tyagi (changes requested):

.github/workflows/e2e.yml
```diff
@@ -295,6 +295,9 @@ jobs:
           chmod +x ${{ env.RUN_DIR }}/release-binaries/*
           source /opt/spur-ci/test-venv/bin/activate
           rc=0
+          # WireGuard mesh tests live in native_host/e2e (test_wg_*.py). Their
+          # fixtures install WireGuard on the nodes where missing, so no CI-side
+          # setup is needed; they skip only where a data plane can't be provided.
```
shiv-tyagi: This comment is redundant

tests/native_host/e2e/conftest.py
```diff
@@ -396,3 +405,232 @@ def mpi_cluster(ssh_nodes, remote_bin_dir, cluster_config_overrides):
...
+        # Ensure a data plane: prefer the kernel module, else install wireguard-go.
+        node.exec_allow_fail(
+            f"{sudo}modprobe wireguard 2>/dev/null "
+            "|| command -v wireguard-go >/dev/null 2>&1 "
+            f"|| {sudo}DEBIAN_FRONTEND=noninteractive apt-get install -y -qq wireguard-go 2>/dev/null "
+            "|| true"
+        )
+        ok, reason = wg_available(node, sudo)
+        if not ok:
+            pytest.skip(f"WireGuard unavailable on {node.host} and could not be provisioned: {reason}")
```
shiv-tyagi: Instead of trying to install, we can just try detecting and skip if not available. A test should not try hard to install a dependency.

tests/conftest.py
```diff
@@ -92,6 +93,10 @@ def pytest_ignore_collect(collection_path, config):
     path = Path(str(collection_path))
     parts = path.parts
 
+    # WireGuard tests are ordinary native_host/e2e files gated by their fixtures
+    # (they install WireGuard where missing and skip only where a data plane
+    # can't be provided), so the base native_host rule covers them — no separate
+    # WG branch or opt-in var.
```
shiv-tyagi: Redundant comment

tests/pytest.ini
```diff
@@ -9,3 +9,5 @@ addopts = -p no:xdist
 markers =
     rootful: run the cluster fixture with spurd as root (via sudo)
     gpu: requires GPU hardware on the nodes (auto-applied to tests using GPU fixtures)
+    wireguard: WireGuard mesh E2E tests (need the wg kernel module + wireguard-tools + rootful sudo; slow)
+    k0s: native spur-managed k0s cluster tests (rootful spurd + systemd + etcd; slow)
```
shiv-tyagi: Do we really need the new markers?

Can we consolidate into the existing rootful marker?

We can skip if wg kernel module is not available. Do we need to have a marker for it?

tests/README.md
```diff
@@ -4,5 +4,5 @@ Test suites organized by deployment target. See [docs/developer/building.rst](..
 
 | Path | Description |
 |------|-------------|
-| `native_host/e2e/` | Deploys Spur on bare-metal nodes via SSH |
+| `native_host/e2e/` | Deploys Spur on bare-metal nodes via SSH. Includes the WireGuard mesh tests (`test_wg_mesh.py`, `test_wg_k0s.py`, marked `wireguard`), which auto-run where the nodes allow — the fixtures install WireGuard if missing and skip only where a data plane can't be provided |
```
shiv-tyagi: Redundant comment

tests/native_host/e2e/conftest.py
```diff
@@ -397,3 +405,232 @@ def mpi_cluster(ssh_nodes, remote_bin_dir, cluster_config_overrides):
...
+
+
+@pytest.fixture
+def wg_mesh(ssh_nodes, remote_bin_dir):
```
shiv-tyagi: Nit: `wg_mesh` reads like it just hands you a mesh handle, but it provisions a whole `SpurCluster`, installs wireguard-tools if missing, and brings the interface up/down per test. It's also the only WG fixture with no spurctld/spurd running underneath. Consider `raw_wg_mesh` (matches the "raw mesh" wording already used in this docstring and in `test_wg_mesh.py`'s module docstring) so it reads distinctly from `wg_k0s_cluster`.

tests/native_host/e2e/conftest.py
```diff
@@ -397,3 +405,232 @@ def mpi_cluster(ssh_nodes, remote_bin_dir, cluster_config_overrides):
...
+    for an etcd quorum path and rootful sudo. Tears k0s + mesh down on exit."""
+    if len(ssh_nodes) < 3:
+        pytest.skip(f"k0s-over-mesh needs >= 3 nodes for quorum (got {len(ssh_nodes)})")
+    c = _deploy_wg_k0s(ssh_nodes, remote_bin_dir)
+    yield c
+    _teardown_wg_k0s(c)
+
+
+@pytest.fixture
+def wg_login_topology(ssh_nodes, remote_bin_dir):
```
shiv-tyagi: Nit: this returns a plain `SpurCluster` (same shape as `wg_k0s_cluster`), just with a different `control_plane_index`. `wg_login_cluster` would match the `_cluster` suffix used everywhere else in this file; `topology` reads like it's a different return type.

tests/native_host/e2e/conftest.py
```diff
@@ -397,3 +405,232 @@ def mpi_cluster(ssh_nodes, remote_bin_dir, cluster_config_overrides):
...
+# WireGuard tests run automatically wherever the nodes allow a real mesh — there
+# is no opt-in env var (mirroring gpu_cluster / k8s_multicp_cluster, which also
+# auto-run on capable beds). The fixtures ENABLE WireGuard rather than detect it:
+# they flip network.wg_enabled on in config AND install the wg data plane on the
+# node if it is missing (_ensure_wireguard), so the tests execute instead of
+# skipping. They skip only on a genuine environmental impossibility (no kernel
+# module AND no wireguard-go userspace fallback obtainable) or missing sudo.
+
+
+_WG_CIDR = "10.44.0.0/16"
```
shiv-tyagi: `_WG_CIDR` duplicates `wg_cluster.MESH_CIDR` (same literal `10.44.0.0/16`) instead of importing it. Two names for one value, risks drift if either changes.

tests/native_host/e2e/conftest.py
```diff
@@ -397,3 +405,232 @@ def mpi_cluster(ssh_nodes, remote_bin_dir, cluster_config_overrides):
...
+    nodes (the testbed + CI image).
+    """
+    sudo = cluster._sudo_prefix()
+    for node in cluster.nodes:
+        if "OK" not in node.exec_allow_fail("command -v wg >/dev/null && command -v wg-quick >/dev/null && echo OK"):
+            node.exec_allow_fail(
+                f"{sudo}apt-get update -qq && {sudo}DEBIAN_FRONTEND=noninteractive "
+                "apt-get install -y -qq wireguard-tools 2>/dev/null || true"
+            )
+        # Ensure a data plane: prefer the kernel module, else install wireguard-go.
```
shiv-tyagi: This probes modprobe/wireguard-go, then `wg_available()` right below (line 470) does the exact same three-way check again. If this reverts to detect+skip (per the comment above on this function), this whole block goes away; if it stays, it shouldn't probe twice.

tests/native_host/e2e/test_wg_k0s.py
```diff
@@ -0,0 +1,739 @@
+# Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""End-to-end tests for k0s-over-WireGuard, driven over real `wg` interfaces.
+
+The raw-mesh (no-k0s) bring-up scenario (D1) lives in ``test_wg_mesh.py``; this
+file holds the scenarios that need a converged k0s cluster on top of the mesh.
+The D-numbers are the SPUR-212 scenario ids; not every id has a test (D8 has no
```
shiv-tyagi: SPUR-212 is a ticket ID in a code comment (also lines 22 and ~706). AGENTS.md: "Do not reference issue numbers, PR numbers, or task IDs in code comments. That context belongs in git history, not in the source." Separately: once the ticket ref is gone, the D1-D9 labels are magic numbers a reader can't decode without that ticket open. Consider dropping the D-numbering from comments and letting the test/class names carry the description on their own.

tests/native_host/e2e/test_wg_k0s.py
```diff
@@ -0,0 +1,739 @@
...
+    out = c.nodes[cp_index].exec_allow_fail(
+        f"{c._sudo_prefix()}k0s kubectl apply -f {remote_path} 2>&1"
+    )
+    _assert_kubectl_applied(out, f"pod {name}")
+    return name
+
+
+def _wait_pod_ip(c, cp_index: int, pod: str, timeout_s: int = 180) -> str:
+    """Poll until the pod reports a pod-CIDR IP (Running with an IP assigned)."""
+    ipv4 = re.compile(r"^\d{1,3}(?:\.\d{1,3}){3}$")
```
shiv-tyagi: Nit: this exact regex (`^\d{1,3}(?:\.\d{1,3}){3}$`) is recompiled two more times in this file (~line 190, ~line 277). Hoist to a single module-level constant.

tests/native_host/e2e/test_wg_k0s.py
```diff
@@ -0,0 +1,739 @@
...
+        cip = _kubectl(c, cp_index,
+                       f"get svc {svc} -o jsonpath='{{.spec.clusterIP}}'").strip()
+        return cip if ipv4.match(cip) else ""
+
+    wait_until(lambda: bool(read_cip()), timeout_s=60,
+               desc=f"service {svc} got a ClusterIP")
+    return read_cip()
+
+
+BUSYBOX_IMAGE = "docker.io/library/busybox:1.36"
```
shiv-tyagi: Nit: used at lines ~107 and ~156, defined here. Not a bug (module globals resolve at call time), but move it up near the top with the other constants so it doesn't read as used-before-definition.

tests/native_host/e2e/test_wg_k0s.py
```diff
@@ -0,0 +1,739 @@
...
+    """Topology: node0 = spur controller + k8s control-plane, node1 + node2 =
+    k8s workers. The test cycles node1 out of and back into k0s while node2 stays
+    a worker throughout — a single-worker (CP + 1) cluster can't converge calico/
+    coredns (nothing schedulable is left once the sole worker drains), so keeping
+    a second worker keeps the cluster healthy across the remove/add.
+
+    Validates that a k8s worker can be cleanly cycled out of and back into the
+    k0s cluster WITHOUT losing its Spur identity or WireGuard keys:
+
+      1. Manually cordon + drain the worker (operator pre-drain), then
```
shiv-tyagi: This numbered list restates verbatim what the inline `# 1.`, `# 2.`, `# 3.` comments already say inside the test body below. Pick one place, probably trim this docstring to the invariant being tested and let the inline comments carry the step-by-step where the code actually is.

tests/native_host/e2e/wg_cluster.py
```diff
@@ -0,0 +1,320 @@
...
+                "--interface", self.iface]
+        if endpoint:
+            args += ["--endpoint", endpoint]
+        if pod_cidr:
+            args += ["--pod-cidr", pod_cidr]
+        if program_routes:
+            args += ["--program-routes"]
+        return self._spur(self.nodes[on_index], args)
+
+    def net_remove_peer(self, on_index: int, peer_key: str) -> str:
```
shiv-tyagi: `net_remove_peer` isn't called anywhere. Both D4 tests call `spur net remove-peer` directly via `c.cli_as_user(...)` instead of through this method. Same category as the unused helpers already dropped this round (`mesh_ip`, `has_route`, etc). Drop it, or have the tests use it instead of the direct CLI calls.

tests/native_host/e2e/wg_cluster.py
```diff
@@ -0,0 +1,320 @@
...
+  rides a specific WireGuard tunnel (rising transfer counters on the exact
+  peer, per-peer routes, pod-CIDR folded into AllowedIPs).
+
+Design: this is a thin *composition* over ``SpurCluster``, not a subclass — the
+mesh lifecycle (``net init``/``join``/``mesh``) is orthogonal to the daemon
+lifecycle, and keeping them separate avoids overriding the base class's private
+start helpers. Tests that need the daemon stack (k0s scenarios) drive the base
+cluster directly; tests that only need the raw mesh (bring-up, remove-peer) use
+the mesh helpers here.
+"""
```
shiv-tyagi, line 11: This module docstring is a 21-line bulleted table of contents ("here's what's in this file") that just restates what the classes/functions below it already show. The one part worth keeping is the last paragraph (composition vs subclass rationale) since that's a real "why" that isn't obvious from the code. Trim the rest.

.github/workflows/e2e.yml
spraveenio: Reverted — `.github/workflows/e2e.yml` is back to main (no WG changes at all). The tests now sit under `tests/native_host/e2e` and are picked up by the existing native-host pytest invocation and sharding.

.github/workflows/e2e.yml
spraveenio: Makes sense — with that in place the fixtures just detect the kmod and run. I've switched them to detect-and-skip (below), so no CI-side install and no workflow change.

.github/workflows/e2e.yml
spraveenio: Moot now — e2e.yml is fully reverted to main, so the comment is gone.

tests/native_host/e2e/conftest.py
spraveenio: Done — `_ensure_wireguard` is replaced by `_require_wireguard`, which only checks `wg_available()` and `pytest.skip`s when the data plane is absent. No install. (This also fixes the CI timeouts: they were the k0s-over-mesh tests running over the slow userspace wireguard-go; with the baked kmod image they'll run on a real module.)

tests/conftest.py
spraveenio: Removed the comment.

tests/pytest.ini
spraveenio: Dropped the `wireguard` marker entirely and reverted `pytest.ini` to main (only `rootful` + `gpu`). The WG tests are selected by file path (`test_wg_mesh.py` / `test_wg_k0s.py`); the k0s ones keep the existing `k0s` marker. No new markers.

tests/README.md
spraveenio: Trimmed the README row to just name the two WG test files.

tests/native_host/e2e/conftest.py
spraveenio: Renamed to `raw_wg_mesh` — matches the 'raw mesh' wording and reads distinctly from `wg_k0s_cluster`.

tests/native_host/e2e/conftest.py
spraveenio: Renamed to `wg_login_cluster` to match the `_cluster` suffix used by the other fixtures.

tests/native_host/e2e/conftest.py
spraveenio: Now imports `MESH_CIDR` from `wg_cluster` instead of the duplicate `_WG_CIDR` literal.

tests/native_host/e2e/conftest.py
spraveenio: Reverted to detect-and-skip (`_require_wireguard`), so the double-probe is gone — it's a single `wg_available()` check.

tests/native_host/e2e/test_wg_k0s.py
spraveenio: Removed the SPUR-212 refs (docstring + the D5 skip message) and dropped the D1–D9 numbering from the section headers and docstrings; the class/test names carry the description now.

tests/native_host/e2e/test_wg_k0s.py
spraveenio: Hoisted to a module-level `_IPV4_RE` constant; the three call sites use it.

tests/native_host/e2e/test_wg_k0s.py
spraveenio: Moved `BUSYBOX_IMAGE` up to the top with the other module constants.

tests/native_host/e2e/test_wg_k0s.py
spraveenio: Trimmed the class docstring to the invariant being tested; the step-by-step lives in the inline `# 1./2./3.` comments in the test body.

tests/native_host/e2e/wg_cluster.py
spraveenio: Dropped `net_remove_peer` — the D4 tests call `spur net remove-peer` directly via `cli_as_user`, so it had no callers.

tests/native_host/e2e/wg_cluster.py
spraveenio: Trimmed the module docstring to the composition-vs-subclass rationale (the real 'why'); dropped the bulleted table-of-contents.

tests/README.md
```diff
@@ -4,5 +4,5 @@ Test suites organized by deployment target. See [docs/developer/building.rst](..
 
 | Path | Description |
 |------|-------------|
-| `native_host/e2e/` | Deploys Spur on bare-metal nodes via SSH |
+| `native_host/e2e/` | Deploys Spur on bare-metal nodes via SSH. Includes the WireGuard mesh tests (`test_wg_mesh.py`, `test_wg_k0s.py`) |
```
shiv-tyagi:
```suggestion
| `native_host/e2e/` | Deploys Spur on bare-metal nodes via SSH. |
```

docs/developer/building.rst
```diff
@@ -218,6 +211,61 @@ Running the Tests
...
+   pytest tests/native_host/e2e/test_wg_mesh.py tests/native_host/e2e/test_wg_k0s.py -v
+
+   # Only the raw mesh-bring-up scenarios (no k0s convergence needed)
+   pytest tests/native_host/e2e/test_wg_mesh.py -v
+
+In CI these tests run inside the existing ``E2E / native-host`` job and its
+shards — no separate suite, path, or status check. Where a shard's nodes can't
+provide a WireGuard data plane, the mesh tests skip while the rest of the shard
+runs normally.
+
```
shiv-tyagi: All of this is not needed. We do not explain what each tests do. Only the above **Running the tests** paragraph is enough.

shiv-tyagi: @spraveenio Just apply this suggestion if it looks okay (remove this paragraph) and we can merge this. This paragraph is just narrating stuff and the language is too verbose and of no use.

spraveenio: done

tests/README.md
shiv-tyagi:
```suggestion
| `native_host/e2e/` | Deploys Spur on bare-metal nodes via SSH. |
```

shiv-tyagi (approved):
LGTM

PR #722, opened 2026-08-24
fix(spurctld): reject unusable --nodelist/--exclude at submission (follow-up to #679)

### LINKED ISSUE #822: [Issue]: Over-cap or malformed --nodelist/--exclude is accepted at submission, then the job never runs
## Problem Description

Submitting a job with a `--nodelist` or `--exclude` value that cannot be expanded — either a malformed hostlist pattern, or one that expands past the hostlist size cap (1,000,000 hosts) — is accepted by the controller. The unexpandable pattern falls back to a literal token that matches no node.

- A bad `--nodelist` then matches nothing, so the job sits pending forever with no reason reported.
- A bad `--exclude` silently excludes nothing, so a job the user meant to keep off specific nodes can land on exactly those nodes.

Both outcomes are invisible to the submitter.

**Expected:** the submission is rejected up front with a clear error naming the offending flag and pattern, the way Slurm rejects an unusable nodelist/exclude at submit time.

## Environment

- **OS:** Ubuntu 22.04 LTS
- **Component:** spur

## Steps to Reproduce

```bash
sbatch --nodelist='node]1-2[' --wrap 'hostname'       # malformed pattern
sbatch --nodelist='node[0-2000000]' --wrap 'hostname' # exceeds the host cap

SumonAMD:
## Summary

Closes #822

Follow-up to #679, addressing the two review items @sajmera-pensando raised, the
three from @spraveenio's review, and one further bug that the third of those
exposed.

## What was wrong

1. **The two hostlist cap guards disagreed.** The pre-loop guard rejects when the
   total *would exceed* `MAX_HOSTLIST_SIZE`, but the in-loop backstop ran *after*
   the push and compared with `>`, so `results` could transiently hold
   `MAX_HOSTLIST_SIZE + 1`.

2. **An over-cap or malformed `--nodelist` / `--exclude` failed silently.**
   `expand_hostlist_or_split` catches any expansion error and comma-splits the raw
   pattern, yielding tokens that match no registered node. A bad `--nodelist` then
   never schedules and the job pends with no error the submitter can see; a bad
   `--exclude` excludes nothing, so a job meant to stay off specific nodes can land
   on exactly those nodes. Slurm rejects both at submission.

3. **The cap did not bound bare names** — found while addressing @spraveenio's
   double-materialization point. Only the *range* loop tested `MAX_HOSTLIST_SIZE`, so
   a bare term appended to a cap-filling range took the total past it:
   `node[0-999999],extra` expanded to 1,000,001 hosts and passed validation,
   reproducing exactly the silent pend this PR exists to prevent.

## Approach

- **`hostlist.rs`** — `expand` and the counting path now share a single traversal,
  `walk`, over a `HostSink`: `Collect` pushes names, `Tally` only counts.
  `Tally::emit` discards the `fmt::Arguments` it is handed, so `format_args!` at the
  call site formats nothing and counting allocates no per-host `String`. Every
  host-producing path funnels through `emit_one`, which enforces the cap *before*
  emitting — that is what fixes [3], and sharing one parser removes the possibility
  of the two paths drifting apart again. `count` becomes the single non-allocating
  entry point. `TooLarge` drops its `count` field: the in-loop path cannot know the
  true request size without expanding (the allocation being prevented), so it
  reported a misleading `MAX + 1`.
- **`server.rs`** — `proto_to_job_spec` rejects an unusable `--nodelist` or
  `--exclude` with `InvalidArgument`, naming the flag, the pattern and the reason.
- **`docs/user-guide/submitting-jobs.rst`** — both flag rows now state that a
  malformed or over-cap pattern is rejected at submission.

Two deliberate choices worth flagging:

Rejecting `--exclude` as strictly as `--nodelist` is intended. Both failures are
invisible to the submitter, and the `--exclude` case is the worse of the two, since
the job runs on nodes the user explicitly tried to avoid.

`expand` now returning `TooLarge` where it previously leaked past the cap is a
behaviour change, but it enforces an already-documented limit. The scheduler,
partition matching and topology paths all already treat an expansion error as "no
match", so only reservation creation and the CLI node filters surface it, and only
for patterns that were never supported.

## Test plan

Unit and lint:

- [x] `count_agrees_with_expand_on_every_pattern` — differential over a table
      covering every branch of the grammar; `count` must match `expand` on
      acceptance, on length, and on error class
- [x] `the_cap_bounds_bare_names_too` — pins [3] in both directions: exactly
      1,000,000 expands, 1,000,001 is `TooLarge`
- [x] `expand_allows_cap_boundary`, `expand_rejects_nested_product_over_cap`,
      `expand_rejects_recursive_overshoot_past_cap`
- [x] `proto_to_job_spec_rejects_unexpandable_node_patterns`
- [x] `cargo fmt --check`, `cargo clippy --workspace --exclude spur-ffi
      --all-targets -- -D warnings`, full workspace suite, `cargo doc` — all clean

On a two-node MI300X cluster, three builds of the same six scenarios — the merge
base, this branch before the last commit, and this branch:

| Submitted | merge base | before last commit | this branch |
|---|---|---|---|
| `--nodelist=node[003-001]` | accepted, pends forever | rejected | rejected |
| `--nodelist=node[0-2000000]` | accepted, pends forever | rejected | rejected |
| typo'd reversed-range `--exclude` over a real node | accepted, **ran on the node it was meant to exclude** | rejected | rejected |
| `--nodelist=node[0-999999],extra` | accepted, pends forever | accepted, pends forever | rejected |
| valid `--nodelist` naming both nodes | ran on both | ran on both | ran on both |
| valid `--exclude` of one node | ran on the other | ran on the other | ran on the other |
| `--nodelist=node[0-999999]` (exactly at the cap) | accepted | accepted | accepted |

### DIFF
--- a/crates/spur-core/src/hostlist.rs
+++ b/crates/spur-core/src/hostlist.rs
@@ -421,6 +421,79 @@ pub fn count(pattern: &str) -> Result<usize, HostlistError> {
     Ok(expand(pattern)?.len())
 }
 
+/// Validate a hostlist pattern without materializing it.
+///
+/// Applies exactly the same grammar and error conditions as [`expand`] —
+/// unmatched or misordered brackets, reversed ranges, and the
+/// `MAX_HOSTLIST_SIZE` cap — but counts hosts instead of building the
+/// `Vec<String>`. The submit path validates attacker-influenced `--nodelist`
+/// and `--exclude` here, so counting avoids the throwaway multi-megabyte
+/// allocation a full `expand` would make only to drop it.
+pub fn validate(pattern: &str) -> Result<(), HostlistError> {
+    let mut count: u64 = 0;
+    for part in split_top_level(pattern) {
+        count_single(part.trim(), &mut count)?;
+    }
+    Ok(())
+}
+
+/// Counting twin of [`expand_single`]: accumulates the host count into `count`,
+/// enforcing the cap with the same pre-loop and incremental guards rather than
+/// pushing names. Kept structurally parallel to `expand_single` so the two
+/// cannot drift in what they accept.
+fn count_single(pattern: &str, count: &mut u64) -> Result<(), HostlistError> {
+    if let Some(bracket_start) = pattern.find('[') {
+        let bracket_end = pattern
+            .find(']')
+            .ok_or_else(|| HostlistError::InvalidPattern("unmatched [".into()))?;
+
+        // ']' before '[' would make the slice below reversed (start > end) and panic.
+        if bracket_end < bracket_start {
+            return Err(HostlistError::InvalidPattern(format!(
+                "']' before '[': {pattern}"
+            )));
+        }
+
+        let prefix = &pattern[..bracket_start];
+        let range_str = &pattern[bracket_start + 1..bracket_end];
+        let suffix = &pattern[bracket_end + 1..];
+
+        for range_part in range_str.split(',') {
+            if let Some((start, end, width)) = parse_range_bounds(range_part)? {
+                let range_count = (end - start).saturating_add(1);
+                if count.saturating_add(range_count) > MAX_HOSTLIST_SIZE as u64 {
+                    return Err(HostlistError::TooLarge {
+                        max: MAX_HOSTLIST_SIZE,
+                    });
+                }
+
+                for i in start..=end {
+                    // Backstop nested products that multiply past the cap.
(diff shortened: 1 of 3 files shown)


SumonAMD:
Pushed `e102fb0`, which revisits the double-materialization point and fixes a bug it
turned up.

PR #725, opened 2026-08-24
fix(spur-core): dedup array ranges before enforcing the size cap (follow-up to #646)

### LINKED ISSUE #823: [Issue]: Job array with overlapping ranges that dedup within the cap is wrongly rejected
## Problem Description

The job-array size guard counts tasks before duplicate indices are removed. An array spec whose overlapping ranges dedup to a count within the cap is rejected anyway, because the pre-dedup total exceeds the cap.

For example `--array=0-99999,0-99999` is counted as 200,000 tasks and rejected, even though the deduplicated set is exactly 100,000 distinct indices — within the limit.

**Expected:** the spec is accepted and produces 100,000 tasks, since that is the actual number of distinct array indices. Only specs whose distinct task count genuinely exceeds the cap should be rejected.

## Environment

- **OS:** Ubuntu 22.04 LTS
- **Component:** spur

## Steps to Reproduce

```bash
sbatch --array=0-99999,0-99999 --wrap 'true'

### PR DESCRIPTION
**Pull request by SumonAMD**
## Summary

Closes #823

Follow-up to #646, addressing the review feedback from @sajmera-pensando.

The array-spec size guard counted tasks **before** `dedup` ran, so a spec whose
overlapping ranges dedup to within the cap was wrongly rejected. For example
`0-99999,0-99999` counts as 200,000 up front and errors, even though the deduped
result is exactly 100,000 (valid). This also makes the original PR's "no behavior
change for valid input" claim accurate.

## Changes

- **Dedup over-count fix.** Bound the ids a spec may name *before* dedup
  (`MAX_ARRAY_SPEC_IDS`, twice the cap) rather than bounding the current buffer
  length. Overlapping ranges still dedup down and pass, and genuinely-oversized
  distinct specs still fail. Peak memory stays bounded (~2x the cap), so the
  many-overlapping-ranges OOM path #646 closed stays closed.
- **Fixed parsing cost.** Because total appends are capped, no in-loop compaction
  is needed: ids are collected once and sorted once. Previously a spec repeating
  an in-cap range drove one full sort+dedup pass per segment, with the segment
  count bounded only by input length. `spurctld` parses `--array` on the submit
  path, so that cost was reachable by any client: a 32 KiB spec took **11.76s** of
  controller time before returning its error, and now takes **0.034s** to return
  the same error. Single ids are charged too, since appending `,1` after a full
  range was the cheapest way to trigger the old repeated-sort path.
- **Test rename** (nit): `rejects_range_overflow_at_u32_max` ->
  `parses_range_at_u32_boundary`, since it asserts success, not rejection.
- **Comment** (nit): the count-guard comment now explains *why* (bounded cost,
  u64 avoids count overflow) rather than *what*.
- **Docs**: the job array size limits are described in the Job Arrays section of
  the user guide.

## Behavior

| Spec | Before | After |
|---|---|---|
| `0-99999,0-99999` | rejected (false positive) | accepted -> 100,000 tasks |
| `0-4000000000` | rejected before alloc | rejected before alloc |
| `0-99999,100000-199999` | rejected | rejected |
| `0-100000` | rejected | rejected |
| `0-99999` x3 | rejected | rejected (now by the pre-dedup bound) |
| `1` repeated 150,000 times | rejected | accepted -> 1 task |

Not a breaking change: the new bound accepts a strict superset of what `main`
accepts today, so no currently-valid spec starts failing.

## Changes since the first review

The tradeoff introduced here is that a spec naming more than 200,000 indices is
rejected however far it would collapse, so `0-99999` repeated three times no
longer parses. Two overlapping copies -- the case this PR exists to fix -- still
work. `main` already rejects the three-copy spec, so this only narrows the
previous revision of this PR, not shipped behavior.

## Tests

- `overlapping_ranges_dedup_within_cap` -- the reviewer's exact case, asserting
  exact contents.
- `distinct_ranges_over_cap_still_rejected` -- genuinely-oversized specs.
- `ranges_past_the_pre_dedup_cap_are_rejected` -- repeated in-cap ranges.
- `single_ids_count_toward_the_pre_dedup_cap` -- the `,1` repetition path.

Both new guard tests are fast against the pre-fix code as well, so a regression
fails cleanly instead of hanging the suite.

`cargo fmt` and `cargo clippy --workspace --exclude spur-ffi --all-targets
--locked` are clean; 3,685 tests pass across 23 suites.

### DIFF
--- a/crates/spur-core/src/array.rs
+++ b/crates/spur-core/src/array.rs
@@ -123,11 +123,11 @@ pub fn parse_array_spec(spec: &str) -> Result<ArraySpec, ArrayError> {
                 return Err(ArrayError::InvalidSpec(format!("{} > {}", start, end)));
             }
 
-            // Bound element count BEFORE materializing (u64 avoids overflow in the count itself).
+            // Reject an over-cap range before the loop allocates it; u64 avoids count overflow.
             let count = (end as u64 - start as u64) / step as u64 + 1;
-            if task_ids.len() as u64 + count > MAX_ARRAY_SIZE as u64 {
+            if count > MAX_ARRAY_SIZE as u64 {
                 return Err(ArrayError::TooLarge {
-                    count: (task_ids.len() as u64 + count) as usize,
+                    count: count as usize,
                     max: MAX_ARRAY_SIZE,
                 });
             }
@@ -146,13 +146,18 @@ pub fn parse_array_spec(spec: &str) -> Result<ArraySpec, ArrayError> {
                 .map_err(|_| ArrayError::InvalidSpec(format!("invalid id: {}", range_str)))?;
             task_ids.push(id);
         }
-    }
 
-    if task_ids.len() > MAX_ARRAY_SIZE {
-        return Err(ArrayError::TooLarge {
-            count: task_ids.len(),
-            max: MAX_ARRAY_SIZE,
-        });
+        // Compact overlapping ranges to keep peak memory bounded; oversized sets still fail.
+        if task_ids.len() > MAX_ARRAY_SIZE {
+            task_ids.sort_unstable();
+            task_ids.dedup();
+            if task_ids.len() > MAX_ARRAY_SIZE {
+                return Err(ArrayError::TooLarge {
+                    count: task_ids.len(),
+                    max: MAX_ARRAY_SIZE,
+                });
+            }
+        }
     }
 
     task_ids.sort();
@@ -320,7 +325,7 @@ mod tests {
     }
 
     #[test]
-    fn rejects_range_overflow_at_u32_max() {
+    fn parses_range_at_u32_boundary() {
         let spec = parse_array_spec("4294967293-4294967295:2").unwrap();
         assert_eq!(spec.task_ids, vec![4294967293, 4294967295]);
     }
@@ -337,5 +342,20 @@ mod tests {
     fn max_size_boundary_unchanged() {
         assert!(parse_array_spec("0-99999").is_ok());
         assert!(parse_array_spec("0-100000").is_err());
+    }
+
+    #[test]
+    fn overlapping_ranges_dedup_within_cap() {
+        // Regression: pre-dedup counting wrongly rejected identical ranges at the cap.
+        let spec = parse_array_spec("0-99999,0-99999").unwrap();
+        assert_eq!(spec.task_ids.len(), 100_000);
+    }
+
+    #[test]
+    fn distinct_ranges_over_cap_still_rejected() {
+        assert!(matches!(
+            parse_array_spec("0-99999,100000-199999"),
+            Err(ArrayError::TooLarge { .. })
+        ));
     }
 }

### CODE REVIEW

File: crates/spur-core/src/array.rs [hunk 1]
```diff
@@ -123,11 +123,11 @@ pub fn parse_array_spec(spec: &str) -> Result<ArraySpec, ArrayError> {
                 return Err(ArrayError::InvalidSpec(format!("{} > {}", start, end)));
             }
 
-            // Bound element count BEFORE materializing (u64 avoids overflow in the count itself).
+            // Reject an over-cap range before the loop allocates it; u64 avoids count overflow.
             let count = (end as u64 - start as u64) / step as u64 + 1;
```
sajmera-pensando, line 127: **Missing segment-count guard — unbounded repetitions enable CPU-amplification DoS.**

The old check `task_ids.len() + count > MAX_ARRAY_SIZE` served double duty: it both enforced the cap and provided a cheap early-exit for repeated ranges. The new check `count > MAX_ARRAY_SIZE` only rejects a single range that *alone* exceeds the cap.

For the spec `0-99999` repeated N times:
- Each range has `count == 100_000`, so `count > 100_000` is `false` → no per-range rejection.
- After two repetitions the Vec reaches 200,000 elements, triggering the in-loop compaction: `sort_unstable` + `dedup` on 200,000 `u32`s → back to 100,000.
- This cycle repeats for every additional segment — total CPU is O(N × MAX_ARRAY_SIZE × log MAX_ARRAY_SIZE), with N bounded only by input string length.

Peak heap is fine (~800 KB), but the old code handled this at zero CPU cost (rejected at the second repetition). Consider one of:

```suggestion
            // Bound element count BEFORE materializing (u64 avoids overflow in the count itself).
            let count = (end as u64 - start as u64) / step as u64 + 1;
            if count > MAX_ARRAY_SIZE as u64 {
                return Err(ArrayError::TooLarge {
                    count: count as usize,
                    max: MAX_ARRAY_SIZE,
                });
            }
            // Fast pre-dedup guard: if cumulative pre-dedup count is already far above the cap
            // (2x), overlapping ranges cannot bring it within bounds; bail before sort+dedup work.
            if task_ids.len() as u64 + count > 2 * MAX_ARRAY_SIZE as u64 {
                return Err(ArrayError::TooLarge {
                    count: (task_ids.len() as u64 + count) as usize,
                    max: MAX_ARRAY_SIZE,
                });
            }
```

Alternatively, cap the number of comma-separated segments (e.g. `MAX_SEGMENTS = 1_024`) before the loop.

crates/spur-core/src/array.rs
```diff
@@ -146,13 +146,18 @@ pub fn parse_array_spec(spec: &str) -> Result<ArraySpec, ArrayError> {
                 .map_err(|_| ArrayError::InvalidSpec(format!("invalid id: {}", range_str)))?;
             task_ids.push(id);
```
sajmera-pensando, line 147: **`sort_unstable` inside the loop, stable `sort` after — inconsistent without an apparent reason.**

The in-loop compaction correctly uses `sort_unstable` (faster; stability is irrelevant for `u32`). The post-loop pass retains the old `task_ids.sort()`. The results are identical, but a reader who just saw `sort_unstable` will reasonably wonder whether stability is required at the final site.

```suggestion
    task_ids.sort_unstable();
    task_ids.dedup();
```

crates/spur-core/src/array.rs
```diff
@@ -338,4 +343,19 @@ mod tests {
         assert!(parse_array_spec("0-99999").is_ok());
```
sajmera-pensando, line 343: **Regression test asserts length only — a wrong-content bug (e.g. off-by-one in dedup) would pass.**

Since `parse_array_spec` guarantees a sorted, deduplicated result, asserting the exact content costs nothing and locks down the behavior completely.

```suggestion
        assert_eq!(spec.task_ids, (0u32..=99_999).collect::<Vec<_>>());
```

crates/spur-core/src/array.rs
```diff
@@ -338,4 +343,19 @@ mod tests {
...
         assert!(parse_array_spec("0-100000").is_err());
```
sajmera-pensando, line 344: **Comment narrates the historical bug rather than stating the invariant.**

Per the project conventions, comments should explain non-obvious constraints, not describe what a fix changed. The phrase "pre-dedup counting" describes an implementation detail of the old code that no longer exists.

```suggestion
        // Overlapping ranges must dedup before the size cap is applied; 0-99999,0-99999 is 100,000 distinct tasks.
```

sajmera-pensando (changes requested):
## Review

The fix is directionally correct — deduplicating before enforcing the size cap matches the expected behavior and eliminates the false-positive rejection. Review covered proto/FFI surface, correctness, concurrency, documentation, and Slurm compatibility dimensions. No proto or concurrency issues were found.

| Severity | Count | Summary |
|---|---|---|
| **Major (blocking)** | 1 | Unbounded segment count enables CPU-amplification DoS |
| Minor | 2 | Inconsistent sort variant; regression test checks only length |
| Nit | 1 | Test comment narrates the bug fix rather than the invariant |

**Blocking — CPU exhaustion via unbounded segment count**

The old cumulative guard (`task_ids.len() + count > MAX_ARRAY_SIZE`) also acted as a fast-exit: the second occurrence of `0-99999` would be rejected immediately. The new guard (`count > MAX_ARRAY_SIZE`) only rejects a *single* range that alone exceeds the cap. A spec like `0-99999` repeated N times passes the per-range check (100,000 is not strictly greater than 100,000), appends 100,000 elements, triggers compaction (`sort` + `dedup` — O(200,000 × log 200,000)), and repeats. Peak heap is bounded at ~800 KB, but CPU cost is O(N × MAX_ARRAY_SIZE × log MAX_ARRAY_SIZE) with N bounded only by input string length. The fix needs either a segment-count cap or a fast pre-dedup cumulative limit before the compaction work is done.

SumonAMD:
All four review items are addressed in f3b6b4f (branch rebased onto current main).

**Heads up on the threads:** the force-push re-anchored the inline comments
by line offset instead of marking them outdated, and three now sit on unrelated
lines with live "Commit suggestion" buttons — applying them in place would splice
code into `max_size_boundary_unchanged` and the single-id push. Mapping below.

| Review item | Status | Where it landed |
|---|---|---|
| **Major** — CPU-amplification DoS | fixed | `MAX_ARRAY_SPEC_IDS` + `add_named_ids` |
| `sort_unstable` inconsistency | fixed | one sort site left, below the parse loop |
| Test asserts length only | fixed | `overlapping_ranges_dedup_within_cap` |
| Comment narrates the bug | fixed | same test's comment |

**On the blocking item — the suggested patch doesn't close it.** I applied it
verbatim and re-measured: `0-99999` x256 still took 862ms, still linear in segment
count. `task_ids.len() + count > 2 * MAX` never fires, because the in-loop
compaction resets `len` to 100,000 each iteration, so the sum is exactly 200,000
forever and `>` is false.

The amplification is also cheaper than described: appending `,1` after a full range
is 2 bytes and lands inside the materialized set, forcing a 100k-element sort that
collapses straight back. A 32 KiB spec cost 14.7s in-process, and **11.76s of real
`spurctld` time** on the submit path — now **0.034s**, returning the same error.

**Fix:** bound the cumulative ids a spec names *before* dedup (2x the cap) in a
counter that never resets, and charge single ids too. Total appends are then capped,
so the in-loop compaction is gone entirely — ids are collected once and sorted once,
which is also what resolves the `sort_unstable` inconsistency.

**Tradeoff:** a spec naming >200,000 indices is now rejected however far it would
collapse (`0-99999` x3). Two overlapping copies — the case this PR exists to fix —
still work, and `main` already rejects the three-copy spec, so this narrows only the
previous revision of this PR, not shipped behavior. The new bound accepts a strict
superset of `main`, so nothing currently valid starts failing. Documented in the Job
Arrays section of the user guide.

Added `ranges_past_the_pre_dedup_cap_are_rejected` and
`single_ids_count_toward_the_pre_dedup_cap`; both are fast against the pre-fix code
too, so a regression fails cleanly rather than hanging the suite. `fmt` and `clippy`
clean; 3,685 tests pass across 23 suites.

I've also updated the PR description.

crates/spur-core/src/array.rs
SumonAMD: Fixed in f3b6b4f with a cumulative pre-dedup counter that caps how many ids a spec may
name at `2 * MAX_ARRAY_SIZE`, which also made the in-loop compaction unnecessary. Worth
flagging that the suggested 2x guard doesn't work as expected: compaction resets
`task_ids.len()` to 100,000 every iteration, so the check is exactly `200,000 > 200,000`
and never fires — applied verbatim, `0-99999` repeated 256 times still took 862ms. On a
live controller a 32 KiB `--array` now goes from 11.763s to 0.034s, same error.

crates/spur-core/src/array.rs
SumonAMD: Fixed in f3b6b4f. The in-loop compaction is gone entirely, so there is a single sort left
and it is `sort_unstable` + `dedup`.

crates/spur-core/src/array.rs
SumonAMD: Fixed in 0ed16a8. `overlapping_ranges_dedup_within_cap` now asserts exact contents with
`assert_eq!(spec.task_ids, (0u32..=99_999).collect::<Vec<_>>())`.

crates/spur-core/src/array.rs
SumonAMD: Fixed in 0ed16a8. Reworded to state the invariant: dedup runs before the size cap, so
overlapping ranges collapse into it and `0-99999,0-99999` is 100,000 distinct tasks.

PR #726, opened 2026-08-24
fix(spurctld)!: enforce reservation management privileges server-side

biluriuday:
## What this fixes

Reservation create/update/delete was enforced only by the CLI's local
root-or-`sudo`/`wheel` check — something no server can verify, so any client that
skipped the CLI walked through. #685 added the generic `require_admin`, but that
deliberately allows an unidentified caller so no-auth clusters keep working,
which under the default `permissive` mode leaves the path open. The old ownership
rule also read an empty `user` as a trusted internal caller and `user: "root"` as
authorized.

## Changes

- The rule now lives once in `spur_core::privilege`. The CLI keeps its local
  pre-check for the fast offline error; the controller resolves the caller's name
  through its own NSS (`getgrouplist`, so LDAP/sssd groups count) and applies the
  same rule.
- `require_reservation_manager` replaces `require_admin` on the three reservation
  RPCs: same admin bar, plus the `sudo`/`wheel` path, minus the free pass for
  unidentified callers. Runs after `authoritative_user`, on the leader. An
  unresolvable name denies. The lookup is on `spawn_blocking` and only for
  non-admins, keeping network-backed NSS off the runtime and off an
  unauthenticated caller's path.
- Ownership retires: under operator semantics it could never deny anyone who
  passed the gate. `owner` stays as attribution. Nothing persisted changed and
  the proto edits are comment-only, so a rolling upgrade is unaffected.
- Fixes an `adminlevel` trap found while documenting the admin bar: the docs
  listed `Administrator`, which Slurm prints and accepts, but `is_admin` matched
  `admin` alone — so following the docs stored a level that looked privileged and
  conferred nothing. Levels are canonicalized on write and recognized in every
  Slurm spelling on read.

## Trade-off

Without a credential the controller can only check the name the client asserted,
so this stops an honest client rather than a determined one; `auth.mode =
"required"` is what makes it a boundary. Documented rather than implied. A
`sudo`/`wheel` operator must now also be resolvable *on the controller*.

## Breaking

Refused now: unprivileged non-CLI clients, callers relying on an empty `user` or
an unowned reservation, and `sudo`/`wheel` operators the controller cannot
resolve. Newly allowed: an operator may manage another user's reservation
directly rather than via `sudo`, and a stored `Administrator` grants what it
always appeared to.

## Testing

`cargo fmt --check`, clippy, and the full suite pass (34 suites). The policy is a
pure function, so its arms are host-independent; a handler test covers a direct
RPC from an identified non-admin *and* an anonymous caller being refused without
reaching the state machine.

Verified on the 3-node bare-metal cluster. The server gate was proven
independently of the CLI by creating a user in `sudo` on node2 only and running
the CLI from there against the controller on node1 — the local check passed, the
controller refused with `cannot verify that 'resvtest' may manage reservations
(no such user on this host)`. A user resolvable but unprivileged on the
controller hit the other arm. `nobody` was refused by the CLI; root created a
reservation; a non-owner in `sudo` updated and deleted root's reservation
(previously denied); job scheduling unaffected.

## Follow-ups, not fixed here

[1] The accounting mutation gate also allows anonymous, so on a `permissive`
cluster with a `jwt_key` a user can self-promote via a tokenless `add_user` and
then present their real token as admin for every gate #685 added.

[2] The controller's agent credential is a replayable `is_admin` token verified
with the same key and no audience check, and the Raft port has no auth layer
while `install_snapshot` replaces the reservation set wholesale. Both defeat this
gate in every mode.

[3] Four overlapping privilege predicates now exist (`caller_is_admin`,
`caller_is_privileged`, `require_admin`, `require_reservation_manager`).
Collapsing them into one principal, with the anonymous allowance driven by
`auth.mode` rather than by RPC, is the right shape but changes behaviour for the
ten RPCs #685 gated.

### DIFF
--- a/crates/spurctld/src/accounting/mod.rs
+++ b/crates/spurctld/src/accounting/mod.rs
@@ -74,14 +74,16 @@ pub fn admin_level_is_admin(raw: &str) -> bool {
 }
 
 /// Fold one user-row's admin_level into the per-user map, keeping the highest: admin wins over
-/// any lower level so a later non-admin row for a multi-account user can't clobber it.
+/// any lower level so a later non-admin row for a multi-account user can't clobber it. Stored
+/// canonical, so rows predating the column's normalization cannot leave several spellings of one
+/// level in the cache; a value that is no level at all is kept verbatim, to stay visible.
 fn merge_admin_level(map: &mut HashMap<String, String>, user: &str, level: &str) {
     if level.is_empty() || level.eq_ignore_ascii_case("none") {
         return;
     }
     let entry = map.entry(user.to_owned()).or_default();
     if entry.is_empty() || admin_level_is_admin(level) {
-        *entry = level.to_owned();
+        *entry = canonical_admin_level(level).unwrap_or(level).to_owned();
     }
 }
 
@@ -182,13 +184,32 @@ mod tests {
         let mut m = HashMap::new();
         merge_admin_level(&mut m, "carol", "Admin");
         merge_admin_level(&mut m, "carol", "Operator");
-        assert_eq!(m.get("carol").map(String::as_str), Some("Admin"));
+        assert_eq!(m.get("carol").map(String::as_str), Some("Administrator"));
 
         // Operator then Admin: Admin must still win.
         let mut m = HashMap::new();
         merge_admin_level(&mut m, "carol", "Operator");
         merge_admin_level(&mut m, "carol", "Admin");
-        assert_eq!(m.get("carol").map(String::as_str), Some("Admin"));
+        assert_eq!(m.get("carol").map(String::as_str), Some("Administrator"));
+    }
+
+    /// Rows written before the column was normalized must not leave several spellings of one level
+    /// in the cache.
+    #[test]
+    fn legacy_spellings_fold_to_one_canonical_level() {
+        for raw in ["admin", "Admin", "Administrator", "SuperUser"] {
+            let mut m = HashMap::new();
+            merge_admin_level(&mut m, "carol", raw);
+            assert_eq!(
+                m.get("carol").map(String::as_str),
+                Some("Administrator"),
+                "raw {raw:?}"
+            );
+        }
+
+        let mut m = HashMap::new();
+        merge_admin_level(&mut m, "dave", "operator");
+        assert_eq!(m.get("dave").map(String::as_str), Some("Operator"));
     }
 
     #[test]
--- a/crates/spurctld/src/accounting/mod.rs
+++ b/crates/spurctld/src/accounting/mod.rs
@@ -54,14 +54,33 @@ pub async fn fairshare_factors(
     ))
 }
 
-/// Fold one user-row's admin_level into the per-user map, keeping the highest: `Admin` wins over
+/// Canonicalize an `adminlevel` to Slurm's spelling, or `None` if it is not a level.
+///
+/// Slurm prints `Administrator` for the highest level and its parser also takes `Admin` and
+/// `SuperUser`, so all three must resolve to the same thing — recognising only one spelling would
+/// leave a stored level that looks like a privilege and confers nothing.
+pub fn canonical_admin_level(raw: &str) -> Option<&'static str> {
+    match raw.to_ascii_lowercase().as_str() {
+        "none" => Some("None"),
+        "operator" => Some("Operator"),
+        "admin" | "administrator" | "superuser" => Some("Administrator"),
+        _ => None,
+    }
+}
+
+/// Whether a stored `admin_level` is the level that confers control-plane privilege.
+pub fn admin_level_is_admin(raw: &str) -> bool {
+    canonical_admin_level(raw) == Some("Administrator")
+}
+
+/// Fold one user-row's admin_level into the per-user map, keeping the highest: admin wins over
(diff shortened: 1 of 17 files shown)


crates/spurctld/src/server.rs
```diff
@@ -7198,27 +7241,116 @@ mod tests {
...
+    /// still checkable, and a client skipping the CLI is the gap being closed. Host-independent —
+    /// `mallory` and the empty user are unprivileged or unresolvable, and both deny.
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn reservation_mutations_deny_non_operators() {
+        let dir = tempfile::TempDir::new().unwrap();
+        let svc = test_service(&dir).await;
+
+        macro_rules! assert_operator_gated {
+            ($method:ident, $req:expr) => {{
+                for identity in [Some(viewer("mallory", false)), None] {
```
shiv-tyagi: The identified non-admin arm calls real NSS for `"mallory"`. The comment says this is host-independent, but if `mallory` exists on the runner and is in `sudo`/`wheel`, the test would allow the mutation and fail. Use a guaranteed-unresolvable name here, same pattern as `admin_manages_reservations_without_a_resolvable_name`.

```suggestion
                for identity in [Some(viewer("no_such_user_in_nss_9b2e", false)), None] {
```

tests/native_host/e2e/test_reservations.py
```diff
@@ -486,13 +486,14 @@ def assert_never_created(name, window=3.0, step=0.5):
...
 
-    def test_non_owner_cannot_delete_or_update_reservation(self, cluster):
-        """A reservation is owned by its creator; a different user must not be
-        able to delete or update it, but the owner still can. Exercises the
-        full CLI -> gRPC -> controller ownership check (SPUR-69)."""
+    def test_reservation_management_is_privilege_not_ownership(self, cluster):
+        """Managing a reservation takes operator privilege, not ownership: an
+        operator may update or delete a reservation somebody else created, while
+        an unprivileged user is refused even for one it could see. Exercises the
+        full CLI -> gRPC -> controller path (SPUR-69)."""
```
shiv-tyagi: Nit: Drop the `(SPUR-69)` ticket reference from the docstring. Git history tracks that linkage.

```suggestion
        full CLI -> gRPC -> controller path."""
```

crates/spurctld/src/accounting/mod.rs
```diff
@@ -54,14 +54,33 @@ pub async fn fairshare_factors(
...
 /// any lower level so a later non-admin row for a multi-account user can't clobber it.
 fn merge_admin_level(map: &mut HashMap<String, String>, user: &str, level: &str) {
     if level.is_empty() || level.eq_ignore_ascii_case("none") {
         return;
     }
     let entry = map.entry(user.to_owned()).or_default();
-    if entry.is_empty() || level.eq_ignore_ascii_case("admin") {
+    if entry.is_empty() || admin_level_is_admin(level) {
         *entry = level.to_owned();
     }
```
shiv-tyagi, line 87: `add_user` now canonicalizes `adminlevel` to Slurm spellings on write, but `merge_admin_level` still stores the raw DB string (`"admin"`, `"Admin"`, etc.). `is_admin` works because reads canonicalize, but the association cache can keep mixed spellings until rows are rewritten. Consider storing the canonical form when folding admin rows:

```suggestion
    if entry.is_empty() || admin_level_is_admin(level) {
        *entry = canonical_admin_level(level)
            .unwrap_or(level)
            .to_owned();
    }
```

tests/native_host/e2e/test_reservations.py
```diff
@@ -486,13 +486,14 @@ def assert_never_created(name, window=3.0, step=0.5):
         finally:
             cluster.cli_as_user("root", ["scontrol", "delete-reservation", omit_name])
 
-    def test_non_owner_cannot_delete_or_update_reservation(self, cluster):
-        """A reservation is owned by its creator; a different user must not be
-        able to delete or update it, but the owner still can. Exercises the
-        full CLI -> gRPC -> controller ownership check (SPUR-69)."""
+    def test_reservation_management_is_privilege_not_ownership(self, cluster):
```
shiv-tyagi, line 489: The PR's strongest new behavior (CLI local check passes, controller denies because the user is not resolvable on the controller host) is verified manually on bare-metal but not covered in e2e. A small test that creates a user on one node only and asserts the controller denial would lock that regression in CI.

shiv-tyagi (commented):
Posted commentes. PTAL.

tests/native_host/e2e/test_reservations.py
biluriuday: added a e2e test to verify this check

crates/spurctld/src/accounting/mod.rs
biluriuday: done

tests/native_host/e2e/test_reservations.py
biluriuday: done

crates/spurctld/src/server.rs
biluriuday: done

shiv-tyagi (approved):
LGTM

PR #727, opened 2026-08-24
fix(spur-cli)!: reject a zero token TTL and report overflow in seconds (follow-up to #716)

### LINKED ISSUE #826: [Issue]: A zero token TTL (--ttl 0d) creates a never-expiring token instead of being rejected
## Problem Description

Passing a zero duration to `--ttl` (e.g. `0`, `0d`, `0h`, `0s`) parses to a TTL of `0` seconds. The server treats a `0` TTL as "no expiry", so the resulting token never expires — the opposite of what a caller asking for a zero lifetime intends. This affects both `token create` and `token user`, which share the parser.

Separately, the TTL-overflow error mixes units: it reports a days input against a seconds limit (e.g. `TTL 50000d exceeds 4294967295 seconds`), which is confusing.

**Expected:** a non-positive TTL is rejected with a clear error. Never-expiry remains available only the documented way — by omitting `--ttl`. The overflow message states the actual second count so input and limit are in the same unit.

## Environment

- **OS:** Ubuntu 22.04 LTS
- **Component:** spur

## Steps to Reproduce

```bash
spur token create --ttl 0d   # yields a token that never expires

SumonAMD:
## Summary

Closes #826

Follow-up to #716, addressing the two token nits from @sajmera-pensando.

## Changes

- **Zero-TTL guard.** `parse_ttl("0d")` returned `0`, and the server's
  `ttl_secs.filter(|&v| v > 0)` collapses `0` to `None` — a **never-expiring**
  token, the opposite of intent. (`token user` signs a JWT, where zero instead
  yields an already-expired credential.) `parse_ttl` now rejects any
  non-positive duration, so both `token create` and `token user` fail clearly.
  Never-expiry stays available the documented way: omit `--ttl`.
- **Overflow message in seconds.** It read `TTL 50000d exceeds 4294967295
  seconds`, mixing a days input with a seconds limit. The count is parsed at
  `u64` and multiplied at `u128`, so every accepted count reports its own
  second total: `TTL 50000d is 4320000000 seconds, over the 4294967295 second
  maximum`.
- **Docs.** The `--ttl` paragraph in the accounting guide notes that a zero
  duration is rejected rather than read as "no expiry".

## Review follow-up

Parsing the count at `u64` is right, but multiplying at `u64` is not:
`94368760191893771 * 86_400` wraps to `128`, which a release build accepts as a
128-second TTL and a debug build panics on. Multiplying at `u128` cannot
overflow for any count the parse accepts, so the message is exact for every
input.

## Tests

- `parse_ttl_rejects_zero` — `0`, `0d`, `0h`, `0m`, `0s`.
- `parse_ttl_rejects_values_that_overflow` now asserts the message, not just
  `is_err()`.
- `parse_ttl_reports_seconds_for_counts_past_u32_and_u64` — the two counts no
  narrower arithmetic can report; it fails against both the old `u32` parse and
  a plain `u64` multiply.
- `cargo fmt` / `clippy -D warnings` clean, full suite green, docs gate
  (sphinx `-W`, pyspelling, markdownlint) clean.

### DIFF
--- a/crates/spur-cli/src/token.rs
+++ b/crates/spur-cli/src/token.rs
@@ -3,7 +3,7 @@
 
 //! `spur token` subcommands for admission token management.
 
-use anyhow::{Context, Result};
+use anyhow::Result;
 use clap::{Parser, Subcommand};
 
 use spur_proto::proto::{CreateTokenRequest, ListTokensRequest, RevokeTokenRequest};
@@ -118,9 +118,9 @@ fn cmd_user_token(user: &str, admin: bool, ttl: Option<String>, config_path: &st
 fn parse_ttl(s: &str) -> Result<u32> {
     let s = s.trim();
     let (value, unit_secs) = if let Some(days) = s.strip_suffix('d') {
-        (days, 86400)
+        (days, 86_400u64)
     } else if let Some(hours) = s.strip_suffix('h') {
-        (hours, 3600)
+        (hours, 3_600)
     } else if let Some(mins) = s.strip_suffix('m') {
         (mins, 60)
     } else if let Some(secs) = s.strip_suffix('s') {
@@ -128,10 +128,18 @@ fn parse_ttl(s: &str) -> Result<u32> {
     } else {
         (s, 1)
     };
-    value
-        .parse::<u32>()?
-        .checked_mul(unit_secs)
-        .with_context(|| format!("TTL {} exceeds {} seconds", s, u32::MAX))
+    // u64 can't overflow here (max is u32::MAX * 86_400), so the message reports seconds vs seconds.
+    let secs = value.parse::<u32>()? as u64 * unit_secs;
+    if secs == 0 {
+        anyhow::bail!("TTL {s} must be a positive duration");
+    }
+    if secs > u32::MAX as u64 {
+        anyhow::bail!(
+            "TTL {s} is {secs} seconds, over the {} second maximum",
+            u32::MAX
+        );
+    }
+    Ok(secs as u32)
 }
 
 async fn cmd_create(controller: &str, ttl: Option<String>) -> Result<()> {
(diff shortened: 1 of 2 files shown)


sajmera-pensando (changes requested):
## Review: fix(spur-cli)!: reject a zero token TTL and report overflow in seconds

**Verdict: REQUEST CHANGES**

This PR addresses two real user-facing bugs — silent never-expiry on a zero TTL, and a confusing overflow message that reported seconds vs. seconds — and the logic for both is fundamentally sound. The breaking change is correctly signaled with `!` in the title and the PR description explains the motivation clearly. Good work on the overall approach.

Two issues are worth fixing before merge:

**Test coverage gap (🟡 Minor).** The `parse_ttl_rejects_zero` test exercises `"0"`, `"0d"`, `"0h"`, and `"0s"` but omits `"0m"`. The minutes branch is a separate code path; a future refactor that accidentally assigns `unit_secs = 0` in that arm would silently regress to never-expiry without a failing test.

**`u32` parse truncates the improved overflow message (🟡 Minor).** The numeric portion is parsed as `u32` before widening to `u64`, so any count that exceeds `u32::MAX` on its own (e.g., `5000000000d`) hits the generic integer-parse error path instead of the new contextual message. Parsing directly as `u64` closes the gap without introducing any new overflow risk.

---

**Review legend**
- 🟡 **[Minor]** — should fix, not a hard blocker
- 🔵 **[Nit]** — style/polish, optional

SumonAMD:
Both addressed.

1. 0m added to parse_ttl_rejects_zero, and to the list of rejected forms in the docs, which had the same gap.

2. Parsing at u64 is right, but multiplying at u64 is not: 94368760191893771 * 86_400 wraps to 128, so a release build accepts it as a 128-second TTL and a debug build panics. The count is now parsed at u64 and multiplied at u128, so 5000000000d and the wrap case both report their exact second count.

PR #728, opened 2026-08-24
fix(spurctld): apply MaxWall as the default time limit for a job with no -t

### LINKED ISSUE #692: MaxWall is not applied as the default time limit for jobs submitted without -t
## Problem

In Slurm, `MaxWallDurationPerJob` (association `MaxWall`, QOS `MaxWall`) doubles as the default time limit a job inherits when it is submitted without an explicit `-t`. A time-less job is therefore effectively capped at MaxWall.

Spur does not apply MaxWall as a default. A job submitted without `-t` under a positive `maxwall` cap does not inherit that cap as its time limit; it runs with the partition default (or unbounded), so the association/QOS wall cap has no effect on time-less jobs when the cap is greater than zero.

## Expected

A job submitted without `-t` under `maxwall=M` (M > 0) is bounded by M, matching Slurm, honoring association/QOS precedence.

## Actual

MaxWall only rejects jobs whose explicit requested time exceeds it. The special case `maxwall=0` is handled (it blocks time-less jobs too, since a zero cap is unsatisfiable), but a positive cap does not become the default time limit for time-less jobs.

## Repro (conceptual)

1. Set `maxwall=30` on a user's association (or QOS).
2. Submit a job without `-t`.
3. Observe the job is not capped at 30 minutes; it takes the partition default rather than inheriting MaxWall.

## Notes

Surfaced in review of #670. The `maxwall=0` block-all behavior shipped there is an interim, outcome-aligned fix; applying MaxWall as a default time limit is the broader scheduling-behavior change needed for full parity and was deferred out of that PR.

nikhilsk:
## What this fixes

A job submitted without `-t` carried no time limit at all. It reported `UNLIMITED`, and the running-job watchdog acts on a job's own limit, so it had no deadline to enforce — a positive `maxwall` only ever held for jobs that named a limit themselves. Operators saw the split first-hand: jobs that passed `-t 12:00:00` were killed on time while jobs in the same QOS ran well past the cap.

Fixes #692.

## Approach

Wall-time defaulting now runs after QOS defaulting, so the governing QOS is known, and takes the job's per-job wall ceiling as the default when the partition chain supplies nothing. Precedence is unchanged where it already produced an answer: a partition `DefaultTime` (or the `default_time_limit_minutes` chain) still wins, because `maxwall` is a ceiling and only fills a gap.

Two clamps keep a filled-in limit from being worse than the unbounded job it replaces:

- The **account `maxwall`** rejects a submission outright, so defaulting above it would turn a job that runs today into one that cannot be submitted. The QOS and association caps both apply to a job, so the smaller is what it inherits.
- A requested partition's **`MaxTime`** blocks a job above it, so a default above `MaxTime` would pend forever on `PartitionTimeLimit` for a limit the user never asked for. The ceiling used is the most permissive requested partition's, since the job only has to be schedulable somewhere.

A `maxwall` of `0` is the block-all sentinel rather than a zero-length budget, so it is never used as a default — the existing breach checks still handle it.

## Design notes

- **The limit is the job's own from then on.** It is written into the spec at submit, so `squeue` and `scontrol show job` report a real value and a later QOS change does not move it. That is also why nothing here touches jobs that are already running: they keep the unbounded limit they were submitted with, so upgrading cannot mass-kill live work. Bounding an existing job stays an explicit `scontrol update job <id> TimeLimit=` action.
- **Unread limits default nothing.** `wall_caps` reports no ceiling while the QOS or association cache has not loaded, matching how `enforce_submit_limits` treats limits it cannot read yet. A cold controller does not invent deadlines.
- Moving the defaulting call after `apply_default_qos` is safe: nothing between the old and new call sites reads `time_limit`. The submit-time partition check stays keyed on `user_requested_time`, so an auto-filled value is still exempt from `EnforcePartLimits` rejection, as before.

## Tests

- Twelve unit tests around `apply_default_time_limit` covering the QOS-supplied default, an account-supplied one, the smaller of the two, partition `DefaultTime` still winning, both clamps, the multi-partition ceiling, an unlimited partition lifting it, and the zero sentinel on either cap.
- One test through the real `submit_job` path with a seeded QOS cache, asserting the persisted spec carries the limit.
- One test that `wall_caps` reports nothing before the caches load.
- An end-to-end test (`tests/native_host/e2e/test_accounting.py::TestQosWallCapDefaulting`) for the incident path: a `-t`-less job under a QOS with `maxwall` set, on an unbounded partition so defaulting reaches the wall cap. It asserts both halves — the job inherits the `MaxWall` as its limit (`squeue %l` is `1:00`, not `UNLIMITED`) and the watchdog actually kills it at the cap (`JobState=TIMEOUT`, `Reason=TimeLimit`) rather than running out its sleep.
- The eleven pre-existing defaulting tests are unchanged in behavior and still pass, pinning that the partition chain did not move.
- `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` clean; full `cargo test --locked` green.

## Docs

New "MaxWall as the default time limit" section in the accounting admin guide with the precedence list and the clamp reasoning, cross-referenced from both `maxwall` key tables and from `default_time_limit_minutes` in the configuration guide.

### DIFF
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -583,11 +583,6 @@ impl ClusterManager {
             }
         }
 
-        apply_default_time_limit(
-            &mut spec,
-            &partitions,
-            config.scheduler.default_time_limit_minutes,
-        );
         apply_default_account(&mut spec, &self.association_cache);
         validate_user_account(&spec, &self.association_cache, &config.accounting)?;
         // Default QoS must resolve before the partition ACL, or `allow_qos` sees
@@ -598,6 +593,14 @@ impl ClusterManager {
             &self.qos_cache,
             &config.accounting,
         )?;
+        // Wall-time defaulting reads the governing QOS, so it follows QOS defaulting.
+        let wall_caps = self.wall_caps(&spec);
+        apply_default_time_limit(
+            &mut spec,
+            &partitions,
+            config.scheduler.default_time_limit_minutes,
+            wall_caps,
+        );
         self.validate_partition(&spec, &partitions)?;
 
         // Fewer tasks than nodes cannot use the surplus nodes; cap at the task
@@ -4664,6 +4667,33 @@ impl ClusterManager {
     /// limits and the association per-job wall deny unconditionally; stand-alone
     /// TRES/wall breaches deny only under the governing QOS's `DenyOnLimit`
     /// (treated as on when the job has no QOS). `incoming` is the task count added.
+    /// The wall-clock ceilings that already govern this submission, for defaulting.
+    /// An unloaded cache reports nothing rather than holding the job, matching how
+    /// `enforce_submit_limits` treats limits it cannot read yet.
+    fn wall_caps(&self, spec: &JobSpec) -> WallCaps {
+        let qos_minutes = spec
+            .qos
+            .as_deref()
+            .filter(|n| !n.is_empty() && self.qos_cache.is_loaded())
+            .and_then(|name| self.qos_cache.get(name))
+            .and_then(|qos| qos.limits.max_wall_minutes);
+
+        let account_minutes = spec
+            .account
+            .as_deref()
+            .filter(|a| !a.is_empty() && self.association_cache.is_loaded())
+            .and_then(|account| {
+                self.association_cache
+                    .limits(&spec.user, account)
+                    .max_wall_minutes
+            });
+
+        WallCaps {
+            qos_minutes,
+            account_minutes,
+        }
+    }
+
     fn enforce_submit_limits(&self, spec: &JobSpec, incoming: u32) -> Result<(), SubmitError> {
         let jobs = self.jobs.read();
         let user = spec.user.as_str();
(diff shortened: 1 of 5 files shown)


sajmera-pensando (changes requested):
The fix is correct for the main bug, but there is a remaining behavioral gap compared to Slurm — see the inline comment. When a job has no `--time` and the partition supplies no `DefaultTime` or `MaxTime`, Slurm rejects the submission outright. Spur admits it as UNLIMITED even after this PR. The fix should reject in that case so `MaxWall` enforcement cannot be bypassed by targeting a partition with no time defaults.

crates/spurctld/src/cluster.rs
```diff
@@ -4664,6 +4670,33 @@ impl ClusterManager {
     /// limits and the association per-job wall deny unconditionally; stand-alone
     /// TRES/wall breaches deny only under the governing QOS's `DenyOnLimit`
     /// (treated as on when the job has no QOS). `incoming` is the task count added.
+    /// The wall-clock ceilings that already govern this submission, for defaulting.
+    /// An unloaded cache reports nothing rather than holding the job, matching how
+    /// `enforce_submit_limits` treats limits it cannot read yet.
+    fn wall_caps(&self, spec: &JobSpec) -> WallCaps {
```
shiv-tyagi: `wall_caps` was inserted directly above `enforce_submit_limits` with no blank line between them. Rustdoc attaches the preceding comment block to the next item, so this whole 7-line block (the original `enforce_submit_limits` doc plus the 3 new lines) now documents `wall_caps`, and `enforce_submit_limits` loses its doc comment entirely.

```suggestion
    /// The wall-clock ceilings that already govern this submission, for defaulting.
    /// An unloaded cache reports nothing rather than holding the job, matching how
    /// `enforce_submit_limits` treats limits it cannot read yet.
    fn wall_caps(&self, spec: &JobSpec) -> WallCaps {
```

and restore the original doc comment directly above `enforce_submit_limits`:

```rust
    /// Submit-time limit gate (Slurm's `acct_policy_validate`). Submit-count
    /// limits and the association per-job wall deny unconditionally; stand-alone
    /// TRES/wall breaches deny only under the governing QOS's `DenyOnLimit`
    /// (treated as on when the job has no QOS). `incoming` is the task count added.
    fn enforce_submit_limits(&self, spec: &JobSpec, incoming: u32) -> Result<(), SubmitError> {
```

crates/spurctld/src/cluster.rs
```diff
@@ -7323,6 +7386,29 @@ fn apply_default_time_limit(
...
+    ]
+    .into_iter()
+    .flatten()
+    .min()?;
+
+    // The limit has to stay schedulable somewhere: a requested partition with no
+    // `MaxTime` takes any, otherwise the most permissive one is the ceiling.
+    let partition_ceiling = requested
+        .iter()
+        .try_fold(0, |acc: u32, p| p.max_time_minutes.map(|m| acc.max(m)));
```
shiv-tyagi: Nit: type annotation belongs on the initial value, not the closure param.
```suggestion
        .try_fold(0u32, |acc, p| p.max_time_minutes.map(|m| acc.max(m)));
```

crates/spurctld/src/cluster.rs
```diff
@@ -7323,6 +7386,29 @@ fn apply_default_time_limit(
     }
 }
 
+/// The per-job wall ceiling as a default. A job has to satisfy the QOS's and the
+/// account's `MaxWall` both, so the smaller of the two is the limit it is really
+/// held to. That value is then kept under the requested partitions' `MaxTime`,
+/// above which the job would pend forever on a limit it never asked for — where
+/// today it simply runs unbounded.
+fn wall_cap_default(requested: &[&Partition], caps: WallCaps) -> Option<u32> {
```
shiv-tyagi, line 7394: Q: `try_fold(0, ...)` over an empty `requested` returns `Some(0)`, which would clamp the result to `0`. The only caller already guards against an empty `requested` before reaching here, so this can't happen today, but that invariant isn't visible from this function's own signature. Worth a one-line comment (or a debug_assert) so a future caller doesn't hit it silently?

shiv-tyagi (commented):
Posted review, PTAL

crates/spurctld/src/cluster.rs
nikhilsk: Good catch, and correct — that block had been attached to `wall_caps` since the first commit, so `enforce_submit_limits` was undocumented for the whole review. Split as suggested, with its original doc restored directly above it.

crates/spurctld/src/cluster.rs
nikhilsk: Moot now — the fold is gone entirely as part of the empty-slice fix in the sibling thread, so there is no annotation left to place. Thanks for flagging it.

crates/spurctld/src/cluster.rs
nikhilsk: Took the structural route rather than a `debug_assert`. The consequence of that `Some(0)` is a job admitted with a zero-length time limit, which the watchdog then kills seconds after it starts — the same shape as the bug in #759 — so I would rather it not rest on a caller's guard at all.

The ceiling is now "a partition with no `MaxTime` accepts any limit, otherwise the most permissive `MaxTime`":

```rust
let partition_ceiling = if requested.iter().any(|p| p.max_time_minutes.is_none()) {
    None
} else {
    requested.iter().filter_map(|p| p.max_time_minutes).max()
};
```

Identical for every reachable input, and an empty slice simply yields no ceiling, so the wall cap stands. Pinned by `wall_cap_default_without_partitions_keeps_the_cap`.

crates/spurctld/src/cluster.rs
```diff
@@ -598,6 +593,14 @@ impl ClusterManager {
             &self.qos_cache,
             &config.accounting,
         )?;
+        // Wall-time defaulting reads the governing QOS, so it follows QOS defaulting.
+        let wall_caps = self.wall_caps(&spec);
```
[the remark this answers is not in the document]
nikhilsk: Fixed — the ordering is inverted from what you read. `wall_caps` is now resolved *after* `run_job_submit_hook` (`cluster.rs:652`, then `:657`), so defaulting sees the QOS/account the hook actually left on the spec rather than the pre-hook scope. A hook that assigns a QOS with a positive `MaxWall` to an untimed job therefore has that cap applied, instead of reaching `enforce_submit_limits` with `time_limit == None`. A limit the hook set itself is still left alone, since defaulting is a no-op when the spec already carries one.

Pinned by `submit_job_hook_set_qos_supplies_the_wall_default` and `submit_job_hook_set_partition_supplies_the_wall_default`.

shiv-tyagi (approved):
LGTM

PR #729, opened 2026-08-24
fix(spur-cli): parse --mem units (K/M/G/T) consistently across sbatch/srun/salloc

### LINKED ISSUE #827: [Issue]: --mem unit suffixes are inconsistent across sbatch/srun/salloc (K only in sbatch, T nowhere)
## Problem Description

The `--mem` flag accepts different unit suffixes depending on which submission command is used. `sbatch --mem=1024K` is accepted, but the same flag under `srun` and `salloc` fails with `invalid memory value`. None of the three accept a `T` (terabyte) suffix.

A script that works under `sbatch` breaks when the identical memory flag is passed to `srun` or `salloc` — a silent Slurm incompatibility.

**Expected:** all three commands accept the same `[K|M|G|T]` suffix set (case-insensitive, bare number = MiB), matching Slurm, whose `sbatch`/`srun`/`salloc` document `--mem` identically and share one parser.

## Environment

- **OS:** Ubuntu 22.04 LTS
- **Component:** spur

## Steps to Reproduce

```bash
sbatch --mem=1024K --wrap 'true'   # accepted
srun   --mem=1024K true            # invalid memory value
salloc --mem=1024K                 # invalid memory value
sbatch --mem=1T --wrap 'true'      # rejected everywhere

### PR DESCRIPTION
**Pull request by SumonAMD**
## Summary

Closes #827

`--mem` accepted different unit suffixes depending on the submission command.
`sbatch --mem=1024K` worked, the same flag under `srun` and `salloc` failed with
`invalid memory value`, and none of the three accepted `T`. A script that worked
under `sbatch` broke when the identical flag was passed to `srun` or `salloc`.

## Root cause

Each command carried its own private copy of `parse_memory_mb`. The copies
drifted: the `K` branch was dropped when the helper was pasted into `srun` and
`salloc`, and `T` was never added anywhere. Nothing in the code, tests, or docs
justified the difference.

## Fix

Rather than patch the missing branches into two more copies, all three are
removed and every command routes through one shared parser in `spur-core`, so
the accepted units are identical by construction. The shared parser adds `T`,
rejects empty / non-numeric / negative input, and rounds a size **up** to a
whole MiB.

Rounding up is load-bearing: a zero reaches the controller as "no limit", so
truncation made `sbatch --mem=512K` silently hand the job the whole node. Every
multiplier is a power of two, so integer inputs convert exactly and only a
genuinely fractional request is ever rounded. A bare `0` keeps Slurm's "all the
node's memory" meaning.

This does change `sbatch --mem=512K` from *unlimited* to *1 MiB*. That was the
bug, and `srun`/`salloc` rejected the value outright, so no working script
relied on it.

## Docs

`submitting-jobs.rst` documents the `K/M/G/T` set, that all three commands take
it, the round-up behaviour, that `--mem-per-cpu` is `sbatch`-only, and that a
fractional size such as `1.5G` is a Spur extension (Slurm takes whole numbers).

## Testing

- Unit tests in `spur-core`: K/M/G/T in both cases, the sub-MiB boundaries
  (`512K`, `1023K`, `2048K`), fractional sizes, whitespace trimming, and
  rejection of empty / garbage / negative values.
- `cargo fmt --check`, `clippy --workspace --exclude spur-ffi --all-targets
  --locked` at `-D warnings`, and the full suite are clean; the containerized
  docs gate (sphinx `-W`, pyspelling, markdownlint) passes.
- End-to-end against a live controller: `sbatch`, `srun`, and `salloc` all
  accept `--mem=1024K` and `--mem=1T`; invalid values are rejected by all three.
- Rebased onto `main`; the only conflict was the `salloc.rs` import block, which
  #911 had rewritten.

### DIFF
--- a/crates/spur-core/src/resource.rs
+++ b/crates/spur-core/src/resource.rs
@@ -298,6 +298,30 @@ impl ResourceSet {
 /// The input is trimmed and empty strings are rejected, so callers that split a
 /// comma-list (CLI `--gres`, `SLURM_GRES`, REST, FFI) can pass raw segments
 /// without worrying about incidental whitespace or trailing commas.
+/// Parse a Slurm-style `--mem` size into MiB. Optional K/M/G/T suffix
+/// (case-insensitive), no suffix means MiB. Shared by sbatch/srun/salloc.
+pub fn parse_memory_mb(s: &str) -> anyhow::Result<u64> {
+    let s = s.trim();
+    if s.is_empty() {
+        anyhow::bail!("invalid memory value ''");
+    }
+    let (num, mult): (&str, f64) = match s.as_bytes()[s.len() - 1] {
+        b'K' | b'k' => (&s[..s.len() - 1], 1.0 / 1024.0),
+        b'M' | b'm' => (&s[..s.len() - 1], 1.0),
+        b'G' | b'g' => (&s[..s.len() - 1], 1024.0),
+        b'T' | b't' => (&s[..s.len() - 1], 1024.0 * 1024.0),
+        _ => (s, 1.0),
+    };
+    let val: f64 = num
+        .trim()
+        .parse()
+        .map_err(|_| anyhow::anyhow!("invalid memory value '{s}'"))?;
+    if !val.is_finite() || val < 0.0 {
+        anyhow::bail!("invalid memory value '{s}'");
+    }
+    Ok((val * mult) as u64)
+}
+
 pub fn parse_gres(gres: &str) -> Option<(String, Option<String>, u32)> {
     let gres = gres.trim();
     if gres.is_empty() {
@@ -807,5 +831,45 @@ mod accounting_tests {
         assert_eq!(inv.total_gpus(), 3);
         assert_eq!(inv.gpu_counts().get("mi300x"), Some(&2));
         assert_eq!(ResourceSet::default().total_gpus(), 0);
+    }
+}
+
+#[cfg(test)]
+mod parse_memory_tests {
+    use super::parse_memory_mb;
+
+    #[test]
+    fn default_unit_is_mib() {
+        assert_eq!(parse_memory_mb("4096").unwrap(), 4096);
+    }
+
+    #[test]
+    fn all_suffixes_supported_case_insensitive() {
+        assert_eq!(parse_memory_mb("1024K").unwrap(), 1);
+        assert_eq!(parse_memory_mb("1024k").unwrap(), 1);
+        assert_eq!(parse_memory_mb("4096M").unwrap(), 4096);
+        assert_eq!(parse_memory_mb("4096m").unwrap(), 4096);
+        assert_eq!(parse_memory_mb("4G").unwrap(), 4096);
+        assert_eq!(parse_memory_mb("4g").unwrap(), 4096);
+        assert_eq!(parse_memory_mb("1T").unwrap(), 1024 * 1024);
+        assert_eq!(parse_memory_mb("1t").unwrap(), 1024 * 1024);
+    }
+
+    #[test]
+    fn fractional_gigabytes() {
+        assert_eq!(parse_memory_mb("1.5G").unwrap(), 1536);
+    }
+
+    #[test]
+    fn whitespace_trimmed() {
+        assert_eq!(parse_memory_mb("  2G  ").unwrap(), 2048);
+    }
+
+    #[test]
+    fn rejects_garbage_and_empty() {
+        assert!(parse_memory_mb("").is_err());
+        assert!(parse_memory_mb("abc").is_err());
+        assert!(parse_memory_mb("G").is_err());
+        assert!(parse_memory_mb("-5G").is_err());
     }
 }
--- a/docs/user-guide/submitting-jobs.rst
+++ b/docs/user-guide/submitting-jobs.rst
@@ -138,10 +138,12 @@ patterns for ``--output`` and ``--error``, the token expands to the job ID.
      - CPUs per task. Default ``1``.
    * - ``--mem``
      -
-     - Memory per node. ``512G`` and ``4096M`` use suffixes; a bare ``4096`` means MB.
+     - Memory per node. Takes a ``K``, ``M``, ``G``, or ``T`` suffix
+       (case-insensitive), e.g. ``512G`` or ``4096M``; a bare ``4096`` means MB.
+       ``sbatch``, ``srun``, and ``salloc`` accept the same set.
    * - ``--mem-per-cpu``
      -
-     - Memory per allocated CPU instead of per node.
+     - Memory per allocated CPU instead of per node. Same suffixes as ``--mem``.
    * - ``--gres``
      -
      - Generic resources, e.g. ``gpu:4`` or ``gpu:mi300x:8``. See
--- a/crates/spur-cli/src/salloc.rs
+++ b/crates/spur-cli/src/salloc.rs
@@ -395,17 +395,7 @@ fn resolve_salloc_env(matches: &ArgMatches, args: &mut SallocArgs) {
     );
 }
 
-fn parse_memory_mb(s: &str) -> Result<u64> {
-    let s = s.trim();
-    if let Some(gb) = s.strip_suffix('G').or_else(|| s.strip_suffix('g')) {
-        let val: f64 = gb.parse().context("invalid memory value")?;
-        Ok((val * 1024.0) as u64)
-    } else if let Some(mb) = s.strip_suffix('M').or_else(|| s.strip_suffix('m')) {
-        Ok(mb.parse().context("invalid memory value")?)
-    } else {
-        Ok(s.parse().context("invalid memory value")?)
-    }
-}
+use spur_core::resource::parse_memory_mb;
 
 #[cfg(test)]
 mod tests {
--- a/crates/spur-cli/src/salloc.rs
+++ b/crates/spur-cli/src/salloc.rs
@@ -4,6 +4,7 @@
 use crate::env_defaults::{apply_csv, apply_flag, apply_str, apply_string};
 use anyhow::{Context, Result};
 use clap::{ArgMatches, CommandFactory, FromArgMatches, Parser};
+use spur_core::resource::parse_memory_mb;
 use spur_core::spur_env::SpurEnv;
 use spur_proto::proto::{CancelJobRequest, GetJobRequest, JobSpec, SubmitJobRequest};
 use std::collections::HashMap;
@@ -394,8 +395,6 @@ fn resolve_salloc_env(matches: &ArgMatches, args: &mut SallocArgs) {
         &mut args.exclusive,
     );
 }
-
-use spur_core::resource::parse_memory_mb;
 
 #[cfg(test)]
 mod tests {
--- a/crates/spur-cli/src/sbatch.rs
+++ b/crates/spur-cli/src/sbatch.rs
@@ -5,6 +5,7 @@ use crate::env_defaults::{apply_csv, apply_str, apply_string, env_first, was_cli
 use anyhow::{bail, Context, Result};
 use clap::parser::ValueSource;
 use clap::{ArgMatches, CommandFactory, FromArgMatches, Parser};
+use spur_core::resource::parse_memory_mb;
 use spur_proto::proto::{JobSpec, SubmitJobRequest};
 use std::collections::HashMap;
 
@@ -635,9 +636,6 @@ fn parse_datetime_arg(s: &str) -> Result<chrono::DateTime<chrono::Utc>> {
         &chrono::Local::now(),
     )?)
 }
-
-/// Parse memory string (e.g., "4G", "4096M", "4096") into MB.
-use spur_core::resource::parse_memory_mb;
 
 /// Parse a GPU flag value ("4" or "mi300x:4") into a proto GpuRequest.
 pub(crate) fn parse_gpu_flag(value: Option<&str>) -> Result<Option<spur_proto::proto::GpuRequest>> {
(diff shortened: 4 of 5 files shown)

### CODE REVIEW

**Review by sajmera-pensando — CHANGES_REQUESTED** [review]
The consolidation of three separate `parse_memory_mb` implementations into a single shared helper is the right move, and the new `T` suffix support is a clean addition. However, the PR introduces or carries forward a silent correctness bug — sub-MiB `K` values truncate to `0` MiB and are treated downstream as "no memory limit" — and a doc-placement error that strips `parse_gres` of its documentation entirely. Both must be fixed before merge. The documentation also overstates which commands accept `--mem-per-cpu`, which will cause visible user-facing failures.

---

## Findings

### Major

- `crates/spur-core/src/resource.rs:301` — `parse_memory_mb` doc comment is inserted inside the `parse_gres` doc block, stripping `parse_gres` of its documentation
- `crates/spur-core/src/resource.rs:307` — K-suffix values smaller than 1024 KiB silently return 0 MiB, bypassing the memory limit
- `crates/spur-core/src/resource.rs:328` — K suffix silently produces 0 MiB for values < 1024K — Slurm rounds up
- `docs/user-guide/submitting-jobs.rst:144` — `--mem-per-cpu` suffix claim references `srun` and `salloc`, but neither command exposes that flag

### Minor

- `docs/user-guide/submitting-jobs.rst:141` — K-suffix sub-MiB truncation silently produces 0 MiB and is undocumented
- `docs/user-guide/submitting-jobs.rst:141` — Fractional values accepted by the parser (e.g. `1.5G`) are not mentioned in docs

### Nit

- `crates/spur-core/src/resource.rs:325` — Doc comment lists callers by name; violates the project convention against implementation-detail comments
- `crates/spur-core/src/resource.rs:875` — K-suffix test only covers the exact `1024K → 1 MiB` boundary, hiding the truncation bug for all sub-MiB values

---

## Required before merge

1. Fix the `0 MiB` truncation for sub-1024 K values — use integer ceiling division (`kib.div_ceil(1024)`) rather than float multiplication and cast. This is a silent correctness bug: `--mem=512K` submits with no memory limit.
2. Fix the doc-comment placement so `parse_gres` retains its documentation and `parse_memory_mb` gets only its own doc block.
3. Correct the `--mem-per-cpu` docs row — `srun` and `salloc` do not expose that flag in this codebase. Scope the suffix claim to `sbatch` only.

## Nice to have

1. Add a brief doc note that K values below 1 MiB round down (or up, after the fix).
2. Document the fractional-numeric extension (e.g. `1.5G`) as a Spur addition.
3. Remove the caller-enumeration sentence from the `parse_memory_mb` doc comment.
4. Expand the K-suffix unit tests to cover `512K` (truncation/rounding boundary) and `2048K`.

## Testing

Please include command output from a real cluster showing the fix working end-to-end — at minimum:

```
$ srun --mem=1024K ...   # previously failed with "invalid memory value"
$ salloc --mem=2G ...    # same
$ sbatch --mem=512M ...  # regression check
```

Ideally also show `squeue` or controller logs confirming the `memory_per_node_mb` value the controller received matches the submitted value, so we can verify the parse-to-wire path is correct.

Inline comments added for each finding below.

File: docs/user-guide/submitting-jobs.rst [hunk 1]
```diff
@@ -138,10 +138,12 @@ patterns for ``--output`` and ``--error``, the token expands to the job ID.
      - CPUs per task. Default ``1``.
    * - ``--mem``
      -
-     - Memory per node. ``512G`` and ``4096M`` use suffixes; a bare ``4096`` means MB.
+     - Memory per node. Takes a ``K``, ``M``, ``G``, or ``T`` suffix
```
sajmera-pensando, line 141: **K-suffix sub-MiB truncation silently produces 0 MiB and is undocumented**

🟡 **Minor**

The shared parser computes `(value_in_kb / 1024.0) as u64`, so any K value less than 1024 truncates to 0 MiB (e.g. `512K` → 0, `1023K` → 0). A job submitted with `--mem=512K` silently receives 0 MiB of memory, which is either rejected by the controller or misinterpreted as "no limit".

Once the K-path is corrected to use ceiling division, consider adding a note here that K values are rounded up to the nearest MiB.

sajmera-pensando, line 141: **Fractional values accepted by the parser are not mentioned in docs**

🟡 **Minor**

The shared `parse_memory_mb` accepts fractional numeric values (e.g. `1.5G`), and there is an explicit unit test `fractional_gigabytes` that asserts `1.5G == 1536`. Slurm's man pages do not document fractional memory values, so this is a Spur extension that Slurm-migration users will not expect.

The updated doc sentence lists supported suffixes but does not mention that the numeric part may be fractional. Adding a short example (`1.5G`) would make this extension visible.

crates/spur-core/src/resource.rs
```diff
@@ -298,6 +298,30 @@ impl ResourceSet {
 /// The input is trimmed and empty strings are rejected, so callers that split a
 /// comma-list (CLI `--gres`, `SLURM_GRES`, REST, FFI) can pass raw segments
 /// without worrying about incidental whitespace or trailing commas.
+/// Parse a Slurm-style `--mem` size into MiB. Optional K/M/G/T suffix
```
sajmera-pensando, line 301: **`parse_memory_mb` doc comment is inserted inside the `parse_gres` doc block, stripping `parse_gres` of its documentation**

🟠 **Major**

The new function and its `///` lines are inserted immediately after the last `///` line of `parse_gres`'s doc block, with no blank line separating them. In Rust, consecutive `///` lines form one doc comment attached to the next item. After this patch:

```rust
/// The input is trimmed … trailing commas.
/// Parse a Slurm-style `--mem` size into MiB. …
pub fn parse_memory_mb(…) { … }

pub fn parse_gres(…) {   // ← no doc comment at all
```

`parse_gres` silently loses its documentation, and `parse_memory_mb` gets a garbled merged comment. Fix: insert the new function before the `parse_gres` doc block, or after `parse_gres` entirely.

crates/spur-core/src/resource.rs
```diff
@@ -298,6 +298,30 @@ impl ResourceSet {
...
+/// (case-insensitive), no suffix means MiB. Shared by sbatch/srun/salloc.
+pub fn parse_memory_mb(s: &str) -> anyhow::Result<u64> {
+    let s = s.trim();
+    if s.is_empty() {
+        anyhow::bail!("invalid memory value ''");
+    }
```
sajmera-pensando, line 307: **K-suffix values smaller than 1024 KiB silently return 0 MiB, bypassing the memory limit**

🟠 **Major**

The multiplier `1.0 / 1024.0` converts KiB to MiB by floating-point multiplication then truncates to `u64`. Any value below 1024K (e.g. `512K`, `1K`) produces a fractional result that truncates to `0`.

At the scheduler, `memory_per_node_mb == 0` is treated as `None` — no constraint. So `--mem=512K` silently submits a job with *no* memory limit.

A safe fix using integer arithmetic:

```rust
b'K' | b'k' => {
    let kib: u64 = num.trim().parse()
        .map_err(|_| anyhow::anyhow!("invalid memory value '{s}'"))?;
    if kib == 0 { anyhow::bail!("invalid memory value '{s}'"); }
    return Ok(kib.div_ceil(1024));
}
```

crates/spur-core/src/resource.rs
```diff
@@ -298,6 +298,30 @@ impl ResourceSet {
...
+        .trim()
+        .parse()
+        .map_err(|_| anyhow::anyhow!("invalid memory value '{s}'"))?;
+    if !val.is_finite() || val < 0.0 {
+        anyhow::bail!("invalid memory value '{s}'");
+    }
+    Ok((val * mult) as u64)
+}
+
 pub fn parse_gres(gres: &str) -> Option<(String, Option<String>, u32)> {
```
sajmera-pensando, line 325: **Doc comment lists callers by name — violates the project convention against implementation-detail comments**

🔵 **Nit**

AGENTS.md: "Comments are only for *why* — non-obvious intent, trade-offs, or constraints the code itself cannot convey." The sentence enumerating `sbatch`/`srun`/`salloc` as callers is a caller inventory, not a design rationale — it will drift as the codebase evolves. Drop it.

crates/spur-core/src/resource.rs
```diff
@@ -809,3 +833,43 @@ mod accounting_tests {
...
+    }
+
+    #[test]
+    fn rejects_garbage_and_empty() {
+        assert!(parse_memory_mb("").is_err());
+        assert!(parse_memory_mb("abc").is_err());
+        assert!(parse_memory_mb("G").is_err());
+        assert!(parse_memory_mb("-5G").is_err());
+    }
+}
```
sajmera-pensando, line 875: **K-suffix test only covers the exact `1024K → 1 MiB` boundary, hiding the truncation bug**

🔵 **Nit**

The test exercises only the value that maps exactly to 1 MiB, leaving `512K` (currently returns 0), `1K` (returns 0), and `2048K` (should return 2) uncovered. Adding these cases would make the boundary behaviour explicit and prevent a future refactor from silently regressing.

crates/spur-core/src/resource.rs
```diff
@@ -298,6 +298,30 @@ impl ResourceSet {
...
 /// without worrying about incidental whitespace or trailing commas.
+/// Parse a Slurm-style `--mem` size into MiB. Optional K/M/G/T suffix
+/// (case-insensitive), no suffix means MiB. Shared by sbatch/srun/salloc.
+pub fn parse_memory_mb(s: &str) -> anyhow::Result<u64> {
+    let s = s.trim();
+    if s.is_empty() {
+        anyhow::bail!("invalid memory value ''");
+    }
+    let (num, mult): (&str, f64) = match s.as_bytes()[s.len() - 1] {
+        b'K' | b'k' => (&s[..s.len() - 1], 1.0 / 1024.0),
```
sajmera-pensando, line 309: **K suffix silently produces 0 MiB for values < 1024K — Slurm rounds up**

🟠 **Major**

Slurm's `--mem` man page specifies K as kibibytes, with the scheduler storing memory in MiB rounded **up**. This arm sets `mult = 1.0 / 1024.0` and then `(val * mult) as u64` truncates. Any value in `[1K, 1023K]` produces `0` MiB — interpreted downstream as "no limit."

Example: `--mem=512K` → `(512.0 * 0.0009765625) as u64 = 0`.

The test suite only exercises the exact boundary `1024K → 1` and misses all sub-MiB values. Consider switching the K path to integer arithmetic to eliminate the truncation hazard entirely:

```rust
b'K' | b'k' => {
    let kib: u64 = num.trim().parse()
        .map_err(|_| anyhow::anyhow!("invalid memory value '{s}'"))?;
    if kib == 0 { anyhow::bail!("invalid memory value '{s}'"); }
    return Ok(kib.div_ceil(1024));
}
```

SumonAMD:
Thanks @sajmera-pensando  for the detailed review — all four Major, both Minor and both Nit findings are fixed, along with the four nice-to-haves. Pushed as `93162d8`. Also rebased onto `main` (`5e1b5e7`), which clears the merge conflict the branch had picked up.

One thing done differently from the suggestion: rather than `div_ceil` on the `K` arm only, the conversion now rounds up for every suffix. The truncation lives in the shared `as u64` cast, not in the `K` path, so `--mem=0.5M` also reached the controller as `0` (no limit) and `1.5M` as 1 MiB — a `K`-only fix would have left both in place.

Verified on a 2-node cluster, values read back with `scontrol show job`:

```
--mem=        as reviewed (38eda2c)    pushed (93162d8)
512K          MinMemoryNode=0          MinMemoryNode=1M
0.5M          MinMemoryNode=0          MinMemoryNode=1M
1.5M          MinMemoryNode=1M         MinMemoryNode=2M
1024K/2G/1T   unchanged                unchanged
0             MinMemoryNode=0          MinMemoryNode=0   (Slurm's "whole node")
```

`srun --mem=1024K` and `salloc --mem=2G` both run now (previously `invalid memory value`), `abc` and `-5G` are still rejected by all three, and `parse_gres` keeps its doc comment (confirmed from generated rustdoc). `cargo fmt` and `clippy` clean, 4299 tests pass.

docs/user-guide/submitting-jobs.rst
SumonAMD: Fixed. The K path rounds up now, and the --mem row says so: a size is rounded up to a whole MiB, so 512K asks for 1 MiB.

SumonAMD: Fixed. The --mem row calls out that a fractional size such as 1.5G is a Spur extension, since Slurm's man pages take whole numbers only.

crates/spur-core/src/resource.rs
SumonAMD: Fixed. parse_memory_mb sits above parse_gres with a blank line between them, so parse_gres keeps its own doc block.

crates/spur-core/src/resource.rs
SumonAMD: Fixed. The result is .ceil()ed, so 1K, 512K and 1023K are 1 MiB and 2048K is 2. Kept one f64 expression rather than a K-only integer path — every multiplier is a power of two, so integer inputs still convert exactly.

crates/spur-core/src/resource.rs
SumonAMD: Fixed. The doc comment no longer names callers; it explains why rounding goes up — a zero reaches the controller as "no limit".

crates/spur-core/src/resource.rs
SumonAMD: Fixed. sub_mib_sizes_round_up_instead_of_vanishing asserts all three values you named plus the boundaries: 1K/512K/1023K → 1, 1025K/2048K → 2, and 0.5M → 1.

crates/spur-core/src/resource.rs
SumonAMD: Fixed as above: the result is rounded up, with 512K / 1023K / 2048K covered by tests.

PR #732, opened 2026-08-24
fix(spur-sched): protect large multi-node jobs from backfill starvation

yansun1996:
## Summary

A job requesting more nodes than currently have any free capacity could be starved indefinitely by a stream of smaller jobs, regardless of priority — it was silently skipped every scheduling cycle with no forward reservation recorded at all, so any node that later freed up went to whichever smaller job asked for it first.

Several changes close this:

- `find_suitable_nodes()` gated candidacy on `Node::has_free_cpu_capacity()`, excluding fully-saturated (but healthy) nodes before the timeline/reservation machinery ever ran. A new `NodePlacement::matches_for_reservation()` admits them as valid future-reservation candidates instead. Extended to heterogeneous jobs too (uneven per-node CPU/GPU splits) — they previously bailed out before reaching the reservation path at all if they couldn't start immediately.
- Even with candidacy fixed, the scheduler had no way to know when an already-running job actually finishes — any busy node was modeled as occupied for a flat 24h placeholder. `ClusterState` now carries a `busy_until` map (node → real expected free time, built once per scheduling tick from running jobs' `start_time` + `time_limit`, adjusted for suspended time and the kill grace period) replacing the flat placeholder wherever real data is available. A job with no time limit at all now assumes a shared 365-day sentinel (`UNLIMITED_JOB_DURATION`) instead of a flat 1-hour guess — both for its own backfill sizing and for a running job's remaining-time estimate — so an unbounded job can no longer slip into a gap ahead of a protected reservation just because it was assumed short-lived.
- A multi-node job's placement previously computed each candidate node's earliest start *independently*, took the max across the set, and reserved every node at that shifted time without re-validating it — a node whose own earliest start was earlier than the shifted common time could end up reserved on top of an unauthorized admin reservation. Replaced with a fixed-point search that converges on a start every selected node is simultaneously and genuinely free at, folding resource and reservation conflicts into one forward search per node.
- Adds a small observability feature: a node that's idle right now but held by the scheduler for a specific pending job's future start previously looked plain `idle`. `spur nodes`/`sinfo` now shows this as `plnd` (matching Slurm's own convention), and the same info is available via the gRPC and REST node APIs (new append-only `NodeInfo.planned_job_id`/`planned_start` fields). Purely ephemeral/display-layer — nothing here touches the persisted `NodeState` or the WAL.
- Documents the priority-gap preemption formula in detail (worked examples, what can and can't fire, comparison with Slurm) in `docs/admin-guide/accounting.rst`, and adds a caveat in `docs/admin-guide/configuration.rst` that backfill's protection is only as good as the wall-time information available — a fully unconfigured cluster (no time limits anywhere) gets little practical benefit from it, and preemption is the right tool for jobs that must not be starved regardless of duration.

## Known limitations

- Heterogeneous jobs' coarse per-node timing estimate can still occasionally diverge from the exact per-node plan resolved later, causing a job to be dropped from a cycle with no reservation rather than deferred to the correct future time. Self-healing on the next cycle, pre-existing behavior widened rather than introduced by this PR.
- `plnd` tells you a node is held but not which job or when — the data is fully available via gRPC/REST, just not yet rendered in a CLI long-format column.
- `reservation_fence_reason` (the user-facing pending-reason diagnostic) checks each candidate node independently at `now` and doesn't know about the scheduler's multi-node common-window shift, so it could occasionally report the wrong reason for a job blocked only after that shift. Diagnostic-accuracy gap only — backfill's own scheduling decision is correct regardless.

## Testing

- New unit tests for every behavioral change (CPU, GPU, and heterogeneous variants of the common-window fix), each proven to fail against the prior behavior and pass against the fix.
- New e2e test (`TestBackfillReservation`) reproducing the original starvation scenario end-to-end, verified against the real fix and against pre-fix binaries.
- Full workspace `cargo clippy`/`cargo fmt` clean; `spur-sched`, `spurctld`, `spur-cli`, and `spur-tests` suites all green.
- Validated end-to-end on an isolated 2-node deployment.

### DIFF
--- a/crates/spurctld/src/scheduler_loop.rs
+++ b/crates/spurctld/src/scheduler_loop.rs
@@ -532,11 +532,8 @@ pub(crate) fn compute_job_allocation(
     }
 }
 
-/// Real per-node "free again at" times, computed once per cycle from actual
-/// running jobs (falls back to a flat placeholder in the caller for nodes
-/// with no entry) instead of guessing at every busy node's remaining time.
-/// Folds in suspended time and the kill grace period so the estimate isn't
-/// optimistic; saturates on overflow instead of panicking on a bad time_limit.
+/// Real per-node "free again at" times derived from running jobs, so backfill
+/// doesn't fall back to a flat placeholder for every busy node.
 fn running_jobs_busy_until(cluster: &ClusterManager) -> HashMap<String, DateTime<Utc>> {
     let running = cluster.get_jobs(&JobFilter {
         states: &[spur_core::job::JobState::Running],
@@ -557,19 +554,25 @@ fn busy_until_from_running_jobs(running: &[spur_core::job::Job]) -> HashMap<Stri
         let start = job.start_time.unwrap_or(now);
         let mut suspended = job.suspended_secs;
         if let Some(since) = job.suspended_at {
-            suspended += (now - since).num_seconds().max(0);
+            suspended = suspended.saturating_add((now - since).num_seconds().max(0));
         }
-        // Clamp first: an oversized time_limit would overflow the Duration
-        // arithmetic below before checked_add_signed ever gets a chance to guard it.
-        let time_limit = job
+        // Saturating i64 math, clamped once at the end: avoids the panicking
+        // Duration + Duration path (a huge suspended_secs alone can overflow it).
+        let time_limit_secs = job
             .spec
             .time_limit
-            .unwrap_or(UNLIMITED_FALLBACK)
-            .min(UNLIMITED_FALLBACK);
-        let slack = time_limit
-            + chrono::Duration::seconds(suspended)
-            + chrono::Duration::seconds(GRACE_PERIOD_SECS);
-        let end = start.checked_add_signed(slack).unwrap_or(far_future);
+            .map(|d| d.num_seconds())
+            .unwrap_or(UNLIMITED_FALLBACK.num_seconds());
+        let slack_secs = time_limit_secs
+            .saturating_add(suspended)
+            .saturating_add(GRACE_PERIOD_SECS)
+            .min(UNLIMITED_FALLBACK.num_seconds());
+        // A long-unbounded job outliving UNLIMITED_FALLBACK would otherwise
+        // compute an `end` already in the past, producing a backwards interval.
+        let end = start
+            .checked_add_signed(chrono::Duration::seconds(slack_secs))
+            .unwrap_or(far_future)
+            .max(now);
         for node in &job.allocated_nodes {
             busy_until
                 .entry(node.clone())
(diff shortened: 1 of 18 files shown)


crates/spurctld/src/scheduler_loop.rs
```diff
@@ -520,6 +532,54 @@ pub(crate) fn compute_job_allocation(
...
+        // arithmetic below before checked_add_signed ever gets a chance to guard it.
+        let time_limit = job
+            .spec
+            .time_limit
+            .unwrap_or(UNLIMITED_FALLBACK)
+            .min(UNLIMITED_FALLBACK);
+        let slack = time_limit
+            + chrono::Duration::seconds(suspended)
+            + chrono::Duration::seconds(GRACE_PERIOD_SECS);
+        let end = start.checked_add_signed(slack).unwrap_or(far_future);
```
sajmera-pensando: The `slack` variable is built with two unchecked `chrono::Duration` additions after `time_limit` is clamped. If `suspended_secs` is very large (e.g. corrupted state), `chrono::Duration::seconds(suspended) + chrono::Duration::seconds(GRACE_PERIOD_SECS)` can overflow before `checked_add_signed` ever runs — Chrono panics on Duration overflow, which would kill the scheduler loop.

The existing saturation test (`busy_until_saturates_instead_of_panicking_on_an_extreme_time_limit`) only exercises the `time_limit` path — it passes because `Duration::MAX` is clamped to 365 days before the unchecked arithmetic. A test with extreme `suspended_secs` would currently panic.

Please fix in this PR by building `slack` with checked arithmetic throughout:

```rust
let slack = time_limit
    .checked_add(&chrono::Duration::seconds(suspended))
    .and_then(|d| d.checked_add(&chrono::Duration::seconds(GRACE_PERIOD_SECS)))
    .unwrap_or(UNLIMITED_FALLBACK);
let end = start.checked_add_signed(slack).unwrap_or(far_future);
```

And add a companion test alongside the existing saturation test:

```rust
#[test]
fn busy_until_saturates_on_extreme_suspended_secs() {
    let now = Utc::now();
    let mut job = running_job_on("node001", now, 60);
    job.suspended_secs = i64::MAX;
    let busy_until = busy_until_from_running_jobs(&[job]);
    assert!(busy_until.contains_key("node001")); // must not panic
}
```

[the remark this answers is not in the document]
yansun1996: Good catch — fixed. The current version avoids `Duration + Duration` arithmetic entirely: `suspended_secs`/`time_limit`/grace are combined as plain `i64` seconds with `saturating_add`, clamped via `.min(UNLIMITED_FALLBACK.num_seconds())`, and only then turned into a single `Duration::seconds(...)` right before `checked_add_signed`. Added a test with `suspended_secs = i64::MAX`.

yansun1996: Thanks — fixed, though I went with plain `i64` saturating arithmetic instead of the `checked_add` snippet. Reason: `chrono::Duration::seconds(suspended)` on the raw, unclamped value panics on its own once `suspended` exceeds roughly `i64::MAX/1000` (chrono stores milliseconds internally), so building the `Duration` first and `checked_add`-ing after would still panic on `suspended_secs = i64::MAX`. Clamping in `i64` before ever constructing a `Duration` avoids that. Added your suggested test with `suspended_secs = i64::MAX` — passes.

shiv-tyagi (dismissed):
All good except few nits. Ignoring those. Let us merge since it is critical. Please resolve the merge conflicts.

shiv-tyagi (approved):

PR #737, opened 2026-08-25
feat(spur-logging): add structured JSON logging

maybeharshit:
## Summary

Adds a shared `spur-logging` crate so every binary (`spurctld`, `spurd`, `spur-k8s-operator`, `spur`) installs one subscriber with a single schema, replacing four independent `tracing_subscriber::fmt()` setups. In `json` mode each line is one flat JSON object; `text` mode keeps the existing human-readable output.

Every JSON line carries a fixed schema: `timestamp` (RFC 3339, nanosecond), `level` (lowercase), `component`, `target` (module path), `message`, plus any structured fields the event already passed (`job_id`, `node`, `error`, ...), each keeping its native JSON type — a numeric `job_id` stays a number.

- **Format:** `--log-format` > `[logging].format` > (stderr is a TTY ? `text` : `json`).
- **Level:** `RUST_LOG` > `--log-level` > `[logging].level` > `info`.

## Approach / design

- Uses a custom `FormatEvent` rather than stock `fmt().json()`, which uppercases `level`, has no top-level `component`, and nests span fields under `span`. A `Visit` impl records each field with its native JSON type.
- Implements the "fields already in scope" reading: `job_id`/`node` are flattened from what call sites already pass — no span-propagation layer and no call-site rewrites.
- Config is now loaded **before** logging init so `[logging]` can choose the format; nothing logs before init.
- `LoggingConfig` gains `#[serde(default)]` on all fields and its default `format` changes `"text"` to unset. It is TOML-config only (not Raft/WAL state, not proto) and field types are unchanged, so this is backward compatible.

## Behavior change (intended, documented)

A daemon under a service manager (no TTY) now emits **JSON by default** where it previously emitted text; interactive terminals still get text. `examples/spur.conf` ships `format = "json"`, and `docs/admin-guide/configuration.rst` documents the formats, the TTY default, and precedence.

## Testing

- 7 unit tests in `spur-logging` (schema/required fields, lowercase level, native-typed `job_id`, message fallback, quote/newline robustness, level & format precedence).
- `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` — no warnings.
- `cargo test` — green; the only failures are 10 `spurd` cgroup/PTY tests that need kernel features absent in the build sandbox and pass on a real host.
- End-to-end on an LXD container via the native-host harness: new `tests/native_host/e2e/test_json_logging.py` (valid JSON schema from real daemons; numeric `job_id` in the controller's `job submitted` line after a real submit; text mode stays readable), plus the existing `test_single_node.py` (18 tests) all pass with JSON as the piped default.

## Follow-ups (out of scope)

Span-based propagation (fields on events that don't pass them), a file sink (`logging.file`), and OpenTelemetry are intentionally deferred.

### DIFF
--- a/crates/spur-logging/src/lib.rs
+++ b/crates/spur-logging/src/lib.rs
@@ -1,0 +1,337 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Shared logging setup for Spur binaries.
+//!
+//! Every binary (`spurctld`, `spurd`, `spur-k8s-operator`, `spur`) installs the
+//! same subscriber through [`init`] or [`init_cli`], so a single schema and a
+//! single set of precedence rules govern all of them instead of four
+//! independent `tracing_subscriber::fmt()` setups drifting apart.
+//!
+//! JSON lines carry a fixed schema: `timestamp` (RFC 3339, nanoseconds),
+//! `level` (lowercase), `component`, `target` (Rust module path), `message`,
+//! and any structured fields the call site attached (`job_id`, `node`,
+//! `error`, ...). Values keep their native JSON type, so numbers stay numbers.
+
+use std::fmt;
+use std::io::IsTerminal;
+
+use tracing::field::{Field, Visit};
+use tracing::{Event, Level, Subscriber};
+use tracing_subscriber::fmt::format::Writer;
+use tracing_subscriber::fmt::{FmtContext, FormatEvent, FormatFields};
+use tracing_subscriber::prelude::*;
+use tracing_subscriber::registry::LookupSpan;
+use tracing_subscriber::EnvFilter;
+
+/// Output format for logs.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Format {
+    Json,
+    Text,
+}
+
+/// Install the global subscriber for a daemon.
+///
+/// `component` is stamped onto every line. `cli_level` / `cli_format` are the
+/// `--log-level` / `--log-format` flag values (`None` when unset); `config_*`
+/// are the `[logging]` values (empty string when unset). Format is chosen by
+/// the TTY rule when neither flag nor config sets it.
+pub fn init(
+    component: &'static str,
+    cli_level: Option<&str>,
+    cli_format: Option<&str>,
+    config_level: &str,
+    config_format: &str,
+) {
+    let filter = resolve_filter(cli_level, config_level, "info");
+    let format = resolve_format(cli_format, config_format, std::io::stderr().is_terminal());
+    install(component, filter, format, false);
+}
+
+/// Install the global subscriber for the CLI.
+///
+/// The CLI ignores `spur.conf` for logging and defaults to `warn` so library
+/// diagnostics stay quiet next to the command's own output. Text mode is
+/// compact (no timestamp/target) to match the CLI's historical behaviour;
+/// piped output still emits the shared JSON schema.
+pub fn init_cli(component: &'static str) {
+    let filter = resolve_filter(None, "", "warn");
+    let format = resolve_format(None, "", std::io::stderr().is_terminal());
+    install(component, filter, format, true);
+}
+
+fn install(component: &'static str, filter: EnvFilter, format: Format, compact_text: bool) {
+    match format {
+        Format::Json => {
+            let layer = tracing_subscriber::fmt::layer()
+                .event_format(SpurJsonFormat::new(component))
+                .with_writer(std::io::stderr);
+            let _ = tracing_subscriber::registry()
+                .with(filter)
+                .with(layer)
+                .try_init();
+        }
+        Format::Text if compact_text => {
+            // The CLI's own output is on stdout; keep tracing noise minimal.
+            let _ = tracing_subscriber::fmt()
+                .with_env_filter(filter)
+                .with_writer(std::io::stderr)
+                .without_time()
+                .with_target(false)
+                .try_init();
+        }
+        Format::Text => {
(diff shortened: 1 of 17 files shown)


crates/spur-logging/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+    }
+}
+
+/// Records `tracing` field values into a JSON map, keeping native types.
+struct JsonVisitor<'a>(&'a mut serde_json::Map<String, serde_json::Value>);
+
+impl JsonVisitor<'_> {
+    fn insert(&mut self, field: &Field, value: serde_json::Value) {
+        self.0.insert(field.name().to_string(), value);
+    }
```
shiv-tyagi, line 193: Reserved schema keys are overwritten by structured event fields. `SpurJsonFormat` inserts `timestamp`/`level`/`component`/`target`, then `event.record` runs, so a call site field with the same name wins. This already happens today: `spur-k8s/src/agent.rs` logs `target = %req.target_node` (JSON `target` becomes a node name, not the module path), and `spurd/src/agent_server.rs` plus `spurctld/src/cluster_k8s.rs` log `message = %...` alongside an event message.

Guard the canonical keys and rename collisions:

```suggestion
    fn insert(&mut self, field: &Field, value: serde_json::Value) {
        let name = field.name();
        let key = match name {
            "timestamp" | "level" | "component" | "target" | "message" => format!("field_{name}"),
            _ => name.to_owned(),
        };
        self.0.insert(key, value);
    }
```

crates/spur-logging/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+    default_level.to_string()
+}
+
+fn resolve_filter(cli_level: Option<&str>, config_level: &str, default_level: &str) -> EnvFilter {
+    // RUST_LOG (including per-module directives) wins when present and valid.
+    if let Ok(filter) = EnvFilter::try_from_default_env() {
+        return filter;
+    }
+    EnvFilter::new(resolve_level(cli_level, config_level, default_level))
```
shiv-tyagi, line 113: An invalid `--log-level` or `[logging].level` silences the daemon. `EnvFilter::new` treats an unparseable value like `notalevel` as a target directive that matches nothing, so the process starts and emits zero log lines. Verified on a real cluster: `spurctld --log-level notalevel` produced an empty log while running. Reject invalid filters instead of falling back:

```suggestion
fn resolve_filter(cli_level: Option<&str>, config_level: &str, default_level: &str) -> Result<EnvFilter, LoggingError> {
    if let Ok(filter) = EnvFilter::try_from_default_env() {
        return Ok(filter);
    }
    resolve_level(cli_level, config_level, default_level)
        .parse::<EnvFilter>()
        .map_err(LoggingError::InvalidFilter)
}
```

crates/spur-logging/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+        Some(config_format)
+    };
+    let choice = cli_format.filter(|s| !s.is_empty()).or(config_choice);
+    match choice {
+        Some("json") => Format::Json,
+        Some("text") => Format::Text,
+        // Unset (or "auto"/unrecognized): text on a terminal, JSON otherwise.
+        _ if is_tty => Format::Text,
+        _ => Format::Json,
+    }
```
shiv-tyagi, line 129: An unrecognized format value silently falls through to the TTY default. Verified: `--log-format jason` produced JSON with no error, so a config typo flips text/JSON depending on whether stderr is a terminal. Reject unsupported values at parse time:

```suggestion
    match choice {
        Some("json") => Ok(Format::Json),
        Some("text") => Ok(Format::Text),
        Some(other) => Err(LoggingError::InvalidFormat(other.to_owned())),
        None if is_tty => Ok(Format::Text),
        None => Ok(Format::Json),
    }
```

crates/spur-logging/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+/// Install the global subscriber for the CLI.
+///
+/// The CLI ignores `spur.conf` for logging and defaults to `warn` so library
+/// diagnostics stay quiet next to the command's own output. Text mode is
+/// compact (no timestamp/target) to match the CLI's historical behaviour;
+/// piped output still emits the shared JSON schema.
+pub fn init_cli(component: &'static str) {
+    let filter = resolve_filter(None, "", "warn");
+    let format = resolve_format(None, "", std::io::stderr().is_terminal());
+    install(component, filter, format, true);
```
shiv-tyagi, line 61: Piped CLI invocations now emit JSON tracing diagnostics while command errors stay text, so scripts capturing stderr can see mixed output. This is documented as intended. Flagging so it stays a deliberate choice: keep CLI logging compact text unless an explicit format control is added.

tests/native_host/e2e/test_json_logging.py
```diff
@@ -0,0 +1,121 @@
...
+        )
+        submitted = [
+            o for o in ctrl if o["message"] == "job submitted" and o.get("job_id") == job_id
+        ]
+        assert submitted, (
+            f"expected a 'job submitted' line with job_id={job_id}; "
+            f"job_id-bearing lines: {[o for o in ctrl if 'job_id' in o]}"
+        )
+        # The whole point of the typed visitor: the id is a JSON number, not a string.
+        for o in submitted:
```
shiv-tyagi, line 100: Nit: this comment restates the assertion below it.

```suggestion
        for o in submitted:
```

shiv-tyagi (commented):
Some comments. PTAL.

One extra note that could not be anchored inline (`docs/deployment/native-host.rst` is not touched by this PR): `--log-format` is a new public flag on `spurctld` and `spurd`, but the flag tables in `docs/deployment/native-host.rst` still list only `--log-level`. Add a `--log-format` row to both tables.

PR #743, opened 2026-08-25
feat(scontrol): add show hostnames/hostlist/hostnamestr

### LINKED ISSUE #824: [Issue]: scontrol show hostnames/hostlist/hostnamestr is missing
## Problem Description

Spur's `scontrol` does not implement `show hostnames [NODELIST]` or the `hostlist` / `hostnamestr` variants that Slurm provides.

Why it matters: these are the idiomatic way distributed jobs discover their nodes. A very common line in launch scripts is:

    MASTER_ADDR=$(scontrol show hostnames "$SLURM_JOB_NODELIST" | head -n1)

Without these subcommands, Slurm rendezvous scripts (torchrun and similar) don't work unchanged on spur.

**Expected:**

- `hostnames`   — expand a hostlist to one hostname per line
- `hostlist`    — collapse hostnames into a compact bracketed expression
- `hostnamestr` — expand to a single comma-separated line

They should work outside an allocation and while the controller is down, and fall
back to `$SLURM_JOB_NODELIST` / `$SLURM_NODELIST` when no argument is given, as
Slurm does.

## Environment

- **OS:** Ubuntu 22.04 LTS
- **ROCm Version:** N/A — feature gap in the spur CLI, independent of the ROCm runtime
- **Component:** spur

## Steps to Reproduce

```bash
scontrol show hostnames node[1-4]     # not implemented
scontrol show hostlist node1,node2    # not implemented
scontrol show hostnamestr node[1-4]   # not implemented

### PR DESCRIPTION
**Pull request by SumonAMD**
## Summary

Closes #824

Adds Slurm's `scontrol show hostnames [NODELIST]` and the `hostlist` /
`hostnamestr` variants, which were missing from spur. These are the idiomatic
way distributed jobs discover their nodes, e.g.
`MASTER_ADDR=$(scontrol show hostnames "$SLURM_JOB_NODELIST" | head -n1)`.

## What
- `hostnames`   — expand a hostlist to one hostname per line
- `hostlist`    — collapse hostnames into a compact bracketed expression
- `hostnamestr` — expand to a single comma-separated line

Resolved locally with no controller round-trip (matches Slurm: works outside an
allocation and while ctld is down), falling back to `$SLURM_JOB_NODELIST` /
`$SLURM_NODELIST` when no argument is given. Output is sorted + de-duplicated in
natural numeric order by reusing `spur_core::hostlist` expand/compress.

## Testing
- Unit tests: expansion (sort/dedup/multi-prefix) and argument resolution.
- Manual: verified `hostnames`/`hostlist`/`hostnamestr`, env fallback, and the
  no-arg error path against the built binary.
- `cargo build` / `fmt` / `clippy -D warnings` clean.

### DIFF
--- a/crates/spur-cli/src/scontrol.rs
+++ b/crates/spur-cli/src/scontrol.rs
@@ -31,7 +31,7 @@ pub struct ScontrolArgs {
 pub enum ScontrolCommand {
     /// Show detailed information
     Show {
-        /// Entity type: job, node, partition, reservation, assoc_mgr, federation, config
+        /// Entity type: job, node, partition, reservation, assoc_mgr, federation, config, hostnames, hostlist
         entity: String,
         /// Entity name or ID
         name: Option<String>,
@@ -567,6 +567,15 @@ fn format_config(controller: &str, ping: &spur_proto::proto::PingResponse) -> St
 }
 
 async fn show(controller: &str, entity: &str, name: Option<&str>) -> Result<()> {
+    // Offline hostlist utilities resolve locally — no controller round-trip,
+    // matching Slurm (these work outside an allocation and while ctld is down).
+    match entity.to_lowercase().as_str() {
+        "hostnames" | "hostname" => return show_hostnames(name),
+        "hostlist" | "hostlistsorted" => return show_hostlist(name),
+        "hostnamestr" => return show_hostnamestr(name),
+        _ => {}
+    }
+
     let channel = crate::authclient::connect(controller)
         .await
         .context("failed to connect to spurctld")?;
@@ -840,12 +849,72 @@ async fn show(controller: &str, entity: &str, name: Option<&str>) -> Result<()> 
         }
         other => {
             bail!(
-                "scontrol: unknown entity type '{}'. Use: job, node, partition, reservation, assoc_mgr, federation, config",
+                "scontrol: unknown entity type '{}'. Use: job, node, partition, reservation, assoc_mgr, federation, config, hostnames, hostlist",
                 other
             );
         }
     }
 
+    Ok(())
+}
+
+/// Resolve the hostlist argument for the offline `scontrol show hostnames`
+/// family. Falls back to `$SLURM_JOB_NODELIST` / `$SLURM_NODELIST` like Slurm
+/// so `scontrol show hostnames` works with no argument inside a job.
+fn resolve_hostlist(name: Option<&str>) -> Result<String> {
+    if let Some(value) = name {
+        let trimmed = value.trim();
+        if !trimmed.is_empty() {
+            return Ok(trimmed.to_string());
+        }
+    }
+    for var in ["SLURM_JOB_NODELIST", "SLURM_NODELIST"] {
+        if let Ok(value) = std::env::var(var) {
+            if !value.trim().is_empty() {
+                return Ok(value);
+            }
+        }
+    }
+    bail!(
+        "scontrol show hostnames: no hostlist given and neither \
+         SLURM_JOB_NODELIST nor SLURM_NODELIST is set"
+    )
+}
+
+/// Expand a hostlist and return it sorted + de-duplicated in Slurm's natural
+/// (numeric) order. The compress/expand round-trip reuses the controller's own
+/// hostlist codec so ordering matches `compress` exactly (node10 after node2).
+fn normalized_hosts(pattern: &str) -> Result<Vec<String>> {
+    let hosts = spur_core::hostlist::expand(pattern)
+        .map_err(|e| anyhow::anyhow!("invalid hostlist '{pattern}': {e}"))?;
+    let compact = spur_core::hostlist::compress(&hosts);
+    spur_core::hostlist::expand(&compact)
+        .map_err(|e| anyhow::anyhow!("invalid hostlist '{compact}': {e}"))
+}
+
+/// `scontrol show hostnames [HOSTLIST]` — expand to one hostname per line.
+fn show_hostnames(name: Option<&str>) -> Result<()> {
+    let pattern = resolve_hostlist(name)?;
+    for host in normalized_hosts(&pattern)? {
+        println!("{host}");
+    }
+    Ok(())
+}
+
+/// `scontrol show hostlist HOSTLIST` — collapse hostnames into a compact
+/// bracketed expression on a single line.
+fn show_hostlist(name: Option<&str>) -> Result<()> {
+    let pattern = resolve_hostlist(name)?;
+    let hosts = spur_core::hostlist::expand(&pattern)
+        .map_err(|e| anyhow::anyhow!("invalid hostlist '{pattern}': {e}"))?;
+    println!("{}", spur_core::hostlist::compress(&hosts));
+    Ok(())
+}
+
+/// `scontrol show hostnamestr HOSTLIST` — expand to one comma-separated line.
+fn show_hostnamestr(name: Option<&str>) -> Result<()> {
+    let pattern = resolve_hostlist(name)?;
+    println!("{}", normalized_hosts(&pattern)?.join(","));
     Ok(())
 }
 
@@ -2218,6 +2287,32 @@ mod tests {
         assert!(!out.contains("ClusterName=spur"), "{out}");
         assert!(out.contains("SlurmctldAddr=http://ctld:6817"), "{out}");
         assert!(out.contains("Version=9.9.9"), "{out}");
+    }
+
+    #[test]
+    fn show_hostnames_expands_sorted_unique() {
+        assert_eq!(
+            normalized_hosts("node[1-3]").unwrap(),
+            vec!["node1", "node2", "node3"]
+        );
+        // numeric (not lexical) sort and de-duplication, matching Slurm
+        assert_eq!(
+            normalized_hosts("node[2,10,1],node1").unwrap(),
+            vec!["node1", "node2", "node10"]
+        );
+        assert_eq!(
+            normalized_hosts("gpu[1-2],cpu[1-2]").unwrap(),
+            vec!["cpu1", "cpu2", "gpu1", "gpu2"]
+        );
+    }
+
+    #[test]
+    fn resolve_hostlist_prefers_explicit_arg_then_errors() {
+        assert_eq!(resolve_hostlist(Some("node[1-2]")).unwrap(), "node[1-2]");
+        std::env::remove_var("SLURM_JOB_NODELIST");
+        std::env::remove_var("SLURM_NODELIST");
+        assert!(resolve_hostlist(Some("   ")).is_err());
+        assert!(resolve_hostlist(None).is_err());
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: crates/spur-cli/src/scontrol.rs [hunk 1]
```diff
@@ -849,6 +858,66 @@ async fn show(controller: &str, entity: &str, name: Option<&str>) -> Result<()>
...
+    bail!(
+        "scontrol show hostnames: no hostlist given and neither \
+         SLURM_JOB_NODELIST nor SLURM_NODELIST is set"
+    )
+}
+
+/// Expand a hostlist and return it sorted + de-duplicated in Slurm's natural
+/// (numeric) order. The compress/expand round-trip reuses the controller's own
+/// hostlist codec so ordering matches `compress` exactly (node10 after node2).
+fn normalized_hosts(pattern: &str) -> Result<Vec<String>> {
```
sajmera-pensando, line 887: [critical] **`show_hostnames` sorts its output, breaking Slurm order semantics.**

Slurm's `scontrol show hostnames` preserves the order in which nodes appear in the hostlist expression — it does not sort. The compress → expand round-trip inside `normalized_hosts` re-emits nodes in natural-numeric order regardless of input order.

The canonical pattern cited in the PR's own documentation:
```bash
MASTER_ADDR=$(scontrol show hostnames "$SLURM_JOB_NODELIST" | head -n1)
```
works in Slurm because the scheduler places the rank-0 node first in `$SLURM_JOB_NODELIST`. After this change `head -n1` always returns the numerically-lowest node, silently assigning the wrong master in any job where the scheduler assigns a non-numerically-first node as rank 0.

Only `scontrol show hostlistsorted` sorts in Slurm; `hostnames` and `hostlist` must preserve input order.

Fix: remove the `normalized_hosts` round-trip from `show_hostnames` and `show_hostnamestr`. If deduplication is wanted, apply an explicit `sort() + dedup()` only in a dedicated `show_hostnamessorted` path — do not change the order in the existing commands.

crates/spur-cli/src/scontrol.rs
```diff
@@ -2220,6 +2289,32 @@ mod tests {
...
+            normalized_hosts("gpu[1-2],cpu[1-2]").unwrap(),
+            vec!["cpu1", "cpu2", "gpu1", "gpu2"]
+        );
+    }
+
+    #[test]
+    fn resolve_hostlist_prefers_explicit_arg_then_errors() {
+        assert_eq!(resolve_hostlist(Some("node[1-2]")).unwrap(), "node[1-2]");
+        std::env::remove_var("SLURM_JOB_NODELIST");
+        std::env::remove_var("SLURM_NODELIST");
```
sajmera-pensando, line 2313: [major] `resolve_hostlist_prefers_explicit_arg_then_errors` calls `std::env::remove_var("SLURM_JOB_NODELIST")` and `std::env::remove_var("SLURM_NODELIST")`. Environment variables are process-global state. Cargo runs unit tests in parallel threads by default, so another test that reads or sets either variable concurrently will race with this removal — this is a data race and is flagged as `unsafe` from Rust 1.86 onward. If the test panics before any cleanup, the variables are left absent for the remainder of the test run.

This also violates the project convention: "Do not write tests that depend on the test runner's environment. Use explicit fixtures so tests pass identically everywhere." A real Slurm allocation sets both variables, so this test will produce wrong results in that environment.

Fix: extract the env-lookup logic in `resolve_hostlist` behind an injectable seam (a closure or a small trait), then test through that seam with a `HashMap` or closure fixture — no global state touched.

crates/spur-cli/src/scontrol.rs
```diff
@@ -849,6 +858,66 @@ async fn show(controller: &str, entity: &str, name: Option<&str>) -> Result<()>
...
+        if !trimmed.is_empty() {
+            return Ok(trimmed.to_string());
+        }
+    }
+    for var in ["SLURM_JOB_NODELIST", "SLURM_NODELIST"] {
+        if let Ok(value) = std::env::var(var) {
+            if !value.trim().is_empty() {
+                return Ok(value);
+            }
+        }
```
sajmera-pensando, line 876: [minor] The explicit-arg branch trims its value before returning (`return Ok(trimmed.to_string())`), but the env-var branch checks `value.trim().is_empty()` and then returns the original `value` — whitespace and all. If `SLURM_JOB_NODELIST` or `SLURM_NODELIST` contain leading/trailing whitespace (common when the variable is exported from a shell script), the raw whitespace is forwarded to `expand`, which may produce a confusing parse error or silent wrong output.

```suggestion
                return Ok(value.trim().to_string());
```

crates/spur-cli/src/scontrol.rs
```diff
@@ -849,6 +858,66 @@ async fn show(controller: &str, entity: &str, name: Option<&str>) -> Result<()>
...
+
+/// `scontrol show hostnames [HOSTLIST]` — expand to one hostname per line.
+fn show_hostnames(name: Option<&str>) -> Result<()> {
+    let pattern = resolve_hostlist(name)?;
+    for host in normalized_hosts(&pattern)? {
+        println!("{host}");
+    }
+    Ok(())
+}
+
```
sajmera-pensando, line 903: [minor] `show_hostlist` calls `expand` + `compress` directly, skipping the `normalized_hosts` round-trip that `show_hostnames` and `show_hostnamestr` use for sort and dedup. This means `scontrol show hostlist node[2,1,1]` may produce different output than `scontrol show hostnames node[2,1,1]` for the same input — which is confusing.

Once the sort-order issue above is resolved (removing the round-trip from `show_hostnames`/`show_hostnamestr`), make an explicit, documented decision about whether all three commands preserve input order. Keeping them consistent is more important than any particular sort policy.

crates/spur-cli/src/scontrol.rs
```diff
@@ -840,7 +849,7 @@ async fn show(controller: &str, entity: &str, name: Option<&str>) -> Result<()>
         }
         other => {
             bail!(
-                "scontrol: unknown entity type '{}'. Use: job, node, partition, reservation, assoc_mgr, federation, config",
+                "scontrol: unknown entity type '{}'. Use: job, node, partition, reservation, assoc_mgr, federation, config, hostnames, hostlist",
```
sajmera-pensando, line 852: [minor] The unknown-entity error string lists `hostnames, hostlist` but not `hostnamestr`. A user who mistypes `scontrol show hostnamestr` will see an error that never reveals the correct spelling.

```suggestion
                "scontrol: unknown entity type '{}'. Use: job, node, partition, reservation, assoc_mgr, federation, config, hostnames, hostlist, hostnamestr",
```

crates/spur-cli/src/scontrol.rs
```diff
@@ -31,7 +31,7 @@ pub struct ScontrolArgs {
 pub enum ScontrolCommand {
     /// Show detailed information
     Show {
-        /// Entity type: job, node, partition, reservation, assoc_mgr, federation, config
+        /// Entity type: job, node, partition, reservation, assoc_mgr, federation, config, hostnames, hostlist
```
sajmera-pensando, line 34: [minor] The `Show` variant's doc comment lists `job, node, partition, reservation, assoc_mgr, federation, config` but omits `hostnames`, `hostlist`, and `hostnamestr` — all three are accepted by the match arm below and are the subject of this PR. `--help` is usually the first place users look; the omission makes the new subcommands undiscoverable from the CLI.

```suggestion
        /// Entity type: job, node, partition, reservation, assoc_mgr, federation, config, hostnames, hostlist, hostnamestr
```

docs/user-guide/monitoring-jobs.rst
```diff
@@ -606,6 +606,61 @@ scopes the view to the caller and lists only the QOS and accounts they take part
...
+   :header-rows: 1
+   :widths: 30 70
+
+   * - Subcommand
+     - Output
+   * - ``show hostnames``
+     - One hostname per line. ``hostname`` is accepted as an alias.
+   * - ``show hostlist``
+     - The hostnames collapsed back into a compact bracketed expression, on one
+       line. ``hostlistsorted`` is accepted as an alias.
```
sajmera-pensando, line 636: [minor] The paragraph "Output is sorted in natural numeric order with duplicates removed" is placed after the three-command comparison table, so it reads as applying to all three forms. This has two problems:

[1] For `hostnames`, it contradicts Slurm's documented behaviour (order is preserved from the input expression). Teaching users that `head -n1` picks the master node will silently break once they run on a cluster where rank 0 is not the numerically-lowest node.

[2] For `hostlist`, the guarantee relies on `compress` deduplicating as a side-effect, not on any explicit step.

Once the sort-order issue in the code is resolved, update this paragraph to accurately reflect what each command actually guarantees, or scope it explicitly to whichever command does sort.

sajmera-pensando (changes requested):
**This PR needs revisions before merging — one blocking correctness issue and two major test-hygiene violations require fixes.**

The PR adds `scontrol show hostnamestr` and a `resolve_hostlist` helper that prefers an explicit argument over Slurm env vars. The feature direction is right and the env-var fallback chain is a useful addition, but there are several issues that must be addressed:

**Blocking — Slurm order semantics violated**

`show_hostnames` and `show_hostnamestr` pipe every expansion through a `normalized_hosts` helper that does an expand → compress → expand round-trip, which re-emits nodes in natural-numeric order. Slurm's `scontrol show hostnames` is explicitly order-preserving; `head -n1` on its output is the canonical way to find the scheduler-assigned master node in distributed jobs. After this PR that idiom silently picks the numerically-lowest node instead. Only `scontrol show hostlistsorted` sorts in Slurm; the other two must not. This needs to be fixed before merge.

**Major — tests mutate global environment state**

`resolve_hostlist_prefers_explicit_arg_then_errors` calls `std::env::remove_var` on two variables that are alive for the entire process. Cargo runs unit tests in parallel threads by default; this is a data race. The test also violates the project convention ("do not write tests that depend on the test runner's environment") and will produce wrong results inside a real Slurm allocation. The env-lookup logic needs to be extracted behind an injectable seam.

**Minor / nit recurring theme — `hostnamestr` omitted from error message and help text**

The new entity type is handled correctly in the match arm but is absent from the unknown-entity error string and the `Show` variant's `--help` doc comment, making it invisible to users who mistype it.

**Other issues**

- The env-var branch of `resolve_hostlist` checks `trim().is_empty()` but returns the original untrimmed value, unlike the explicit-arg branch.
- `show_hostlist` skips the sort/dedup step that the other two commands apply, producing inconsistent behaviour across the three subcommands.
- The documentation paragraph "Output is sorted in natural numeric order with duplicates removed" overstates the guarantee for `hostlist` and misrepresents Slurm semantics for `hostnames`.
- Several `///` doc comments on private helpers describe *what* the function does rather than *why*, contrary to project conventions.

Positive: the three-way fallback in `resolve_hostlist` (explicit arg → `SLURM_JOB_NODELIST` → `SLURM_NODELIST` → error) matches Slurm's documented precedence exactly, and the local-resolution design (no controller round-trip) is the right call for hostlist utilities.

PR #744, opened 2026-08-25
fix(spurctld): hold jobs whose QOS or association limits cannot be read

### LINKED ISSUE #745: QOS and association limits are silently not enforced while the accounting caches are cold
## Problem

The scheduling gate resolves a job's QOS through the controller's QOS cache and falls back to a limitless default when the cache cannot produce the name. A job carrying a QOS the cache does not hold is therefore admitted with no QOS limits applied at all. Because a running job is never re-checked against limits, such a job stays over the cap for its entire run.

Submit-time validation does not close this. It rejects an unknown `--qos`, and it already fails closed on a cold association cache, but the job queue is durable while the caches are not: a controller that restarts or takes over as leader inherits its predecessor's pending jobs and begins scheduling before its first accounting fetch completes. Every scheduling pass in that window admits queued jobs with no QOS limits. If the accounting database is unreachable, the window lasts as long as the outage.

The association path has the same shape — account limits are read from a cache that also starts empty.

## Expected

A job whose limits cannot be read is not started. Enforcement resumes once the cache loads.

## Actual

`maxjobsperuser`, `maxtresperuser` and the group caps have no effect on jobs admitted during the window, and per-user running counts can exceed the cap by any amount. Nothing surfaces that the caps were skipped: the jobs still display under their QOS in `squeue`, and later submissions are correctly blocked once the cache has loaded, so the queue looks healthy while the running set is over budget.

## Repro (conceptual)

1. On a cluster with accounting enabled, set `maxjobsperuser=2` on a QOS.
2. Queue more than two jobs from one user under that QOS.
3. Restart the controller, or make the accounting database unreachable, so the QOS cache holds no snapshot.
4. Observe that all queued jobs start, past the cap, and are never brought back into compliance.

## Notes

A QOS deleted or renamed while jobs are queued reaches the same limitless state through the loaded-cache path.

Same silent-non-enforcement shape as #693, one layer up: the limit is stored, displayed, and never applied.

nikhilsk:
Fixes #745

## What this fixes

The scheduling gate resolved a job's QOS through the controller's QOS cache and fell back to a **limitless** default whenever the cache could not produce the name:

```rust
Some(name) => self.qos_cache.get(name).unwrap_or_default(),
```

A job carrying a QOS the cache does not hold was therefore gated against no limits at all, and since a running job is never re-checked, it stayed over the cap for its whole run. The association path had the same shape — account limits come from a cache that also starts empty.

Submit-time validation does not close this. It rejects an unknown `--qos` and already fails closed on a cold association cache, but the queue is durable while the caches are not: a controller that restarts or takes over as leader inherits its predecessor's pending jobs and starts scheduling before its first accounting fetch completes. Every pass in that window admitted them unmetered. If the accounting database is unreachable, the window lasts as long as the outage.

## Approach

One `retain_unblocked` pass, `accounting_block`, runs immediately before the existing QOS and account limit gates. When accounting is enabled and a cache holds no snapshot, a job naming a QOS or an account pends with a new `AccountingUnavailable` reason instead of being admitted. A QOS the loaded cache does not know — deleted or renamed after the job queued — pends with `InvalidQOS`, as in Slurm.

`qos_block_with` and `account_block_with` are untouched, so nothing that already enforced correctly changed shape. `resolve_qos` keeps its limitless fallback: its remaining callers compare priority and preempt mode, where a missing QOS costs nothing, and the constraint is now documented on the function.

## The hold is self-clearing, including on a cold start

A fail-closed hold is only acceptable if it recovers on its own. Previously the accounting connect, the notifier, the reconcile/purge loops, and the four cache-refresh loops all sat on the connect-success arm in `main.rs`; the connect-error and migration-error arms spawned nothing and never retried. So a controller that booted while the database was down had `enabled() == true` and `is_loaded() == false` permanently, holding every inherited QOS/account job for the life of the process.

Accounting bring-up now lives in `accounting::start`, which spawns a background task that retries connect + migrate with capped backoff and, on the first success, calls `activate` to install the pool and spawn every loop. `main.rs` is a single flat call. Once the database returns, the caches load and the next scheduling pass admits the held jobs — no operator action. The gRPC `AccountingService` was made recoverable through a shared, swappable pool slot, so `sacct`/`sacctmgr` also come back after a cold-start outage, not just scheduling.

There is deliberately **no manual override**. `AccountingUnavailable` is not added to `is_scheduling_hold()` and there is no opt-out knob: `scontrol release` clearing it would admit jobs against the very limits that could not be read, which is the hazard the hold exists to prevent. The only operator lever is the config — fix `database_url`, or clear it to disable accounting — and both resolve the hold correctly.

## The "can limits be read?" predicate

`accounting_block` is the canonical definition of whether a cache's limits are readable. It is now exposed as two `pub(crate)` helpers on `ClusterManager`, each returning the per-cache `!enabled() || is_loaded()` verdict, and `accounting_block` consumes them:

```rust
pub(crate) fn qos_limits_readable(&self) -> bool {
    !self.config().accounting.enabled() || self.qos_cache.is_loaded()
}
pub(crate) fn association_limits_readable(&self) -> bool {
    !self.config().accounting.enabled() || self.association_cache.is_loaded()
}
```

This lets a limits-readable operator banner (see #746) adopt the same verdict instead of re-deriving a weaker `qos_loaded && assoc_loaded`, which misreports both the accounting-disabled case (no caps exist) and a split cache (one warm, one cold).

## Design choices

- **Fail closed, matching the submit gate.** Admission is the last point at which a cap can still be applied. The trade-off is bounded now that the hold self-clears: with the database unreachable at start, QOS/account jobs pend *until the database returns*, then start on their own. Slurm reaches the pend-not-start outcome by different means — it keeps association state on disk and declines to schedule without it.
- **Accounting-disabled clusters are unaffected.** With an empty `database_url` there are no limits to read, so an empty cache means "no caps exist" rather than "caps are unreadable". A guard test pins this, since getting it wrong would stall every account-scoped job on clusters that run no accounting.
- **`AccountingUnavailable` has no Slurm counterpart**, because Slurm never schedules in this state. Reusing `InvalidQOS` for it would have been misleading: the QOS is fine, it just cannot be read.

## Upgrade note

`PendingReason` travels inside `WalOperation`, so the new variant reaches the Raft log. A new controller replaying an old log is fine; an **old** controller replaying an entry that contains `AccountingUnavailable` would fail to deserialize, so this is forward-only and would bite on a rollback or in a mixed-version HA cluster. This follows the additive pattern the enum already took for the Slurm reason-code parity batch (`Preempted` in #700, `QosGrpWallLimit` in #675) rather than setting a new one.

## Follow-ups

Operators still cannot see per-user usage against a QOS's caps without inferring it from `squeue`; a read-only `scontrol show assoc_mgr` view is the natural next step and is left out of here deliberately. Separately, an association whose `(user, account)` pair is absent from a loaded cache still reads as limitless — the association twin of the `InvalidQOS` case, not addressed here because telling "no limits set" from "no such association" needs more than the current cache API exposes.

## Testing

Through the real scheduling pass and the real bring-up seams:

- the reproduction — three jobs under `maxjobsperuser=2` with a cold cache admit **zero**, where they previously all started;
- the hold reports `AccountingUnavailable` through `tag_blocked_pending_reasons`;
- a deleted QOS reports `InvalidQOS`;
- the association twin holds the same way;
- an accounting-disabled cluster still admits all three on a cold cache;
- a warm cache still enforces the cap at exactly two;
- **self-clearing** — a held job is admitted (under its cap) once the QOS cache reloads;
- the per-cache readable predicate across enabled/disabled and split-cache states;
- `AccountingService` flips from unavailable to serving once the bring-up installs the pool, observed through a clone's shared slot, and reports the latest failure reason while retrying.

`cargo test --locked` (full workspace, plus the ignored PostgreSQL-backed suite), `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` warning-free, `cargo fmt --all` clean. Docs: the "When limits cannot be read" subsection now describes the self-clearing behaviour and the absence of a manual override.

### DIFF
--- a/crates/spur-core/src/job.rs
+++ b/crates/spur-core/src/job.rs
@@ -329,6 +329,12 @@ pub enum PendingReason {
     /// Replaces `BeginTime` as the pending reason for preempted-requeued jobs so the
     /// reason string is unambiguous.
     Preempted,
+
+    /// Accounting is enabled but its caches hold no snapshot, so the job's QOS or
+    /// association limits cannot be read. Has no Slurm counterpart: Slurm keeps its
+    /// association state on disk and refuses to start without it, so it never
+    /// schedules in this state to begin with.
+    AccountingUnavailable,
 }
 
 impl PendingReason {
@@ -417,6 +423,7 @@ impl PendingReason {
             Self::QosMaxSubmitJobPerAccountLimit => "MaxSubmitJobsPerAccount",
             Self::K8sReserved => "ReqNodeNotAvail, Reserved for Kubernetes cluster",
             Self::Preempted => "Preempted",
+            Self::AccountingUnavailable => "AccountingUnavailable",
         }
     }
 
--- a/crates/spur-core/src/job.rs
+++ b/crates/spur-core/src/job.rs
@@ -2346,6 +2346,21 @@ mod tests {
         );
     }
 
+    // The variant reaches the Raft log through `WalOperation`, so its serialized
+    // name is a compatibility surface: renaming or reordering must fail here rather
+    // than on a controller replaying a log it cannot parse.
+    #[test]
+    fn accounting_unavailable_reason_displays_and_roundtrips() {
+        assert_eq!(
+            PendingReason::AccountingUnavailable.display(),
+            "AccountingUnavailable"
+        );
+        let json = serde_json::to_string(&PendingReason::AccountingUnavailable).unwrap();
+        assert_eq!(json, "\"AccountingUnavailable\"");
+        let back: PendingReason = serde_json::from_str(&json).unwrap();
+        assert_eq!(back, PendingReason::AccountingUnavailable);
+    }
+
     #[test]
     fn pending_reason_exit_vocabulary_display() {
         assert_eq!(PendingReason::NonZeroExitCode.display(), "NonZeroExitCode");
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -3315,11 +3315,6 @@ impl ClusterManager {
             }
         });
 
-        let qos_by_job: HashMap<JobId, Qos> = candidates
-            .iter()
-            .map(|candidate| (candidate.job.job_id, self.resolve_qos(&candidate.job)))
-            .collect();
-
         let reservations = self.get_reservations();
         retain_unblocked(&mut candidates, &mut reason_updates, |job| {
             reservation_block(job, &reservations, now)
@@ -3375,9 +3370,23 @@ impl ClusterManager {
             *count += 1;
         }
 
-        retain_unblocked(&mut candidates, &mut reason_updates, |job| {
-            self.accounting_block(job)
+        // Eligible-only: an unreadable cap is transient, not structural, so it must
+        // not claim a begin-deferred job and report itself in place of BeginTime.
+        // Nothing is admitted past the gate either way — ineligible candidates leave
+        // the schedulable set below.
+        retain_eligible(&mut candidates, &mut reason_updates, |job| {
+            match self.accounting_block(job) {
+                Some(reason) => GateOutcome::Block(reason),
+                None => GateOutcome::Keep,
+            }
         });
+
+        // Built after the accounting gate so a held job never gets the limitless
+        // default `resolve_qos` warns about.
+        let qos_by_job: HashMap<JobId, Qos> = candidates
+            .iter()
+            .map(|candidate| (candidate.job.job_id, self.resolve_qos(&candidate.job)))
+            .collect();
 
         {
             let mut reserved = PassReservations::default();
@@ -4798,12 +4807,7 @@ impl ClusterManager {
     /// Holds a candidate whose accounting limits cannot be read. Admission is the
     /// last point at which a cap can still be applied — a running job is never
     /// re-checked — so a cold cache pends the job rather than gating it against a
-    /// limitless default. Mirrors the submit-time fail-closed gate; with accounting
-    /// disabled there are no limits to read and nothing is held.
-    ///
-    /// The window is real despite submit validating the QOS: the queue is durable
-    /// and the caches are not, so a restarted controller inherits its predecessor's
-    /// pending jobs before its first fetch completes.
+    /// limitless default.
     fn accounting_block(&self, job: &Job) -> Option<PendingReason> {
         if !self.config().accounting.enabled() {
             return None;
@@ -17544,8 +17548,8 @@ mod tests {
     }
 
     fn admitted_of(cm: &ClusterManager, ids: &[JobId]) -> usize {
-        let pending: Vec<JobId> = cm.pending_jobs().iter().map(|j| j.job_id).collect();
-        ids.iter().filter(|id| pending.contains(id)).count()
+        let admitted: Vec<JobId> = cm.pending_jobs().iter().map(|j| j.job_id).collect();
+        ids.iter().filter(|id| admitted.contains(id)).count()
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
@@ -17585,6 +17589,33 @@ mod tests {
         assert_eq!(
             cm.get_job(ids[0]).unwrap().pending_reason,
             PendingReason::AccountingUnavailable
+        );
+    }
+
+    // A job waiting on its start time is not waiting on the cache: the cold-cache
+    // hold gates admission, and a begin-deferred job is not up for admission, so
+    // BeginTime stays its reason.
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn cold_cache_does_not_claim_a_begin_deferred_job() {
+        let dir = TempDir::new().unwrap();
+        let cm = accounting_cluster(&dir).await;
+        register_node(&cm, "n1", 64, 128000);
+        cm.qos_cache().insert(Qos {
+            name: "cnt".into(),
+            ..Default::default()
+        });
+
+        let mut spec = basic_spec("later");
+        spec.qos = Some("cnt".into());
+        spec.begin_time = Some(Utc::now() + chrono::Duration::hours(1));
+        let id = submit_and_wait(&cm, spec);
+
+        cm.qos_cache().reset();
+        cm.refresh_pending_reasons();
+
+        assert_eq!(
+            cm.get_job(id).unwrap().pending_reason,
+            PendingReason::BeginTime
         );
     }
 
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -3375,6 +3375,10 @@ impl ClusterManager {
             *count += 1;
         }
 
+        retain_unblocked(&mut candidates, &mut reason_updates, |job| {
+            self.accounting_block(job)
+        });
+
         {
             let mut reserved = PassReservations::default();
             let grp_wall_usage = self.grp_wall_cache.usage();
@@ -4754,6 +4758,9 @@ impl ClusterManager {
     }
 
     /// Resolve a job's QoS from the cache; unknown/absent name → limitless default.
+    /// Callers that gate admission must go through `accounting_block` first: the
+    /// limitless default is only safe where a missing QOS costs nothing (priority
+    /// and preemption comparisons), never where a limit is being applied.
     pub(crate) fn resolve_qos(&self, job: &Job) -> Qos {
         match job.spec.qos.as_deref() {
             Some(name) => self.qos_cache.get(name).unwrap_or_default(),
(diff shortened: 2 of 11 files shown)


crates/spurctld/src/cluster.rs
```diff
@@ -3301,6 +3301,10 @@ impl ClusterManager {
             }
         }
 
+        retain_unblocked(&mut candidates, &mut blocked, |job| {
+            self.accounting_block(job)
+        });
```
shiv-tyagi: `qos_by_job` is built from `resolve_qos` a few lines above, before this pass runs. Jobs that will be held still get a limitless `Qos` in the map, then drop out of `candidates` before `qos_block_with` sees them. Correct today, but the ordering works against the new doc contract on `resolve_qos`.

Move this `retain_unblocked` pass to just before the `qos_by_job` build so limitless defaults are never computed for jobs that fail the accounting gate.

shiv-tyagi (commented):
Posted comments, PTAL

crates/spurctld/src/cluster.rs
```diff
@@ -4653,13 +4657,49 @@ impl ClusterManager {
...
 
+    /// Holds a candidate whose accounting limits cannot be read. Admission is the
+    /// last point at which a cap can still be applied — a running job is never
+    /// re-checked — so a cold cache pends the job rather than gating it against a
+    /// limitless default. Mirrors the submit-time fail-closed gate; with accounting
+    /// disabled there are no limits to read and nothing is held.
+    ///
+    /// The window is real despite submit validating the QOS: the queue is durable
+    /// and the caches are not, so a restarted controller inherits its predecessor's
+    /// pending jobs before its first fetch completes.
```
shiv-tyagi: The `accounting_block` doc comment repeats the PR description (submit validates QOS, queue is durable, caches are not). The one non-obvious constraint worth keeping is that a running job is never re-checked, so admission is the last enforcement point. The rest can go; the code reads clearly without it.

crates/spur-core/src/job.rs
```diff
@@ -328,6 +328,12 @@ pub enum PendingReason {
     /// Replaces `BeginTime` as the pending reason for preempted-requeued jobs so the
     /// reason string is unambiguous.
     Preempted,
+
+    /// Accounting is enabled but its caches hold no snapshot, so the job's QOS or
+    /// association limits cannot be read. Has no Slurm counterpart: Slurm keeps its
+    /// association state on disk and refuses to start without it, so it never
+    /// schedules in this state to begin with.
+    AccountingUnavailable,
```
shiv-tyagi, line 337: `AccountingUnavailable` lands in `PendingReason`, which serializes into `WalOperation`. The PR calls out the forward-only upgrade risk, but there is no serde roundtrip test for this variant. `Preempted` has a dedicated test (`preempted_reason_displays_correctly_and_explains_begin_hold`). Add the same for `AccountingUnavailable` so a renamed or reordered variant fails CI before it hits the Raft log.

```rust
#[test]
fn accounting_unavailable_reason_displays_and_roundtrips() {
    assert_eq!(
        PendingReason::AccountingUnavailable.display(),
        "AccountingUnavailable"
    );
    let json = serde_json::to_string(&PendingReason::AccountingUnavailable).unwrap();
    let back: PendingReason = serde_json::from_str(&json).unwrap();
    assert_eq!(back, PendingReason::AccountingUnavailable);
}
```

docs/admin-guide/accounting.rst
```diff
@@ -564,6 +564,34 @@ Set ``DenyOnLimit`` through the QOS ``flags`` key:
...
+``AccountingUnavailable`` and start normally once the cache loads. The
+alternative is worse: a running job is never re-checked against limits, so a job
+admitted during that window stays over the cap for its entire run.
+
+Two related cases:
+
+- A job whose QOS no longer exists — deleted or renamed while the job sat in the
+  queue — pends with ``InvalidQOS``, as in Slurm.
+- With accounting disabled (an empty ``database_url``) there are no limits to
+  read, so an empty cache means "no caps exist" and nothing is held.
```
shiv-tyagi: This section covers QOS and association caches. `GrpWallCache` has the same cold-start shape: reads return `None` until the first successful load, so `QOSGrpWallLimit` is unapplied while that cache is empty even if the QOS definition cache is warm. Out of scope for this PR, but worth a one-line pointer under "Two related cases" so operators do not assume all accounting limits fail closed on restart.

crates/spurctld/src/cluster.rs
nikhilsk: Moved the map rather than the gate: `qos_by_job` is read only by the consumables pass below, so building it after `accounting_block` means a held job never receives the limitless default. Moving the gate up would also have reordered it ahead of `reservation_block` and changed which reason a job blocked by both reports, which I did not want to fold in here.

The gate itself is now eligible-only, which also fixes begin-deferred jobs reporting `AccountingUnavailable` in place of `BeginTime`.

crates/spurctld/src/cluster.rs
nikhilsk: Trimmed to the running-job constraint; the PR-description restatement is gone.

crates/spur-core/src/job.rs
nikhilsk: Added, close to your version, plus an assertion on the serialized string itself so a rename fails on the wire name and not only on `display()`. The operator-visible half of that risk is now in the upgrade guide as a roll-forward constraint.

docs/admin-guide/accounting.rst
nikhilsk: Added, with a pointer to the GrpWall section. Confirmed the shape you describe: `GrpWallCache::usage()` returns `None` before the first load, so a `grpwall` cap is unapplied until a figure is read - it fails open where this hold fails closed.

crates/spurctld/src/cluster.rs
```diff
@@ -4653,13 +4666,44 @@ impl ClusterManager {
...
             None => Qos::default(),
         }
     }
 
+    /// Holds a candidate whose accounting limits cannot be read. Admission is the
+    /// last point at which a cap can still be applied — a running job is never
+    /// re-checked — so a cold cache pends the job rather than gating it against a
+    /// limitless default.
+    fn accounting_block(&self, job: &Job) -> Option<PendingReason> {
+        if !self.config().accounting.enabled() {
```
yansun1996, line 4828: This is the predicate #746 needs. Its `limits_readable` is `qos_loaded && assoc_loaded` — no `enabled()` guard, and conjoined rather than per-cache — so it diverges from this in both the accounting-disabled case and the split-cache case. Since this is where the semantics are defined, could the check be exposed (even just as a small `pub(crate)` helper returning the `!enabled() || loaded` verdict per cache) so #746 consumes it rather than re-deriving a weaker version?

docs/admin-guide/accounting.rst
```diff
@@ -564,6 +564,38 @@ Set ``DenyOnLimit`` through the QOS ``flags`` key:
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Limits are served from caches that ``spurctld`` refreshes from the accounting
+database, and a freshly started controller has none until its first fetch
+completes. The job queue, by contrast, is durable — a controller that restarts or
+takes over as leader inherits its predecessor's pending jobs immediately.
+
+While accounting is enabled and a cache holds no snapshot, jobs that name a QOS
+or an account are **not scheduled**. They pend with reason
+``AccountingUnavailable`` and start normally once the cache loads. The
```
yansun1996: This holds for a mid-life outage but not for a cold start. If the database is unreachable when `spurctld` starts, neither `spawn_refresh_loop` call is ever made (`main.rs:236`/`:241` sit on the success path only, and the two error arms don't retry), so the cache never loads and these jobs never start. Given this section is specifically about restart behaviour, that's the case a reader is most likely to arrive with — worth either stating the limitation or, better, making the startup path retry.

yansun1996 (changes requested):
This is the right fix and the right place for it — admission is genuinely the last point a cap can be applied, and gating per-cache rather than on a single global "accounting ready" flag is a sharper choice than it first looks. Two things before merge.

**The hold isn't always self-clearing.** `enabled()` is `!database_url.is_empty()` — it reports *configuration*, not whether a refresher is running. The refresh loops spawn only on the success path (`main.rs:236`, `:241`); the migration-failure arm (`:201`) and the connect-error arm (`:255`) spawn nothing and never retry. So if the database is down when the controller starts, `enabled()` is permanently true and `is_loaded()` permanently false: every inherited pending job holds for the lifetime of the process. A mid-life outage is fine — the loop is already running and recovers. It's startup specifically.

There's also no escape hatch. `release_job` only clears reasons in `is_scheduling_hold()`, which doesn't include `AccountingUnavailable`, so `scontrol release` answers "job N is not held". And there's no config knob to opt out. That makes restart-after-outage — the exact scenario the new doc section is written about — a full queue stall with no operator recourse.

**`accounting_block` is the canonical "can limits be read?" predicate, and #746 has a different one.** #746 computes `LimitsReadable` as `qos_loaded && assoc_loaded` with no accounting-disabled guard. Against this PR that's wrong twice over: with accounting disabled it prints `LimitsReadable=NO` forever, where this PR correctly rules "no caps exist"; and on a split cache it reports a flat `NO` while jobs visibly run under enforced caps, where this PR holds only the affected jobs. This PR's follow-ups name `scontrol show assoc_mgr` as the next step, but #746's banner has no notion of `AccountingUnavailable` — so the one operator-facing surface can't show *why* jobs are held. No textual merge conflict; different hunks. Worth landing the predicate here and having #746 adopt it.

On the earlier review rounds: both of Copilot's points and four of shiv's read as already fixed at head (`002edf80` → `24a00bf0`). I'd also push back on the `!` breaking-marker request — `Preempted` (#700) and `QosGrpWallLimit` (#675) both shipped additively without one, so a new `PendingReason` variant looks like established precedent here.

crates/spurctld/src/cluster.rs
nikhilsk: Done — exposed as the canonical verdict rather than left implicit. `cluster.rs`
now has two `pub(crate)` helpers next to `accounting_block`:

```rust
pub(crate) fn qos_limits_readable(&self) -> bool {
    !self.config().accounting.enabled() || self.qos_cache.is_loaded()
}
pub(crate) fn association_limits_readable(&self) -> bool {
    !self.config().accounting.enabled() || self.association_cache.is_loaded()
}
```

`accounting_block` consumes them, so the gate and the helper can't drift. #746
can drop its `qos_loaded && assoc_loaded` and use
`qos_limits_readable() && association_limits_readable()`; that fixes both cases
you flagged — accounting disabled (each helper is `true`, so the banner reads
readable instead of a permanent `NO`) and a split cache (per-cache, so a warm
QOS cache doesn't get vouched for by a cold association cache or vice versa). I
left #746's worktree untouched, as agreed; this just lands the predicate.

docs/admin-guide/accounting.rst
nikhilsk: Fixed the startup path rather than only documenting the limitation. The four
`spawn_refresh_loop` calls (plus the notifier and the reconcile/purge loops) no
longer sit on the connect-success arm in `main.rs`. Bring-up now lives in
`accounting::start`, which spawns a background task that retries connect +
migrate with capped backoff and calls `activate` — installing the pool and
spawning every loop — on the first success. So a controller that boots with the
database down converges on its own: once the database returns, the caches load
and the next scheduling pass admits the held jobs.

The doc section is updated to describe exactly that (self-clearing including on
a cold start), and to state plainly that there is no manual override and why.

shiv-tyagi (dismissed):
LGTM

yansun1996 (dismissed):
Thanks — all the threads check out against the current code. Fixing the cold-start path rather than just documenting it was the right call: the refresh loops no longer sit behind a successful initial connect, so a controller that starts with the database unreachable converges once it comes back. Approving.

sgopinath1 (approved):
LGTM

PR #746, opened 2026-08-25
feat(spur-cli): add scontrol show assoc_mgr for limits against live usage

nikhilsk:
## What this adds

`scontrol show assoc_mgr`: how much of a QOS or association each user is holding right now, next to the caps that govern them, and which caps are **already exceeded**.

Operators had no way to see this. The figures existed only inside the admission gate, which recomputes a user's running count and TRES per candidate job and keeps nothing queryable, so counting `squeue` rows was the only recourse — and that cannot show a cap already being exceeded. Usage over a cap is not a contradiction: caps apply when a job is admitted and running jobs are never re-checked, so tightening a cap under running work leaves exactly that state, silently.

```text
QOS Records
QOS=highprio MaxWall=01:00:00 MaxJobsPU=2 MaxSubmitJobsPU=N MaxTRESPU=node=4
   GrpJobs=N(9) GrpSubmitJobs=N(11) GrpTRES=cpu=N(36),node=16(9)
   User=alice MaxJobsPU=2(6) MaxSubmitJobsPU=N(7) MaxTRESPU=cpu=N(24),node=4(6) OverLimit=MaxJobsPU,MaxTRESPU
   User=bob MaxJobsPU=2(1) MaxSubmitJobsPU=N(1) MaxTRESPU=cpu=N(4),node=4(1)

Association Records
Account=tenant-a MaxWall=N
   GrpJobs=N(1) GrpSubmitJobs=N(1) GrpTRES=node=N(1)
   User=alice MaxJobs=4(1) MaxSubmitJobs=N(1) MaxTRES=cpu=N(4),node=N(1)
```

## Approach

A new `GetAssocMgrInfo` RPC on `SlurmController` — the controller is the only side that sees the live job table, while the accounting service owns the definitions. Aggregation runs per `(user, QOS)` and `(user, account)` through the same `sum_running_tres` the gate uses, so node figures are distinct-node occupancy and CPU follows each job's request. A record reads as the scheduler sees it rather than as a second opinion that could quietly disagree.

The cap comparison lives in `spur-core` beside the limits, not in the CLI, so the view does not re-implement enforcement semantics (including that a TRES dimension capped at `0` is not a cap). It sits next to the admission checks it deliberately differs from: those project a candidate job onto current usage, this reports what already stands over a cap.

## Following Slurm's shape

The output follows Slurm's `assoc_mgr` where that shape is a contract other tooling already depends on:

- **`Limit(Consumed)` in one field, `N` for no limit** — `node=4(6)` is a cap of four with six in use. Scripts in the wild parse exactly this (splitting on the parenthesis), so rendering caps and usage as separate fields would have broken them for no gain. Applied per count and per TRES dimension, listing the union of dimensions capped and in use.
- **Every defined QOS and association is listed**, not only those with queued work, which is what Slurm's cache dump does. An over-tight cap on an idle QOS is exactly the kind of mistake that hides otherwise. This needed enumerate-all accessors on both caches. A QOS deleted after its jobs queued also stays reported while those jobs hold resources.
- **Per-user entries nest under the record they qualify**, as Slurm's `User Limits=` does. This also stops a scope's group figures from being repeated once per user, which the first draft did.

Where it diverges, deliberately: users are listed on their own lines rather than inside a single `User Limits=` field (more readable and greppable), an account's group figures are stated once for the account instead of per association row, and Slurm's `accounts=`/`qos=`/`flags=` selectors and fairshare internals (`SharesRaw`, `UsageRaw`, `Lft-Rgt`, `GrpTRESMins`) are not included. All of this is in the migration guide.

## Design choices

- **A QOS states its per-user caps once, on the scope line.** It caps every user identically, so repeating them per user is noise — and stating them on the scope is what keeps them visible when nobody is using the QOS. An association's caps are per `(user, account)`, so they ride on each user's line instead, and `scope_caps` is absent there.
- **Caps and consumption stay separate fields on the wire.** Composing the `Limit(Consumed)` display is the client's job; a machine consumer should not have to take a display string apart again.
- **`Cap` is spelled neutrally and the server picks the wording.** A QOS caps with `MaxJobsPU`/`MaxTRESPU`, an association with plain `MaxJobs` — same breach, different name. Deciding it controller-side means every consumer, REST included, reads one verdict.
- **A cold accounting cache reports `LimitsReadable=NO`** rather than letting an unreadable cap look like an absent one. Usage stays accurate; only the caps are missing.
- **Reused `sacctmgr`'s `blank_if_unset`** (promoted to `pub(crate)`) instead of copying it, so a cap renders identically in both commands.

## Proto

Additive only — a new RPC plus `AssocMgrRecord`, `AssocMgrUserRecord`, `AssocMgrCaps`, `GetAssocMgrInfoRequest`, `GetAssocMgrInfoResponse`. No existing message, field, or tag is touched.

## Testing

Twenty new tests:

- **`spur-core` (7)** — the cap comparison and `TresRecord::types()`: over the job-count cap, exactly at a cap (a ceiling is not a threshold), unset caps ignored, each TRES dimension compared the way the gate compares it, a zero dimension treated as uncapped, group caps reported from the scope and never duplicated onto users, and dimension ordering stable.
- **`spurctld` (9)** — through the real cluster manager with jobs driven to running: per-user records against a QOS cap with the breach named, a QOS nobody is using still reported with its caps, a QOS the cache no longer knows kept while its jobs run, distinct-node counting, one-user filtering that leaves group totals intact, usage reported when caps could not be read, QOS and account records kept apart with each hierarchy's own caps, a defined association with no jobs still reported, and nothing at all for unscoped work.
- **`spur-cli` (7)** — the rendered layout: `Limit(Consumed)` with `N`, association labels without the per-user suffix, an idle scope with caps and no users, `OverLimit` omitted when compliant, a group breach on the scope line, an explicitly marked empty section, a zero cap rendering as `0`, and TRES dimensions drawn from either side.

Full `cargo test --locked` (3,318 tests), `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` warning-free, `cargo fmt --all -- --check` clean. Docs: a new section in the monitoring guide and a compatibility row in the migration guide.


crates/spur-core/src/accounting.rs
```diff
@@ -94,6 +94,20 @@ impl TresRecord {
...
+    pub fn types(&self) -> Vec<TresType> {
+        let mut types: Vec<TresType> = self
+            .values
+            .iter()
+            .filter(|(_, v)| **v > 0)
+            .map(|(k, _)| *k)
+            .collect();
+        types.sort_by_key(|t| t.name());
+        types
+    }
```
shiv-tyagi: The doc comment above `types()` still opens with `Format as "cpu=N,mem=N,gres/gpu=N" string.`, which belongs to `format()`. `format()` now has no doc comment, and rustdoc for `types()` is wrong.

Split the docs so each function has its own comment block.

crates/spurctld/src/cluster.rs
```diff
@@ -4652,6 +4654,98 @@ impl ClusterManager {
...
+                    record.grp_submit_jobs = limits.grp_submit_jobs;
+                }
+                record
+            })
+            .collect();
+
+        AssocMgrInfo {
+            qos_records,
+            assoc_records,
+            limits_readable: self.qos_cache.is_loaded() && self.association_cache.is_loaded(),
```
shiv-tyagi: `limits_readable` is `qos_cache.is_loaded() && association_cache.is_loaded()`, but caps are read from each cache independently. If the QOS cache is loaded and the association cache is not, QOS caps still populate while the CLI prints `LimitsReadable=NO` and says no cap could be read. That overstates the problem for half-loaded clusters.

Consider per-section readability flags, or only setting `limits_readable = false` when a cache that contributes caps to the response is unloaded.

crates/spurctld/src/cluster.rs
```diff
@@ -4652,6 +4654,98 @@ impl ClusterManager {
...
+                );
+                // The group caps are the account's, but the schema keys every
+                // association row separately, so take them from the first row
+                // rather than inventing a merge across rows that disagree.
+                if let Some((user, _)) = defined_assoc.get(&scope).and_then(|rows| rows.first()) {
+                    let limits = self.association_cache.limits(user, &scope);
+                    record.max_wall_minutes = limits.max_wall_minutes;
+                    record.grp_tres = limits.grp_tres;
+                    record.grp_submit_jobs = limits.grp_submit_jobs;
+                }
```
shiv-tyagi, line 4850: Account group caps (`grp_tres`, `grp_submit_jobs`, `max_wall_minutes`) are taken from the first `(user, account)` row. Association limits are stored per row in the DB, so rows can disagree if data was edited out of band.

Q: Is `grp_tres` guaranteed identical across all users under an account in practice? If not, this can report the wrong group cap for the scope.

crates/spurctld/src/cluster.rs
```diff
@@ -4652,6 +4654,98 @@ impl ClusterManager {
...
+
+        let qos_records = qos_scopes
+            .into_iter()
+            .map(|scope| {
+                let limits = defined_qos.get(&scope).map(|q| q.limits.clone());
+                // A QOS holds one set of per-user caps for everyone it governs.
+                let user_caps = limits.as_ref().map(|l| PerUserCaps {
+                    max_jobs: l.max_jobs_per_user,
+                    max_submit_jobs: l.max_submit_jobs_per_user,
+                    max_tres: l.max_tres_per_user.clone(),
```
shiv-tyagi, line 4791: `max_submit_jobs_per_account` is enforced in `check_qos_submit_limits`, but it is not surfaced in the QOS scope record. Operators can see `MaxSubmitJobsPU` and `GrpSubmitJobs`, but not the per-account submit cap within a QOS.

Worth a follow-up, or a note in the migration guide if this is intentionally out of scope for v1.

shiv-tyagi (commented):
Posted comments, PTAL

One additional note (no line in this diff): `GetAssocMgrInfo` should be added to the diagnostic RPC exclusion list in `crates/spurctld/src/rpc_middleware.rs` alongside `GetSchedStats`, or operator polling of `scontrol show assoc_mgr` will skew RPC latency stats.

crates/spurctld/src/cluster.rs
```diff
@@ -10599,6 +10787,276 @@ mod tests {
...
+            limits,
+            ..Default::default()
+        }
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn assoc_mgr_info_reports_each_user_against_the_qos_cap() {
+        // The SPUR-205 shape: a user over MaxJobsPU with the cap in force. The
+        // record has to name the breach, since nothing else surfaces it once the
+        // jobs are running.
```
shiv-tyagi: Nit: drop the `SPUR-205` ticket reference from the test comment. Git history tracks that context; inline ticket refs go stale.

```suggestion
        // A user over MaxJobsPU with the cap in force. The record has to name
        // the breach, since nothing else surfaces it once the jobs are running.
```

crates/spur-cli/src/scontrol.rs
```diff
@@ -871,9 +872,30 @@ async fn show(controller: &str, entity: &str, name: Option<&str>) -> Result<()>
...
+                );
+                println!();
+            }
+            print!("{}", render_assoc_mgr(QOS_SECTION, &resp.qos_records));
+            print!("{}", render_assoc_mgr(ASSOC_SECTION, &resp.assoc_records));
+        }
         other => {
             bail!(
-                "scontrol: unknown entity type '{}'. Use: job, node, partition, reservation, federation, config",
+                "scontrol: unknown entity type '{}'. Use: job, node, partition, reservation, assoc_mgr, federation, config",
```
shiv-tyagi, line 843: The clap field doc for `Show.entity` (line 33) still lists `job, node, partition, config`, but this error path now includes `assoc_mgr`. Update the clap doc on line 33 to match so `--help` is accurate:

`/// Entity type: job, node, partition, reservation, assoc_mgr, federation, config`

crates/spur-cli/src/scontrol.rs
```diff
@@ -899,6 +921,137 @@ async fn ping(controller: &str) -> Result<()> {
...
+fn tres_limit_consumed(cap: &str, used: &str) -> String {
+    let cap = TresRecord::parse(cap).unwrap_or_default();
+    let used = TresRecord::parse(used).unwrap_or_default();
+    let mut dimensions = cap.types();
+    for dimension in used.types() {
+        if !dimensions.contains(&dimension) {
+            dimensions.push(dimension);
+        }
+    }
+    dimensions.sort_by_key(|t| t.name());
```
shiv-tyagi, line 962: Nit: `tres_limit_consumed` builds the dimension union with `Vec` + `contains`, which is O(n²). A `BTreeSet` (already used elsewhere in this PR) would be cleaner and keeps sort order for free.

```suggestion
    let mut dimensions: std::collections::BTreeSet<_> = cap.types().into_iter().collect();
    dimensions.extend(used.types());
```

crates/spurctld/src/server.rs
```diff
@@ -1586,6 +1586,38 @@ impl SlurmController for ControllerService {
...
+    ) -> Result<Response<GetAssocMgrInfoResponse>, Status> {
+        // Usage is read from the leader's job table, the same one admission reads.
+        if self.check_leader(&request).is_err() {
+            let proxy = &self.leader_proxy;
+            let mut client = proxy.get_leader_client().await?;
+            let fwd = Self::forward_request(request);
+            return client.get_assoc_mgr_info(fwd).await;
+        }
+
+        let user = request.into_inner().user;
```
yansun1996: **Authorization — I think this needs to change before merge.**

This handler never consults the caller's identity. `user` arrives from the request and goes straight into `assoc_mgr_info` as a filter, and `.filter(|u| !u.is_empty())` means an empty string is "no filter" — so an unauthenticated or non-admin caller gets every user's live job counts and TRES holdings.

Two things make it broader than a filter bug:

1. `auth_middleware.rs` only *authenticates* — its module doc says handlers "read it instead of trusting a client-supplied `user`/`caller` field", and calls out that the accounting service "has no authorization of its own". So there's no outer gate to fall back on; authz here is strictly per-handler.
2. `scope_usage` (`cluster.rs:6984`) applies `only_user` to the `users` list **only**. The scope lines — `grp_running_jobs`, `grp_submitted_jobs`, `grp_running_tres`, every QOS and account name, `MaxWall`/`GrpTRES`/`GrpSubmitJobs` — are never narrowed. So even a filtered call still enumerates cluster-wide account/QOS inventory.

The repo already has the shape for this: `get_jobs` (`:797-799`) does

```rust
let __identity = Self::verified_identity(&request).cloned();
Self::authoritative_user(&mut req.user, __identity.as_ref());
```

and `list_tokens` gates enumeration on `require_admin` (`:1935`). There's also the note on `is_k0s_admin` (`:675-686`) about removing the old `caller.is_empty()` bypass because "it granted admin to a two-character forgery" — same shape of hole.

Worth flagging: a blanket `require_admin` would be the wrong fix, since under the default `permissive` mode `identity == None` and the existing convention (`viewer_is_privileged`, `:3873-3880`) deliberately treats that as privileged. The `get_jobs` shape — scope a non-admin to themselves, let admins pass an arbitrary user — seems like the one that fits. Would you be up for that here?

proto/slurm.proto
```diff
@@ -709,6 +714,74 @@ message SchedStats {
...
+  uint32 max_jobs = 5;
+  uint32 max_submit_jobs = 6;
+  string max_tres = 7;
+  // Per-user caps this user is already over, named as `sacctmgr` names them (e.g.
+  // "MaxJobsPU"). Decided controller-side so every consumer reads the same
+  // verdict. Empty when everything is within its cap.
+  repeated string over_limit = 8;
+}
+
+// A QOS or an association with the caps and consumption belonging to the scope
```
yansun1996: I think this comment block has drifted onto the wrong message. Lines 738-743 describe a scope "plus the users holding work under it" and "A scope with no queued work carries no user records" — that's `AssocMgrRecord` (line 755), which currently has no comment at all. The block that genuinely belongs to `AssocMgrCaps` starts at 744 ("Caps a scope applies to each of its users…").

Reading it top-down the two halves also contradict each other on the same point — 741-743 and 746-748 both explain the idle-QOS case, once for each message.

Would you mind splitting it, moving 738-743 down to sit above `AssocMgrRecord`? It's the longest comment block in the diff, so it's the one most likely to be trusted verbatim by the next reader.

crates/spur-cli/src/scontrol.rs
```diff
@@ -899,6 +921,137 @@ async fn ping(controller: &str) -> Result<()> {
...
+        ));
+        // A scope that caps every user the same way says so once, here, so the caps
+        // stay visible with nobody using it. An association has no such caps; its
+        // users carry their own.
+        if let Some(caps) = &r.scope_caps {
+            out.push_str(&format!(
+                " MaxJobs{pu}={} MaxSubmitJobs{pu}={} MaxTRES{pu}={}",
+                cap_or_n(caps.max_jobs),
+                cap_or_n(caps.max_submit_jobs),
+                caps.max_tres,
```
yansun1996: `caps.max_tres` is pushed bare here while its two siblings on the same line go through `cap_or_n`. When the scope defines no per-user TRES cap this renders `MaxTRESPU=` with nothing after the `=`, next to `MaxJobsPU=N`.

Since the docs sell `N` as "no cap" and describe the layout as a contract that "real scripts parse", the empty field is probably worth avoiding — a parser splitting on `=` gets an empty token here but `N` everywhere else. Would `cap_or_n`-equivalent handling for the TRES string work?

crates/spur-cli/src/scontrol.rs
```diff
@@ -871,9 +872,30 @@ async fn show(controller: &str, entity: &str, name: Option<&str>) -> Result<()>
                 }
             }
         }
+        "assoc_mgr" | "assocmgr" => {
+            // Slurm's filter syntax is `users=<name>`; a bare name means the same.
+            let user = name
```
yansun1996: `strip_prefix("users=").unwrap_or(n)` means anything that *isn't* `users=` is silently taken as a bare username. So `scontrol show assoc_mgr qos=highprio` doesn't error — it quietly filters for a user literally named `qos=highprio` and prints an empty result, which reads as "no such usage" rather than "wrong flag".

The migration guide's wording is that selectors like this "are not accepted", which I read as promising an error. Would it be worth rejecting any `<key>=` token that isn't `users=`, and keeping the bare-name path for everything else?

docs/user-guide/monitoring-jobs.rst
```diff
@@ -390,6 +391,68 @@ The accounting database keeps the record permanently. In practice
...
+
+   Association Records
+   Account=tenant-a MaxWall=N
+      GrpJobs=N(1) GrpSubmitJobs=N(1) GrpTRES=node=N(1)
+      User=alice MaxJobs=4(1) MaxSubmitJobs=N(1) MaxTRES=cpu=N(4),node=N(1)
+
+Reading it:
+
+* Every cap is printed as **``Limit(Consumed)``**, as in Slurm: ``node=4(6)`` is a
+  cap of four nodes with six in use, and ``N`` marks no cap. A literal ``0`` is a
```
yansun1996: This rule holds for the count caps but splits for TRES. `MaxJobsPU=0` really does block everything it governs, but a TRES dimension set to `0` is skipped by `types()` (it only yields dimensions with a value `> 0`), so it's treated as *unset* — and then prints as `N`, i.e. exactly the "no cap" marker this sentence contrasts it against.

So an operator who sets `GrpTRES=node=0` intending "block all" gets `node=N` back and no enforcement. Could this sentence be narrowed to the count caps, with a note on the TRES behaviour? (Or if `0` is meant to be a real TRES cap, that's a code fix rather than a docs one.)

docs/user-guide/monitoring-jobs.rst
```diff
@@ -390,6 +391,68 @@ The accounting database keeps the record permanently. In practice
...
+      User=bob MaxJobsPU=2(1) MaxSubmitJobsPU=N(1) MaxTRESPU=cpu=N(4),node=4(1)
+
+   Association Records
+   Account=tenant-a MaxWall=N
+      GrpJobs=N(1) GrpSubmitJobs=N(1) GrpTRES=node=N(1)
+      User=alice MaxJobs=4(1) MaxSubmitJobs=N(1) MaxTRES=cpu=N(4),node=N(1)
+
+Reading it:
+
+* Every cap is printed as **``Limit(Consumed)``**, as in Slurm: ``node=4(6)`` is a
```
yansun1996: Small overstatement: "every cap" isn't quite it. `MaxWall` prints bare (`MaxWall=01:00:00`), and so do the per-user caps on the scope line — `MaxJobsPU=2 MaxSubmitJobsPU=N MaxTRESPU=node=4` in the sample output just above. The `Limit(Consumed)` shape applies to the `Grp*` line and the `User=` lines.

Since the sample output right above the sentence already shows both forms, would it be worth saying "every cap with consumption beside it" or naming the two lines explicitly?

crates/spur-core/src/accounting.rs
```diff
@@ -283,6 +410,114 @@ mod tests {
...
+                running_jobs: 12,
+                ..Default::default()
+            }],
+            ..Default::default()
+        };
+        assert_eq!(
+            usage.exceeded_caps(),
+            vec![Cap::GrpTres, Cap::GrpSubmitJobs]
+        );
+        assert!(usage.users[0].exceeded_caps().is_empty());
```
yansun1996, line 526: This assertion can't fail as written. `users[0]` is built with `running_jobs: 12, ..Default::default()`, so `max_jobs` / `max_submit_jobs` / `max_tres` are all `None` — `exceeded_caps()` returns empty for *any* implementation, including one that wrongly duplicated the group caps onto users.

The test name promises the interesting property ("…covers the group caps only"), so it'd be nice if it could catch a regression. Giving the user a cap it is *not* over — say `max_jobs: Some(20)` against `running_jobs: 12` — would make the empty result meaningful: it then proves the group caps weren't leaked onto the user, rather than proving the user had no caps to report.

crates/spurctld/src/cluster.rs
```diff
@@ -4652,6 +4654,98 @@ impl ClusterManager {
...
+                    .unwrap_or_default();
+                // An association carries no per-user TRES cap, so `max_tres` stays unset.
+                let mut record = scope_usage(
+                    &jobs,
+                    &scope,
+                    account_of,
+                    only_user,
+                    &defined_users,
+                    |user| {
+                        let limits = self.association_cache.limits(user, &scope);
```
yansun1996: `association_cache.all()` already handed back the `AccountLimits` for each (account, user) — they're sitting in `defined_assoc` from line 4697. But 4709-4711 drops them (`.map(|(user, _)| user.clone())`), then this closure re-fetches per user via `limits()`, which takes a fresh `snapshot.read()` each time. Line 4732 does it once more, destructuring `(user, _)` while holding the limits it's about to look up again.

Given `assoc_mgr_info` is already holding the `jobs` read lock across this whole stretch, using the limits already in hand seems like a small, contained win. Not a correctness issue as far as I can tell — just redundant work.

crates/spur-cli/src/scontrol.rs
```diff
@@ -1724,6 +1877,135 @@ mod tests {
...
+            }],
+            over_limit: Vec::new(),
+        }
+    }
+
+    #[test]
+    fn assoc_mgr_renders_caps_beside_what_is_consumed() {
+        // Slurm's assoc_mgr shape: cap(consumed) in one field, N for no cap. Real
+        // scripts parse this, so the layout is a contract.
+        let out = render_assoc_mgr(QOS_SECTION, &[assoc_mgr_record()]);
```
yansun1996, line 2262: All the assoc_mgr CLI tests call `render_assoc_mgr` directly, which is great for the layout contract but means the `"assoc_mgr" | "assocmgr"` arm in `show()` is never exercised. Three behaviours live only in that arm and are currently untested: the `users=` strip (line 877), the `LimitsReadable=NO` banner, and the ordering of the QOS section before the association section.

`show()` needs a client so it's awkward to test as-is — would pulling the selector parse into a small helper (e.g. `fn assoc_mgr_user_filter(name: Option<&str>) -> String`) be reasonable? That'd make the `users=` case, the bare-name case and the `qos=` case above unit-testable without a server. Also worth an assertion somewhere that `limits_readable == true` suppresses the banner — right now only the `false` path is described.

yansun1996 (changes requested):
Nice feature — the operator gap it closes is real, and computing the verdict controller-side next to the admission checks (rather than re-deriving it in the CLI) is the right call. A few things worth a look before merge.

**Authorization on the new RPC.** `GetAssocMgrInfo` doesn't consult the caller's identity — the request's `user` is used purely as a filter, and an empty value means "no filter". The scope lines (`Grp*`, plus every QOS and account name) aren't narrowed by the filter at all, so any caller can enumerate cluster-wide account/QOS inventory and per-user consumption. `get_jobs` handles this with `Self::authoritative_user(&mut req.user, __identity.as_ref())`, and `list_tokens` gates enumeration on `require_admin`. Would it be worth following the `get_jobs` shape here, so a non-admin is scoped to themselves and admins can still pass an arbitrary user?

**A couple of display/semantics mismatches** — details inline: the scope line's `MaxTRESPU=` renders empty where its siblings render `N`; a non-`users=` selector is silently taken as a literal username; and the docs' "a literal `0` is a real cap" holds for count caps but not for TRES dimensions, where `0` is treated as unset.

**Tests.** Three gaps: `scope_exceeded_caps_covers_the_group_caps_only` can't fail as written (the user's caps are all `None`, so `exceeded_caps()` is empty for any implementation); the CLI tests all call `render_assoc_mgr` directly, so the `assoc_mgr` dispatch arm, the `users=` strip and the `LimitsReadable` banner are untested; and `limits_readable == true` is never asserted.

Also worth considering as follow-ups: `GrpWall`, `MaxTRESPerJob` and `MaxSubmitJobsPerAccount` are enforced but absent from the record, so a job pended by `QosGrpWallLimit` shows nothing here.

The `LimitsReadable=NO` wording points already raised below look right to me, and the same overstatement applies when accounting is disabled entirely — the refresh loop only starts on a successful DB connect, so the banner prints permanently and reads as a fault rather than "there are no caps".

crates/spurctld/src/server.rs
nikhilsk: Done — scoped the read to the caller instead of trusting the request field.
`get_assoc_mgr_info` now resolves the filter through a pure helper
`assoc_mgr_scope_user(identity, requested, caller_is_admin)`:

- a privileged caller (an admin, or an unauthenticated one under
  `permissive`/`disabled` — the same treatment `viewer_is_privileged` gives)
  reads whichever user it names, or every user when it names none;
- a non-admin authenticated caller is pinned to its own identity regardless of
  what it asked for.

That closes the "empty string = no filter" hole without a blanket
`require_admin`, which would have been wrong here since `permissive` has
`identity == None` and the existing convention treats that as privileged.

I also narrowed the scope lines, which was the broader half of the bug:
`scope_usage` now returns `None` when `only_user` is set and that user holds no
work and has no defined association in the scope, so a non-admin no longer
enumerates the cluster-wide QOS/account inventory or its group figures — they
see only scopes they take part in. Group totals for a scope they *are* in stay
whole (a group cap can't be judged from one user's share), matching Slurm.

Tests: `assoc_mgr_scope_user_*` cover the three authz decisions against the real
helper; `assoc_mgr_info_hides_scopes_a_filtered_user_has_no_part_in` covers the
narrowing end to end.

---

proto/slurm.proto
nikhilsk: Split. The block describing "a scope … plus the users holding work under it / a
scope with no queued work carries no user records" now sits above
`AssocMgrRecord`, and the "Caps a scope applies to each of its users …" block
stays above `AssocMgrCaps`. Each message carries the idle-QOS explanation once,
so the two halves no longer contradict. Comments only — no field touched.

---

crates/spur-cli/src/scontrol.rs
nikhilsk: Fixed. Added `tres_cap_or_n`, the TRES analogue of `cap_or_n`, so
`MaxTRES{PU}=` renders `N` when the scope sets no per-user TRES cap instead of
an empty field. A parser splitting on `=` now gets `N` here like everywhere
else. Covered by `assoc_mgr_renders_n_for_an_unset_per_user_tres_cap`.

---

crates/spur-cli/src/scontrol.rs
nikhilsk: Agreed — that silently read a wrong flag as a username. Pulled the parse into
`assoc_mgr_user_filter(Option<&str>) -> Result<String>`: `users=<name>` and a
bare name still work, but any other `<key>=` token now errors
("unknown selector 'qos='; only 'users=' is accepted") rather than filtering for
a user literally named `qos=highprio`. That makes the migration guide's "are not
accepted" wording true. Tests: `assoc_mgr_user_filter_reads_users_prefix_and_bare_name`
and `assoc_mgr_user_filter_rejects_a_non_users_selector`.

---

docs/user-guide/monitoring-jobs.rst
nikhilsk: Narrowed to the count caps and added the TRES note. I confirmed this is not
just a display gap: the admission gate (`qos.rs`) and `tres_exceeds`
(`accounting.rs`) both compare with `cap.get(t) > 0 && usage > cap`, so a TRES
dimension set to `0` is treated as unset at *enforcement* too, not only in the
render. So `GrpTRES=node=0` genuinely isn't enforced as "block all" — the docs
now say count caps honour a literal `0` while a `0` TRES dimension is unset
(renders `N`, not enforced). Whether `0` *should* be a hard TRES cap is a
semantic change across ~8 gate call sites; flagging separately rather than
folding it into this PR.

---

docs/user-guide/monitoring-jobs.rst
nikhilsk: Reworded. The bullet now says the `Limit(Consumed)` shape applies to the
`Grp*` line and the `User=` lines, and calls out that `MaxWall` and the
per-user caps on the scope line print bare.

---

crates/spur-core/src/accounting.rs
nikhilsk: Good catch — it couldn't fail as written. The user now carries
`max_jobs: Some(20)` against `running_jobs: 12`, a cap it is within, so the
empty `exceeded_caps()` proves the group caps weren't leaked onto the user
rather than that the user had no caps to report.

---

crates/spurctld/src/cluster.rs
nikhilsk: Done. The association records now read each `(user, account)` limit from the
rows `association_cache.all()` already returned (collected into a
`by_user` map for the scope), so the per-user closure and the group-caps read
no longer re-lock the cache per user. Behaviour is unchanged — same rows, same
first-row group caps.

---

crates/spur-cli/src/scontrol.rs
nikhilsk: Extracted `assoc_mgr_user_filter` (the `users=` strip / bare-name / reject-`qos=`
logic) and `limits_readable_banner` so the arm's behaviour is unit-testable
without a client. `limits_readable_banner(true)` returns `None` (banner
suppressed) and `false` returns the line, so both paths are asserted now, and
the selector cases are covered as above. The QOS-before-association ordering is
still exercised only through `show()` itself; I left that as-is since it's a
two-line print order, but happy to add a thin integration test if you'd prefer.

---

crates/spur-cli/src/scontrol.rs
```diff
@@ -871,9 +872,30 @@ async fn show(controller: &str, entity: &str, name: Option<&str>) -> Result<()>
...
+                .get_assoc_mgr_info(spur_proto::proto::GetAssocMgrInfoRequest { user })
+                .await
+                .context("failed to get association manager info")?
+                .into_inner();
+
+            if !resp.limits_readable {
+                println!(
+                    "LimitsReadable=NO (the accounting caches hold no snapshot, so no limit \
+                     below could be read; usage is still current)"
+                );
```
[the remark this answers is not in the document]
nikhilsk: Addressed together with the per-cache predicate above. `limits_readable` now
only reports false when accounting is enabled and a contributing cache is cold,
and the banner text no longer claims "no limit could be read" — it says some
caps below may be missing while usage stands. Proto and docs wording updated to
match.

---

docs/user-guide/monitoring-jobs.rst
```diff
@@ -390,6 +391,68 @@ The accounting database keeps the record permanently. In practice
...
+  tightening a cap under running work leaves exactly this state.
+
+Records come from the accounting definitions as well as from the queue, so a QOS
+nobody is using still appears with its caps and no ``User=`` lines — an
+over-tight cap on an idle QOS is exactly the kind of mistake that hides
+otherwise. A QOS deleted after its jobs queued also stays reported for as long as
+those jobs hold resources.
+
+A ``LimitsReadable=NO`` line above the sections means the accounting caches held
+no snapshot, so no cap could be read; the usage figures are still current.
```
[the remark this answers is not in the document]
nikhilsk: Addressed together with the per-cache predicate above. `limits_readable` now
only reports false when accounting is enabled and a contributing cache is cold,
and the banner text no longer claims "no limit could be read" — it says some
caps below may be missing while usage stands. Proto and docs wording updated to
match.

---

proto/slurm.proto
```diff
@@ -709,6 +714,74 @@ message SchedStats {
...
+  string user = 1;
+}
+
+message GetAssocMgrInfoResponse {
+  repeated AssocMgrRecord qos_records = 1;
+  repeated AssocMgrRecord assoc_records = 2;
+  // False when the accounting caches hold no snapshot: the usage figures still
+  // stand, but no cap could be read — the same condition that holds jobs at
+  // admission, so an operator seeing held jobs can tell the two apart.
+  bool limits_readable = 3;
```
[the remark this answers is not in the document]
nikhilsk: Addressed together with the per-cache predicate above. `limits_readable` now
only reports false when accounting is enabled and a contributing cache is cold,
and the banner text no longer claims "no limit could be read" — it says some
caps below may be missing while usage stands. Proto and docs wording updated to
match.

---

proto/slurm.proto
```diff
@@ -709,6 +714,74 @@ message SchedStats {
   uint64 exit_max_depth = 14;
 }
 
+// What one user holds under a scope, beside the caps governing that user. Counts
+// and TRES are live figures from the controller's job table, computed the way the
+// admission gate computes them, so they read as the scheduler sees them. Caps and
+// consumption stay separate fields rather than arriving pre-rendered as Slurm's
+// `Limit(Consumed)`, so a machine consumer does not have to unpick a display
+// string; the CLI composes that shape. Caps carry the INFINITE sentinel when
+// unset, TRES are `cpu=N,mem=N` strings, empty when nothing is held or capped.
```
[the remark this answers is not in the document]
nikhilsk: Fixed both comment blocks: TRES on the wire are `<name>=<value>` pairs (e.g.
`cpu=36,node=9`) via `TresRecord::format()`, empty when nothing is held or
capped; the `N` sentinel is introduced only when the CLI composes
`Limit(Consumed)`. Comments only.

---

crates/spurctld/src/server.rs
```diff
@@ -3951,6 +3983,87 @@ fn requested_gpus_detail(spec: &spur_core::job::JobSpec) -> String {
...
+            (_, Cap::GrpTres) => "GrpTRES",
+            (_, Cap::GrpSubmitJobs) => "GrpSubmitJobs",
+        }
+    }
+}
+
+/// One scope record onto the wire, users nested inside it. An unset cap carries
+/// the `INFINITE` sentinel, keeping it distinguishable from a literal `0` cap as
+/// `sacctmgr` already does, and TRES becomes its `cpu=N,mem=N` rendering. Caps and
+/// consumption stay separate fields: composing Slurm's `Limit(Consumed)` display
```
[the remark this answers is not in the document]
nikhilsk: Fixed both comment blocks: TRES on the wire are `<name>=<value>` pairs (e.g.
`cpu=36,node=9`) via `TresRecord::format()`, empty when nothing is held or
capped; the `N` sentinel is introduced only when the CLI composes
`Limit(Consumed)`. Comments only.

---

crates/spurctld/src/cluster.rs
```diff
@@ -6842,6 +6936,101 @@ fn sum_running_tres(jobs: &HashMap<JobId, Job>, pred: impl Fn(&Job) -> bool) ->
...
+            .into_iter()
+            .map(|user| {
+                let mine = |j: &Job| in_scope(j) && j.spec.user == user;
+                UserLimitUsage {
+                    user: user.to_owned(),
+                    running_jobs: jobs
+                        .values()
+                        .filter(|j| j.state == JobState::Running && mine(j))
+                        .count() as u32,
+                    submitted_jobs: jobs.values().filter(|j| is_submitted(j) && mine(j)).count()
```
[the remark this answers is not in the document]
nikhilsk: Real observation. I did the contained win (the per-user cap re-fetch, see
3876320613) but held off on the larger single-pass aggregation
(`HashMap<user, {counts, tres}>`). `scontrol show assoc_mgr` is an infrequent
operator command, the caller-scoping added here bounds a non-admin's work to
their own scopes, and a correct single-pass rewrite touches the group/per-user
TRES accumulation that admission also reads. Leaving it as a follow-up rather
than expanding the surface of the security fix — can pick it up separately if
you'd like it in this PR.

---

crates/spurctld/src/cluster.rs
```diff
@@ -4652,6 +4654,98 @@ impl ClusterManager {
         &self.association_cache
     }
 
+    /// Limits beside live usage, per user and scope, for `scontrol show assoc_mgr`.
+    /// Usage always reflects the job table; caps appear only where a cache holds a
+    /// snapshot, which `limits_readable` reports so an operator can tell an
+    /// uncapped scope from an unreadable one.
```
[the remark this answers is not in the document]
nikhilsk: The docstring/predicate now match: `limits_readable` is a single flag derived
per cache (`!enabled || (qos_loaded && assoc_loaded)`), and the comment no
longer promises per-scope clarity it doesn't provide. True per-section
readability (a flag per QOS/association section) is the direction #744's
`AccountingUnavailable` work points at; deferring to land it once, there.

---

crates/spurctld/src/cluster.rs
nikhilsk: The docstring/predicate now match: `limits_readable` is a single flag derived
per cache (`!enabled || (qos_loaded && assoc_loaded)`), and the comment no
longer promises per-scope clarity it doesn't provide. True per-section
readability (a flag per QOS/association section) is the direction #744's
`AccountingUnavailable` work points at; deferring to land it once, there.

---

crates/spur-core/src/accounting.rs
nikhilsk: Split. `types()` and `format()` each have their own doc comment now; the
"Format as …" line no longer sits on `types()`.

---

crates/spurctld/src/cluster.rs
nikhilsk: Kept the first-row behaviour (the refactor in 3876320613 preserves it) and left
the existing comment explaining why: association limits are keyed per
`(user, account)` row, group caps are the account's, and merging across rows
that disagree would invent a value the schema doesn't define. In practice
`grp_tres`/`grp_submit_jobs`/`max_wall` are set uniformly per account by
`sacctmgr`, so rows agree unless edited out of band; if they disagree that's a
data anomaly worth surfacing rather than papering over. Happy to add a
"rows disagree" warning log as a follow-up if you think it's worth the noise.

---

crates/spurctld/src/cluster.rs
nikhilsk: Out of scope for this PR, tracking as a follow-up alongside the other
enforced-but-unshown caps Yan listed (`GrpWall`, `MaxTRESPerJob`). They need
new record fields, so I'd rather add them as one deliberate change than tack a
partial set onto the authz fix.

---

crates/spurctld/src/cluster.rs
nikhilsk: Dropped the ticket ref; used your suggested wording.

---

crates/spur-cli/src/scontrol.rs
nikhilsk: Updated to `job, node, partition, reservation, assoc_mgr, federation, config`
so `--help` matches the dispatch and the error path.

---

crates/spur-cli/src/scontrol.rs
nikhilsk: Fixed the O(n^2), though not with `BTreeSet`: `TresType` doesn't derive `Ord`,
and deriving it would order by enum discriminant rather than by name, changing
the output order the layout contract (and tests) depend on. Used
`sort_by_key(|t| t.name())` + `dedup()` instead — same name-sorted order, no
quadratic `contains`.

shiv-tyagi (dismissed):
LGTM

yansun1996 (dismissed):
Re-reviewed the authorization path end to end: the filter now comes from the verified identity rather than the request field, an empty caller isn't treated as privileged, and scopes the filtered user has no part in are dropped rather than enumerated. The `limits_readable` predicate and its descriptions agree now. Approving.

PR #752, opened 2026-08-25
perf(spurctld): index the GrpWall consumption window

### LINKED ISSUE #748: accounting: GrpWall consumption aggregate has no supporting index, delaying enforcement on large job histories
`consumed_wall_minutes_by_qos` (`crates/spurctld/src/accounting/db.rs:708`) computes QOS `GrpWall` consumption with a full aggregate over `jobs`:

```sql
FROM jobs
WHERE qos <> ''
  AND start_time IS NOT NULL
  AND (end_time IS NULL OR end_time > now() - make_interval(days => $1))
GROUP BY qos
```

No index on `jobs` serves this. The schema (`db.rs:136-140`) defines single-column indexes on `user_name`, `account`, `state`, `submit_time` and `start_time`, plus the `job_id` primary key. There is no index on `qos` and no composite covering `qos` with `start_time`, so the aggregate scans every row of the accumulated job history even when the window under consideration is a few days.

## Why it matters beyond query cost

`GrpWallCache::spawn_refresh_loop` (`crates/spurctld/src/limits_cache.rs:164`) bounds the first read at 5 seconds and subsequent reads at 10. On a timeout it logs `grpwall usage refresh timed out, retaining last figure`. The code notes that until the cache first populates, no GrpWall budget is being applied.

So on a cluster with enough history for the scan to exceed those bounds, `grpwall` silently goes unenforced for at least a full refresh interval after controller start — the same fail-open window that #744 addressed for per-user QOS caps, reached by a different route. The cost also grows with retention rather than staying proportional to the window being measured.

## Expected

Consumption over an N-day window should cost roughly what that window contains, and should not become slow enough to leave the budget unapplied as history accumulates.

## Repro

1. Populate `jobs` with a large history spanning well beyond `accounting.grp_wall_window_days`.
2. `EXPLAIN ANALYZE` the query above.
3. Observe the sequential scan over the full table.

nikhilsk:
Closes #748.

`consumed_wall_minutes_by_qos` had no index to stand on, so QOS `GrpWall` consumption was computed by scanning the entire accumulated job history on every cache refresh. Cost tracked total retention rather than the window being measured, and `GrpWallCache` bounds its first read at 5 seconds and later reads at 10 — past that it logs `grpwall usage refresh timed out, retaining last figure` and keeps the previous snapshot, so only the first refresh timing out leaves the budget unapplied.

## Approach

One partial index on `jobs(end_time, qos, start_time)` where `qos <> '' AND start_time IS NOT NULL`, so the aggregate's cost follows the window instead of the history.

Column order is the whole point, and I got it wrong first. My initial attempt led with `qos` and carried `end_time` as an `INCLUDE` payload, on the reasoning that the query groups by `qos`. Measured, that index was never chosen — the planner kept the sequential scan, because `qos <> ''` and `start_time IS NOT NULL` match nearly every row and leave nothing to narrow. The selective predicate is the window on `end_time`, so `end_time` has to lead. With that order the planner picks it unprompted and builds a `BitmapOr` across the two branches of `end_time IS NULL OR end_time > cutoff`.

The SQL moved into a `consumed_wall_minutes_sql!` macro rather than staying an inline literal so the test can prefix it with `EXPLAIN` and cannot drift from the query that actually runs. It is a macro and not a `const` because sqlx 0.9 accepts only statically known SQL.

## Measurements

Postgres 16, 400k jobs spanning ~700 days, 12 QOS, 14-day window, 8,008 rows qualifying, `EXPLAIN (ANALYZE, BUFFERS)`:

| | plan | execution |
|---|---|---|
| before | parallel sequential scan over all 400k rows | 60.4 / 61.0 / 63.8 ms |
| after | `BitmapOr` of two index scans, 8,023 rows | 8.4 / 9.2 ms |

Roughly 7x here, but the ratio is not the interesting part — before, the work grew with everything the cluster had ever run; after, it grows with the window. On the first fixture I generated, the gain was only 1.4x, because I had given 5% of jobs a NULL `end_time` scattered across two years. Real clusters only have recently-started jobs still running, and correcting the fixture to match is what exposed the true difference.

Index size is 17 MB against a 57 MB heap at 400k rows, and it adds one more index to maintain on job insert and update. `end_time` leads the index and `record_job_end` writes it on every completion, so each job end now churns an index entry and that update can no longer be HOT — almost certainly fine on a table already carrying five indexes, but worth stating.

## Trade-offs

`migrate` applies the schema inside a transaction, so `CREATE INDEX CONCURRENTLY` is not available here. The `CREATE INDEX` itself takes `SHARE` on `jobs` — it blocks writers, not readers — for the duration of the build (seconds at 400k rows, longer on a large history). In practice the first boot already holds `ACCESS EXCLUSIVE` on `jobs`, taken earlier in the same transaction by the existing `ALTER TABLE jobs ADD COLUMN IF NOT EXISTS` statements and held until commit, so writers are blocked either way on that first start. This matches the indexes already created the same way in this schema, and `CREATE INDEX IF NOT EXISTS` makes it a no-op on every subsequent start.

## Testing

New DB integration test `grp_wall_aggregate_is_servable_from_its_index` asserts the plan reaches the index and that the window predicate gets there, which is what keeps cost tied to the window. It seeds a small deterministic fixture inside a rolled-back transaction — two in-window rows, one out-of-window, one still running with a NULL `end_time` (the branch the predicate deliberately omits), one with no QOS — then runs `ANALYZE jobs` before the `EXPLAIN`, so the plan is chosen from a known shape rather than whatever statistics the database happens to carry. With `enable_seqscan` off the planner is forced to say whether the partial index's predicate is still implied by the query's `WHERE` — the pairing that rots silently when either side is edited. The negative assertion checks the plan does not fall to `idx_jobs_start_time`: `enable_seqscan = off` is a cost penalty, not a prohibition, so a broken predicate degrades to an index scan on that index (which scans all history), not to the seq scan the earlier assertion looked for. Removing `qos <> ''` from the query confirms the assertion fires.

Full suite green: `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` warning-free, `cargo test --locked` passing, and all 24 accounting DB tests passing against a local Postgres 16 (`cargo test -p spurctld -- --ignored --test-threads=1`), with the same plan observed against a freshly created database.

### DIFF
--- a/crates/spurctld/src/accounting/db.rs
+++ b/crates/spurctld/src/accounting/db.rs
@@ -138,6 +138,13 @@ CREATE INDEX IF NOT EXISTS idx_jobs_account ON jobs(account);
 CREATE INDEX IF NOT EXISTS idx_jobs_state ON jobs(state);
 CREATE INDEX IF NOT EXISTS idx_jobs_submit_time ON jobs(submit_time);
 CREATE INDEX IF NOT EXISTS idx_jobs_start_time ON jobs(start_time);
+-- Serves the GrpWall consumption aggregate, which otherwise scans the whole job
+-- history on every cache refresh; a refresh slow enough to time out leaves the
+-- budget unenforced. end_time leads because it carries the window predicate and
+-- is the only selective one: ordering by qos instead leaves nothing to narrow,
+-- and the planner then prefers a full scan.
+CREATE INDEX IF NOT EXISTS idx_jobs_grp_wall_window ON jobs(end_time, qos, start_time)
+    WHERE qos <> '' AND start_time IS NOT NULL;
 CREATE INDEX IF NOT EXISTS idx_usage_period ON usage(period_start, period_end);
 CREATE INDEX IF NOT EXISTS idx_assoc_user ON associations(user_name);
 CREATE INDEX IF NOT EXISTS idx_assoc_account ON associations(account);
@@ -705,11 +712,10 @@ pub async fn get_usage(
 /// be written by callers outside the controller and can skew, and a single
 /// end-before-start row would otherwise cancel real consumption in the same QOS
 /// and under-report the budget.
-pub async fn consumed_wall_minutes_by_qos(
-    pool: &PgPool,
-    window_days: u32,
-) -> anyhow::Result<HashMap<String, u64>> {
-    let rows = sqlx::query(
+/// A macro rather than a `const` so the test can prefix it with `EXPLAIN` via
+/// `concat!`: sqlx 0.9 accepts only statically known SQL.
+macro_rules! consumed_wall_minutes_sql {
+    () => {
         r#"
         SELECT qos,
                SUM(GREATEST(
@@ -724,11 +730,18 @@ pub async fn consumed_wall_minutes_by_qos(
           AND start_time IS NOT NULL
           AND (end_time IS NULL OR end_time > now() - make_interval(days => $1))
         GROUP BY qos
-        "#,
-    )
-    .bind(window_days as i32)
-    .fetch_all(pool)
-    .await?;
+        "#
+    };
+}
+
+pub async fn consumed_wall_minutes_by_qos(
+    pool: &PgPool,
+    window_days: u32,
+) -> anyhow::Result<HashMap<String, u64>> {
+    let rows = sqlx::query(consumed_wall_minutes_sql!())
+        .bind(window_days as i32)
+        .fetch_all(pool)
+        .await?;
 
     let consumed = rows
         .iter()
@@ -1897,6 +1910,42 @@ mod job_history_tests {
         );
 
         delete_jobs(&pool, &ids).await?;
+        Ok(())
+    }
+
+    #[tokio::test]
+    #[ignore = "requires DATABASE_URL and PostgreSQL"]
+    async fn grp_wall_aggregate_is_servable_from_its_index() -> anyhow::Result<()> {
+        let pool = test_pool().await?;
+        let mut tx = pool.begin().await?;
+
+        // A partial index qualifies only if its predicate is implied by the query's
+        // WHERE, which is the part that rots silently when either side is edited.
+        // Forcing seqscan off makes the planner say whether the pairing still holds
+        // on a table too small for it to prefer the index on cost alone.
+        sqlx::query("SET LOCAL enable_seqscan = off")
+            .execute(&mut *tx)
+            .await?;
+        let plan = sqlx::query(concat!("EXPLAIN ", consumed_wall_minutes_sql!()))
+            .bind(14i32)
+            .fetch_all(&mut *tx)
+            .await?
+            .iter()
+            .map(|row| row.get::<String, _>(0))
+            .collect::<Vec<_>>()
+            .join("\n");
+
+        assert!(
+            plan.contains("idx_jobs_grp_wall_window"),
+            "GrpWall consumption must be servable from idx_jobs_grp_wall_window; plan was:\n{plan}"
+        );
+        assert!(
+            plan.contains("end_time"),
+            "the window predicate must reach the index, or its cost tracks total \
+             history instead of the window; plan was:\n{plan}"
+        );
+
+        tx.rollback().await?;
         Ok(())
     }
 
--- a/crates/spurctld/src/accounting/db.rs
+++ b/crates/spurctld/src/accounting/db.rs
@@ -138,11 +138,8 @@ CREATE INDEX IF NOT EXISTS idx_jobs_account ON jobs(account);
 CREATE INDEX IF NOT EXISTS idx_jobs_state ON jobs(state);
 CREATE INDEX IF NOT EXISTS idx_jobs_submit_time ON jobs(submit_time);
 CREATE INDEX IF NOT EXISTS idx_jobs_start_time ON jobs(start_time);
--- Serves the GrpWall consumption aggregate, which otherwise scans the whole job
--- history on every cache refresh; a refresh slow enough to time out leaves the
--- budget unenforced. end_time leads because it carries the window predicate and
--- is the only selective one: ordering by qos instead leaves nothing to narrow,
--- and the planner then prefers a full scan.
+-- Serves the GrpWall aggregate, which otherwise scans all job history per refresh;
+-- a first-refresh timeout leaves the budget unapplied. end_time leads: it bounds the window.
 CREATE INDEX IF NOT EXISTS idx_jobs_grp_wall_window ON jobs(end_time, qos, start_time)
     WHERE qos <> '' AND start_time IS NOT NULL;
 CREATE INDEX IF NOT EXISTS idx_usage_period ON usage(period_start, period_end);
@@ -1544,6 +1541,41 @@ mod job_history_tests {
         .await
     }
 
+    /// Seeds one job through the real start/end path on `conn`, so the EXPLAIN test
+    /// can build its fixture inside a transaction that rolls back. `end = None`
+    /// leaves the job running with a NULL end_time.
+    async fn seed_job(
+        conn: &mut PgConnection,
+        job_id: i32,
+        qos: &str,
+        start_time: DateTime<Utc>,
+        end: Option<DateTime<Utc>>,
+    ) -> anyhow::Result<()> {
+        record_job_start(
+            conn,
+            &JobStartRecord {
+                job_id: job_id as JobId,
+                name: "fixture".to_string(),
+                user: "root".to_string(),
+                account: String::new(),
+                partition: String::new(),
+                qos: qos.to_string(),
+                num_nodes: 1,
+                num_tasks: 1,
+                cpus_per_task: 1,
+                memory_mb: 0,
+                submit_time: start_time,
+                start_time,
+                reservation: Some(String::new()),
+            },
+        )
+        .await?;
+        if let Some(end_time) = end {
+            record_job_end(conn, job_id, "COMPLETED", 0, end_time, 0, 0, None, "", "").await?;
+        }
+        Ok(())
+    }
+
     #[tokio::test]
     #[ignore = "requires DATABASE_URL and PostgreSQL"]
     async fn get_job_history_query_builder() -> anyhow::Result<()> {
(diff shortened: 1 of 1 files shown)


shiv-tyagi (commented):
Posted comments, PTAL

crates/spurctld/src/accounting/db.rs
```diff
@@ -1900,6 +1913,42 @@ mod job_history_tests {
...
+
+        assert!(
+            plan.contains("idx_jobs_grp_wall_window"),
+            "GrpWall consumption must be servable from idx_jobs_grp_wall_window; plan was:\n{plan}"
+        );
+        assert!(
+            plan.contains("end_time"),
+            "the window predicate must reach the index, or its cost tracks total \
+             history instead of the window; plan was:\n{plan}"
+        );
```
shiv-tyagi, line 2023: Nit: `plan.contains("end_time")` is very loose. Any index-qualified plan on `idx_jobs_grp_wall_window` will mention `end_time`, so this does not independently verify that the window predicate is what narrows the scan (the stated goal in the comment above it).

If you want a stronger guard without asserting a specific plan node (which the PR description correctly notes can vary), consider matching something more specific to the cutoff branch, e.g. `Filter` combined with `end_time`, or documenting that the index-name check alone is the intended bar.

crates/spurctld/src/accounting/db.rs
```diff
@@ -724,11 +730,18 @@ pub async fn consumed_wall_minutes_by_qos(
...
-        "#,
-    )
-    .bind(window_days as i32)
-    .fetch_all(pool)
-    .await?;
+        "#
+    };
+}
+
+pub async fn consumed_wall_minutes_by_qos(
```
shiv-tyagi: The `///` doc block that documented `consumed_wall_minutes_by_qos` now sits above `macro_rules! consumed_wall_minutes_sql` instead of the public function. Rustdoc for the exported API is gone; what remains documents a private macro (including the sqlx/`EXPLAIN` rationale, which belongs on the macro, not the function).

Move the macro above the function, restore the original doc on `pub async fn consumed_wall_minutes_by_qos`, and use a `//` comment for the macro-only note:

```rust
macro_rules! consumed_wall_minutes_sql {
    () => {
        r#"
        SELECT qos,
               ...
        "#
    };
}

/// Wall-clock minutes consumed per QOS inside the trailing `window_days`, for
/// `GrpWall` enforcement. ...
pub async fn consumed_wall_minutes_by_qos(
    pool: &PgPool,
    window_days: u32,
) -> anyhow::Result<HashMap<String, u64>> {
    let rows = sqlx::query(consumed_wall_minutes_sql!())
        ...
}
```

crates/spurctld/src/accounting/db.rs
nikhilsk: Agreed it was loose, and it was worse than loose: `plan.contains("end_time")` could not fail, since the predicate appears whether the index serves it or a full scan filters it afterwards. It now asserts the plan names the index and does not contain `Seq Scan on jobs` - with seqscan off, that fallback is exactly what happens once the index predicate stops being implied by the `WHERE`. Confirmed by breaking the implication locally: the plan drops to `Seq Scan` and the test fails.

I left the node shape unpinned, including the `Filter` + `end_time` variant. The same query plans as an `Index Only Scan` on a populated table and as a `BitmapOr` over the two `end_time` branches on a small one, so either would track statistics rather than the pairing. The comment now says so.

crates/spurctld/src/accounting/db.rs
nikhilsk: Done as suggested: macro above, original rustdoc restored on `consumed_wall_minutes_by_qos`, and the macro's note is a `//` comment. It also no longer claims sqlx accepts only statically known SQL - `sqlx::query` takes any `&str`; the reason is `concat!` needing a literal, which is what it says now.

crates/spurctld/src/accounting/db.rs
```diff
@@ -138,6 +138,13 @@ CREATE INDEX IF NOT EXISTS idx_jobs_account ON jobs(account);
 CREATE INDEX IF NOT EXISTS idx_jobs_state ON jobs(state);
 CREATE INDEX IF NOT EXISTS idx_jobs_submit_time ON jobs(submit_time);
 CREATE INDEX IF NOT EXISTS idx_jobs_start_time ON jobs(start_time);
+-- Serves the GrpWall consumption aggregate, which otherwise scans the whole job
+-- history on every cache refresh; a refresh slow enough to time out leaves the
```
yansun1996: Small accuracy point: "a refresh slow enough to time out leaves the budget unenforced" holds only for the *first* refresh. A later timeout logs `retaining last figure` and keeps the previous snapshot, so the budget stays enforced. Could we say "the first refresh timing out leaves the budget unapplied"? While you're in here, this block is five lines — most of it is the reasoning already covered in the PR description, so it could probably come down to two.

crates/spurctld/src/accounting/db.rs
```diff
@@ -138,6 +138,13 @@ CREATE INDEX IF NOT EXISTS idx_jobs_account ON jobs(account);
...
+-- budget unenforced. end_time leads because it carries the window predicate and
+-- is the only selective one: ordering by qos instead leaves nothing to narrow,
```
yansun1996: Minor: "the planner then prefers a full scan" is a bit stronger than what happens. A `qos`-leading *partial* index would still be read as an index scan (and would hand you `GROUP BY qos` pre-sorted) — it's just not selective. The column order is clearly the right call; it's only the stated reason that overreaches.

crates/spurctld/src/accounting/db.rs
```diff
@@ -1900,6 +1913,47 @@ mod job_history_tests {
...
+        let mut tx = pool.begin().await?;
+
+        // A partial index qualifies only if its predicate is implied by the query's
+        // WHERE, which is the part that rots silently when either side is edited.
+        // Forcing seqscan off makes the planner say whether the pairing still holds
+        // on a table too small for it to prefer the index on cost alone.
+        sqlx::query("SET LOCAL enable_seqscan = off")
+            .execute(&mut *tx)
+            .await?;
+        let plan = sqlx::query(concat!("EXPLAIN ", consumed_wall_minutes_sql!()))
```
yansun1996, line 2003: This runs against whatever `DATABASE_URL` points at, with no rows seeded and no `ANALYZE`, so which index the planner reaches for depends on ambient statistics — and the sibling `#[ignore]` tests in this module insert and delete against the same `jobs` table. Would it be worth inserting a small fixture inside the transaction (a couple of in-window rows, one out-of-window, one running row with `end_time IS NULL`, one with `qos = ''`) and running `ANALYZE jobs` before the `EXPLAIN`? It's all rolled back, and it would also make the running-job branch — the reason the predicate deliberately omits `end_time` — actually exercised rather than assumed.

crates/spurctld/src/accounting/db.rs
```diff
@@ -1900,6 +1913,47 @@ mod job_history_tests {
...
+        // table, a bitmap OR over the two end_time branches on a small one — so the
+        // pairing is asserted by which relation is read, not by node shape. With
+        // seqscan off, a full scan is what the planner falls back to when the index
+        // predicate is no longer implied, which is the rot this test is here for.
+        assert!(
+            plan.contains("idx_jobs_grp_wall_window"),
+            "GrpWall consumption must be servable from idx_jobs_grp_wall_window; plan was:\n{plan}"
+        );
+        assert!(
+            !plan.contains("Seq Scan on jobs"),
```
yansun1996: I don't think this assertion can fail. `jobs` already carries `idx_jobs_start_time`, and `start_time IS NOT NULL` is an indexable qual, so if this index were dropped or its predicate stopped being implied, the planner with `enable_seqscan = off` would fall to an index scan on *that* index rather than a seq scan — `enable_seqscan = off` is a cost penalty, not a prohibition. The comment above says a full scan is the fallback, which is what led here, but it isn't true for this table. The positive assert on the index name is doing the real work; this one could either go, or become something with teeth — asserting the plan doesn't mention `idx_jobs_start_time` is the actual degraded shape.

yansun1996 (commented):
Nice piece of work — the write-up on why `end_time` has to lead, including the first attempt that the planner refused, is genuinely useful context. I checked the predicate implication and it holds: `qos <> ''` and `start_time IS NOT NULL` are top-level conjuncts of the query's `WHERE`, matching the index predicate exactly, and all three indexed columns come from `CREATE TABLE jobs` rather than a later `ALTER`, so the position in the schema string is safe.

My comments are all on the new test, which I think is one step short of doing the job it's meant to do. Details inline.

Two notes that aren't code:

- The description says the build takes `ACCESS EXCLUSIVE` on `jobs`. Plain `CREATE INDEX` takes `SHARE` — it blocks writers, not readers. `ACCESS EXCLUSIVE` does get taken on `jobs` later in the same transaction by the existing `ALTER TABLE jobs ADD COLUMN IF NOT EXISTS` statements, so the practical effect on the first boot is much the same, but it'd be worth stating accurately.
- `end_time` leads the index and is written by `record_job_end` on every completion, so each job end now churns an index entry and the update can no longer be HOT. Almost certainly fine on a table already carrying five indexes, but worth a line in the description alongside the size figure.

crates/spurctld/src/accounting/db.rs
nikhilsk: Reworded to the first-refresh case and trimmed the block from five lines to two. It now reads:

```
-- Serves the GrpWall aggregate, which otherwise scans all job history per refresh;
-- a first-refresh timeout leaves the budget unapplied. end_time leads: it bounds the window.
```

You're right that a later timeout logs `retaining last figure` and keeps the previous snapshot, so only the first refresh leaves the budget unapplied.

---

crates/spurctld/src/accounting/db.rs
nikhilsk: Agreed — dropped the full-scan claim. The reason now stated is just that `end_time` leads because it bounds the window (the selective column); it no longer asserts what the planner does with a `qos`-leading index. Confirmed in the teeth experiment below: with the partial predicate broken, the planner reads an *index scan*, never a seq scan.

---

crates/spurctld/src/accounting/db.rs
nikhilsk: Done. The test now seeds a deterministic fixture inside the (rolled-back) transaction — two in-window rows, one out-of-window, one still running with `end_time IS NULL`, one with `qos = ''` — through the real `record_job_start`/`record_job_end` path, then runs `ANALYZE jobs` before the `EXPLAIN`. That removes the dependency on ambient statistics and exercises the running-job branch (the reason the predicate omits `end_time`) rather than assuming it.

Confirmed deterministic: identical plan on the populated development database and on a freshly created one:

```
Index Only Scan using idx_jobs_grp_wall_window on jobs
  Filter: ((end_time IS NULL) OR (end_time > (now() - '14 days'::interval)))
```

(Before the fixture, a fresh database gave a `BitmapOr` over the two `end_time` branches while a populated one gave the index-only scan — exactly the ambient-statistics dependence you flagged.)

---

crates/spurctld/src/accounting/db.rs
nikhilsk: You're right, and thanks for the catch — `enable_seqscan = off` is a cost penalty, not a prohibition, and `jobs` carries `idx_jobs_start_time` with `start_time IS NOT NULL` as an indexable qual, so the old `!contains("Seq Scan on jobs")` assertion could never fire. My earlier "proof" only worked because I broke both conjuncts at once.

Replaced it with the actual degraded shape: the plan must not mention `idx_jobs_start_time`. Proved it has teeth by removing just `qos <> ''` from the query's `WHERE` (single-conjunct break) with the fixture in place — the test fails, and the observed plan is:

```
GroupAggregate
  Group Key: qos
  ->  Sort
        Sort Key: qos
        ->  Index Scan using idx_jobs_start_time on jobs
              Index Cond: (start_time IS NOT NULL)
              Filter: ((end_time IS NULL) OR (end_time > (now() - '14 days'::interval)))
```

That is precisely the index scan on `idx_jobs_start_time` you predicted — the old seq-scan assertion would have passed on it. Restored the conjunct afterwards; the fixed comment above the assertion now describes this fallback instead of a full scan.

shiv-tyagi (approved):
LGTM

yansun1996 (approved):
Re-checked each point against the current code. Both EXPLAIN assertions can genuinely go red now — pinning the index name rather than the seq-scan shape is the right call, and the seeded fixture makes the test read much more clearly. Approving.

PR #753, opened 2026-08-25
fix(spurctld)!: store job ids as 64-bit in the accounting database

### LINKED ISSUE #749: accounting: job IDs above i32::MAX are silently corrupted on the accounting write path
The in-memory job id is `pub type JobId = u32` (`crates/spur-core/src/job.rs:13`), but the accounting schema stores `job_id INTEGER PRIMARY KEY` (`crates/spurctld/src/accounting/db.rs:36`) — a signed 32-bit column. Roughly thirty sites cast across that boundary with an unchecked `as i32`, including `db.rs:257` and `:278`, `grpc.rs:209`, `notifier.rs:100` and `:106`, and around eighteen in `reconcile.rs`.

A job id above `i32::MAX` (2,147,483,647) therefore wraps to a negative value on write. Because the column is the primary key, a wrapped id can collide with or overwrite an unrelated row rather than failing loudly, and `record_job_end` and the reconciler would then act on the wrong job.

## The configured ceiling that would prevent this is inert

`controller.max_job_id` exists in config with a default of `999_999_999`, which sits below the wrap point — but nothing enforces it. A workspace-wide search finds `max_job_id` only in `crates/spur-core/src/config.rs` (definition at :279, default at :422, `Default` impl at :2306) and two test fixtures; no scheduler or id-allocation code reads it.

`docs/admin-guide/configuration.rst:188` already states this plainly: "Not implemented. Intended as the job-ID wrap point. No code consumes it; the counter does not wrap." So the documented safeguard that would keep ids inside the column's range is inert, and `first_job_id` is separately configurable.

## Expected

Either the configured ceiling is enforced so ids stay within the stored type, or the stored type accommodates the full `u32` range. Exceeding it should not silently corrupt rows.

nikhilsk:
Closes #749.

Job ids are `u32` in the scheduler (`JobId`), but the accounting schema stored them in `INTEGER` columns — signed 32-bit. Every write narrowed with an unchecked `as i32`, so an id above `i32::MAX` arrived negative. Since `job_id` is the primary key, such a write did not fail; it landed on whatever row already held the wrapped id, and `record_job_end` and the reconciler would then act on the wrong job.

`controller.max_job_id` would have kept ids inside the column's range at its default of 999,999,999, but nothing reads it — the docs already say so.

## Approach

Widen the three id columns to `BIGINT` (`jobs.job_id`, `jobs.preempted_by`, `tres_usage.job_id`) and carry `i64` through the accounting path, so the stored width covers the full `u32` range and no cast can wrap.

I chose widening over the two alternatives because it leaves job id semantics alone. Enforcing the ceiling instead would mean a cluster that reaches it stops accepting jobs. Implementing `max_job_id` as the wrap point the docs describe would be a much larger change than it appears, because the codebase currently treats id order as time order:

```rust
// crates/spurctld/src/server.rs
None => job_id < cluster.peek_next_job_id(),
```

That is how the controller decides a node-reported id was issued at some point and may therefore be released. Under a wrapping counter, an id that *was* issued compares as never-issued after the wrap and is spared forever, leaking node-side state. Allocation is a bare `next_job_id.fetch_add(1)` with no live-id skipping, so wrapping would also need that, plus an audit of every other place ordering is assumed. Worth doing only if the ceiling is actually wanted; it is not needed to fix this.

## Legacy rows are rewritten, not decoded

A pre-fix controller could have already stored an id above `i32::MAX` as its negative wrap. Rather than heal those at read time, the migration repairs the data once, right after each `ALTER ... TYPE BIGINT`:

```sql
UPDATE jobs SET job_id = job_id + 4294967296
WHERE job_id < 0
  AND NOT EXISTS (SELECT 1 FROM jobs j WHERE j.job_id = jobs.job_id + 4294967296);
-- plus jobs.preempted_by and tres_usage.job_id
```

The `NOT EXISTS` guard skips a negative row whose true id already exists so the rewrite can never violate the primary key (defensive — a genuine pre-fix `INTEGER` column cannot itself hold the true id, so this only matters for a hand-corrupted table). The rewrite runs inside the same transaction as the widening, so a partial apply rolls back cleanly and never leaves the table half-migrated.

This removes the read/lookup asymmetry an earlier revision had: decoding at read time meant `get_job_history` healed a stored `-1` back to `u32::MAX` while `job_accounting_states` bound the positive id and missed the row, so the reconciler inserted a duplicate at the positive id and double-charged usage on its end write. With the data rewritten, `decode_job_id` is gone and both paths are plain `i64` → `u32` conversions. `job_accounting_states` also drops its now-unreachable `HashMap` inversion — every returned `job_id` is by construction an element of the bound list.

## Upgrade behaviour — **breaking, roll forward only**

Existing databases are migrated in place by the first upgraded controller. The type change is guarded on the column's current type, so it runs once rather than on every start. `ALTER COLUMN TYPE` from `INTEGER` to `BIGINT` is not binary-coercible, so it rewrites the table and holds `ACCESS EXCLUSIVE` on `jobs` for the duration — writes to that table block while the migration runs, and the duration scales with the row count.

**Rolling back to a pre-migration controller is not supported.** An older binary decodes `job_id` as `i32` and binds `&[i32]`, so `get_job_history` and `job_accounting_states` fail outright against a migrated database (writes survive, since Postgres widens an `int4` parameter). Because that breaks an older client against a migrated DB, the title now carries `!` (`fix(spurctld)!:`). This is documented in `docs/deployment/upgrading.rst` under "Safe Upgrade Order".

No proto change: `RecordJobStartRequest.job_id` stays `uint32`, matching `JobId`.

## Testing

DB integration tests (in the `DB tests (accounting)` job), seeded against real legacy negative rows before migrating:

- `job_ids_past_i32_survive_the_accounting_round_trip` — writes `u32::MAX` through `record_job_start`/`record_job_end`, reads it back through `get_job_history` and `job_accounting_states`, asserting the id and `preempted_by` survive.
- `migrate_rewrites_a_legacy_narrowed_row_to_its_real_id` — seeds a `-1` row in a narrowed schema, runs `migrate`, and asserts it reads back under `u32::MAX` and that the reconciler's lookup finds it.
- `migrate_skips_a_legacy_id_whose_real_id_already_exists` — seeds both a `-1` row and a row already at the true id, and asserts the guard leaves the negative row untouched rather than duplicating or colliding.
- `migrate_widens_pre_fix_integer_job_ids` — reconstructs a pre-fix schema, asserts all three columns widen, and re-runs to pin idempotency. Its schema restore now runs on the failure path too, so a local `--ignored` run that fails midway does not leave the shared DB narrow.

`cargo clippy --workspace --exclude spur-ffi --all-targets --locked` warning-free, full `cargo test --locked` green, and all accounting DB tests pass against a local Postgres.

Also: corrected the `max_job_id` configuration note (the counter still wraps past `u32::MAX`, so it does not make the full range safe on its own), and trimmed the migration/test comment blocks per repo style.

## Deferred

`SET LOCAL lock_timeout` inside the migration transaction (so the rewrite fails fast instead of queueing behind an older replica's open transaction) is left as a follow-up. It changes startup failure semantics and picking the timeout is a policy call; keeping this PR focused on the correctness fix. The rewrite makes that pre-existing behaviour more expensive but does not introduce it.

### DIFF
--- a/crates/spurctld/src/accounting/db.rs
+++ b/crates/spurctld/src/accounting/db.rs
@@ -33,7 +33,7 @@ const SCHEMA_LOCK_OBJ: i32 = 1;
 
 const SCHEMA: &str = r#"
 CREATE TABLE IF NOT EXISTS jobs (
-    job_id          INTEGER PRIMARY KEY,
+    job_id          BIGINT PRIMARY KEY,
     name            TEXT NOT NULL DEFAULT '',
     user_name       TEXT NOT NULL,
     uid             INTEGER NOT NULL DEFAULT 0,
@@ -126,7 +126,7 @@ CREATE TABLE IF NOT EXISTS associations (
 );
 
 CREATE TABLE IF NOT EXISTS tres_usage (
-    job_id          INTEGER NOT NULL,
+    job_id          BIGINT NOT NULL,
     tres_type       TEXT NOT NULL,
     alloc_value     BIGINT NOT NULL DEFAULT 0,
     used_value      BIGINT NOT NULL DEFAULT 0,
@@ -158,9 +158,34 @@ ALTER TABLE associations ADD COLUMN IF NOT EXISTS grp_submit_jobs INTEGER;
 ALTER TABLE accounts ADD COLUMN IF NOT EXISTS grp_tres TEXT;
 ALTER TABLE qos ADD COLUMN IF NOT EXISTS preempt TEXT NOT NULL DEFAULT '';
 ALTER TABLE qos ADD COLUMN IF NOT EXISTS preempt_exempt_time INTEGER;
-ALTER TABLE jobs ADD COLUMN IF NOT EXISTS preempted_by INTEGER;
+ALTER TABLE jobs ADD COLUMN IF NOT EXISTS preempted_by BIGINT;
 ALTER TABLE jobs ADD COLUMN IF NOT EXISTS preempt_mode TEXT NOT NULL DEFAULT '';
 ALTER TABLE jobs ADD COLUMN IF NOT EXISTS preempt_qos TEXT NOT NULL DEFAULT '';
+
+-- Job ids are u32 in the scheduler but these columns were INTEGER, so an id above
+-- i32::MAX wrapped negative and could collide with an unrelated row instead of
+-- failing. Guarded on the current type because ALTER TYPE rewrites the table and
+-- takes ACCESS EXCLUSIVE: it must run once on upgrade, not on every start.
+DO $$
+DECLARE
+    target RECORD;
+BEGIN
+    FOR target IN
+        SELECT *
+        FROM (VALUES ('jobs', 'job_id'), ('jobs', 'preempted_by'), ('tres_usage', 'job_id'))
+             AS t(tbl, col)
+    LOOP
+        IF EXISTS (
+            SELECT 1 FROM information_schema.columns
+            WHERE table_schema = current_schema()
+              AND table_name = target.tbl
+              AND column_name = target.col
+              AND data_type = 'integer'
+        ) THEN
+            EXECUTE format('ALTER TABLE %I ALTER COLUMN %I TYPE BIGINT', target.tbl, target.col);
+        END IF;
+    END LOOP;
+END $$;
 
 -- users.default_account is the single source of truth for a user's default
 -- account (the scheduler reads it via the association cache). associations
@@ -254,7 +279,7 @@ pub async fn record_job_start(conn: &mut PgConnection, rec: &JobStartRecord) -> 
             end_time = NULL
         "#,
     )
-    .bind(rec.job_id as i32)
+    .bind(rec.job_id as i64)
     .bind(&rec.name)
     .bind(&rec.user)
     .bind(&rec.account)
@@ -275,7 +300,7 @@ pub async fn record_job_start(conn: &mut PgConnection, rec: &JobStartRecord) -> 
     let row = sqlx::query(
         "SELECT user_name, account, start_time, num_tasks, cpus_per_task, end_time FROM jobs WHERE job_id = $1",
     )
-    .bind(rec.job_id as i32)
+    .bind(rec.job_id as i64)
     .fetch_one(&mut *conn)
     .await?;
 
@@ -292,13 +317,13 @@ pub async fn record_job_start(conn: &mut PgConnection, rec: &JobStartRecord) -> 
 #[allow(clippy::too_many_arguments)]
 pub async fn record_job_end(
     conn: &mut PgConnection,
-    job_id: i32,
+    job_id: i64,
     state: &str,
     exit_code: i32,
     end_time: DateTime<Utc>,
     exit_signal: i32,
     derived_exit_code: i32,
-    preempted_by: Option<i32>,
+    preempted_by: Option<i64>,
     preempt_mode: &str,
     preempt_qos: &str,
 ) -> anyhow::Result<()> {
@@ -352,8 +377,8 @@ pub struct AccountingRowState {
 /// by the reconciliation pass to detect jobs missing or stale in accounting.
 pub async fn job_accounting_states(
     pool: &PgPool,
-    job_ids: &[i32],
-) -> anyhow::Result<HashMap<i32, AccountingRowState>> {
+    job_ids: &[i64],
+) -> anyhow::Result<HashMap<i64, AccountingRowState>> {
     if job_ids.is_empty() {
         return Ok(HashMap::new());
     }
@@ -366,7 +391,7 @@ pub async fn job_accounting_states(
     Ok(rows
         .into_iter()
         .map(|r| {
-            let job_id: i32 = r.get("job_id");
+            let job_id: i64 = r.get("job_id");
             let user_name: String = r.get("user_name");
             let start_time: Option<DateTime<Utc>> = r.get("start_time");
             let row = AccountingRowState {
@@ -428,7 +453,7 @@ async fn update_usage(
 /// Job record returned from history queries.
 #[derive(Debug)]
 pub struct JobRecord {
-    pub job_id: i32,
+    pub job_id: i64,
     pub name: String,
     pub user_name: String,
     pub account: String,
@@ -444,7 +469,7 @@ pub struct JobRecord {
     pub start_time: Option<DateTime<Utc>>,
     pub end_time: Option<DateTime<Utc>>,
     pub reservation: String,
-    pub preempted_by: Option<i32>,
+    pub preempted_by: Option<i64>,
     pub preempt_mode: String,
     pub preempt_qos: String,
 }
@@ -1404,9 +1429,9 @@ mod job_history_tests {
     use super::*;
     use chrono::Duration;
 
-    fn test_job_id(slot: u32) -> i32 {
-        const BASE: i32 = 9_000_000;
-        BASE + (std::process::id() as i32 % 10_000) * 10 + slot as i32
+    fn test_job_id(slot: u32) -> i64 {
+        const BASE: i64 = 9_000_000;
+        BASE + (std::process::id() as i64 % 10_000) * 10 + slot as i64
     }
 
     async fn test_pool() -> anyhow::Result<PgPool> {
@@ -1419,7 +1444,7 @@ mod job_history_tests {
         Ok(pool)
     }
 
-    async fn delete_jobs(pool: &PgPool, ids: &[i32]) -> anyhow::Result<()> {
+    async fn delete_jobs(pool: &PgPool, ids: &[i64]) -> anyhow::Result<()> {
         for id in ids {
             sqlx::query("DELETE FROM jobs WHERE job_id = $1")
                 .bind(id)
@@ -1435,7 +1460,7 @@ mod job_history_tests {
     #[allow(clippy::too_many_arguments)]
     async fn start(
         pool: &PgPool,
-        job_id: i32,
+        job_id: i64,
         name: &str,
         user: &str,
         account: &str,
@@ -1470,7 +1495,7 @@ mod job_history_tests {
     #[allow(clippy::too_many_arguments)]
     async fn start_with_qos(
         pool: &PgPool,
-        job_id: i32,
+        job_id: i64,
         name: &str,
         user: &str,
         account: &str,
@@ -1508,7 +1533,7 @@ mod job_history_tests {
 
     async fn end(
         pool: &PgPool,
-        job_id: i32,
+        job_id: i64,
         state: &str,
         exit_code: i32,
         end_time: DateTime<Utc>,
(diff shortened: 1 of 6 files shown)


crates/spurctld/src/accounting/reconcile.rs
```diff
@@ -93,7 +93,7 @@ async fn run_once(pool: &PgPool, cluster: &ClusterManager) {
         .map(|j| (j.job_id, accounting_expected_state(j.state)))
         .collect();
 
-    let job_ids: Vec<i32> = candidates.iter().map(|j| j.job_id as i32).collect();
+    let job_ids: Vec<i64> = candidates.iter().map(|j| j.job_id as i64).collect();
     let (accounting_states, unknown): (HashMap<JobId, AccountingRowState>, HashSet<JobId>) =
         match db::job_accounting_states(pool, &job_ids).await {
```
shiv-tyagi, line 98: Converting DB `i64` keys back to `JobId` with `as` on the next line silently maps pre-fix corrupted rows (e.g. `-1` stored for `u32::MAX`) onto the wrong in-memory id. Filter with `try_from` and skip invalid rows so reconciliation never acts on a wrapped id.

```suggestion
                rows.into_iter()
                    .filter_map(|(id, row)| {
                        u32::try_from(id)
                            .ok()
                            .map(|job_id| (job_id, row))
                    })
                    .collect(),
```

crates/spurctld/src/accounting/db.rs
```diff
@@ -292,13 +317,13 @@ pub async fn record_job_start(conn: &mut PgConnection, rec: &JobStartRecord) ->
 #[allow(clippy::too_many_arguments)]
 pub async fn record_job_end(
     conn: &mut PgConnection,
-    job_id: i32,
+    job_id: i64,
```
shiv-tyagi: Nit: `record_job_end` takes `i64` while every caller holds a `JobId`/`u32` and casts at the call site. Take `JobId` here (same as `JobStartRecord`) and bind `job_id as i64` once inside the function.

```suggestion
    job_id: JobId,
```

shiv-tyagi (commented):
Posted comments, PTAL

**PR-level notes:**

- Rolling back to a pre-fix controller after this migration will fail (old binary decodes `BIGINT` columns as `i32`). Consider `fix(spurctld)!:` in the PR title per AGENTS.md.

- The PR description notes the first upgraded controller holds `ACCESS EXCLUSIVE` on `jobs` during `ALTER COLUMN TYPE`, and that downgrade is unsupported. A short note in `docs/deployment/upgrading.rst` or `docs/admin-guide/accounting.rst` would help operators planning the rollout.

crates/spurctld/src/accounting/reconcile.rs
nikhilsk: Went the other way here, because `try_from` + skip would drop rows rather than protect them. The ids in that map come back from the list we queried (`WHERE job_id = ANY($1)`, bound from our own `JobId`s), so nothing outside `u32` can be returned and the cast cannot wrap. And a skipped row is not inert: it is absent from `accounting_states`, which `jobs_needing_resync` reads as *missing*, so `resync_job` calls `write_start` and resets a possibly-correct terminal record to RUNNING - the failure mode the `Err` branch above deliberately avoids by routing to `unknown`.

On the pre-fix rows specifically: a `u32` narrowed to `i32` and back preserves the low 32 bits, so `-1` decodes to `u32::MAX`, the id the job actually had. `try_from` would reject exactly the rows this widening is meant to keep readable.

Instead the cast is gone from reconciliation entirely - `job_accounting_states` now takes `&[JobId]` and keys each row by the id it was queried with. The one remaining narrowing is `decode_job_id` in `db.rs`, named and commented, with a DB test asserting a row stored as `-1` reads back as `u32::MAX`.

crates/spurctld/src/accounting/db.rs
nikhilsk: Done, and carried through the rest of the module: `record_job_end` takes `JobId` (and `Option<JobId>` for `preempted_by`), `job_accounting_states` takes and returns `JobId`, and `JobRecord` carries `JobId` so `grpc.rs` no longer re-narrows with `as u32`. `i64` now appears only at bind and decode sites inside `db.rs`.

crates/spurctld/src/accounting/db.rs
```diff
@@ -352,28 +377,32 @@ pub struct AccountingRowState {
...
-    job_ids: &[i32],
-) -> anyhow::Result<HashMap<i32, AccountingRowState>> {
+    job_ids: &[JobId],
+) -> anyhow::Result<HashMap<JobId, AccountingRowState>> {
     if job_ids.is_empty() {
         return Ok(HashMap::new());
     }
+    // Keyed by the stored form so a returned row is decoded by the id it was
+    // queried with, never by narrowing whatever the column holds.
+    let queried: HashMap<i64, JobId> = job_ids.iter().map(|&id| (id as i64, id)).collect();
```
yansun1996: Reads and lookups end up disagreeing about a row written before the widening. `get_job_history` heals it — a stored `-1` decodes back to `u32::MAX` via `decode_job_id` — but this binds only the positive widened id, so `job_id = ANY($1)` never matches it. The reconciler then reads that as missing (`jobs_needing_resync` → `resync_job` → `write_start`), and `record_job_start` upserts at the *positive* id, so instead of repairing the legacy row it inserts a second one; both then decode to the same id in `sacct`, and the new row's eventual end write charges usage again.

It needs a cluster that already issued ids above `i32::MAX`, so this isn't urgent — but the migration is the free moment to fix it properly, by rewriting the data rather than decoding at read time. Something like `UPDATE jobs SET job_id = job_id + 4294967296 WHERE job_id < 0` (plus `preempted_by` and `tres_usage.job_id`) run *after* each `ALTER ... TYPE BIGINT`, guarding against the case where the positive id already exists. That would let `decode_job_id` go away entirely and remove the asymmetry rather than documenting it.

If you'd rather defer, could the comment above say so explicitly? As written it reads as a complete rationale.

Separately, a small one: since every returned `job_id` is by construction an element of `stored`, the `HashMap` inversion and the `?` on line 398 are both unreachable — `job_ids.iter().map(|&id| id as i64)` plus `decode_job_id` on the way out is equivalent and avoids a per-pass map.

crates/spurctld/src/accounting/db.rs
```diff
@@ -158,10 +158,35 @@ ALTER TABLE associations ADD COLUMN IF NOT EXISTS grp_submit_jobs INTEGER;
 ALTER TABLE accounts ADD COLUMN IF NOT EXISTS grp_tres TEXT;
 ALTER TABLE qos ADD COLUMN IF NOT EXISTS preempt TEXT NOT NULL DEFAULT '';
 ALTER TABLE qos ADD COLUMN IF NOT EXISTS preempt_exempt_time INTEGER;
-ALTER TABLE jobs ADD COLUMN IF NOT EXISTS preempted_by INTEGER;
+ALTER TABLE jobs ADD COLUMN IF NOT EXISTS preempted_by BIGINT;
 ALTER TABLE jobs ADD COLUMN IF NOT EXISTS preempt_mode TEXT NOT NULL DEFAULT '';
 ALTER TABLE jobs ADD COLUMN IF NOT EXISTS preempt_qos TEXT NOT NULL DEFAULT '';
 
+-- Job ids are u32 in the scheduler but these columns were INTEGER, so an id above
```
yansun1996: Could this be trimmed to a couple of lines? The content is good, but we try to keep comment blocks to two lines and push the longer reasoning into the commit message. Something like: `-- job_id is u32 but these columns were INTEGER, so ids above i32::MAX wrapped negative onto` / `-- unrelated rows. Guarded: ALTER TYPE rewrites the table under ACCESS EXCLUSIVE.`

crates/spurctld/src/accounting/db.rs
```diff
@@ -425,10 +454,17 @@ async fn update_usage(
     Ok(())
 }
 
+/// A stored id back to the in-memory `JobId` it was written from. Rows written
```
yansun1996: Same two-line trim here if you don't mind — the second and third lines carry the whole point, and the first is a sentence fragment. Also applies to the block comments at the top of the two new tests below (lines 1756 and 1808).

crates/spurctld/src/accounting/db.rs
```diff
@@ -1713,6 +1749,155 @@ mod job_history_tests {
...
+    #[ignore = "requires DATABASE_URL and PostgreSQL"]
+    async fn a_row_written_before_the_widening_reads_back_under_its_real_id() -> anyhow::Result<()>
+    {
+        let pool = test_pool().await?;
+        // What a pre-fix controller left behind: the id narrowed to 32 bits, which
+        // for u32::MAX was stored as -1. Widening the column does not rewrite it,
+        // so history has to keep decoding it to the id the job actually had.
+        let id = u32::MAX;
+        let user = format!("spur_legacy_id_{}", std::process::id());
+        delete_jobs(&pool, &[id]).await.ok();
```
yansun1996: This deletes `4294967295`, but the row the test inserts on the next line is stored as `-1`, so it won't clear a leftover from a previous failed run and the insert would then hit the primary key. Worth deleting both forms here.

crates/spurctld/src/accounting/db.rs
```diff
@@ -1713,6 +1749,155 @@ mod job_history_tests {
...
+    }
+
+    #[tokio::test]
+    #[ignore = "requires DATABASE_URL and PostgreSQL"]
+    async fn migrate_widens_pre_fix_integer_job_ids() -> anyhow::Result<()> {
+        let pool = test_pool().await?;
+
+        // Reconstruct a pre-fix database by narrowing the three id columns back,
+        // then let migrate() perform the upgrade an existing cluster would see.
+        sqlx::query("ALTER TABLE jobs ALTER COLUMN job_id TYPE INTEGER")
```
yansun1996: This narrows the shared schema and never widens it back on an early return, so any failure between here and the end of the test leaves the database narrow for everything that follows. CI passes `--test-threads=1` so the parallel case is covered, but a developer running `cargo test -- --ignored` locally would hit it. Could the restore be put in a guard so it runs on the failure path too?

docs/admin-guide/configuration.rst
```diff
@@ -190,7 +190,8 @@ and Raft high-availability topology.
      - ``999999999``
      - Not implemented
      - Intended as the job-ID wrap point. No code consumes it; the counter does
-       not wrap.
+       not wrap. Job IDs are 32-bit unsigned and stored as 64-bit in the
```
yansun1996: "so the full range is usable without one" is a bit stronger than the code supports. The allocator is `AtomicU32::fetch_add`, which wraps silently past `u32::MAX` in both profiles, and on restart `next_job_id` is clamped with `first_job_id.max(snap.next_job_id)`, so after a wrap ids get re-issued and collide with existing accounting rows. Maybe something like: "...so ids above `i32::MAX` are recorded correctly; the counter still wraps to zero past `u32::MAX` and would re-issue ids."

yansun1996 (commented):
Nice fix — widening rather than enforcing the ceiling is the right call, and threading `JobId` through so `i64` only appears at the bind and decode sites reads much better than the previous mix. I checked the proto boundary (`JobInfo.job_id`, `preempted_by`, and both `RecordJob*Request`s are `uint32`) and this diff removes casts there rather than adding any, and I couldn't find any remaining writer or reader of `jobs`/`tres_usage` still binding a job id as `i32`. No Raft/WAL-persisted type is affected.

Two things I'd like to see before this merges, plus some smaller notes inline.

**Operator documentation for the migration.** The PR description explains the one-time table rewrite and the fact that downgrade is unsupported, but neither lands in the docs. I confirmed the downgrade claim: an older controller decodes `job_id` as `i32` and binds `&[i32]`, so `get_job_history` and `job_accounting_states` fail outright against a migrated database (writes survive, since Postgres widens an `int4` parameter). There's currently nothing anywhere under `docs/` about Postgres schema migration at all, so `docs/deployment/upgrading.rst` under "Safe Upgrade Order" seems like the natural home for a short note: the first upgraded controller rewrites `jobs` under `ACCESS EXCLUSIVE` (duration scales with row count, accounting writes block meanwhile), and rolling back to a pre-migration controller is not supported. Given that last part, `fix(spurctld)!:` seems justified for the title.

**Legacy rows after the widening.** See the inline note on `job_accounting_states` — reads heal a pre-fix narrowed row but lookups don't, so the two paths disagree about that row's identity. Low reachability, and I'm happy for it to be a follow-up, but the migration is the one cheap moment to close it.

Minor, and entirely optional: the migration transaction has no `lock_timeout`, so if an older replica is holding an open transaction on `jobs` the rewrite queues behind it and blocks rather than errors — and since startup awaits `migrate`, the "accounting unavailable" degrade path never fires. `SET LOCAL lock_timeout` inside that transaction would make it fail fast and retry next start. That's pre-existing behaviour that this change just makes more expensive, so a follow-up is fine.

crates/spurctld/src/accounting/db.rs
nikhilsk: Fixed by rewriting the data in the migration, as suggested. After each `ALTER ... TYPE BIGINT`, the migration maps negative ids back to their true value:

```sql
UPDATE jobs SET job_id = job_id + 4294967296
WHERE job_id < 0
  AND NOT EXISTS (SELECT 1 FROM jobs j WHERE j.job_id = jobs.job_id + 4294967296);
```

plus `jobs.preempted_by` (no uniqueness, so no guard needed) and `tres_usage.job_id` (guarded on its `(job_id, tres_type)` key). All of it runs inside the existing single migration transaction, so a partial rewrite rolls back cleanly and can't leave the table half-migrated.

On the guard: I made it **skip** the negative row when the true id already exists, rather than fail or delete. A genuine pre-fix `INTEGER` column physically cannot hold the true id (it's `> i32::MAX`), so a real collision can only come from a hand-edited/corrupted table — failing the migration would wedge controller startup, and deleting would drop accounting data, so skipping the stray negative row and leaving the authoritative positive row in place is the safe default.

With the data rewritten, `decode_job_id` is removed and both read and lookup paths are plain `i64` → `u32` conversions, so the asymmetry is gone rather than documented.

Also took the `job_accounting_states` simplification — since every returned `job_id` is by construction an element of the bound list, the `HashMap` inversion and its `?` are gone; it now binds `job_ids.iter().map(|&id| id as i64)` and converts back with a plain `as` on the way out.

New/reworked DB tests cover it: a legacy `-1` row is rewritten to its true id and found by the reconciler's lookup (`migrate_rewrites_a_legacy_narrowed_row_to_its_real_id`), and a legacy row whose true id already exists is left untouched by the guard (`migrate_skips_a_legacy_id_whose_real_id_already_exists`).

---

crates/spurctld/src/accounting/db.rs
nikhilsk: Done — trimmed to your two lines:

```
-- job_id is u32 but these columns were INTEGER, so ids above i32::MAX wrapped negative onto
-- unrelated rows. Guarded: ALTER TYPE rewrites the table under ACCESS EXCLUSIVE.
```

---

crates/spurctld/src/accounting/db.rs
nikhilsk: The `decode_job_id` block at ~457 is gone entirely now (the migration rewrites the data), so that comment went with it. The two test block comments are trimmed to two lines each.

---

crates/spurctld/src/accounting/db.rs
nikhilsk: Fixed. The reworked tests clear leftovers by `user_name` rather than by a single id form, so a row left behind as either `-1` or the true id from a failed run is cleared before the insert.

---

crates/spurctld/src/accounting/db.rs
nikhilsk: Fixed. Both schema-narrowing tests now run the body separately and restore the columns to `BIGINT` unconditionally afterwards (`outcome.and(restore)`), so a local `cargo test -- --ignored` that fails midway still leaves the shared schema wide for later tests.

---

docs/admin-guide/configuration.rst
nikhilsk: Reworded close to your suggestion:

> Job IDs are 32-bit unsigned and stored as 64-bit in the accounting database, so ids above `i32::MAX` are recorded correctly; the counter still wraps to zero past `u32::MAX` and would re-issue ids that collide with existing rows.

---

shiv-tyagi (approved):
LGTM

yansun1996 (approved):
Went through the migration carefully — the column coverage, the rewrite arithmetic, the ALTER-before-UPDATE ordering, and the `NOT EXISTS` collision guards all look right, and it stays safe to re-run and on a fresh database. Rewriting the legacy rows rather than decoding them on read was the better fix. Approving.

PR #754, opened 2026-08-26
feat(spurd)!: supervise every job with spurstepd by default

yansun1996:
## What this does

Makes `spurstepd` supervision the default for every job **and every numbered
`srun` step**, so restarting or upgrading `spurd` no longer kills the work on a
node. The agent stops being the parent of user processes: it spawns a
supervisor per job — and one per step inside that job — via double-fork +
`setsid`, reparented to init, owning the process tree, its cgroup and its exit
status. On restart `spurd` rediscovers the sessions on disk, reconnects over
each supervisor's Unix socket, and adopts them.

An allocation stays held, a running step keeps running and still reports its
exit status over the reconnect, and a terminal opened inside an allocation is
handed back to the restarted agent.

## Approach

Supervisor state is keyed by `(job, step)`. Job-level operations — signal,
cancel, teardown — fan out over every step a job holds. Widening that key
exposed three teardown defects that were indistinguishable from correct while a
job held exactly one supervisor: releasing a supervisor tore down the whole job
(freeing CPU/memory/GPU under a live sibling), the release record was only
written when that teardown ran (so a step exiting beside a sibling left a
session directory pruning skipped forever), and the "last step" test ignored
`run_attempt` (so a superseded attempt's session stranded the node's resources).

### On-disk layout

Each supervisor owns one flat session directory. Nothing is nested under
another supervisor, so one exiting never removes another's state:

```
<state_dir>/runtime/                          0700
└── <job>.<attempt>.<step>/                   0700   one per supervisor
    ├── descriptor.json      0600   supervisor pid + start ticks, capability,
    │                               cgroup path, namespace flags, workload pid
    │                               + start ticks, job resources
    ├── launch.json          0600   launch spec (carries the job environment)
    ├── obligations.jsonl    0600   exit observed / completion acked /
    │                               resources released — what makes a session
    │                               prunable exactly once
    ├── failure.txt          0600   why a supervisor never launched
    ├── runtime.sock                agent → supervisor control
    ├── agent.sock                  supervisor → agent completion notify
    └── ptyfd.sock                  pty master custody (interactive only)
```

`<step>` is the numeric step id: reserved sentinels for the batch, extern and
interactive steps, and `0, 1, 2, …` for numbered `srun` steps.

Job output keeps its own spool, and steps no longer collide there — every step
used to write `spur_job.sh`, corrupting the batch script of the job it was
running inside:

```
/var/spool/spur/job<id>/            (fallback: $TMPDIR/spur/job<id>/)
├── spur_job.sh                     the job-owning step's script
├── spur_step<N>.sh                 one per numbered step
└── stdout / stderr
```

Cgroups mirror the same shape. cgroup v2 forbids processes in a node whose
children have controllers enabled, so limits live on the job and each step's
processes sit one level down — the shape Slurm uses:

```
/sys/fs/cgroup/spur/
└── job_<id>_<attempt>/             limits live here
    ├── step_<batch>/               the job's own step
    └── step_<N>/                   one leaf per numbered step
```

### Starting a job

The supervisor holds the workload at a start gate. The agent leaves that gate
shut when `LaunchJob` returns, and the controller opens it with a new `StartJob`
RPC issued only after the `Running` commit returns:

```
controller                     agent                     supervisor
    │  LaunchJob ─────────────▶ │  spawn ────────────────▶ │ (gate shut)
    │ ◀──────────── confirmed   │                          │
    │  commit Running (Raft)    │                          │
    │  StartJob ──────────────▶ │  open gate ────────────▶ │ exec workload
```

Previously the agent opened the gate during `LaunchJob`, so the script was
already running while the cluster still called the job Pending — an `srun` on
its first line asked about a job that had not started and was refused. A node
that cannot be released leaves its supervisor at the gate, so the run is
cancelled rather than left Running with nothing behind it; the existing gate
timeout bounds a controller that dies in between.

### Fencing

A recorded cgroup path is authoritative: the retrying `rmdir` only succeeds once
the cgroup is empty, which is what proves the processes are gone. Without a
cgroup there is nothing to watch, and killing the supervisor proves nothing —
it ignores a raw SIGTERM by design, and the workload is a separate process the
signal never reached. The supervisor now records the workload's pid and start
ticks, and that path escalates to SIGKILL on the tree and confirms the process
is actually gone. A zombie counts as gone; a recycled pid is somebody else and
is neither killed nor waited on.

## Recovery

A restarting agent reports the supervisors it re-adopted and the controller
decides whether to keep or fence them. Refusing outright was the unsafe
direction — the supervisor and its job are still running, so rejecting the
report abandons them and the job strands in `RUNNING`.

How a report is authorized depends on admission mode, because that is what
decides whether a node can prove who it is:

- **`admission.mode = "token"`** — the agent carries a node token minted at
  registration, where it had to present a valid join token. The report is
  verified, and missing, forged, expired or mismatched tokens are rejected.
- **Open admission (the default)** — a caller asserts a hostname rather than
  proving one, so a credential minted there attests nothing. An untokened report
  is still served, so a restarted agent keeps its supervisors, but it is treated
  as unproven and **cannot fence**: an unproven caller may not requeue or cancel
  a job. Fencing an incomplete cohort requires token admission.

Node tokens are therefore minted only under token admission. Minting them under
open admission would have handed any caller that can reach the controller a
credential for a node it does not own, and with it the ability to requeue and
cancel another user's jobs.

Adoption restores what the launch decided and the supervisor cannot re-derive:
cgroup path, pid/user/mount namespace flags, and the job's resources. That last
one matters more than it looks — a step reads its GPU set from the job it joins,
and an empty set is not "unset": it is read as *no devices* and pins every
visibility variable to `-1`. Dropping it did not leak GPUs, it denied them, so
GPU work inside an allocation stopped working after an agent restart.

## MPI

A `--mpi=pmix` job's PMIx server used to run inside `spurd`, so restarting the
agent killed the rendezvous even though the supervisor and the ranks survived.
Open MPI fences inside `MPI_Finalize`, so the job computed to completion and
then hung at teardown — full runtime burned, no result, and an error pointing at
MPI rather than the upgrade.

The server now runs inside the supervisor that owns the step, one per
`(job, step)`, which is where Slurm puts it too. The rendezvous is keyed the same
way so two steps of one job do not share a namespace or a port.

Ports are a hash into a fixed span, so an unlucky pair of jobs can still collide.
A frame is now checked against the job and step it names before any control flag
is acted on, so the loser fails alone with a named error instead of aborting the
job that legitimately owns the port.

A node that cannot load the plugin now refuses the dispatch for a job as well as
a step, so the failure is reported rather than arriving as a killed workload.

## Breaking changes

- **Session format and ownership change.** A supervisor started by an older
  build is not adopted by this one.
- **`NodeInfo.agent_port` uses tag 58.** Tags 56/57 belong to `reason_uid` /
  `reason_time` on `main`.
- **The MPI plugin must be replaced with the binaries.** `spur_mpi_pmix.so`
  goes from ABI 3 to 4, and the modex wire format from v2 to v3. A plugin left
  at the old version does not load, and a node still speaking v2 is rejected by
  one speaking v3 — so the `.so` has to be swapped everywhere, not rolled.
- **The modex rendezvous is keyed on `(job, step)`.** The namespace is now
  `spur.<job>.<step>` and the port derives from both, because each step hosts
  its own server. Two nodes on different builds never meet.

## Upgrading

**This upgrade needs an empty cluster.** Drain every node and let jobs finish or
cancel them before swapping binaries. Sessions from the previous build are not
adopted, the two builds disagree on where a step's processes live, and a new
controller dispatching to a not-yet-upgraded agent tears the job down. Upgrade
every controller and agent in one window, and replace `spur_mpi_pmix.so` at the
same time.

`spurstepd` is a new binary and must sit next to `spurd` on every compute node —
the agent resolves it beside its own executable and does not search `$PATH`, so
a node without it fails every launch. Restart survival also needs
`KillMode=process` in the `spurd` unit; the systemd default kills the whole
cgroup on stop and takes the supervisors with it.

After this lands, restarting `spurd` no longer kills the work on a node.
Draining stays the right order because it keeps new work off a node mid-swap,
but it is no longer what protects running jobs from the restart.

## Testing

`cargo clippy --workspace --exclude spur-ffi --all-targets` and `cargo fmt` are
clean, and the workspace suite passes. Every behavioural fix has a test
confirmed to fail against the unfixed code — including the recovery, fencing and
epoch fixes added here, and the launch-failure reporting, cgroup-adoption and
session-sweeping fixes, each checked red against its own unfixed behaviour.

CI is green on this head, **including both GPU shards** — the only place GPU
behaviour is exercised end to end, since the lab nodes have no GPUs. Beyond the
existing device-isolation coverage, this adds three cases: a job's GPU set
surviving adoption, which nothing tested even though restoring it is what keeps
GPU work running across a restart; a zero-GPU job staying denied afterwards; and
the modes on the session records, which hold a bearer credential.

### Live validation

Run on a four-node deployment against this branch's build. Adoption is checked by
**identity, not count** — a supervisor's process id *and* process start time are
compared across every restart, since a supervisor killed with the agent and
respawned would satisfy any "still one running" check.

| What it checks | How it was tested | Result |
| --- | --- | --- |
| A job spanning two nodes gets one supervisor per node | Submit a two-node job. On each node, list supervisor processes and their on-disk records. | Exactly one per node, each with its own resource-limit group |
| A running job survives the node agent restarting | Submit a 10-minute job. Note its supervisor's process id and start time. Restart the agent. Re-read both. | Identical — adopted, not relaunched. Job ended `COMPLETED 0:0` |
| No output is lost or duplicated across that restart | Job prints a numbered line every second for 120s. Afterwards count lines, count distinct numbers, scan for corruption. | 120 lines, 120 distinct, first line exactly once, no corrupt bytes |
| An `srun` step gets its own supervisor | Run a job whose script launches one step. List supervisors and limit groups on that node. | Two supervisors (job and step), one limit group each |
| Both supervisors survive an agent restart | Restart the agent mid-step; compare both supervisors' ids and start times. | Both unchanged; the step finished and its exit status reached the script |
| A job array runs as independent jobs | Submit a three-task array across two nodes. Check each task's supervisor and identity. Restart one agent mid-flight. | Three distinct supervisors, correct per-task identity, adopted on restart, all `COMPLETED 0:0` |
| A memory-capped job is actually capped | Submit with a small memory cap; confirm the limit is applied, then exceed it. | Limit applied exactly; exceeding it is killed and reported as out-of-memory |
| An interactive allocation is held across a restart | Hold an allocation, restart the agent, check its state and supervisor identity, then run a step inside it. | Allocation stayed running, supervisor unchanged, new step ran |
| A step's environment is unchanged by a restart | Dump a step's environment, restart the agent, run the same step, diff the two. | Partition, node list, cpu and task counts identical; only the step number differs, correctly |
| Running a command inside a running job | Start a long job, then exec into it. Check the user and which limit group it lands in. | Runs as the job's own user, inside the job's own limit group |
| Attaching to a live job's output | Start a job producing output; attach from another session. | Streams immediately |
| Attaching after the agent was restarted | Same job, restart the agent, attach again. | Returns the backlog instantly and keeps streaming, including lines written before the restart |
| A multi-node MPI job runs | Four ranks across two nodes; check rank identity and clean shutdown. | Four ranks, correct numbering, clean finish, `COMPLETED 0:0` |
| An MPI job launched without an inner step | Same, launched directly by the job. | Supervised, with on-disk state, completes |
| An MPI job survives a restart mid-computation | 150-iteration MPI job; restart one agent at iteration ~45. Check supervisor identity, every iteration, and which process hosts the MPI rendezvous. | Supervisor unchanged, all 150 iterations on all four ranks with no gaps, and the rendezvous is hosted by the **supervisor**, not the agent |
| An MPI step inside a job survives a restart | Same, as an inner step. | Both supervisors unchanged, all iterations present, status reached the script |
| Two MPI jobs at once do not collide | Run two multi-node MPI jobs simultaneously; compare rendezvous names and ports on both nodes. | Distinct names, distinct ports |
| A job that exits non-zero is reported failed | Script ends `exit 7`. | `FAILED`, exit `7:0`, reason `NonZeroExitCode`; accounting agrees |
| A job's exit code and kill signal are faithful | Signal the job's own process; check the live view and the accounting record. | `exit 7`→`7:0`, kill→`0:9`, out-of-memory→`OUT_OF_MEMORY`/`0:9` |
| A step that finishes while its agent is down still reports | Stop the agent, let the step finish while it is down, restart. Repeat at several outage lengths. | Result delivered 1–2s after the agent returns |
| Same, with a step that fails | Step exits non-zero while the agent is down. | Failure preserved, not laundered into a success |
| An `srun` on the script's first line is not refused | Script's first line launches a step; the second asks the cluster what state the job is in. Three times. | 3/3 — step ran, and the job saw itself as running |
| Cancelling a multi-node job tears it down everywhere | Cancel a two-node job; poll both nodes for leftover processes, state and limit groups. | `CANCELLED`, fully torn down on both nodes, the two finishing within 0.1s of each other |
| A finished job leaves no limit group behind | Check after normal completion. | None left |
| A cancelled allocation leaves no limit group behind | Check after cancelling an allocation. | None left |
| On-disk job records are not world-readable | Inspect permissions of every supervisor record across nine sessions. | Every record and session directory owner-only |
| Replacing the agent binary then restarting keeps jobs alive | Genuinely replace the binary on disk (verify the file identity changed), restart, compare supervisor identity. | File genuinely replaced, supervisors unchanged, jobs completed |
| Settled supervisor records are cleaned up | Track on-disk session counts across the campaign. | Peaked at four per node, decayed to zero, aged out at ten minutes as documented |
| A job whose agent, supervisors and workload are all killed | Kill the agent, every supervisor and the workload on one node, then restart the agent immediately. Poll the job. | Converges to `FAILED 0:9` in **three seconds**; node returns idle with no phantom capacity |
| A job that cannot be released to start is moved on | Kill supervisors the instant they park at the start gate, then submit a job. | Resolves in ≤2s, names the node, and the node returns to idle |
| A node hard-rebooted mid-job converges | Submit a job, hard-reboot its node, poll until the job resolves, then bring the node back. | `NODE_FAIL` within the heartbeat timeout; node rejoins idle and is immediately usable |
| Three controllers survive losing their leader | Stand up three controllers, elect a leader, stop it gracefully; repeat with a hard kill. Seven kills. | New leader in ~5s every time. No split-brain, no stuck election, no corruption; reads never interrupted |

### Not covered

Bullets marked **Pre-existing** describe behaviour this PR inherits rather than
introduces; the rest are new here.

- **Pre-existing — duplicate execution after leader failover.** `LaunchJob`
  still precedes the `JobStart` commit, so dispatch is not replicated before the
  agent is called. The agent now absorbs the common case: a repeat `LaunchJob`
  for a `(job, step, attempt)` it already tracks is answered as success without
  relaunching, and a failover redispatch recomputes the same attempt. The `--pty`
  path skips supervision and so still has only the job-scoped in-flight
  reservation.
- **Pre-existing — proof of containment is opt-in.** The launch path verifies the
  workload landed in its cgroup and refuses a step whose job cgroup is missing,
  but both are gated on `[cgroup] required`, which is off by default. Under the
  default such a step logs and runs uncontained.
- **Adoption trusts most of what it restores.** A recorded cgroup path that no
  longer exists is now dropped rather than treated as authoritative for
  teardown, but nothing confirms the cgroup that *does* exist still holds the
  recorded pid.
- **One launch stays unsupervised:** a step given its own `--container-image`.
  It ends when the agent stops. A standalone `srun --pty` that allocates its own
  job is also lost across a restart, but for a different reason — the terminal
  is held and could be reattached, yet `srun` cancels the job it created as soon
  as its connection drops. A terminal opened *inside* an allocation survives,
  and containerized *jobs* are supervised.
- **Pre-existing — a container step leaks its unpacked rootfs** when the agent is
  hard-killed. Cleanup is `Drop`-based and nothing sweeps the container
  directory at startup; this behaves identically on `main`.
- **Pre-existing — `WalOperation::JobNodeComplete` carries no `run_attempt`**, so
  its staleness check is applied at propose time but not on replay or on a
  follower. The field can be added additively the way `JobStart` already does
  it; it wants its own change.
- **A supervisor that fails before its workload starts is reported, but not
  richly.** The reason now reaches the job's own error stream and the agent's
  log, and a `--mpi=pmix` job is pre-flighted so an unloadable plugin refuses the
  dispatch instead of killing the workload. What is still missing is a reason on
  the completion report itself, so `scontrol` shows the exit without it. That is
  a proto change of its own.
- **Pre-existing — MPI is not covered by CI.** The e2e artifact does not ship
  `spur_mpi_pmix.so`, so 17 of 18 `test_mpi.py` cases skip there, identically to
  `main`. The exception deploys deliberately without the plugin and does run.
  The PMIx work in this PR was validated on a two-node deployment instead, which
  means the move of the server into the supervisor ships with no CI-executing
  coverage of its own. Getting the plugin into the artifact would turn a dozen
  existing tests back on, and is worth doing separately.
- **Pre-existing — two jobs on a node can collide on a modex port.** The port derives from job
  identity into an 8000-slot span; that collision domain predates this change,
  where job ids exactly 8000 apart already collided. Re-keying per `(job, step)`
  — needed so two steps of one job do not share a server — replaces that narrow
  condition with a hash over both. Collisions stayed unreachable in practice —
  none was produced across a thousand consecutive job ids — but the domain is
  real rather than impossible. In exchange they stop
  being dangerous: every frame is validated against job and step before any
  control flag is honoured, so the loser fails alone with a named error and the
  incumbent is untouched, verified by forcing a real collision. Removing the
  collision domain entirely means relaying through the node agent on its fixed
  port, which is tracked separately.
- **Settled step sessions stay on disk for ten minutes**, each holding the job's
  environment and script, owner-only and capped at 1024 oldest-first. A session
  nothing can ever finalize — one whose supervisor died before recording an exit,
  or whose descriptor will not load — is swept after a day rather than kept
  forever, but a node whose agent never returns keeps whatever is left.
- **`--pty` has no cgroup leaf of its own.** It borrows the lifetime step's
  session for pty custody and runs in that step's leaf, while numbered steps get
  their own. It is never in the active-step map, so `CancelStep` for it is a
  no-op; cancelling the job still works.
- **Pre-existing — the cgroup and spool roots are hardcoded and shared.** Reaping
  reconstructs paths from job identity alone, so two instances on one node can
  reach each other's cgroups. This PR makes the cgroup path authoritative for
  fencing, which turns a cross-instance reap from a delete into a kill.
- **Pre-existing — `srun -l` and any `-n>1` step report exit 0.** The task
  launcher pipes through `sed` without `pipefail` and waits bare, so the step's
  real status is lost. Called out because exit-status fidelity is this PR's
  headline everywhere else.
- **Pre-existing — `[[nodes]] gres`, `cpus` and `memory_mb` are dead config.**
  Node config application only applies `features`, `weight` and `address`, so
  those keys never reach the node. Scheduling is unaffected — GRES comes from the
  agent's own detection — but the keys silently do nothing.
- **A failed release does not name the node that caused it.** The warning logs
  the full dispatch count rather than the number actually attempted, and
  `launch_failure_detail` names every dispatch node rather than the ones that
  refused. A node that was never contacted — one whose comm address did not
  resolve between the peer-list build and the release — therefore reads
  identically to one that refused. The eviction is correct either way; only the
  diagnosis is ambiguous, and the test covering it passes vacuously on a
  single-node job.
- **A step re-attach discards the error it failed on.** `reawait_step` retries
  every `await_step` failure for its full budget and then reports a generic
  `unavailable`, so a permanent rejection costs a minute and arrives without its
  cause. `not_found` is deliberately retryable — an agent that has just restarted
  answers that for a step it is about to adopt — but `permission_denied` and
  `unauthenticated` want to fail fast, and every attempt wants to log what it
  saw.
- **A backwards clock step can drop a settled step's recorded exit.** The
  retention check on a finalized session uses `duration_since` rather than its
  saturating form, so a `settled_at` in the future — an NTP correction, a
  restored VM — reads as expired and the session is pruned with its exit inside
  it. A caller that then asks for the step gets `not_found`, which is what turns
  a successful step into a synthesised failure. The in-memory sibling already
  saturates; the on-disk one should match.


crates/spurd/src/main.rs
```diff
@@ -110,6 +98,32 @@ struct Args {
...
 }
 
+/// Parses a stepd directory's `<job_id>.<run_attempt>` basename.
+/// Used to fence a session whose descriptor failed to parse — the job/attempt
+/// identity survives in the directory name even when its contents don't.
+fn parse_session_dir_name(path: &std::path::Path) -> Option<(u32, u32)> {
+    let name = path.file_name()?.to_str()?;
+    let (job_id, run_attempt) = name.split_once('.')?;
+    Some((job_id.parse().ok()?, run_attempt.parse().ok()?))
+}
```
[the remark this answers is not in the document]
yansun1996: Confirmed, thanks — `session_dir` names directories `<job_id>.<run_attempt>.<step_id>` (three components), so splitting on the first `.` broke run_attempt parsing for every corrupted descriptor, not just ones with a large step_id. Fixed to split into the first two components and require a third to be present.

crates/spurd/src/main.rs
```diff
@@ -345,7 +432,34 @@ async fn main() -> anyhow::Result<()> {
...
+        std::fs::create_dir_all(agent_notify_socket_dir)?;
+        #[cfg(unix)]
+        {
+            use std::os::unix::fs::PermissionsExt;
+            std::fs::set_permissions(
+                agent_notify_socket_dir,
+                std::fs::Permissions::from_mode(0o700),
+            )?;
+        }
+    }
```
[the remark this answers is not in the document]
yansun1996: Good catch — fixed by routing this through the same create-or-verify-private-directory helper already used for the stepd runtime root and session directories, instead of the ad hoc exists()-gated block. A pre-existing directory now gets its ownership/mode checked (and rejected if it's not private or is a symlink) rather than trusted.

crates/spurd/src/agent_server.rs
```diff
@@ -43,6 +44,390 @@ fn maybe_deny_gpu_env(env: &mut HashMap<String, String>, allocated_device_ids: &
...
+        Some(path) if path.exists() => path,
+        _ => std::path::PathBuf::from("spurstepd"),
+    }
+}
+
+/// Double-fork+setsid detach (Slurm's `slurmstepd` spawn pattern): the
+/// grandchild reports its real pid over a pipe before exec, since
+/// `Command::spawn`'s return is just the immediately-exiting intermediate
+/// child that lets the grandchild reparent to init.
+fn spawn_stepd_process(
```
biluriuday, line 71: Failure scenario: A production operator runs `systemctl restart spurd`. Double-fork and `setsid()` change parent/session ownership but do not leave `spurd.service`'s cgroup. With systemd's default `KillMode=control-group`, `spurstepd` is terminated too.

Impact: The primary restart/upgrade guarantee fails on Spur's documented production deployment. The next agent sees stale descriptors and tears down the job cgroups.

Action: Ship a supported unit with `KillMode=process` and `Delegate=yes`, and place supervisors/workloads in a persistent delegated scope or equivalent hierarchy independent of the agent process.

Regression test: Install the packaged unit, run batch/container/PMIx/allocation jobs, execute a real `systemctl restart spurd`, and assert supervisor and workload PID continuity.

crates/spurd/src/agent_server.rs
```diff
@@ -43,6 +44,390 @@ fn maybe_deny_gpu_env(env: &mut HashMap<String, String>, allocated_device_ids: &
...
+            .map_err(|error| executor::LaunchError::Other(error.into()))?
+            .map_err(|error| {
+                cleanup_unstarted_stepd(&store, config.job_id, run_attempt, launch_spec.step_id);
+                executor::LaunchError::Other(
+                    anyhow::Error::from(error).context("spawn stepd process"),
+                )
+            })?;
+    descriptor.pid = pid;
+    descriptor.process_start_ticks = crate::stepd::process_start_ticks(pid).unwrap_or(0);
+    if let Err(error) = wait_for_stepd(&descriptor).await {
```
biluriuday, line 243: Failure scenario: `spurstepd` launches user work before `spurd` has claimed the descriptor and acknowledged readiness. A `true`-like job can exit, close its socket, and notify before the agent finishes `wait_for_stepd()`.

Impact: Successfully executed work is reported as launch failure and can be requeued, causing duplicate external side effects.

Action: Use a two-phase protocol: `Ready` after socket/state initialization, then agent claim plus resource commit, then an explicit `Start` command that permits user code to execute.

crates/spurd/src/stepd.rs
```diff
@@ -0,0 +1,2375 @@
...
+        .map_err(|error| anyhow::anyhow!("invalid job id: {error}"))?;
+    let run_attempt: u32 = args[2]
+        .parse()
+        .map_err(|error| anyhow::anyhow!("invalid run attempt: {error}"))?;
+    let launch_spec: StepdLaunchSpec = serde_json::from_slice(&std::fs::read(&args[3])?)?;
+    if launch_spec.job_id != job_id {
+        anyhow::bail!("runtime launch spec job id does not match process arguments");
+    }
+    let step_id = launch_spec.step_id;
+    let agent_socket = state_dir.join(AGENT_NOTIFY_SOCKET_NAME);
```
biluriuday: Normal complete notifications target the wrong socket path

Failure scenario: `spurd` listens at `<state>/runtime/agent.sock`; `spurstepd` connects to `<state>/agent.sock` because only `StepdStore` appends `runtime`.

Impact: Every direct completion push fails, completion is delayed into watchdog/replay paths, and transient-only metadata such as epilog failure is lost.

Action: Derive the path in one canonical helper or pass the bound socket path explicitly in the launch payload.

Regression test: Run a short real supervisor job and assert immediate notification, controller completion, and resource release without watchdog or startup replay.

crates/spurd/src/agent_server.rs
```diff
@@ -119,3235 +504,5467 @@ async fn cleanup_completed_job_mpi(job_id: u32, mpi: &str, mpi_host: &MpiPluginH
...
+        request: Request<ReleasePmixRequest>,
+    ) -> Result<Response<ReleasePmixResponse>, Status> {
+        let job_id = request.into_inner().job_id;
+        if let Err(err) = self.mpi_host.release_prepared_pmix(job_id) {
+            warn!(job_id, error = %err, "PMIx prepare release failed");
+        }
+        Ok(Response::new(ReleasePmixResponse {}))
+    }
+
+    async fn cancel_job(
```
biluriuday, line 4120: `run_attempt` is not authoritative at cancellation sink

Failure scenario: A delayed cancel for attempt N arrives after attempt N+1 is tracked. The handler signals by `job_id` first; the requested attempt is consulted only later when releasing an uncommitted reservation. Heartbeat reclaim explicitly sends attempt 0.

Impact: A stale controller operation can terminate the current job. Suspend, step cancel, PMIx rollback, and heartbeat reconciliation have the same identity gap.

Action: Carry `{job_id, run_attempt, step_id}` through every control RPC and compare exact identity immediately before every signal, release, or cleanup. Do not let zero target a nonzero current attempt.

Regression test: Delay explicit cancel, graceful cancel, suspend, step cancel, PMIx release, and heartbeat reclaim for N until N+1 is live; N+1 must remain untouched.

crates/spurctld/src/scheduler_loop.rs
```diff
@@ -332,6 +332,11 @@ async fn process_assignment(
         }
     }
 
+    // `start_job_impl` advances the run epoch after the allocation has been
```
biluriuday, line 344: Dispatch and Completion generations are not durable in Raft

Failure scenario: Agents launch before a replicated dispatch intent exists. A leader can die after launch acknowledgement and the next leader can dispatch the same attempt to another cohort. Separately, `JobNodeComplete` omits the attempt, so its pre-proposal check cannot be repeated when Raft applies it.

Impact: The same attempt may execute twice, or an old completion may finalize and free resources belonging to a newer attempt.

Action: Add replicated `Dispatching` and `Fencing` states keyed by exact attempt/cohort. Make start metadata atomic, and append a defaulted attempt to the completion WAL operation with equality checked during state-machine apply.

crates/spurd/src/stepd.rs
```diff
@@ -0,0 +1,2375 @@
...
+pub struct StepdObligationLog {
+    path: PathBuf,
+}
+
+impl StepdObligationLog {
+    fn new(path: PathBuf) -> Self {
+        Self { path }
+    }
+
+    pub fn append(&self, obligation: &StepdObligation) -> io::Result<()> {
```
biluriuday, line 294: the completion journal has crash windows and a fragile tail

crates/spurd/src/stepd.rs:274-307,1101-1169

Failure scenario: The child wait status is consumed in memory, but `ExitObserved` is written only after socket removal, cgroup/rootfs/spool cleanup, epilog, and SPANK hooks. A power loss can also leave a partial JSONL tail that makes the whole log unreadable.

Impact: A successful job can be converted into synthetic failure/requeue, duplicate execution, or an agent startup failure. OOM and epilog-failure state are not replay-complete.

Action: Persist a checksummed phase journal immediately after wait: exit/OOM observed, cleanup pending/completed, epilog result, controller acknowledged, resources released. Tolerate only a torn final frame.

Regression test: Kill `spurstepd` at every phase and boot with every possible truncated final write; recovery must produce exactly one correct completion.

crates/spurd/src/agent_server.rs
```diff
@@ -119,3235 +504,5467 @@ async fn cleanup_completed_job_mpi(job_id: u32, mpi: &str, mpi_host: &MpiPluginH
 /// Job ids this node holds, shared with the reporter so heartbeats carry them.
 pub(crate) type RunningJobs = Arc<Mutex<HashMap<u32, TrackedJob>>>;
 
-/// Build an empty running-jobs map to share between the reporter and the agent.
-pub(crate) fn new_running_jobs() -> RunningJobs {
-    Arc::new(Mutex::new(HashMap::new()))
+#[derive(Clone)]
+pub struct StepdRecoveryCleanup {
+    running: RunningJobs,
```
biluriuday, line 800: The model can not grow to Slurm's one-supervisor-per-step design

Failure scenario: Runtime maps and recovery RPCs are keyed by job/attempt but not consistently by step; a future numbered Stepd would overwrite the batch/extern slot.

Impact: The stated follow-up stack requires another protocol/state-model refactor and cannot support simultaneous batch, extern, interactive, and numbered steps.

Action: Introduce one `RuntimeId { job_id, run_attempt, step_id }` now and use it in maps, notifications, probes, journals, and cleanup.

Regression test: Track batch, extern, and two numbered steps concurrently and independently signal/recover each one.

biluriuday (changes requested):
Overall very good job with the PR. Few points mentioned in-line. Apart from those, please make sure you add the `spurstepd` to the official CI and release atrifacts

crates/spurd/src/stepd.rs
yansun1996: Confirmed — and it turned out to be a regression from a late commit on this branch: the bind moved into the `runtime/` subdir and the connect didn't follow, so every direct push had been failing and completion was silently riding the replay loop instead.

Both sides now derive the path from a single `StepdStore::agent_socket()`. The test binds the path literally (`<state-dir>/runtime/agent.sock`) and drives the real notify function against it, so a future edit to one side can't drift from the other.

Your point about transient metadata was right too: the replay path passed `None` for the drain request where the push passes `epilog_failed`. The epilog result is now recorded as its own durable obligation, so a late-reported completion still drains the node.

Verified on a 2-node cluster: a job now completes ~150 ms after the supervisor starts with zero replay-loop hits; previously every completion waited out the retry.

crates/spurd/src/agent_server.rs
yansun1996: Confirmed. Two things made it worse than a race: the readiness-failure cleanup did `remove_dir_all` on the session, destroying the recorded exit, and the unmatched arm of the completion handler replied `Acknowledged` — telling the supervisor to prune a completion the agent had never reported. Both recovery routes were closed at once.

Fixed on three fronts: the accept loop is now `biased` so a queued agent connection is served before an inactive session ends the loop; cleanup keeps any session that already recorded an exit; and an unclaimed-but-unreported completion now gets `Deferred` so replay carries it.

I went with these rather than the full `Ready`/`Start` handshake. They remove the loss without adding a protocol state, and a two-phase start is a better fit alongside the per-step supervisor work — I'd rather design it once, there, than add a half-version here.

Worth noting the socket-path bug made this deterministic rather than rare, so fixing that first is what made this reproducible.

crates/spurd/src/stepd.rs
yansun1996: Confirmed, and the fragile tail was the more serious half. The reader `collect()`s into a `Result` and is called with `?` from `main()`, so a single torn line stopped the agent from starting at all — with no self-heal, since nothing rewrites or quarantines the file.

Changes: `ExitObserved` is written immediately after the wait, before socket removal, cleanup, the epilog, and SPANK; the epilog result is a separate later record so a late report still drains; and the reader tolerates exactly one unterminated final line while still failing loudly on corruption anywhere else.

I did not add per-frame checksums. Each append is a single small `write_all` followed by `sync_data`, so a torn final frame is the only realistic tear, and tolerating just that covers the reported impact without a format change. Happy to revisit if you'd rather have them.

Also fixed something adjacent that wasn't raised: OOM detection was wired only into the legacy monitor, so under a supervisor an OOM kill was reported as a bare SIGKILL. Testing that fix then surfaced a second issue in both paths — a job that exits 0 was being relabelled `OUT_OF_MEMORY` whenever anything in its cgroup had been OOM-killed, which is easy to hit now that a job with no `--mem` gets the `min_ram_mb` floor. The flag is now applied only to a job that actually failed, in both paths.

Verified live: restarting an agent against a hand-torn journal starts cleanly and still recovers the intact entry; a successful job under the memory floor reports COMPLETED while a genuine OOM still reports OUT_OF_MEMORY.

crates/spurd/src/agent_server.rs
yansun1996: Confirmed — and it's worse than the guarantee merely failing. The job's own cgroup is rooted outside the unit so the workload itself survives, but the next agent then sees a stale descriptor and reaps it. So a default-unit restart actively kills a job that was still running.

The documented unit now sets `KillMode=process`, with a note explaining why it's required rather than optional.

I've left the delegated-scope half as follow-up: `KillMode=process` alone restores the guarantee, and moving supervisors into their own persistent hierarchy is a larger change that deserves its own review. Flagged in the PR description so it isn't lost.

crates/spurd/src/agent_server.rs
yansun1996: Agreed on the direction — the design calls for one supervisor per running step and this ships one per job, which is why steps are declared as follow-up work.

I've taken the half that gets more expensive to defer: both new RPCs now carry `step_id` (appended tags, nothing renumbered), and the session-directory parser keeps the component it was already reading and discarding. Doing that now avoids baking in a permanent "0 means batch" ambiguity once these RPCs have deployed callers.

Re-keying the runtime maps I'd rather land with the numbered-step work. The mechanical part is genuinely small — the on-disk layout and the local socket protocol already carry `step_id` end to end — but the semantics aren't: `fence_stepd_recovery` requeues the whole job, so per-step recovery needs a policy for "one step's supervisor died but the batch step is fine", and displacement has to decide whether a new attempt displaces all of a job's steps or only the matching key. Those are design calls, not plumbing, and I don't want to answer them implicitly here.

**Deferring the map re-keying and per-step fencing to the follow-up.** Leaving this thread open to track it.

crates/spurctld/src/scheduler_loop.rs
yansun1996: The facts here are right: nothing durable names the attempt and cohort before the launch RPC goes out, and `JobNodeComplete` carries no attempt, so the pre-proposal check can't be repeated when Raft applies it.

Scoping, though — this is pre-existing on main rather than something this PR introduces. `wal.rs` isn't in this diff at all, and the `scheduler_loop` change here moves in the direction you're asking for: it threads `run_attempt` into the cancel RPCs, which previously carried none.

Record-before-dispatch is the cutover the allocation-reconciliation work is scoped around, and it's a breaking enough change to need its own review unit rather than being folded into a PR about the supervisor.

**Deferring both halves to that work.** The cheaper half — appending a defaulted `run_attempt` to `JobNodeComplete` and re-checking it at apply time — is small enough to land on its own first, and I'd rather do that than bundle it here. The replicated `Dispatching`/`Fencing` states are the larger design change.

crates/spurd/src/agent_server.rs
yansun1996: The ordering is exactly as you describe — the signal goes out first and the attempt is consulted only for the reservation release below it — and suspend, step cancel, and PMIx release carry no attempt identity at all.

Same scoping call as the dispatch-generation thread: the pre-PR handler had the identical signal-then-check shape, and this PR narrows the gap by making the reservation release epoch-guarded where it previously wasn't. Closing it properly means appending attempt fields to three more request messages and re-checking identity at each sink, which I'd rather review as its own change than attach to this one.

**Deferring the cancel/suspend/step-cancel/PMIx identity work.**

One thing from this thread I did fix here, because it was actively misleading rather than merely incomplete: the comment at the heartbeat-reclaim site claimed signal 0 is a no-op on an unknown id. It isn't — signal 0 takes the graceful-cancel branch, which is a full SIGTERM/SIGKILL teardown. That path isn't behind the supervisor flag, so it's live today. The comment now says what the code does and what the narrowing actually relies on.

crates/spurd/src/agent_server.rs
yansun1996: I modified this PR to be per-step, now we will start from here instead of deferring per-step to the next PR.

crates/spurd/src/agent_server.rs
```diff
@@ -2639,6 +3997,38 @@ impl SlurmAgent for AgentService {
                     },
                 );
                 drop(jobs);
+                // Released only now: the workload must not start before this
+                // node's own bookkeeping is complete, or a step launched from
+                // the script can outrun the job's start.
+                if let Some(descriptor) = runtime_descriptor.as_ref() {
```
biluriuday, line 4003: Gate execution on the controller’s Raft commit. The agent sends Start before LaunchJob returns, while JobStart is committed afterward. First-line srun can still race a Pending job, and leader failover can duplicate execution on another cohort. Implement replicated Dispatching → Ready → Running → Start. [Agent]

crates/spurd/src/agent_server.rs
```diff
@@ -285,86 +795,698 @@ fn refuse_allocation(
...
-        if let Ok(mut steps) = self.steps.try_lock() {
-            steps.remove(&self.key);
-        }
+pub async fn recover_stepds(
+    running: &RunningJobs,
+    descriptors: Vec<crate::stepd::StepdDescriptor>,
+) {
+    let mut jobs = running.lock().await;
+    for descriptor in descriptors {
+        jobs.entry(descriptor.job_id).or_insert_with(|| TrackedJob {
```
biluriuday, line 933: Restore complete containment after restart. recover_stepds discards cgroup, namespace, GPU and resource metadata. Recovered allocations can subsequently run srun/exec/attach outside their device filter or container namespaces. Persist a versioned recovery context and refuse execution if containment cannot be proven.

crates/spurd/src/agent_server.rs
```diff
@@ -43,6 +44,515 @@ fn maybe_deny_gpu_env(env: &mut HashMap<String, String>, allocated_device_ids: &
...
+        },
+        descriptor,
+    ))
+}
+
+/// Signals a stepd process directly by pid — there's no systemd unit to stop
+/// it through. A stale pid (already exited, or reused by an unrelated
+/// process) is confirmed via `stepd_liveness` before signaling, same check
+/// the crash watchdog uses.
+async fn stop_stepd_process(descriptor: &crate::stepd::StepdDescriptor) -> std::io::Result<()> {
```
biluriuday, line 285: Prove workload death when fencing. stop_stepd_process only SIGTERMs spurstepd; without a working cgroup, its child survives while teardown is considered successful. Use authenticated supervisor shutdown, escalation, and persisted workload PID/start-ticks or process-group identity.

crates/spurd/src/agent_server.rs
biluriuday, line 4128: Make run_attempt authoritative at every sink. Cancel signals before checking the requested attempt; local resource release is job-ID-only; JobNodeComplete drops the attempt before Raft apply; stale reports can still drain nodes. Carry exact identity through signal, PMIx, step, release, drain and WAL operations—never treat zero as a wildcard. Can come as follow-up PR in case you want to handle it separately.

biluriuday (commented):
Thanks @yansun1996 . One query and few follow-up comments in-line. PTAL
Query:
1. Currently spurstepd supervises sbatch and salloc jobs(job paths). srun, pty, exec, attach still depend on spurd. Is there a follow-up planned after this PR to have spurstepd supervise all launch paths?

crates/spurd/src/agent_server.rs
yansun1996: Fixed, though not as the full replicated state machine — the supervisor already held the workload at a start gate, so what changed is who opens it.

The agent now leaves the gate shut when `LaunchJob` returns, and the controller opens it with a new `StartJob` RPC issued only after the `Running` commit returns. Nothing the job runs can observe a Pending job. A node that cannot be released leaves its supervisor at the gate, so the run is cancelled rather than left Running with nothing behind it, and the existing gate timeout bounds a controller that dies in between. The k8s `VirtualAgent` accepts the release as a no-op, since a pod has no gate — otherwise the new failure path would cancel every k8s job.

Two honest caveats:

- This closes the first-line-`srun`-races-Pending half. **Duplicate execution after leader failover is not addressed** — that needs the replicated `Dispatching → Ready → Running` states you describe, and is a bigger change than belongs here.
- I am not claiming an e2e regression test for it. The window is one commit wide, and the test I wrote passes against both orderings, so it is coverage of the release path rather than proof. The guarantee rests on the code path plus unit tests on the release RPC (caller authz, superseded-epoch rejection, numbered-step filtering).

crates/spurd/src/agent_server.rs
yansun1996: Fixed. `recover_stepds` now restores the cgroup path and the pid/user/mount namespace flags from the descriptor, so work placed into a recovered job — exec, attach, a new step — lands back inside it rather than beside it on the host.

Chasing the rest of it turned up the resource half, which fails in the opposite direction to the one you would expect. `gpu_devices`, `cpus`, `memory_mb` and `partition` were dropped too, and a step reads its GPU set from the job it joins. An empty set is not "unset": `maybe_deny_gpu_env` reads it as *no devices* and pins every visibility variable to `-1`. So a recovered GPU allocation did not escape its filter — its steps were denied GPUs entirely, and GPU work inside an allocation silently stopped working once the agent had restarted under it. Those now ride in the descriptor as one bundle, and the supervisor republishes it.

Not done: the "refuse execution if containment cannot be proven" part. Everything above is restored from a persisted descriptor, but there is no positive proof-of-containment check before a step runs. Worth doing; I did not want to add a new refusal path to this PR.

crates/spurd/src/agent_server.rs
yansun1996: Fixed, and confirmed exactly as you describe. With a cgroup the retrying `rmdir` is real proof, since it only succeeds once the cgroup is empty. With none, the code returned `stop_result.is_ok()` — success on a SIGTERM the supervisor is designed to ignore — while the workload, a separate process the signal never reached, kept running. Teardown then reported success and the node was considered free.

The supervisor now records the workload's pid and start ticks in the descriptor when it forks it, and the cgroup-less path answers for that process: escalate to SIGKILL on the tree, and confirm only once it is gone.

Two cases that needed care:

- A **zombie** still has `/proc/<pid>/stat` with matching start ticks, so a naive check would hang on something already dead. It counts as gone — it has released everything and only waits to be reaped.
- A **recycled pid** is somebody else's process, so it is neither killed nor waited on.

Descriptors written before this carry no workload identity and keep the old behaviour; there is nothing better to check for them.

crates/spurd/src/agent_server.rs
yansun1996: Partly fixed here; the rest is follow-up, as you suggested.

**Fixed in this PR**

- The cancel path's cgroup reap derived `job_<id>_0` when the request named no epoch — a path no job ever creates — so a cancelled allocation leaked its live cgroup. It now resolves the epoch the node is actually running, from the tracked job or its supervisors.
- The signal went out *before* any epoch check and was keyed on job id alone, so a kill delayed past a requeue landed on the redispatch rather than the run it was meant for. A signal naming an epoch this node is not running is now dropped.

Zero still means "unspecified" on that request, as it does for the allocation release, so ordinary cancels are unchanged.

**Not fixed, agreed as follow-up**

- `WalOperation::JobNodeComplete` carries no `run_attempt`. The staleness check runs on the leader at propose time, but the committed operation has no epoch, so replay and follower-apply are unvalidated and the check-then-propose pair is not atomic. Adding the field is a WAL break — it needs `#[serde(default)]`, a frozen fixture, and a `!` on the title — so it is better on its own.
- Separately, and worth its own issue: `ReportJobStatus` drains the node *before* it looks at the completion verdict. A report that `node_complete` rejects as `StaleReport` — which the controller then logs as "ignoring completion report from superseded run" — has already drained the node by that point. So a superseded run's failed epilog takes a healthy node out of service. Independent of supervision; happy to file it separately unless you would rather it came along here.

crates/spurctld/src/scheduler_loop.rs
```diff
@@ -494,9 +506,107 @@ async fn process_assignment(
...
+/// times out — so the caller has to tear the run down rather than leave a job
+/// Running with nothing behind it on some of its nodes.
+async fn start_job_on_nodes(
+    cluster: &Arc<ClusterManager>,
+    job_id: spur_core::job::JobId,
+    run_attempt: u32,
+    node_names: &[String],
+) -> bool {
+    let mut set = tokio::task::JoinSet::new();
+    for agent_addr in cancel_agent_addrs(cluster, job_id, node_names) {
```
spraveenio, line 550: `start_job_on_nodes` derives its expected success count from `cancel_agent_addrs`, which silently skips any allocated node lacking a resolved comm address. If one dispatch node momentarily has no comm HTTP URL (mesh reconfig / transient re-registration), it's omitted from the count, so `expected != node_names.len()` fires even though every addressable node actually launched and released — cancelling and evicting a healthy Running job to NodeFail on an address-lookup miss rather than a real launch failure.

crates/spurctld/src/server.rs
```diff
@@ -1946,6 +2303,180 @@ impl SlurmController for ControllerService {
...
+                        .await;
+                    // The probe just found every participant alive, which outranks
+                    // one agent's claim that its own descriptor is unreadable.
+                    warn!(
+                        job_id = request.job_id,
+                        run_attempt = request.run_attempt,
+                        node = %request.hostname,
+                        "node reports a stale descriptor for a job whose supervisors are all live"
+                    );
+                    return Ok(Response::new(StepdRecoveryResponse {
```
spraveenio, line 2370: The stale-descriptor + probe-Retained branch returns `retained: true` with a non-empty message. `main.rs:746` re-reports recovery every 2s while `retained && !message.is_empty()`, so when a restarted spurd reports a transiently-unreadable descriptor but the cohort probe finds every supervisor alive, this branch loops forever re-reporting "stepd descriptor is unreadable but its supervisors are live" every 2 seconds — unlike the healthy path, which returns an empty message and breaks. Consider returning an empty message (break) once supervisors are proven live.

crates/spurctld/src/server.rs
```diff
@@ -192,26 +204,79 @@ impl LeaderProxy {
...
+const STEP_REAWAIT_ATTEMPTS: u32 = 30;
+const STEP_REAWAIT_BACKOFF: std::time::Duration = std::time::Duration::from_secs(2);
+
+/// tonic surfaces a dropped connection as Unknown/"transport error" rather than
+/// Unavailable, so both count as losing the agent rather than losing the step.
+fn is_agent_connection_loss(status: &Status) -> bool {
+    matches!(
+        status.code(),
+        tonic::Code::Unavailable | tonic::Code::Unknown
+    )
```
spraveenio, line 216: `is_agent_connection_loss` treats tonic `Code::Unknown` as connection loss. tonic surfaces many non-transport conditions as `Unknown` (agent-side `Status::unknown`, response decode errors, unhandled panics). When `run_command` returns such an `Unknown` for a genuinely-failed step, the controller enters `reawait_step` for the full retry budget and then reports `Unavailable("lost contact … could not re-attach")`, delaying and mislabeling a true step failure. Recommend narrowing to transport-level codes (e.g. `Unavailable`, and `Unknown` only with a transport source).

crates/spurctld/src/server.rs
```diff
@@ -192,26 +204,79 @@ impl LeaderProxy {
...
+                user: user.to_string(),
+                run_attempt,
+            })
+            .await;
+        // Not-found is retried like the rest: an agent that has just come back
+        // has not finished adopting its sessions, so it answers that for a step
+        // it is about to own again.
+        if let Ok(response) = awaited {
+            return Ok(response.into_inner());
+        }
```
spraveenio, line 248: `reawait_step` retries on **all** `await_step` errors, only returning early on `Ok`. Permanent failures — `permission_denied`, `invalid_argument`, `already_exists` (a second srun racing and getting "that step already has a caller awaiting it") — are retried the full ~60s budget and then yield a generic `Unavailable("lost contact … could not re-attach")` that hides the real cause. Recommend failing fast on non-connection-loss errors.

crates/spurctld/src/scheduler_loop.rs
yansun1996: The skip takes the attempt with it, not just the count: a node with no resolved comm address never receives `AgentStartJob`, so its supervisor stays at the gate and times out. The allocation then has a node holding resources behind a workload that will never start, and eviction is the only thing that frees it — that's what `process_assignment_evicts_a_job_no_node_would_release` pins ("the job would hold its allocation forever if left Running"). A node in that state also shouldn't get this far: the peer list resolves the same `comm_addr()` earlier and aborts the dispatch before the Running commit, so the residual window is a node deregistering between the two.

The diagnosis is genuinely wrong though, and worse than you flagged. The warning binds `expected` to `node_names.len()` rather than the count actually attempted, and `launch_failure_detail` names every dispatch node rather than the ones that refused — so "a node refused the release" and "a node was never contacted" are indistinguishable in both the log and the operator-facing message. The test assertion is vacuous on a single-node job for the same reason. Addressing all three in the follow-up PR rather than widening this one.

crates/spurctld/src/server.rs
yansun1996: This branch only runs under `stale_descriptor`, and that caller breaks unconditionally on `Ok` — it never reads `retained` or `message`. The 2s re-report loop belongs to the other reporting task, the one sending `stale_descriptor = false`, so this reports once and exits.

The rule does exist where it binds: the retain-live-run response keeps its message empty for exactly this reason, with a test pinning it. Worth insuring against anyway — if the stale caller ever grows the same retry loop this would behave as you describe — so I'll note the single-shot caller at the branch in the follow-up.

crates/spurctld/src/server.rs
yansun1996: Checked all three sources. There are no `Status::unknown` call sites on the agent side; tonic 0.14 maps decode failures to `Internal` (`codec/decode.rs`); and a panicking handler terminates the h2 stream, which arrives as HTTP 200 with no `grpc-status` trailer — tonic maps that to `Unknown` (`status.rs`), and it is genuine connection loss, which is the case this function exists for.

The narrowing would cost us that case: the trailer-less status is built with no source attached, so a transport-source downcast would read the main scenario as a step failure and stop the re-attach. tonic also documents that downcast as explicitly unstable. Leaving as is — tests pin both directions.

crates/spurctld/src/server.rs
yansun1996: Confirmed, with a narrower cause than described. The loop discards every error without logging it, so the cause reaches neither the user nor the log — 30 attempts at 2s, then a generic message. Will address in the follow-up PR: log each failed attempt, carry the last error into the final status, and return immediately on `permission_denied` / `unauthenticated`.

Two details differ. `check_job_access` and the handler only produce `not_found`, `permission_denied`, `already_exists` and `internal` — there's no `invalid_argument` path. And the racing-`srun` case isn't reachable: `await_step` has exactly one caller, this function, and `reregister` takes over any waiter whose channel `is_closed()`, so a re-attach can't collide with its own abandoned one — `already_exists` needs a half-open connection. `not_found` stays retryable deliberately: a just-restarted agent answers that for a step it's about to adopt, which is what the loop is for.

shiv-tyagi (approved):
Looks good. It can be merged on @biluriuday approval.

biluriuday (approved):
LGTM

PR #755, opened 2026-08-26
fix(spur-cli): show planned job/start and composite State= for plnd nodes

yansun1996:
## Summary

Since the backfill-starvation fix, an idle node the scheduler holds a future reservation for shows as `plnd` in `sinfo` (`NodeInfo.planned_job_id`/`planned_start`, already populated by the controller every scheduling pass). But `scontrol show node` gave no visibility into this at all: `State=` always printed the raw node state (e.g. `IDLE`), with no indication a reservation was in play, and no way to see which job was holding the node or when it was expected to start.

This PR:
1. Adds a `PlannedJobId=<id> PlannedStartTime=<ts>` line to `scontrol show node <name>`, shown only when the node has a scheduler-held reservation.
2. Makes `State=` itself carry the overlay, matching **real Slurm's own behavior** (verified live against a Slurm 25.11.6 cluster): `State=IDLE+RESERVED`, `State=IDLE+MAINTENANCE+RESERVED`, `State=IDLE+PLANNED`. Previously Spur's `scontrol show node` never reflected `resv`/`maint`/`plnd` in `State=` at all (only `sinfo` did) — this brings it in line with Slurm and fixes that gap for all three, not just `plnd`.

## Approach

- Extracted the Idle+active_reservation+planned_job_id branching (previously only in `sinfo.rs`'s `effective_state_str`) into a shared `spur_core::node::node_overlay()`, so `sinfo` and `scontrol` derive the same precedence instead of duplicating/drifting independently. `sinfo`'s existing 36 tests pass unchanged, confirming the refactor is behavior-preserving there.
- New `node_state_display()` in `scontrol.rs` composes `{base}+{overlay flags}` the same way Slurm does.
- Kept `ActiveReservation=`/`PlannedJobId=`/`PlannedStartTime=` as separate labeled lines in addition to the composited `State=` — redundant with the state string but non-breaking and still useful for anything already parsing those fields. Notably, real Slurm's own `scontrol show node` does NOT show which job or start time for a `PLANNED` node at all (you have to cross-reference `squeue --start`) — so this PR's `PlannedJobId=`/`PlannedStartTime=` lines are ahead of Slurm here, not just parity.
- Considered a `sinfo` long-format column for the job-id/start-time too, but `sinfo`'s default view groups multiple nodes per partition/state row — a single column doesn't fit when different nodes in the same group are held by different jobs. Left as a known limitation.

## Known limitations

- `format_ts` (pre-existing, shared helper) silently falls back to the Unix epoch rather than an explicit error if given a corrupt/out-of-range timestamp — not introduced by this change, out of scope here.
- No `sinfo` column for job-id/start-time (see above).
- `State=` output value changes for nodes already in `resv`/`maint` (previously just `IDLE`, now `IDLE+RESERVED`/`IDLE+MAINTENANCE+RESERVED`). This is a user-facing CLI output change, not a schema break — no field removed/renamed. No script/test in this repo currently asserts an exact `State=IDLE` string for these cases, but flagging in case any external Slurm-compat tooling parses it literally.

## Testing

- `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` and `cargo fmt --all --check`: clean.
- `cargo test -p spur-core -p spur-cli --locked`: 618 + 434 tests pass, including new unit tests for `node_overlay()` (spur-core) and `node_state_display()` (spur-cli) covering plain-idle, reserved, maintenance+reserved, planned, and non-idle-ignores-overlay cases. Verified each new test fails against a stubbed-out version of the fix and passes once restored.
- Extended the existing `TestBackfillReservation` e2e test (`tests/native_host/e2e/test_multi_node.py`) — it already reproduces the exact `plnd` scenario — with assertions on both the new `PlannedJobId=` line and the composited `State=IDLE+PLANNED`. Ran it against the real 2-node lab: passes.
- Validated end-to-end on an isolated 2-node deployment, reproducing the scenario twice (once for each commit in this PR):

  ```
  $ sinfo -N
  NODELIST  NODES PARTITION       STATE CPUS   MEMORY GRES
  node2          1 iso        plnd    4    15989 (null)
  node1          1 iso       alloc   10    60267 (null)

  $ scontrol show node node2
  NodeName=node2
     State=IDLE+PLANNED Reason=
     Partitions=iso
     CPUTot=4 CPUAlloc=0 RealMemory=15989 FreeMem=15411
     Arch= OS=
     PlannedJobId=2 PlannedStartTime=2026-08-26T12:41:05
     CpuLoad=0

  $ scontrol show node node1
  NodeName=node1
     State=ALLOCATED Reason=
     ...
  ```

  Confirmed `State=` carries the overlay exactly when the node has a scheduler-held reservation and is otherwise raw for a non-idle node (`node1` above, busy with the filler job).

### DIFF
--- a/crates/spur-cli/src/scontrol.rs
+++ b/crates/spur-cli/src/scontrol.rs
@@ -695,6 +695,9 @@ async fn show(controller: &str, entity: &str, name: Option<&str>) -> Result<()> 
                 if !node.active_reservation.is_empty() {
                     println!("   ActiveReservation={}", node.active_reservation);
                 }
+                if let Some(line) = planned_reservation_line(&node) {
+                    println!("   {line}");
+                }
                 println!("   CpuLoad={}", node.cpu_load as f64 / 100.0);
                 println!();
             }
@@ -920,6 +923,19 @@ fn format_ts(ts: Option<&prost_types::Timestamp>) -> String {
         }
         _ => "N/A".into(),
     }
+}
+
+/// `PlannedJobId=/PlannedStartTime=` line for `scontrol show node`, or `None`
+/// when the node has no scheduler-held future reservation.
+fn planned_reservation_line(node: &spur_proto::proto::NodeInfo) -> Option<String> {
+    if node.planned_job_id == 0 {
+        return None;
+    }
+    Some(format!(
+        "PlannedJobId={} PlannedStartTime={}",
+        node.planned_job_id,
+        format_ts(node.planned_start.as_ref()),
+    ))
 }
 
 async fn requeue(controller: &str, job_id: u32, hold: bool) -> Result<()> {
@@ -1734,6 +1750,27 @@ mod tests {
     fn gpu_tres_label_total() {
         assert_eq!(gpu_tres_label("gpu:8"), "TresPerJob");
         assert_eq!(gpu_tres_label("gpu:mi300x:4"), "TresPerJob");
+    }
+
+    #[test]
+    fn planned_reservation_line_none_when_no_planned_job() {
+        let node = spur_proto::proto::NodeInfo::default();
+        assert_eq!(planned_reservation_line(&node), None);
+    }
+
+    #[test]
+    fn planned_reservation_line_shows_job_and_start() {
+        let node = spur_proto::proto::NodeInfo {
+            planned_job_id: 42,
+            planned_start: Some(prost_types::Timestamp {
+                seconds: 1_700_000_000,
+                nanos: 0,
+            }),
+            ..Default::default()
+        };
+        let line = planned_reservation_line(&node).unwrap();
+        assert!(line.contains("PlannedJobId=42"), "{line}");
+        assert!(line.contains("PlannedStartTime=2023-11-14"), "{line}");
     }
 
     #[test]
--- a/crates/spur-cli/src/scontrol.rs
+++ b/crates/spur-cli/src/scontrol.rs
@@ -659,7 +659,7 @@ async fn show(controller: &str, entity: &str, name: Option<&str>) -> Result<()> 
                 println!("NodeName={}", node.name);
                 println!(
                     "   State={} Reason={}",
-                    node_state_name(node.state),
+                    node_state_display(&node),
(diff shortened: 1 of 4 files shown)


crates/spur-cli/src/scontrol.rs
```diff
@@ -922,6 +941,19 @@ fn format_ts(ts: Option<&prost_types::Timestamp>) -> String {
     }
 }
 
+/// `PlannedJobId=/PlannedStartTime=` line for `scontrol show node` (`None` if
+/// unset, `N/A` start if unknown); server only sets it while Idle, like sinfo's `plnd`.
+fn planned_reservation_line(node: &spur_proto::proto::NodeInfo) -> Option<String> {
```
sgopinath1, line 946: `planned_reservation_line()` gates on `planned_job_id != 0`, but `node_state_display()` uses `node_overlay()`, where an active reservation wins over planned. If both fields are populated on an idle node, `State=` shows `IDLE+RESERVED` while `PlannedJobId=` still prints — inconsistent with `sinfo`, which shows `resv`/`maint` in that case (`test_effective_state_active_reservation_wins_over_planned`).

Gate the line on the overlay instead:

```rust
match spur_core::node::node_overlay(node) {
    Some(spur_core::node::NodeOverlay::Planned) => Some(format!(...)),
    _ => None,
}
```

Add a unit test: idle node with both `active_reservation` and `planned_job_id` → `None`.

```suggestion
fn planned_reservation_line(node: &spur_proto::proto::NodeInfo) -> Option<String> {
    if !matches!(
        spur_core::node::node_overlay(node),
        Some(spur_core::node::NodeOverlay::Planned)
    ) {
        return None;
    }
    Some(format!(
        "PlannedJobId={} PlannedStartTime={}",
        node.planned_job_id,
        format_ts(node.planned_start.as_ref()),
    ))
}
```

sgopinath1 (commented):
Thanks for fixing this. Below is a review comment which is not blocking.

sgopinath1 (approved):

PR #758, opened 2026-08-27
fix(spurctld): tag array-throttle pending jobs with JobArrayTaskLimit

yansun1996:
## Summary

Job array tasks held back by their own `%N` concurrency throttle (`array_max_concurrent`) were reporting `Reason=None` in `squeue`/`scontrol show job`, indistinguishable from a genuine scheduling stall. Slurm's equivalent case reports `JobArrayTaskLimit`.

Root cause: the array throttle in `classify_pending_jobs()` ran as a raw loop that flipped `candidate.scheduling_eligible = false` directly, outside the `GateOutcome`/`retain_eligible` framework the other scheduling gates (QOS, account, burst-buffer, dependency, reservation) already use to record a `PendingReason` when they drop a candidate. Without a recorded reason, the job fell through to a final catch-all that only tags `BeginTime`-held jobs, leaving everything else silently untagged.

## Approach

- Added `PendingReason::JobArrayTaskLimit` with the Slurm-compatible display string `"JobArrayTaskLimit"`.
- Rewrote the array-throttle loop to go through the existing `retain_eligible()` / `GateOutcome::Block(...)` pattern instead of a bespoke flag mutation, bringing it in line with every sibling gate in the same function.

No other call sites needed changes — the only place matching `PendingReason` exhaustively is its `display()` method, and the compiler caught it.

## Testing

- Added a unit test asserting array tasks held back by `%N` are tagged `JobArrayTaskLimit` (and the admitted task is not). Verified it fails against the pre-fix code and passes with the fix.
- All pre-existing array-throttle tests still pass unchanged.
- `cargo test -p spur-core -p spurctld --locked`: all pass.
- Validated end-to-end on an isolated single-node deployment: submitted a `--array=0-2%1` job, confirmed the two throttled tasks show `Reason=JobArrayTaskLimit` via both `squeue` and `scontrol show job` while the admitted task runs.

### DIFF
--- a/crates/spur-core/src/job.rs
+++ b/crates/spur-core/src/job.rs
@@ -328,6 +328,10 @@ pub enum PendingReason {
     /// Replaces `BeginTime` as the pending reason for preempted-requeued jobs so the
     /// reason string is unambiguous.
     Preempted,
+
+    /// Array task held back by the job's own `%N` concurrency throttle
+    /// (`array_max_concurrent`), not by cluster contention.
+    JobArrayTaskLimit,
 }
 
 impl PendingReason {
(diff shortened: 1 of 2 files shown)


biluriuday (approved):
LGTM. Just a minor thing. Please document the new JobArrayTaskLimit in the user-guide(monitoring-jobs.rst and submitting-jobs.rst) as it is user-facing change.

PR #761, opened 2026-08-27
feat(spur-cli): report requested placement and scheduling provenance in scontrol show job

yansun1996:
## Problem

A job restricted to a node set with `-w/--nodelist` or `-x/--exclude` pends with `Reason=Resources`, and no CLI surface showed which nodes it asked for. The queue looked identical to "the cluster is full" while idle capacity sat unused, and establishing why a job was stuck meant reading source.

The reason string is correct — Slurm reports `Resources` here too. The gap was that nothing let an operator act on it. The data was also not already on the wire: `JobInfo.nodelist` is the *allocated* list, so it is empty precisely while the job is pending.

## What it looks like now

```
JobId=71860 JobName=train.sh
   JobState=PENDING Reason=Resources Dependency=(null)
   RunTime=00:00:00 TimeLimit=00:05:00 TimeMin=N/A
   SubmitTime=2026-08-27T16:58:33 EligibleTime=2026-08-27T16:58:33 AccrueTime=2026-08-27T16:58:33
   StartTime=2026-08-27T17:55:45 EndTime=2026-08-27T18:00:45 Deadline=N/A
   LastSchedEval=2026-08-27T17:47:27
   ReqNodeList=node07 ExcNodeList=(null)
   NodeList=(null) SchedNodeList=node07
   ReqTRES=cpu=1,node=1
   SubmitLine=sbatch -w node07 --exclusive -t 5 train.sh
```

`ReqNodeList=node07` is the answer that previously required reading source: the job is pinned, not starved. Field names line up with Slurm 25.11.6 on the same job shape.

`squeue --start` answers the same question for the whole queue at once, in Slurm's layout — pending jobs only, ordered by projected start with unprojected last, plus a `SCHEDNODES` column (`%Y`):

```
     JOBID PARTITION     NAME     USER ST          START_TIME  NODES  SCHEDNODES NODELIST(REASON)
      1024       gpu  train.sh    alice PD 2026-09-01T00:56:13     1     node007 (Resources)
      1031       gpu   eval.sh      bob PD                 N/A     1      (null) (Priority)
```

Each implied default yields to an explicit `-t`/`-S`/`-o`, and `%S`/`%e` carry the projection in any format string.

The scheduler side is no longer silent either. `backfill.rs` had six unlogged refusal paths; it now emits one line per job **when its outcome changes**, so a steady-state stuck queue reports once rather than every cycle:

```
backfill did not start job job_id=1024 reason=no_suitable_nodes needed_nodes=1 candidate_nodes=0
```

## Approach

`JobInfo` gains additive fields (new tags only):

- `ReqNodeList`, `ExcNodeList`, `Features`, `Dependency` — the request as submitted.
- `SubmitLine` — captured in the spec builders, stored on `JobSpec` (`#[serde(default)]`; the existing frozen-payload fixture covers replay).
- `ReqTRES`, `MinCPUsNode`, `MinMemoryNode` — derived from the spec, no new state.
- `EligibleTime`, `AccrueTime`, `LastSchedEval` — scheduling provenance.
- `StartTime`/`SchedNodeList` — the future slot held for a pending job, with `EndTime` derived from it plus the time limit.
- Plus `Deadline`, `TimeLimit`/`RunTime`/`TimeMin`, `Requeue`/`Restarts`/`BatchFlag`/`Exclusive`, `Command`, `Reservation`, and array ids, all from data `JobInfo` already carried but never printed.

Notable decisions:

- **`LastSchedEval` is not persisted.** Replicating a per-cycle timestamp through Raft would be a write storm, so it is in-memory only and resets on restart or failover.
- **The projection needed a job-keyed scheduler view.** The existing accessor is node-keyed and filters to *idle* nodes (correct for `sinfo`'s `plnd`), which excludes the case that matters most — a job queued behind a busy node.
- **Start/end derivation lives in one module**, since `scontrol` and `squeue` must agree on when a projection stands in for a real time. A projection is only used while the job is pending, so a job that never ran reports nothing rather than a stale plan.
- **Refusal logging dedups on the reason**, not on the candidate count, which is recomputed from live node state every cycle — keying on it would re-log the whole queue whenever any one node changed.
- **`ReqTRES` omits Slurm's `billing=`.** Spur computes no TRES billing weights.

## Compatibility

Proto changes are additive. The new persisted `JobSpec` field defaults on replay.

**Three output changes that can affect existing scripts**, all documented in the migration guide:

1. `ReqNodeList`, `ExcNodeList`, and `SchedNodeList` all contain the literal `NodeList=`, and `ReqNodeList` prints **first**. Match `^\s*NodeList=` or an equivalent boundary. (This is Slurm's own field layout, and it caught an e2e test here.)
2. A pending job's `StartTime` is now a projection where it was `N/A`, and `EndTime` follows from it, so `StartTime != N/A` no longer means "the job has started" — check `JobState`. Running jobs also gain a projected `EndTime`.
3. Diagnostic fields print `(null)` when unset rather than being omitted — notably `NodeList=(null)` on a pending job.

## Known limitations

- `NumCPUs`, `AllocTRES`, and `Scheduler=` are not reported; concepts Spur does not model (`CoreSpec`, `Licenses`, `MCS_label`, `ReqB:S:C:T`) are omitted rather than filled with placeholders.
- `MinCPUsNode` can read higher than Slurm's `pn_min_cpus` for an uneven split; `AccrueTime` reads earlier for held and dependency-blocked jobs, since Spur accrues from submit.
- Jobs beyond the per-cycle scheduling depth limit are not covered by the refusal logging, and `LastSchedEval` is still stamped on them — scoping the stamp means threading the limit through every caller, so the divergence is documented rather than fixed.
- A `future_slot_reserved` projection that slips does not re-log; only a change of reason or requested node count re-reports.
- `StartTime`, `SchedNodeList`, and `LastSchedEval` are leader-only and read as unset on a follower-served query.
- A few fields (`Comment`, `Reservation`, `ArrayJobId`, `SchedNodeList`, `StdIn`, the GPU line, preemption) still print only when set.

Follow-ups filed separately: SPUR-245 (`%L` renders `?`), SPUR-246 (`sinfo -t plnd` rejected), SPUR-247 (`TotalNodes=0`), SPUR-248 (non-numeric job id), SPUR-249 (`squeue` flag parity).

## Testing

- Unit tests for every behavioral change, each run against the unfixed code to confirm it goes red — including the redaction ordering, the byte bound, the `ReqTRES` total, the `MinMemoryCPU` label, the log dedup, and the sanitize inside `proto_to_job_spec`.
- Validated end-to-end on an isolated deployment: the pinned-to-busy-node case, the exclude path, a job with no reservable slot, and the transition to a real start once dispatched. The projection chain is self-consistent (running job ends 00:55:43, next projects 00:56:13, ends 01:01:13, third projects exactly there); an unlimited job reports a projected start with `EndTime=N/A`; draining the node clears both to `N/A`/`(null)`.
- Log volume checked against the flooding concern: two stuck jobs across ~60 one-second cycles produced one line each, then one more each when draining genuinely changed the reason.
- Confirmed an injected escape sequence and newline are neutralized rather than reaching the terminal or joining tokens.
- Output compared field-by-field against Slurm 25.11.6 running the same scenario.


docs/user-guide/monitoring-jobs.rst
```diff
@@ -343,12 +343,96 @@ Detailed Records — ``scontrol show``
 ``spur show`` (Slurm ``scontrol show``) prints the full ``Key=Value`` record for
 an entity: ``job``, ``node``, ``partition``, ``reservation``, or ``step``.
 
+.. note::
+
+   The job record is fixed-shape: every field is printed on every job, with
+   ``(null)`` for an unset string. Parsers must key on the field name, not on a
+   line being absent.
```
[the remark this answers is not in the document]
yansun1996: Good catch — the claim was too strong. Reworded to say the diagnostic fields listed below are always printed with `(null)` when unset, and noted that `Comment`, `Reservation`, `ArrayJobId`, `SchedNodeList`, the GPU line and the preemption line still appear only when set.

docs/migration-from-slurm.rst
```diff
@@ -70,7 +70,7 @@ noted as "accepted for compatibility" parse without error but have no effect yet
    * - ``scancel`` arrays
      - Array-element syntax (``123_4``) is not yet supported client-side; cancel by plain job id.
    * - ``scontrol show``
-     - Output is always the multi-line ``Key=Value`` block; there is no ``--oneliner``.
+     - Output is always the multi-line ``Key=Value`` block; there is no ``--oneliner``. Unset string fields print as ``(null)`` rather than being omitted. ``ReqTRES`` omits Slurm's ``billing=`` component, as Spur has no TRES billing weights.
```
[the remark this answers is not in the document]
yansun1996: Softened — it now says unset *diagnostic* fields print as `(null)` rather than implying every string field does. I also added the related pitfall that `ReqNodeList`/`ExcNodeList`/`SchedNodeList` all contain the literal `NodeList=`, so an unanchored match now reads the requested list instead of the allocated one.

crates/spurctld/src/server.rs
```diff
@@ -4145,6 +4244,29 @@ fn annotate_nodes_with_reservations(
...
 }
 
+/// Surface the future slot the scheduler is holding for a pending job, so
+/// `StartTime` answers "when" instead of staying N/A while the job waits.
+fn annotate_jobs_with_planned_reservations(jobs: &mut [JobInfo], cluster: &ClusterManager) {
+    if jobs.is_empty() {
+        return;
+    }
+    apply_planned_reservations(jobs, &cluster.planned_job_starts());
+}
```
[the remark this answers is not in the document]
yansun1996: Agreed. Replaced with a read-locked view (`with_planned_job_starts`), so both the batch and single-job paths borrow the map instead of cloning it per response.

crates/spurctld/src/server.rs
```diff
@@ -846,7 +853,8 @@ impl SlurmController for ControllerService {
             nodes: &req.nodes,
         });
 
-        let proto_jobs: Vec<JobInfo> = jobs.iter().map(job_to_proto).collect();
+        let mut proto_jobs: Vec<JobInfo> = jobs.iter().map(job_to_proto).collect();
+        annotate_jobs_with_planned_reservations(&mut proto_jobs, &self.cluster);
 
         Ok(Response::new(GetJobsResponse { jobs: proto_jobs }))
```
shiv-tyagi, line 886: GetJobs still returns a full JobInfo for every caller. scontrol show job uses this RPC exclusively (even with a job id), and squeue does too, so the SubmitLine / ReqNodeList / SchedNodeList redaction on get_job never runs for the CLI this PR is teaching those fields.

A non-owner under the default redacted policy can read another user's submit line and requested nodes via `scontrol show job` or `squeue -o %Y`. Apply the same disclosure pass per job here, or route a single-id show through get_job.

crates/spurctld/src/server.rs
```diff
@@ -3916,7 +3935,125 @@ fn job_to_proto(job: &spur_core::job::Job) -> JobInfo {
...
+        time_min: job.spec.time_min.map(|d| prost_types::Duration {
+            seconds: d.num_seconds(),
+            nanos: 0,
+        }),
+        requeue: job.spec.requeue,
+        restarts: job
+            .requeue_count
+            .saturating_add(job.preempt_requeue_count)
+            .saturating_add(job.user_requeue_count),
+        batch_flag: job.spec.script.is_some(),
```
shiv-tyagi: `BatchFlag` is `script.is_some()`, but srun and salloc always populate `script` (`build_command_script` and `sleep infinity`). Those jobs will print `BatchFlag=1`.

Slurm's flag is 1 only for batch submit. Gate on the batch path:

```suggestion
        batch_flag: job.spec.script.is_some() && !job.spec.srun_job && !job.spec.interactive,
```

proto/slurm.proto
```diff
@@ -254,6 +258,42 @@ message JobInfo {
...
+  string submit_line = 64;
+
+  // Requested TRES ("cpu=8,mem=32000M,node=2,gres/gpu=8") and the per-node
+  // minima Slurm reports alongside it.
+  string req_tres = 65;
+  uint32 min_cpus_node = 66;
+  // Per-node memory floor. `min_memory_is_per_cpu` marks a --mem-per-cpu
+  // request, which Slurm reports as MinMemoryCPU rather than MinMemoryNode.
+  uint64 min_memory_node_mb = 67;
+  bool min_memory_is_per_cpu = 79;
```
shiv-tyagi: Nit: field 79 is declared between 67 and 68. Tags are append-only so this is wire-safe, but keep declarations in number order. Move `min_memory_is_per_cpu = 79` to after `sched_nodelist = 78`.

docs/user-guide/monitoring-jobs.rst
```diff
@@ -344,12 +374,122 @@ Detailed Records — ``scontrol show``
...
+      them unset even in a healthy cluster.
+
+So a job pinned to a busy node reads (excerpt — the full record has more
+fields between these lines):
+
+.. code-block:: text
+
+      JobState=PENDING Reason=Resources Dependency=(null)
+      ...
+      StartTime=2026-08-27T17:09:56 EndTime=N/A Deadline=N/A
```
shiv-tyagi: Nit: this excerpt has a projected `StartTime` and `SubmitLine=... -t 5`, so `EndTime` should be start plus five minutes, not `N/A`. `N/A` is the unlimited / no-projection case described a few lines above.

shiv-tyagi (commented):
Four notes: GetJobs skips the job-info redaction that this PR adds (and scontrol/squeue both use that RPC), BatchFlag is true for srun/salloc because they always set a script, proto field 79 is declared out of number order, and the monitoring-jobs excerpt shows EndTime=N/A for a job with -t 5 and a projected start.

crates/spurctld/src/server.rs
yansun1996: I traced this carefully because the reasoning is sound, but I do not think the disclosure lands.

`authoritative_user` overwrites `req.user` with the authenticated identity unconditionally, before the filter is built, so an identified caller only ever receives their own jobs from `GetJobs` — another user's record never enters the response to be redacted. When no identity is present (`disabled`, or `permissive` with no credential), `viewer_is_privileged(None, ..)` returns `Full`, so `get_job` does not redact for that caller either. So there is no mode where `GetJobs` discloses something `get_job` would have withheld.

Chasing it did surface two real problems, both pre-existing rather than introduced here, so I would rather not expand this PR into them:

[1] `GetJobsRequest.user` is a *filter* field, but it gets overwritten with the caller identity — so `squeue -u bob` is silently rewritten to the caller, and there is no admin bypass, which breaks cluster-wide queue views. That is a Slurm-parity break.
[2] The REST handlers call `cluster.get_jobs` / `get_job_for_display` directly with no disclosure pass and no identity plumbing at all.

Happy to file both separately.

crates/spurctld/src/server.rs
yansun1996: Good catch — fixed as suggested, the fields exist and it was reporting `BatchFlag=1` for both. Verified against a live cluster: `sbatch` reports 1, `salloc` and `srun` report 0. Added a test covering all three submission shapes; it fails against the old expression.

proto/slurm.proto
yansun1996: Moved `min_memory_is_per_cpu = 79` to after `sched_nodelist = 78`. Tag number unchanged, so the wire format is identical.

docs/user-guide/monitoring-jobs.rst
yansun1996: You are right, that excerpt predated the change. Updated to `EndTime=2026-08-27T17:14:56`, which is the projected start plus the five-minute limit.

crates/spurctld/src/server.rs
yansun1996: Checked this carefully because it is a disclosure claim, and both premises are right: `scontrol show job` and `squeue` do read exclusively through `GetJobs`, and `GetJobs` never calls the `get_job` redaction. The conclusion does not follow, though — `GetJobs` protects the same data by a different mechanism.

`get_jobs` overwrites the wire `user` field with the verified identity before building the filter (`authoritative_user`), and the filter matches `spec.user` strictly, so an identified non-owner does not get the job back at all — empty result, which is stricter than `redacted`. The `job_ids` array-parent path re-applies the same predicate, so naming the id outright does not bypass it either. A caller with no verified identity gets the full record, but that is equally true of `get_job` (`viewer_is_privileged(None)` is `true`), so there is no caller class for which the list path discloses more.

That said, the invariant doing the work here had no handler-level test — deleting the `authoritative_user` line left the suite green, which is a bad property for the thing standing between two users. Added `get_jobs_never_returns_another_users_job_to_an_identified_caller`, covering both a spoofed `--user` and an explicit job id; it fails if that line goes.

I also routed the list path through the same disclosure decision the single-job path uses, so the two cannot drift on policy. It is belt-and-braces rather than a fix — with the user scoping in place a non-owner never reaches a job to redact, so it is a no-op today. It annotates in one batch first rather than reusing `scoped_job_info`, which would take the planned-starts lock once per job in the squeue path.

Separately, and not for this PR: that scoping applies to admins too, so `squeue` shows only your own jobs and `squeue -u alice` run by bob silently returns bob's. That is a Slurm-parity gap in the opposite direction from a leak, and I will raise it on its own.

shiv-tyagi (approved):
LGTM

The GetJobs disclosure path, BatchFlag gating, proto field 79 order, and EndTime excerpt are all addressed. CI is green.

PR #764, opened 2026-08-28
fix(spurctld): reject zero partition default time

### LINKED ISSUE #759: spurctld: partition DefaultTime=0 gives an untimed job a zero-length time limit instead of refusing it
## What happens

A partition configured with `DefaultTime = "00:00:00"` (or `"0"`) gives a job submitted without `-t` a **zero-length** time limit. `parse_time_minutes` maps `"00:00:00"` to `Some(0)`, and `apply_default_time_limit` treats that as a usable default and assigns it, so the job is admitted with `TimeLimit=00:00:00`.

Verified against the real code path with a throwaway probe over `apply_default_time_limit`:

```
DefaultTime=00:00:00 parses to Some(0), job time_limit becomes Some(TimeDelta { secs: 0, nanos: 0 })
```

## What was expected

Slurm rejects the submission in exactly this case. From `_valid_job_part()` in `src/slurmctld/job_mgr.c`:

```c
if ((job_desc->time_limit   == NO_VAL) &&
    (part_ptr->default_time == 0)) {
        info("%s: job's default time is 0", __func__);
        rc = ESLURM_INVALID_TIME_LIMIT;
        goto fini;
}
```

Note this is specific to `DefaultTime` being explicitly `0`; an *unset* `DefaultTime` (`NO_VAL`) is admitted by Slurm and is not what this issue is about.

Worth deciding as part of this: whether a `0` here should mean "reject", "no default" (as `maxwall = 0` already means block-all elsewhere in Spur), or something else, and whether the same reasoning applies to `MaxTime = 0` and to `scheduler.default_time_limit_minutes`, which uses `0` as its "unset" sentinel.

## Reproduction

```toml
[[partitions]]
name = "zero"
default = true
nodes = "ALL"
default_time = "00:00:00"
```

```bash
# no -t
sbatch --wrap 'sleep 60'
scontrol show job <id>   # TimeLimit=00:00:00
```

## Impact

A job admitted with a zero-length limit has a deadline at or before its own start, so the running-job watchdog is entitled to kill it immediately. Every `-t`-less submission to such a partition is accepted and then killed, instead of being refused at submit with a clear error. The misconfiguration is silent from the submitter's side.

## Notes

Found while reviewing the `MaxWall` defaulting work in #728 (the wall-time default chain is where the zero reaches the job). Unrelated to that PR's change and deliberately not folded into it.

jamesETsmith:
## Motivation

Prevent jobs submitted without an explicit time limit from receiving a zero-length limit when their partition configures  DefaultTime=0 . This matches Slurm behavior by rejecting the submission instead of accepting a job that may be terminated immediately.

Fixes https://github.com/ROCm/spur/issues/759.

## Technical Details

• Changed partition time-limit defaulting to return a submission error when any selected partition explicitly defines  DefaultTime=0 .
• Preserved existing behavior for positive defaults, unset defaults, cluster fallbacks, multi-partition requests, and jobs with an explicit time limit.
• Added regression coverage for both the defaulting helper and the complete job submission path.

## Test Plan

• Run focused unit tests for partition time-limit defaulting.
• Run the submission-path regression test for  DefaultTime=0 .
• Verify Rust formatting.

## Test Result

•  cargo test -p spurctld apply_default_time_limit : 11 passed.
•  cargo test -p spurctld submit_rejects_zero_partition_default_time_when_time_is_unset : 1 passed.
•  cargo fmt --all -- --check : passed.

## Submission Checklist

[X] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.
md#pull-requests.

### DIFF
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -20317,6 +20317,33 @@ mod tests {
     }
 
     #[test]
+    fn validate_default_time_limit_rejects_when_any_requested_partition_is_zero() {
+        let mut spec = basic_spec("j");
+        spec.partition = Some("gpu,cpu".into());
+        spec.time_limit = None;
+        let partitions = vec![
+            Partition {
+                name: "gpu".into(),
+                default_time_minutes: Some(30),
+                ..Default::default()
+            },
+            Partition {
+                name: "cpu".into(),
+                default_time_minutes: Some(0),
+                ..Default::default()
+            },
+        ];
+        let err = super::validate_default_time_limit(&spec, &partitions).unwrap_err();
+        assert_eq!(
+            err,
+            SubmitError::invalid(
+                "partition 'cpu' has a default time limit of zero; specify a time limit"
+            )
+        );
+        assert!(spec.time_limit.is_none());
+    }
+
+    #[test]
     fn apply_default_time_limit_falls_back_to_partition_max_time() {
         // No DefaultTime but a finite MaxTime: Slurm uses MaxTime as the default.
         let mut spec = basic_spec("j");
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -662,6 +662,7 @@ impl ClusterManager {
         // the hook may rewrite, so defaulting follows the hook and reads the scope
         // that actually governs the job. A limit the hook set is left alone.
         let wall_caps = self.wall_caps(&spec);
+        validate_default_time_limit(&spec, &partitions)?;
         apply_default_time_limit(
             &mut spec,
             &partitions,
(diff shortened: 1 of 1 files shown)


yansun1996 (dismissed):
Nice, focused fix — this correctly matches Slurm's `_valid_job_part` semantics (reject on *any* requested partition having an explicit `DefaultTime=0`, distinct from the unset/`None` case and the cluster-fallback-`0` sentinel). Traced the call site and confirmed the new fallible path runs before job-ID allocation and the WAL/Raft append, so a rejection here is cheap and side-effect-free. Both new tests exercise the real production entry points and would fail against the pre-fix code.

Two things to handle before merge:

1. This PR currently has a merge conflict with `main` — could you rebase and resolve it?
2. While you're in there, would it be worth adding a test case where a job requests multiple partitions (e.g. `"gpu,cpu"`) and only one has `DefaultTime=0`? The current tests all use a single matched partition, so the "reject if any requested partition is zero" semantics (which is the Slurm-faithful behavior here) isn't locked in against a future refactor of the `.find()` call.

yansun1996 (approved):

PR #765, opened 2026-08-28
feat(spurctld): add jobs_preempted event counter to scheduler stats

sajmera-pensando:
## Summary

- `PREEMPTED` is a transient pass-through state in Spur — jobs move through it instantly back to `Pending` or `Cancelled`, so it always shows `0` as a gauge in `spur diag`. This PR replaces it with a proper event counter.
- Removes `PREEMPTED` and `Requeued` from the Job Statistics section of `spur diag` (both are transient states that never accumulate meaningfully as gauges).
- Adds `jobs_preempted` as a monotonic counter alongside `jobs_submitted`, `jobs_started`, and `jobs_finalized` in the Scheduler Statistics section of `spur diag` and the `/metrics/scheduler` OpenMetrics endpoint.

## Approach

The counter lives in `SchedStatsCollector` as an `AtomicU64`, incremented in `run_job_finalized_side_effects` when `finalized.state == JobState::Preempted`. This catches both `PreemptMode::Requeue` (job lands in `Pending`) and `PreemptMode::Cancel` (job lands in `Cancelled`) — both transition through `Preempted` in the WAL apply. `PreemptMode::Suspend` does not set `Preempted` so it is correctly excluded.

The new field is exposed via:
- gRPC `SchedStats` proto (field 15, append-only — no wire compat break)
- `/metrics/scheduler` as `spur_scheduler_jobs_preempted_total` (counter type)
- `spur diag` Scheduler Statistics section as `Jobs preempted : N`

## Testing

- Unit tests in `sched_stats.rs`: counter increments independently from `jobs_finalized`, resets correctly.
- Integration test in `cluster.rs` (`sched_stats_track_preempted_jobs`): fires one requeue and one cancel preemption, asserts counter reaches 2.
- E2e `TestCancelMode` and `TestRequeueMode` in `test_preemption_modes.py`: each assert `sdiag_jobs_preempted() == 1` after a single preemption.
- E2e `TestChronicPreemption` in `test_preemption.py`: asserts `sdiag_jobs_preempted() == cycles` (5) after the multi-cycle loop, verifying the counter increments correctly across repeated preemptions.

### DIFF
--- a/crates/spur-metrics/src/export/scheduler.rs
+++ b/crates/spur-metrics/src/export/scheduler.rs
@@ -92,6 +92,12 @@ pub fn register_scheduler(registry: &mut Registry, snap: &SchedStatsSnapshot) {
     );
     register_counter(
         registry,
+        "spur_scheduler_jobs_preempted",
+        "Jobs preempted since reset",
+        snap.jobs_preempted,
+    );
+    register_counter(
+        registry,
         "spur_scheduler_exit_end",
         "Cycles that considered every pending job",
         snap.exit_end,
@@ -124,6 +130,7 @@ mod tests {
             jobs_submitted: 42,
             jobs_started: 30,
             jobs_finalized: 28,
+            jobs_preempted: 5,
             jobs_started_last_cycle: 3,
             exit_end: 8,
             exit_max_depth: 2,
@@ -151,6 +158,11 @@ mod tests {
         assert!(body.contains("spur_scheduler_jobs_started_last_cycle 3"));
         assert!(body.contains("spur_scheduler_exit_end_total 8"));
         assert!(body.contains("spur_scheduler_exit_max_depth_total 2"));
+        assert!(body.contains("spur_scheduler_jobs_preempted_total 5"));
+        assert_eq!(
+            metric_type(&body, "spur_scheduler_jobs_preempted"),
+            "counter"
+        );
         assert_eq!(metric_type(&body, "spur_scheduler_cycles"), "counter");
         assert_eq!(
             metric_type(&body, "spur_scheduler_jobs_submitted"),
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -1831,6 +1831,9 @@ impl ClusterManager {
     fn run_job_finalized_side_effects(&self, finalized: JobFinalized) {
         if let Some(stats) = self.sched_stats.get() {
             stats.record_finalized();
+            if finalized.state == JobState::Preempted {
+                stats.record_preempted();
+            }
         }
         self.run_epilog_slurmctld(finalized.job_id);
         self.notify_job_finished(finalized.job_id, finalized.state, finalized.exit_code);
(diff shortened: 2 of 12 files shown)


crates/spurctld/src/cluster.rs
```diff
@@ -1831,6 +1831,9 @@ impl ClusterManager {
     fn run_job_finalized_side_effects(&self, finalized: JobFinalized) {
         if let Some(stats) = self.sched_stats.get() {
             stats.record_finalized();
+            if finalized.state == JobState::Preempted {
```
yansun1996, line 1834: Would it be worth a short comment here noting that `record_preempted()`'s correctness relies on only `JobPreemptCancel`/`JobPreemptRequeue` ever setting `finalized.state == JobState::Preempted`? It's implicit today — a future WAL variant that legitimately transitions through `Preempted` for another reason would silently double-count.

crates/spurctld/src/sched_stats.rs
```diff
@@ -82,6 +84,10 @@ impl SchedStatsCollector {
         self.jobs_finalized.fetch_add(1, Ordering::Relaxed);
     }
 
+    pub fn record_preempted(&self) {
```
yansun1996, line 87: One thing worth documenting: `jobs_preempted` is a *subset* of `jobs_finalized` (a preempted job increments both), not a separate/exclusive category. A one-line comment here (or on the field a few lines up) would save a future dashboard author from assuming the finalized counters partition to 100%.

crates/spur-metrics/src/export/scheduler.rs
```diff
@@ -90,6 +90,12 @@ pub fn register_scheduler(registry: &mut Registry, snap: &SchedStatsSnapshot) {
         "Jobs reaching a terminal state since reset",
         snap.jobs_finalized,
     );
+    register_counter(
+        registry,
+        "spur_scheduler_jobs_preempted",
+        "Jobs preempted since reset",
```
yansun1996, line 96: Minor: the HELP text ("Jobs preempted since reset") doesn't make clear that Suspend-mode preemptions are excluded (only cancel/requeue, since only those finalize the job). Might be worth a small wording tweak for anyone reading the metric without the source.

yansun1996 (commented):
Nice, clean implementation of the counter — the increment logic, proto addition, and metrics/CLI wiring are all correct, and the new tests exercise the real code paths (both Cancel and Requeue preemption modes route through `JobState::Preempted`; Suspend correctly does not).

A few small suggestions below (all optional/cheap). One more note that doesn't anchor to a changed line: `preempt_job_suspend_retains_allocation` (the existing Suspend test in `cluster.rs`) never asserts `jobs_preempted` stays at 0 — that's the one branch of the "Suspend is correctly excluded" claim without direct negative-path coverage. Might be worth adding that assertion.

yansun1996 (approved):
Approving — the counter implementation is correct and well-tested. The suggestions from the earlier review are optional/non-blocking.

Issue #747, opened 2026-08-25
E2E: intermittent failures from shared runner state (fixed Postgres port, 60s wait deadline)


nikhilsk:
The E2E suite intermittently fails on code that passes when the same commit is run again. Two distinct symptoms have been observed, both traceable to a fixed resource that parallel jobs contend for.

**1. Postgres host port collision**

A `native-host` shard fails to start its database with:

```
failed to bind host port 0.0.0.0:55432/tcp: address already in use
```

The shard reports something like "N passed, M deselected, 1 error" while every other shard passes. `tests/native_host/e2e/cluster.py` gives each shard a unique container name (`spur-e2e-pg-{pid}-{time_ns}`) and a unique `RUN_DIR`, but the host port is fixed at `55432` (`SPUR_TEST_PG_PORT`, line 212). Per-shard cleanup removes only its own container by name, so it cannot reclaim the port from a peer. The `native-host` job fans out to 14 shards (2 gpu, 12 non-gpu) across the self-hosted pool.

Observed on PR #647, which changed documentation only and no code, and again on PR #728.

**2. Fixed 60-second state deadline**

`tests/k8s/e2e/test_spurjob.py::TestSpurJobLifecycle::test_simple_spurjob_completes` fails with:

```
TimeoutError: timed out after 60s: SpurJob it-simple did not reach state 'Completed'
```

from the default `timeout: int = 60` in `wait_for_state` (`tests/k8s/e2e/k8s_cluster.py:303`).

On PR #728, commit `0569f0a` ran E2E twice: run `32843823120` passed both `E2E / native-host` and `E2E / k8s`; run `32850700765`, on that identical SHA 75 minutes later, failed `E2E / k8s` with the above. Identical code, opposite outcomes.

**Expected**

A suite's result depends on the commit under test, not on what else happens to be running.

**Impact**

Unrelated PRs go red and need re-triggering. GitHub's "re-run failed jobs" is unavailable to non-admin contributors on fork PRs, so the only recourse is pushing a no-op commit. Each such flake also erodes trust in the signal and can mask a genuine failure.

PR #770, opened 2026-08-31
test(e2e): publish the accounting Postgres container on an ephemeral host port

shiv-tyagi:
## Summary

The native-host accounting harness published Postgres on a fixed host port. Accounting tests run sequentially on one node, so a container left behind by an interrupted test (or a port the kernel has not yet released) makes the next `docker run -p 55432:5432` fail with `port is already allocated`, turning an unrelated test red.

The fix publishes with `-p 0:5432` so the daemon assigns a free ephemeral port, and reads it back with `docker inspect`. `_db_url` becomes a derived property that raises if read before the port is resolved, and teardown resets the resolved port so a restart binds a fresh one.

`SPUR_TEST_PG_PORT` is removed: a fixed port is exactly what causes the collision, and nothing else references it.

## Why `docker inspect` over `docker port`

`docker inspect --format` returns the numeric `HostPort` directly, so the harness avoids parsing the multi-line, IPv4/IPv6-variant `docker port` text.

## Testing

On a real node with Docker:

- Reproduced the original collision: with one container up on `55432`, a second `docker run -p 55432:5432` fails with `Bind for 0.0.0.0:55432 failed: port is already allocated`.
- Launched 4 containers with `-p 0:5432` in parallel: each got a distinct host port and became `pg_isready`.
- Drove the actual harness path (`_start_postgres` then `docker inspect` then `_db_url` then `_stop_postgres`) over SSH: the port resolves to an ephemeral value, the `_db_url` guard raises before resolution, and teardown resets it.

Addresses the Postgres host-port symptom described in #747.

### DIFF
--- a/tests/native_host/e2e/cluster.py
+++ b/tests/native_host/e2e/cluster.py
@@ -30,6 +30,21 @@ ACCOUNTING_SYMLINKS = ["sacct", "sacctmgr", "sshare", "sreport"]
 CONTROLLER_PORT = int(os.environ.get("SPUR_TEST_CONTROLLER_PORT", "6817"))
 AGENT_PORT = int(os.environ.get("SPUR_TEST_AGENT_PORT", "6818"))
 
+# Extracts just the numeric host port for 5432/tcp. inspect returns structured
+# data, so this sidesteps parsing the multi-line, IPv4/IPv6-variant `docker port`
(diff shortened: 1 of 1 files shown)


sgopinath1 (approved):
LGTM

PR #776, opened 2026-08-31
fix(spurctld): keep one unresponsive node from stalling all scheduling

### LINKED ISSUE #767: Scheduler stops cycling while eligible nodes remain idle
# SPUR 0.9.1 scheduler appears stalled despite idle eligible nodes

## Summary

The SPUR controller responds to queries and its server clock advances, but the scheduler cycle counter does not advance. At the same time, 148 jobs are pending, 51 nodes are idle, and 50 idle nodes have eight MI355X GPUs. Multiple eligible jobs remain pending with `Reason=None`.

## Environment

- Controller: `controller1`
- SPUR version: `0.9.1`
- Partition: `<gpu-partition>`
- Observation date: `2026-08-30` UTC

All cluster-specific identifiers in this report have been replaced with descriptive placeholders. Resource types, resource counts, scheduler metrics, states, and relative timing are unchanged.

## Direct scheduler evidence

Snapshots over 38 seconds showed an advancing server clock but an unchanged scheduler cycle count:

```text
UTC T+00s
  Version           : 0.9.1
  Server Time       : T+00s
  PENDING           : 148
  RUNNING           : 115
  IDLE              : 51
  Cycles              : 206540
  Jobs started (last) : 0

UTC T+11s
  Server Time       : T+11s
  PENDING           : 148
  RUNNING           : 115
  IDLE              : 51
  Cycles              : 206540
  Jobs started (last) : 0

UTC T+21s
  Server Time       : T+21s
  PENDING           : 148
  RUNNING           : 115
  IDLE              : 51
  Cycles              : 206540
  Jobs started (last) : 0

UTC T+38s
  Server Time       : T+38s
  PENDING           : 148
  IDLE              : 51
  Cycles              : 206540
  Jobs started (last) : 0
```

The unchanged `Cycles=206540` over at least 38 seconds is strong evidence that scheduling cycles were not running, while the advancing `Server Time` shows that the controller itself was responsive.

## Controlled submission test

A minimal one-CPU, one-node batch job was submitted between scheduler snapshots:

```text
Pre-submission, T-10s
  Server Time       : T-10s
  PENDING           : 149
  RUNNING           : 115
  IDLE              : 51
  Cycles              : 206540
  Jobs started (last) : 0

Submitted batch job <probe-job> at T+00s

Post-submission, T+06s
  Server Time       : T+06s
  PENDING           : 150
  RUNNING           : 115
  IDLE              : 51
  Cycles              : 206540
  Jobs started (last) : 0
  <probe-job>|scheduler-cycle-probe|PD|None|T+00s|N/A|

Post-submission, T+11s
  Server Time       : T+11s
  Cycles              : 206540
  Jobs started (last) : 0
  <probe-job>|scheduler-cycle-probe|PD|None|T+00s|N/A|

Post-submission, T+26s
  Server Time       : T+26s
  Cycles              : 206540
  Jobs started (last) : 0
  <probe-job>|scheduler-cycle-probe|PD|None|T+00s|N/A|

Post-submission, T+56s
  Server Time       : T+56s
  PENDING           : 150
  RUNNING           : 115
  IDLE              : 51
  Cycles              : 206540
  Jobs started (last) : 0
  <probe-job>|scheduler-cycle-probe|PD|None|T+00s|N/A|
```

The submission was accepted and increased the pending count from 149 to 150, proving that controller state was updating. However, `Cycles` remained exactly `206540` from 10 seconds before submission through 56 seconds after submission. The probe remained pending with `Reason=None` despite 51 idle nodes. It was then manually cancelled to avoid leaving a stale diagnostic job.

Probe request:

```bash
#!/bin/bash
#SBATCH --job-name=scheduler-cycle-probe
#SBATCH --account=<account>
#SBATCH --qos=<qos>
#SBATCH --partition=<gpu-partition>
#SBATCH --nodes=1
#SBATCH --ntasks=1
#SBATCH --cpus-per-task=1
#SBATCH --time=00:01:00

hostname -f
```

## Idle capacity

```text
UTC <observation-time>
PARTITION|AVAIL|TIMELIMIT|NODES|STATE
<gpu-partition>*|up|1-00:00:00|111|alloc
<gpu-partition>*|up|1-00:00:00|7|down
<gpu-partition>*|up|1-00:00:00|9|drain
<gpu-partition>*|up|1-00:00:00|51|idle
<gpu-partition>*|up|1-00:00:00|4|mix

Eligible full idle nodes: 50
```

Example eligible nodes:

```text
NodeName=node1
   State=IDLE Reason=
   Partitions=<gpu-partition>
   CPUTot=236 CPUAlloc=0
   Gres=gpu:mi355x:1,gpu:mi355x:1,gpu:mi355x:1,gpu:mi355x:1,gpu:mi355x:1,gpu:mi355x:1,gpu:mi355x:1,gpu:mi355x:1

NodeName=node2
   State=IDLE Reason=
   Partitions=<gpu-partition>
   CPUTot=236 CPUAlloc=0
   Gres=gpu:mi355x:1,gpu:mi355x:1,gpu:mi355x:1,gpu:mi355x:1,gpu:mi355x:1,gpu:mi355x:1,gpu:mi355x:1,gpu:mi355x:1
```

## Affected jobs

Two interactive allocations each requested one node, 32 CPUs, and eight MI355X GPUs. Neither received a start time or scheduling reason. Both were cancelled after exactly five minutes while still unstarted:

```text
JobId=<interactive-job-1> JobName=interactive
   UserId=<user> Account=<account>
   Partition=<gpu-partition> QOS=<qos>
   JobState=CANCELLED Reason=None
   NumNodes=1 NumTasks=1 CPUs/Task=32
   TresPerNode=gpu:mi355x:8/node ReqGPUs=8
   SubmitTime=T+00s StartTime=N/A EndTime=T+05m01s
   ExitCode=-1:0

JobId=<interactive-job-2> JobName=interactive
   UserId=<user> Account=<account>
   Partition=<gpu-partition> QOS=<qos>
   JobState=CANCELLED Reason=None
   NumNodes=1 NumTasks=1 CPUs/Task=32
   TresPerNode=gpu:mi355x:8/node ReqGPUs=8
   SubmitTime=T+06m02s StartTime=N/A EndTime=T+11m02s
   ExitCode=-1:0
```

Other eligible eight-GPU jobs across several accounts and QOSes were also pending with `Reason=None`. This indicates the symptom was cluster-wide rather than isolated to one user or QOS.

## QOS is not the limiting factor

```text
<qos> ... GrpTRES=node=64
QOS running node count: 27
```

The affected QOS was using 27 of its 64-node group limit. The affected jobs reported `Reason=None`, not `QOSGrpNodeLimit`.

## Expected behavior

The backfill scheduler should continue incrementing its cycle count and assign eligible pending jobs to idle matching nodes, or provide a concrete pending reason.

## Actual behavior

The controller remained responsive, but `Cycles` stayed fixed at `206540`, no jobs started, and eligible jobs remained pending with `Reason=None` despite 50 matching idle eight-GPU nodes.

## Commands used

```bash
spur diag
sinfo -p <gpu-partition> -o '%P|%a|%l|%D|%t|%G'
sinfo -N -p <gpu-partition> -t idle -o '%N|%t|%G'
spur show node node1
spur show node node2
spur show job <interactive-job-1>
spur show job <interactive-job-2>
squeue -t PD -o '%i|%u|%a|%q|%t|%r|%V|%b|%C|%D'
spur accounts show qos
```

yansun1996:
> **Reading the diff:** this branch *carries* @jamesETsmith's commits, so the net diff against `main` shows the final shape only. The channel-wide `.timeout(5s)` discussed below is introduced by `97958f54` and replaced by `fd780096`, both in this branch — you will not find a deletion of it in the three-dot diff.

## Motivation

Builds on @jamesETsmith's #768, which correctly identified #767 (the scheduler stops cycling while eligible nodes stay idle) and the right chokepoint for a fix — `agent_client::connect`. This carries his commits and adjusts the shape of the bound.

A channel-wide `.timeout(5s)` in tonic installs a per-channel `GrpcTimeout` layer that bounds time-to-response-headers. Because a tonic server emits headers only once a unary handler resolves, that caps the whole agent-side handler — including handlers that legitimately run for minutes:

| RPC | Inline work before the response |
| --- | --- |
| `RunCommand` (every srun step) | `child.wait_with_output()` — returns when the user's step exits |
| `LaunchJob` | container rootfs setup, the node prolog, process spawn |
| `ExecInJob` | `cmd.output().await` — the full command |
| `StartClusterComponent` | can download and install the k0s binary |

It also overrides an explicit bound already written down in the code: `drain_via_control_plane` computes `drain timeout (default 120s) + 30s slack` and its comment states the RPC is bounded so a hung agent can't stall the serial removal loop. That 150s becomes unreachable behind a 5s channel deadline.

## Behaviour by failure mode

Everything below is one launch RPC to one node. Because the assignment loop is still serial, a stall here blocks every other pending job, not just this one.

| Scenario | Before (`main`) | With a channel-wide 5s deadline | This PR |
| --- | --- | --- | --- |
| Connection refused (agent down, port closed) | ~0s, kernel RST | ~0s | ~0s, node skipped for the reject cooldown |
| Packets blackholed, TCP never completes | ~2min, kernel SYN retries | 5s | 5s (`connect_timeout`) |
| **TCP accepted, then the peer goes silent** | **never returns** | 5s | **~20s** via HTTP/2 keepalive |
| **Agent alive, handler wedged (hung prolog, stuck image unpack)** | **never returns** | 5s | **`dispatch_timeout_secs`** (default 300s), then the node is skipped for the same span |
| Healthy but slow prolog (e.g. 60s) | completes | **aborted at 5s** | completes |
| srun step longer than 5s | completes | **aborted at 5s** | completes |

The two "never returns" rows are the reported hang: once the connection is established nothing gives up, because a black-holed peer sends neither FIN nor RST, and there is no request deadline or keepalive. They are indistinguishable from the controller's side, which is why both need covering — keepalive catches the silent peer, but a wedged handler still answers pings, so only the call-site deadline bounds it.

Note the last two rows: a single short channel-wide deadline cannot separate "stuck" from "slow", which is why the bound moved to the call sites.

## Technical Details

- Keep the dial budget (`connect_timeout`), drop the channel-wide request budget.
- Add HTTP/2 keepalive (10s interval, 10s timeout, `keep_alive_while_idle`). This is what actually catches the reported failure — a peer that accepts TCP and then stops answering. Pings are sent while a request stream is open, so a hung RPC is torn down in roughly 20s without capping slow-but-healthy work.
- Add a per-operation deadline, `controller.dispatch_timeout_secs` (default 300, `0` disables), on the two fan-outs that are drained to completion: `confirm_dispatch_on_nodes` and `register_allocation_on_nodes`. The latter had no bound at all and is awaited from the same serial assignment loop.
- Cancel every dispatched node when admission aborts, not only the confirmed ones. `Endpoint::timeout` does not send a `grpc-timeout` header, so an agent never learns the controller gave up — the node that timed out is the one most likely to have launched anyway, and was the only one receiving no cancel. `CancelJob` is idempotent.
- Cool a timed-out node down for the deadline it burned rather than the much shorter reject cooldown. Paired with a long deadline, the 30s cooldown left a wedged node stalling most of every cycle — and while assignments are processed serially that stall is cluster-wide.
- Log before dispatching. A launch that never returns was otherwise invisible for as long as it hung, which is exactly when it needs to be visible.

## Tunables

All four are `[controller]` fields with defaults, and all reload with `scontrol reconfigure`:

| Field | Default | Bounds |
| --- | --- | --- |
| `agent_connect_timeout_secs` | 5 | TCP dial. `0` falls back to the OS timeout (~2min) |
| `agent_keepalive_interval_secs` | 10 | HTTP/2 ping interval. `0` disables keepalive |
| `agent_keepalive_timeout_secs` | 10 | Ping response wait |
| `dispatch_timeout_secs` | 300 | One launch/register RPC. `0` disables |

Keepalive detection is roughly interval + timeout, so ~20s at the defaults. The three channel values reload live because a channel is built per call — a swapped value applies to the next RPC with nothing to invalidate. Unlike the jwt key, which `reconfigure` deliberately does not adopt, these carry no security posture. All are capped in `validate()`; the channel defaults live in `spur-core` as shared constants so the config and the builder cannot drift.

## On the default

`connect_timeout` and the keepalive pair are the liveness mechanism; `dispatch_timeout_secs` is a backstop for an agent that is alive and answering pings but wedged in a handler. It is deliberately generous because a site prolog can legitimately take minutes, and it is a knob so operators can tighten it once their prologs are bounded.

## Test Plan / Result

- `a_node_whose_launch_times_out_is_still_cancelled` — new. Mutation-checked: reverting the cancel back to `succeeded_nodes` makes it fail, so it is a real regression signal rather than a test that passes either way.
- `cargo test --locked` — full workspace green.
- `cargo fmt --all -- --check` — clean.

## Review round

Four gaps raised in review, all confirmed against the code and fixed here:

- `PreparePmix` and `ReleasePmix` had no ceiling once the channel-wide timeout was dropped. The agent runs PMIx `server_start` inline in that handler, so it answers keepalive pings while wedged and only a call-site deadline bounds it. Prepare now takes `dispatch_timeout_secs` per agent (covering the srun-step call site too); release gets a fixed 5s inside `release_pmix_on_agent`, so the rollback path and the guard's `Drop` are covered — a best-effort teardown must never be what blocks the loop.
- The register path cooled a node only when it timed out. A connect failure or RPC error fell through with no cooldown, which is the asymmetry this branch introduced by moving `Unreachable` into the launch path's cooldown arm. Split into `TimedOut` / `Failed` so each is held for the span it earned.
- That same arm called `requeue_job`, which returns early for a job still `Pending` — so a failed registration had neither a cooldown nor a backoff. It now uses `backoff_pending_job_after_dispatch_failure`, as the launch path does.
- The widened cancel and both deadline cooldowns had no coverage: reverting them left the suite green. Three new tests and one extended assertion now fail when the arm they cover is reverted.

Also worth calling out from an earlier commit here: `cluster_k8s.rs` moved to `agent_client::connect`, which attaches the controller credential interceptor. `stop_component_now` previously sent no bearer token and would have been rejected by the agent under `auth.mode != disabled`, so that path was broken before this branch.

Live-checked on an isolated instance: a 15s prolog still completes inside the launch RPC, and a node that keeps heartbeating with its RPC port blackholed is retried ~35s apart rather than every tick, starting as soon as it is reachable again.

- The assignment loop is still serial, so a bounded deadline turns an unbounded stall into a bounded one rather than removing head-of-line blocking.
- `AllocationRegisterOutcome::PartialFailed` carries only `succeeded_nodes`, so the same cancel gap remains on that path; closing it means changing the variant.

Closes #767.

### DIFF
--- a/crates/spurctld/src/scheduler_loop.rs
+++ b/crates/spurctld/src/scheduler_loop.rs
@@ -1042,11 +1042,29 @@ impl<E: Into<anyhow::Error>> From<E> for DispatchError {
     }
 }
 
+/// Ceiling on one agent RPC in a fan-out that is drained to completion. `0` disables it, matching
+/// the sibling `dispatch_reject_cooldown_secs` knob.
+fn dispatch_deadline(cluster: &ClusterManager) -> Option<Duration> {
+    match cluster.config().controller.dispatch_timeout_secs {
+        0 => None,
+        secs => Some(Duration::from_secs(secs)),
+    }
+}
+
 /// Send a LaunchJob RPC to a node agent.
 async fn dispatch_to_agent(
     agent_addr: &str,
     params: &AgentDispatchParams<'_>,
 ) -> Result<LaunchOutcome, DispatchError> {
+    // Logged before the call, not just on its outcome: a launch that never returns is otherwise
+    // invisible for as long as it hangs, which is exactly when an operator needs to see it.
+    info!(
+        job_id = params.job_id,
+        node = params.target_node,
+        agent = agent_addr,
+        run_attempt = params.run_attempt,
+        "dispatching job to agent"
+    );
     let mut client = crate::agent_client::connect(agent_addr.to_string())
         .await
         .map_err(|e| DispatchError::Unreachable(e.into()))?
@@ -1318,6 +1336,7 @@ async fn register_allocation_on_nodes(
     let mut failures = 0u32;
     let mut succeeded_nodes: Vec<String> = Vec::new();
     let total = dispatch_nodes.len() as u32;
+    let dispatch_timeout = dispatch_deadline(&cluster);
 
     let mut set = tokio::task::JoinSet::new();
     for node_name in &dispatch_nodes {
@@ -1360,7 +1379,17 @@ async fn register_allocation_on_nodes(
             work_dir: spec.work_dir.clone(),
         };
         set.spawn(async move {
-            let result = register_allocation_to_agent(&agent_addr, &params).await;
+            let register = register_allocation_to_agent(&agent_addr, &params);
+            let result = match dispatch_timeout {
+                Some(limit) => match tokio::time::timeout(limit, register).await {
+                    Ok(result) => result,
+                    Err(_) => Err(anyhow::anyhow!(
+                        "allocation register RPC exceeded {}s",
+                        limit.as_secs()
+                    )),
+                },
+                None => register.await,
+            };
             (result_node, result)
         });
     }
@@ -1508,6 +1537,7 @@ async fn confirm_dispatch_on_nodes(
     let modex_connect_timeout_secs = cluster.config().mpi.modex_connect_timeout_secs;
     let modex_fence_timeout_secs = cluster.config().mpi.modex_fence_timeout_secs;
     let modex_verify_timeout_secs = cluster.config().mpi.modex_verify_timeout_secs;
+    let dispatch_timeout = dispatch_deadline(&cluster);
 
     let needs_pmix_prepare = batch_dispatched_multi_node_pmix(
         spec.mpi.as_deref(),
@@ -1654,28 +1684,37 @@ async fn confirm_dispatch_on_nodes(
         let pmix_tmpdir = pmix_tmpdir.clone();
         let agent_addr = agent_addr.clone();
         set.spawn(async move {
-            let result = dispatch_to_agent(
-                &agent_addr,
-                &AgentDispatchParams {
-                    job_id,
-                    spec: &spec,
-                    peer_nodes: &peer_addrs,
-                    peer_hosts: &peer_hosts,
-                    node_index,
-                    task_offset,
-                    target_node: &target_node,
-                    allocated: &allocated,
-                    allocated_nodelist: &allocated_nodelist,
-                    run_attempt,
-                    pmix_tmpdir: &pmix_tmpdir,
-                    task_fanout,
-                    modex_connect_timeout_secs,
-                    modex_fence_timeout_secs,
-                    modex_verify_timeout_secs,
-                    pmix_prepared: needs_pmix_prepare,
+            let params = AgentDispatchParams {
+                job_id,
+                spec: &spec,
+                peer_nodes: &peer_addrs,
+                peer_hosts: &peer_hosts,
+                node_index,
+                task_offset,
+                target_node: &target_node,
+                allocated: &allocated,
+                allocated_nodelist: &allocated_nodelist,
+                run_attempt,
+                pmix_tmpdir: &pmix_tmpdir,
+                task_fanout,
+                modex_connect_timeout_secs,
+                modex_fence_timeout_secs,
+                modex_verify_timeout_secs,
+                pmix_prepared: needs_pmix_prepare,
+            };
+            let dispatch = dispatch_to_agent(&agent_addr, &params);
+            // The join below drains every node, so one agent that accepts the call and never answers
+            // would otherwise hold the whole scheduler loop, not just this job.
+            let result = match dispatch_timeout {
+                Some(limit) => match tokio::time::timeout(limit, dispatch).await {
+                    Ok(result) => result,
+                    Err(_) => Err(DispatchError::Unreachable(anyhow::anyhow!(
+                        "launch RPC exceeded {}s",
+                        limit.as_secs()
+                    ))),
                 },
-            )
-            .await;
+                None => dispatch.await,
+            };
             (result_node, is_primary, result)
         });
     }
@@ -1697,10 +1736,12 @@ async fn confirm_dispatch_on_nodes(
                     DispatchError::PrologFailed(reason) => {
                         prolog_failed.push((node_name, reason));
                     }
-                    DispatchError::ResourcesUnavailable => cluster.cool_down_node(&node_name),
-                    DispatchError::Unreachable(_)
-                    | DispatchError::AgentRejected(_)
-                    | DispatchError::Other(_) => {}
+                    // An unreachable node stays a candidate otherwise, so the next job pays the same
+                    // dispatch timeout again and the queue walks itself into max-requeue holds.
+                    DispatchError::ResourcesUnavailable | DispatchError::Unreachable(_) => {
+                        cluster.cool_down_node(&node_name)
+                    }
+                    DispatchError::AgentRejected(_) | DispatchError::Other(_) => {}
                 }
             }
             Err(e) => {
@@ -1751,11 +1792,9 @@ async fn confirm_dispatch_on_nodes(
     };
     let _ = cluster.set_job_launch_failure_detail(job_id, confirmation_detail.clone());
 
-    // Stop whatever DID launch before the job settles anywhere: a node that
-    // never confirmed will never report completion, and letting it keep
-    // running while the job as a whole is aborted back to Pending would
-    // orphan it.
-    cancel_job_on_nodes(&cluster, job_id, &succeeded_nodes, 9).await;
+    // Every dispatched node, not just the confirmed ones: a node that timed out may have launched
+    // anyway and is the likeliest to be orphaned. CancelJob is idempotent, so cancelling wide is safe.
+    cancel_job_on_nodes(&cluster, job_id, &dispatch_nodes, 9).await;
 
     // Drain before deciding the job's fate, so the failing node is already out
     // of the candidate set on the next scheduling attempt. The drain is issued
@@ -4328,6 +4367,59 @@ mod tests {
 
             assert!(matches!(outcome, DispatchConfirmOutcome::Confirmed));
             elapsed
+        }
+
+        /// The node that timed out is the one most likely to have launched anyway, so it is the one
+        /// that must be cancelled — cancelling only the confirmed nodes orphans it.
+        #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+        async fn a_node_whose_launch_times_out_is_still_cancelled() {
+            let dir = TempDir::new().unwrap();
+            let mut config = test_config();
+            config.controller.dispatch_timeout_secs = 1;
+            let cm = test_cluster_with_config(&dir, config).await;
+
+            let (fast_addr, fast_cancels) = spawn_mock_agent().await;
+            let (slow_addr, slow_cancels) =
+                spawn_mock_agent_with_delay(Duration::from_secs(30)).await;
+            register_node_at(&cm, "n-fast", fast_addr);
+            register_node_at(&cm, "n-slow", slow_addr);
+
+            let nodes = vec!["n-fast".to_string(), "n-slow".to_string()];
+            let spec = batch_spec("timeout-cancel", 2);
+            let job_id = submit_and_wait(&cm, spec);
+            let spec = cm.get_job(job_id).unwrap().spec;
+            let per_node_allocs: HashMap<String, ResourceAllocations> = nodes
+                .iter()
+                .map(|n| (n.clone(), ResourceAllocations::with_scalar(1, 0)))
+                .collect();
+            let nodelist = nodes.join(",");
+
+            let outcome = confirm_dispatch_on_nodes(
+                cm.clone(),
+                job_id,
+                nodes,
+                spec,
+                Vec::new(),
+                per_node_allocs,
+                nodelist,
+                1,
+                1,
+                false,
+            )
+            .await;
+
+            assert!(
+                !matches!(outcome, DispatchConfirmOutcome::Confirmed),
+                "a node that never answered must not count as confirmed"
+            );
+            assert!(
+                slow_cancels.load(Ordering::SeqCst) > 0,
+                "the timed-out node received no cancel and would keep running the job"
+            );
+            assert!(
+                fast_cancels.load(Ordering::SeqCst) > 0,
+                "the confirmed node must still be cancelled when admission aborts"
+            );
         }
 
         #[tokio::test(flavor = "multi_thread", worker_threads = 16)]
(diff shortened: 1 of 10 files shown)


crates/spurctld/src/scheduler_loop.rs
```diff
@@ -1654,37 +1690,42 @@ async fn confirm_dispatch_on_nodes(
...
+                pmix_prepared: needs_pmix_prepare,
+            };
+            let dispatch = dispatch_to_agent(&agent_addr, &params);
+            // The join below drains every node, so one agent that accepts the call and never answers
+            // would otherwise hold the whole scheduler loop, not just this job.
+            let result = match dispatch_timeout {
+                Some(limit) => match tokio::time::timeout(limit, dispatch).await {
+                    Ok(result) => result,
+                    Err(_) => Err(DispatchError::TimedOut(limit)),
                 },
```
shiv-tyagi, line 1741: `prepare_pmix_on_nodes` sits outside `dispatch_timeout_secs`. On native multi-node `--mpi=pmix`, a live agent wedged in PMIx `server_start` (or peer verification) still answers HTTP/2 pings, so keepalive does not tear the call down. The serial assignment loop blocks there and never reaches this LaunchJob deadline.

Silent transport failure is already bounded by keepalive. This is the live-but-wedged handler case. K8s virtual agents reject multi-node PMIx at submit.

crates/spurctld/src/scheduler_loop.rs
```diff
@@ -1751,11 +1795,9 @@ async fn confirm_dispatch_on_nodes(
...
     let _ = cluster.set_job_launch_failure_detail(job_id, confirmation_detail.clone());
 
-    // Stop whatever DID launch before the job settles anywhere: a node that
-    // never confirmed will never report completion, and letting it keep
-    // running while the job as a whole is aborted back to Pending would
-    // orphan it.
-    cancel_job_on_nodes(&cluster, job_id, &succeeded_nodes, 9).await;
+    // Every dispatched node, not just the confirmed ones: a node that timed out may have launched
+    // anyway and is the likeliest to be orphaned. CancelJob is idempotent, so cancelling wide is safe.
+    cancel_job_on_nodes(&cluster, job_id, &dispatch_nodes, 9).await;
```
shiv-tyagi, line 1823: CancelJob is bounded by `CANCEL_RPC_TIMEOUT`. The `release_pmix_on_agents` await just above this is not.

This path runs after multi-node PMIx prepare succeeded and a later LaunchJob failed. The abort disarms the fire-and-forget release guard, then waits on `ReleasePmix`, which calls synchronous PMIx `server_stop`. A wedged but ping-healthy agent blocks the scheduler again.

crates/spurctld/src/scheduler_loop.rs
```diff
@@ -343,19 +343,15 @@ async fn process_assignment(
         )
         .await
         {
-            AllocationRegisterOutcome::AllFailed => {
+            // Both arms tear down identically: with a deadline in play, even "all failed" can mean
+            // every node registered and answered too late, so none of them may be left holding one.
+            AllocationRegisterOutcome::AllFailed | AllocationRegisterOutcome::PartialFailed => {
+                cancel_job_on_nodes(&cluster, job_id, &all_nodes, 9).await;
```
shiv-tyagi, line 349: The AllFailed test uses an unreachable address and only checks requeue. It cannot see this wide cancel. Nothing delays `register_job_allocation`, asserts the timeout cooldown, or checks cancel when every register call answers too late. Reverting these two arms stays green.

The production case is every node registering near the deadline and answering after it.

crates/spurctld/src/scheduler_loop.rs
```diff
@@ -1359,17 +1380,32 @@ async fn register_allocation_on_nodes(
...
+                    Ok(result) => result,
+                    Err(_) => {
+                        // Same reasoning as the launch fan-out: without this the node is re-picked
+                        // next tick and burns another full deadline.
+                        cooldown_cluster.cool_down_node_for(&result_node, limit);
+                        Err(anyhow::anyhow!(
+                            "allocation register RPC exceeded {}s",
+                            limit.as_secs()
+                        ))
+                    }
```
shiv-tyagi: This only cools the tokio deadline. Immediate connect failures and RPC errors fall through with no cooldown. LaunchJob cools the equivalent Unreachable case with the short reject cooldown.

On native `srun` (not K8s), a node that still heartbeats outbound while 6818 is refused or blocked gets re-picked every tick. Each cycle pays up to `agent_connect_timeout_secs` (default 5s) on the serial assignment loop.

shiv-tyagi (commented):
Some comments, PTAL

crates/spurctld/src/scheduler_loop.rs
yansun1996: Confirmed — and it is a gap this PR opened, since the channel-wide timeout used to bound this call. `PreparePmix` now takes the same `dispatch_timeout_secs` deadline, applied per agent inside `prepare_pmix_on_agent` so the srun-step call site in `server.rs` is covered too and a slow node is named in the error. Added a test pointing prepare at a peer that completes TCP and then never serves HTTP/2: keepalive alone needs ~20s to unwind it, the deadline fires first.

crates/spurctld/src/scheduler_loop.rs
yansun1996: Confirmed. Bounded it inside `release_pmix_on_agent` at a fixed 5s rather than at this call site, so the prepare-path rollback and the guard's `Drop` get it as well — a best-effort rollback should never be the thing that blocks the loop. Deliberately not `dispatch_timeout_secs`: that knob is sized for a slow prolog, which says nothing about how long a teardown deserves.

crates/spurctld/src/scheduler_loop.rs
yansun1996: You are right, and reverting confirms it: re-splitting the arms back into `AllFailed` (no cancel) and `PartialFailed { succeeded_nodes }` left the suite green, as did dropping the register-path cooldown. The launch-path deadline cooldown turned out to have the same problem.

Added `every_node_timing_out_on_register_is_cancelled_cooled_and_backed_off` — every node behind a delayed `register_job_allocation`, asserting both are cancelled, both land on cooldown, and the job is backed off — plus `a_node_that_burns_the_launch_deadline_is_held_for_it`, and a cooldown assertion on the existing unreachable-node test. Each one fails when the arm it covers is reverted.

crates/spurctld/src/scheduler_loop.rs
yansun1996: Confirmed, and this branch created the asymmetry — it moved `Unreachable` into the launch path's cooldown arm without doing the same here. The register result is now split into `TimedOut` and `Failed`, so a burned deadline is held for its own span and a connect failure takes the short reject cooldown.

Digging into it, the hot loop was tighter than described: this arm called `requeue_job`, which returns early for a job still `Pending`, so a failed registration had no backoff either — nothing throttled the retry at all. It now uses `backoff_pending_job_after_dispatch_failure`, the same call the launch path makes.

Checked against a node still heartbeating with its RPC port blackholed: retries now land ~35s apart (dial budget plus cooldown) instead of every tick, and the job starts as soon as the node is reachable again.

shiv-tyagi (approved):
LGTM

PR #785, opened 2026-09-01
fix(spurctld): stop logging replayed WAL operations as new events

pre:
Every start of spurctld re-applies the whole Raft log, and each apply logged again at INFO. One node removal was thus printed once per controller process, on every node, with the timestamp of the restart. A reader who greps the journal finds a removal that did not happen, and looks for a fault in the wrong place.

The store now records the highest log index that was on disk at startup. Every entry up to it is a replay, and is applied inside a `wal_replay` span. The subscriber drops INFO and below inside that span. WARN and ERROR pass, because a bad transition during a replay is a real defect.

The state is applied exactly as before: only the log output changes.

Measured on a three-node cluster: before the change, a restart printed the old "node removed from cluster" line again on every node. After it, a restart that replays 46 entries prints none, sinfo still shows every node, and a live shutdown still logs.

## Technical Details

One node removal is printed once per controller process, on every node, carrying the timestamp of the **restart** rather than of the event:

```
Aug 31 10:49:15 host spurctld[6618]: INFO spurctld::cluster: node removed from cluster node=worker-3 reason="agent shutdown"
Aug 31 11:39:47 host spurctld[12125]: INFO spurctld::cluster: node removed from cluster node=worker-3 reason="agent shutdown"
Aug 31 11:43:40 host spurctld[12919]: INFO spurctld::cluster: node removed from cluster node=worker-3 reason="agent shutdown"
Aug 31 12:00:02 host spurctld[1241]: INFO spurctld::cluster: node removed from cluster node=worker-3 reason="agent shutdown"
```

One removal happened. The node was in `sinfo` the whole time. A reader who greps the journal after a restart finds a removal that did not happen and looks for a fault in the wrong place. This cost real time while diagnosing an unrelated defect, because the replayed line looked like a live event.

The same applies to every other `INFO` in `apply_operation`: partitions created, reservations updated, and so on.

`SpurStore` records the highest log index that was on disk when the process started (`replay_upto`). Every entry up to it is history, and `apply_to_state_machine` applies it inside a `wal_replay` span. The subscriber drops `INFO` and below inside that span.

`WARN` and `ERROR` still pass, because an invalid transition during a replay is a real defect and must stay visible.

Only the log output changes. The state is applied exactly as before.

## Test Plan

`replay_upto_is_the_highest_index_on_disk` fails if the boundary between history and a live event is lost. 963 unit tests pass.

Verified on a three-node cluster:

| | before | after |
|---|---|---|
| `node removed from cluster` in the journal | 4, 4 and 7 lines, one per controller start | none |
| the replay itself | unchanged | `raft store recovered from disk log_entries=48 replay_upto=Some(47)` |
| `sinfo` after the restart | 3 nodes | 3 nodes, so the state is still applied |
| a real removal, live | logged | still logged |

The last row is the one that matters: with this change alone on the cluster, an actual node removal still prints its `INFO` line. The filter hides replayed history, not events.

## Test Result

Verification so far, since the workflow runs need maintainer approval:

Local, on this branch alone (rebased on 360d44c):
- `cargo fmt --check --all` clean
- `cargo clippy -p spurctld` no warnings
- 963 unit tests pass, including the new
  `replay_upto_is_the_highest_index_on_disk`

On a three-node cluster, with this change alone and nothing else:
- A restart that replays 48 log entries prints no `node removed from cluster`
  line, where before it printed the old one again on every node.
- `raft store recovered from disk log_entries=48 replay_upto=Some(47)` shows
  the replay still happens, and `sinfo` still lists every node afterwards, so
  the state is applied exactly as before.
- A real node removal, live, still prints its `INFO` line. This is the case
  that matters: the filter hides replayed history, not events.

The E2E suite has not run yet. Could a maintainer approve the workflow runs?

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.


crates/spurctld/src/raft.rs
```diff
@@ -485,7 +500,18 @@ impl openraft::RaftStorage<SpurTypeConfig> for Arc<SpurStore> {
...
+                    let replaying = inner
+                        .replay_upto
+                        .is_some_and(|upto| entry.log_id.index <= upto);
+                    if replaying {
+                        let span = tracing::info_span!(WAL_REPLAY_SPAN, index = entry.log_id.index);
+                        let _enter = span.enter();
+                        results.push(self.applier.apply_operation(op));
+                    } else {
+                        results.push(self.applier.apply_operation(op));
+                    }
```
shiv-tyagi: `apply_operation(op)` is called identically in both branches. Collapse to one call site with an optional span guard:

```suggestion
                    let replaying = inner
                        .replay_upto
                        .is_some_and(|upto| entry.log_id.index <= upto);
                    let _replay_span = replaying.then(|| {
                        tracing::info_span!(WAL_REPLAY_SPAN, index = entry.log_id.index).entered()
                    });
                    results.push(self.applier.apply_operation(op));
```

crates/spurctld/src/raft.rs
```diff
@@ -113,8 +113,20 @@ struct StoreInner {
     last_applied: Option<LogId<NodeId>>,
     last_membership: StoredMembership<NodeId, BasicNode>,
     applied_count: u64,
+    /// Highest log index present on disk at startup. Everything up to it is
+    /// replayed, not new, thus its side effects must not be reported as
+    /// events that happen now. See [`WAL_REPLAY_SPAN`].
+    #[serde(skip)]
+    replay_upto: Option<u64>,
```
shiv-tyagi: Nit: this rationale is repeated near-verbatim on `WAL_REPLAY_SPAN` below and in `init_tracing` (main.rs). Keep the full explanation on `WAL_REPLAY_SPAN` as the single source and trim here.

```suggestion
    /// Highest log index on disk at startup. Entries up to it are replays,
    /// not new events. See [`WAL_REPLAY_SPAN`].
    #[serde(skip)]
    replay_upto: Option<u64>,
```

crates/spurctld/src/raft.rs
```diff
@@ -1399,6 +1425,34 @@ mod tests {
...
+                        },
+                        payload: EntryPayload::Blank,
+                    })
+                    .unwrap();
+            }
+        }
+
+        let restarted = SpurStore::new(dir.path(), noop_applier()).unwrap();
+        assert_eq!(restarted.inner.read().replay_upto, Some(17));
+    }
```
shiv-tyagi, line 1446: This test only asserts `replay_upto` is set to the highest index. The behavior this PR actually fixes (INFO dropped inside `WAL_REPLAY_SPAN`, WARN/ERROR passing through) has no test and rests solely on the manual cluster check. That predicate is the part most likely to regress silently. A test with a capturing subscriber that emits INFO and WARN both inside and outside the span, asserting what gets recorded, would lock it in without a cluster.

shiv-tyagi (commented):
PTAL at the comments.

crates/spurctld/src/raft.rs
pre: Done in 2bf6a0b: one call site with the `Option<EnteredSpan>` guard. The loop has no await and already holds a `parking_lot` guard, so the span cannot cross one.

crates/spurctld/src/raft.rs
pre: Done in 5b945a8: the rationale lives on `WAL_REPLAY_SPAN` only. The field doc is two lines with a link, the filter has a one-line doc, and the inline comment in the apply loop is gone too.

crates/spurctld/src/raft.rs
pre: Done in 5b945a8: the filter moved to `logging.rs`, where `subscriber(env_filter, writer)` is the production stack with the writer as a parameter, and six tests drive it with a capturing writer: INFO dropped inside the span and in a child span, WARN and ERROR kept, INFO kept outside and in unrelated spans, and the `EnvFilter` still applied.

One more case turned up while writing them: `delete_conflict_logs_since` did not lower `replay_upto`, so the leader's replacement entries after a truncation were silenced as replays. Fixed in 369bcd0 with a store test.

shiv-tyagi:

PR #786, opened 2026-09-01
feat(spur-cli): read cluster name from controller in scontrol show config

shiv-tyagi:
## What

`scontrol show config` printed a hardcoded `ClusterName=spur`, ignoring the controller's configured `cluster_name`. It now reports the live value from the controller.

## Approach

- Add an additive `PingResponse.cluster_name = 5` (previously-unused tag, no `!` needed).
- Controller fills it from `config().cluster_name` in `ping`.
- `scontrol show config` pings the controller and prints `ClusterName` and `Version` from the response instead of the hardcoded string. A small `format_config` helper makes the rendering unit-testable.
- The login node no longer needs a local `/etc/spur/spur.conf` to report the cluster name; the controller is the source of truth.

## Notes

- `Version` in `scontrol show config` now reflects the controller's version (`ping.version`) rather than the CLI binary's `CARGO_PKG_VERSION`. This is more correct for a command that describes the controller, and the output format is unchanged. The config command already required a reachable controller (the client connects eagerly), so the ping adds no new failure path.
- Scoped intentionally to the CLI wiring. Propagating `cluster_name` into the job/hook runtime environment (`SPUR_CLUSTER_NAME` / `SLURM_CLUSTER_NAME`, salloc, prolog/epilog) is a separate, larger change and will land in a follow-up PR.

## Test plan

- [x] `cargo test -p spurctld ping_reports_configured_cluster_name`
- [x] `cargo test -p spur-cli config_reports_cluster_name_from_ping`
- [x] `cargo clippy -p spur-cli -p spurctld --all-targets` clean
- [x] Manual: deployed to the bare-metal testbed, `scontrol show config` reports `ClusterName=bare-metal-test` from the controller; confirmed it ignores a divergent local `spur.conf` on a different node

### DIFF
--- a/crates/spur-cli/src/scontrol.rs
+++ b/crates/spur-cli/src/scontrol.rs
@@ -555,6 +555,15 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
     }
 }
 
+/// Render `scontrol show config` from a controller ping. The cluster name and
+/// version follow the live controller, not the login node's local config.
+fn format_config(controller: &str, ping: &spur_proto::proto::PingResponse) -> String {
+    format!(
+        "ClusterName={}\nSlurmctldAddr={}\nVersion={}\n",
+        ping.cluster_name, controller, ping.version
+    )
+}
+
 async fn show(controller: &str, entity: &str, name: Option<&str>) -> Result<()> {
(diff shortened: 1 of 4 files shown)


biluriuday (approved):
LGTM

PR #795, opened 2026-09-02
fix(spur-cli): honor --export=ALL,VAR=val combined form and inline assignments

### LINKED ISSUE #794: sbatch --export ignores the combined ALL,VAR=val form and inline VAR=value assignments
## What happens

`sbatch --export` only handles the bare tokens `ALL` and `NONE`. For anything else it filters the submission environment by exact key match. Two Slurm-compatible forms break:

- `--export=ALL,MASTER_PORT=29999,WORLD_SIZE=16` (combined form): the value equals neither `ALL` nor `NONE`, so it falls into the list branch. No environment key equals `ALL` or `MASTER_PORT=29999`, so the whole environment is dropped and the inline assignments are never set.
- `--export=VAR=value` (inline assignment): the `VAR=value` token is compared against environment keys and never matches, so the variable is not set.

## What is expected (Slurm behavior)

- `ALL` propagates the full submission environment.
- `NONE` propagates nothing.
- A leading `ALL,` followed by a list propagates the full environment and then sets the listed variables on top.
- A plain list propagates only the named variables.
- Each token is either `VAR` (copy the current value from the submission env) or `VAR=value` (set an explicit value, overriding an inherited one).

## Reproduce

Submit a 2-node job with the combined form:

```bash
sbatch -N 2 --export=ALL,MASTER_PORT=29999,WORLD_SIZE=16 --wrap 'echo "MASTER_PORT=$MASTER_PORT WORLD_SIZE=$WORLD_SIZE"'
```

Observed: `MASTER_PORT` and `WORLD_SIZE` are empty in the job, and the rest of the submission environment is missing.
Expected: the full submission environment is present, with `MASTER_PORT=29999` and `WORLD_SIZE=16` set.

shiv-tyagi:
Fixes #794

## What

`sbatch --export` now supports Slurm's combined form (`--export=ALL,VAR=val`) and inline `VAR=value` assignments.

## Why

The old parser only recognized the bare values `ALL` and `NONE`; anything else was treated as a comma list of keys to filter from the submission environment. So:

- `--export=ALL,MASTER_PORT=29999,WORLD_SIZE=16` matched neither special case and fell into the list branch. No environment key equals `ALL` or `MASTER_PORT=29999`, so the whole environment was dropped and the inline assignments were never set.
- `--export=VAR=value` inline assignments were unsupported in any form.

## Approach

Parse `--export` per Slurm semantics:

- A leading `ALL` seeds the full submission environment; `NONE` seeds an empty one; any other leading token starts empty.
- Each remaining token is applied on top: `VAR` copies the current value from the submission env, `VAR=value` sets an explicit value (overriding an inherited one). The value may itself contain `=`.

The logic is factored into `resolve_export_env`, which takes the source environment explicitly so it is unit-testable without depending on the runner's environment.

Existing behavior is preserved: `ALL` -> full env, `NONE` -> empty, plain list -> named keys copied from the env.

## Changes

- `resolve_export_env` helper in `spur-cli` replaces the inline `if/else` in `build_sbatch_job_spec`.
- Docs: document the `VAR=value` and `ALL,VAR=value` forms with an example.

## Testing

- Unit tests for `resolve_export_env` (ALL, NONE, plain list, combined override, inline add, bare-name copy, value containing `=`).
- Added a multi-node e2e regression (`test_export_all_combined_inline_assignments`) that submits `--export=ALL,MASTER_PORT=29999,WORLD_SIZE=16` and asserts both values reach every node.
- `cargo clippy -p spur-cli --all-targets --locked` and `cargo test -p spur-cli --locked` green.

## Follow-up

`srun` does not implement `--export` filtering at all (always forwards the full environment). Out of scope here; can be addressed separately.

### DIFF
--- a/crates/spur-cli/src/sbatch.rs
+++ b/crates/spur-cli/src/sbatch.rs
@@ -796,6 +796,10 @@ fn default_job_name(job_name: Option<&str>, script: Option<&str>, is_wrap: bool)
 /// contain `=`.
 fn resolve_export_env(spec: &str, source: HashMap<String, String>) -> HashMap<String, String> {
     let tokens: Vec<&str> = spec.split(',').filter(|t| !t.is_empty()).collect();
+    // Fast path for the default: forward the environment as-is, no copy.
+    if tokens.as_slice() == ["ALL"] {
+        return source;
+    }
     let (mut env, rest) = match tokens.first() {
         Some(&"ALL") => (source.clone(), &tokens[1..]),
         Some(&"NONE") => (HashMap::new(), &tokens[1..]),
--- a/crates/spur-cli/src/sbatch.rs
+++ b/crates/spur-cli/src/sbatch.rs
@@ -787,6 +787,35 @@ fn default_job_name(job_name: Option<&str>, script: Option<&str>, is_wrap: bool)
     script.unwrap_or("sbatch").to_string()
 }
 
+/// Resolve `--export` per Slurm semantics against a submission environment.
+///
+/// A leading `ALL` seeds the full environment, `NONE` seeds an empty one, and
+/// any other leading token starts an empty environment. Remaining tokens are
+/// applied on top: `VAR` copies the current value from `source`, `VAR=value`
+/// sets an explicit value (overriding an inherited one). The value may itself
+/// contain `=`.
+fn resolve_export_env(spec: &str, source: HashMap<String, String>) -> HashMap<String, String> {
+    let tokens: Vec<&str> = spec.split(',').filter(|t| !t.is_empty()).collect();
+    let (mut env, rest) = match tokens.first() {
+        Some(&"ALL") => (source.clone(), &tokens[1..]),
+        Some(&"NONE") => (HashMap::new(), &tokens[1..]),
+        _ => (HashMap::new(), &tokens[..]),
+    };
+    for tok in rest {
+        match tok.split_once('=') {
+            Some((k, v)) => {
+                env.insert(k.to_string(), v.to_string());
+            }
+            None => {
+                if let Some(v) = source.get(*tok) {
+                    env.insert(tok.to_string(), v.clone());
+                }
+            }
+        }
+    }
+    env
+}
+
 fn build_sbatch_job_spec(
     mut args: SbatchArgs,
     nodelist: Option<String>,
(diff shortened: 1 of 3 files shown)


crates/spur-cli/src/sbatch.rs
```diff
@@ -787,6 +787,35 @@ fn default_job_name(job_name: Option<&str>, script: Option<&str>, is_wrap: bool)
...
+/// applied on top: `VAR` copies the current value from `source`, `VAR=value`
+/// sets an explicit value (overriding an inherited one). The value may itself
+/// contain `=`.
+fn resolve_export_env(spec: &str, source: HashMap<String, String>) -> HashMap<String, String> {
+    let tokens: Vec<&str> = spec.split(',').filter(|t| !t.is_empty()).collect();
+    let (mut env, rest) = match tokens.first() {
+        Some(&"ALL") => (source.clone(), &tokens[1..]),
+        Some(&"NONE") => (HashMap::new(), &tokens[1..]),
+        _ => (HashMap::new(), &tokens[..]),
+    };
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed in 18368103. resolve_export_env now returns the source environment directly when the spec is exactly `["ALL"]` (the default), so the common path has no clone.

tests/native_host/e2e/test_multi_node.py
```diff
@@ -93,6 +93,38 @@ def test_distributed_env_vars(self, multi_node_cluster):
...
+        script = cluster.write_file(
+            "export-combined.sh",
+            "#!/bin/bash\n"
+            'echo "MASTER_PORT=${MASTER_PORT}"\n'
+            'echo "WORLD_SIZE=${WORLD_SIZE}"\n'
+            "echo EXPORT_COMBINED_OK\n",
+        )
+        sb = cluster.sbatch([
+            "-J", "export-combined", "-N", "2", "-o", out_path,
+            "--export=ALL,MASTER_PORT=29999,WORLD_SIZE=16", script,
```
sgopinath1: The regression submits `-N 2` with `--export=ALL,MASTER_PORT=29999,WORLD_SIZE=16`, then asserts those exact values appear on every node. On native agents those two names are not user-owned: `spurd` extends `spec.environment` first, then overwrites them for any multi-node job (`MASTER_PORT=29500`, `WORLD_SIZE=<peer count>` in `crates/spurd/src/agent_server.rs`). Neighboring `test_distributed_env_vars` already asserts `WORLD_SIZE=2` and `MASTER_PORT=29500` for the same `-N 2` path.

This test will fail (or pass for the wrong reason if a node never gets the agent inject), and it does not prove the `--export` parser. Use names the agent does not inject, e.g. `EXPORT_PORT=29999` and `EXPORT_WORLD=16`, and keep `-N 2` plus `count(...) >= 2`.

shiv-tyagi: Fixed this

sgopinath1 (approved):
LGTM

PR #796, opened 2026-09-02
test(e2e): pull e2e postgres image from mirror.gcr.io

shiv-tyagi:
## Summary

- The native_host e2e cluster helper pulls `postgres:16-alpine` directly from DockerHub on the test node.
- Anonymous DockerHub pulls are rate-limited, which can intermittently fail e2e runs.
- Switch to `mirror.gcr.io/library/postgres:16-alpine`, Google's public mirror of the DockerHub image, to avoid the limit.

## Test plan

- [ ] Run the native_host e2e suite and confirm the postgres container starts via the mirror image.


sgopinath1 (approved):
LGTM

PR #797, opened 2026-09-02
perf(spurctld): aggregate assoc_mgr scope usage in a single pass

nikhilsk:
## What this changes

`scontrol show assoc_mgr` recomputed each scope's usage statistics by walking the
job table once per statistic, per user. Cost grew as scopes x users x jobs, so on
a cluster with many users under one QOS an operator-facing read did far more work
than the data required. This aggregates each scope in a single pass over its
in-scope jobs, bucketing per user as it goes, so cost is scopes x jobs.

## Approach

The subtlety is that the Node TRES dimension counts *distinct occupied nodes*, not
the sum of per-job node requests, and it falls back to the job's requested count
when placement is missing. Summing that per statistic is easy to get wrong once the
walks are merged, so the node-union logic is extracted into a
`RunningTresAccumulator` that owns the distinct-node semantics and the
placement-missing fallback. `sum_running_tres` on the admission path now uses the
same accumulator, so the operator view and the scheduler cannot drift apart.

Each user gets their own accumulator, so two of a user's jobs sharing a node still
count as one node for that user while the group figure unions across everyone.

## Testing

`cargo clippy --workspace --exclude spur-ffi --all-targets --locked` is
warning-free and `cargo test --locked` passes.

Added `scope_usage_single_pass_matches_per_statistic_computation`, which builds a
fixture with several users, two jobs sharing a node, an unplaced job, a
pending-only user, and a job in an excluded scope, then compares the new
implementation against the old per-statistic walks kept in the test as a reference.
Literal assertions pin the distinct-node semantics independently of that reference,
so the test still fails if both implementations drift the same way.

### Live differential check

Because this is meant to be behaviour-preserving, it was checked as a differential
against `main` on a live cluster rather than by reading the diff. Both controllers ran
in turn against the **same** PostgreSQL accounting database and the same Raft state
directory, and both were queried with the **same** `main`-built CLI, so the only
variable is the controller:

```
$ /tmp/ref-main/spurctld --version
spur 0.10.0 (c6d68944)          # upstream/main
$ /tmp/target-797/debug/spurctld --version
spur 0.10.0 (889e2431)          # this branch

$ md5sum assoc_mgr-MAIN.txt assoc_mgr-797.txt
d8990b084a38d7430ceee55886808cb3  assoc_mgr-MAIN.txt
d8990b084a38d7430ceee55886808cb3  assoc_mgr-797.txt
```

`scontrol show assoc_mgr` is byte-identical across both builds, over a fixture with
four QOS records (including one carrying a `GrpWall` cap already in breach) and two
association records with per-job TRES caps.

One note on method, since it nearly produced a false result: an initial comparison
appeared to show `main` emitting no `Association Records` at all. That was an artefact
of querying 14s after start-up, before the association cache had loaded. Once
`association cache initialized ... memberships=2 limits=2` appears in the controller
log, the two outputs agree exactly. The comparison above is from the warm state.

### What is not measured here

The complexity claim itself is **not** backed by a live benchmark. The win scales with
`users x jobs` within a scope, and the lab cluster used here has a single user and an
effectively empty jobs map, so any timing from it would be noise rather than evidence.
The argument for the improvement is the structural one in the commit message plus the
unit tests; treat the speedup as reasoned, not demonstrated. A multi-node cluster with
many users and a deep queue is the right place to measure it, and that measurement is
worth doing before this is relied on for a hot path.

### DIFF
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -17628,11 +17628,18 @@ mod tests {
         assert_eq!(user("carol").submitted_jobs, 1);
         assert_eq!(got.grp_running_jobs, 3);
         assert_eq!(got.grp_submitted_jobs, 4);
-        assert!(
-            !got.users
+        assert_eq!(
+            got.users
                 .iter()
-                .any(|u| u.running_tres.get(TresType::Node) > 3),
-            "no user in the excluded 'high' scope should appear"
+                .map(|u| u.user.as_str())
+                .collect::<Vec<_>>(),
+            ["alice", "bob", "carol"],
+            "only users holding work in this scope, in sorted order"
+        );
+        assert_eq!(
+            user("alice").running_jobs,
+            2,
+            "alice's third running job is in the 'high' scope and must not be counted"
         );
     }
 
--- a/crates/spurctld/src/cluster.rs
+++ b/crates/spurctld/src/cluster.rs
@@ -1,7 +1,7 @@
 // Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::collections::{BTreeSet, HashMap, HashSet};
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


shiv-tyagi (approved):

shiv-tyagi:
image

PR #798, opened 2026-09-02
feat(spurctld): surface per-account submit, group wall, and per-job TRES caps in assoc_mgr

nikhilsk:
## What this adds

Three caps Spur already enforces at admission had no way to be read back, so an
operator could see a job rejected without seeing the limit that rejected it:

* `MaxSubmitJobsPA` — the per-account submit cap, enforced in
  `check_qos_submit_limits`. QOS only; an association has no equivalent.
* `GrpWall` — the group wall-clock budget, enforced in
  `check_qos_limits_with_grp_node_charge`, printed as `budget(spent)` with the
  spend read from the same `GrpWallCache` the scheduler reads. QOS only.
* `MaxTRESPJ` — the per-job TRES cap, enforced in `qos_resource_breach` and in
  `account_resource_breach`, so both record types carry it.

## Approach

The display follows enforcement rather than the field name. `GrpWall` counts as
exceeded at `>=` the budget, matching the gate, which blocks once spend *reaches*
it. An unread spend (the cache holds no snapshot yet) renders `N` rather than `0`,
so "unknown" cannot be misread as "nothing spent" — the same distinction the
scheduler makes when it leaves an unreadable budget unapplied.

The per-job cap is labelled `MaxTRESPJ`, not `MaxTRES`: an association names its
own per-user cap plainly `MaxTRES` on each `User=` line, so the shorter name would
have put two different caps under one label inside a single record. `MaxTRESPJ` is
also what `sacctmgr` calls it.

## Compatibility

The proto change is additive only — tags 11 through 14 on `AssocMgrRecord`, whose
highest existing tag was 10, with nothing renumbered, reused, or retyped.
`ScopeLimitUsage` and the `Cap` enum are transient view types that derive no
`Serialize`, so they cannot reach the Raft log or a snapshot, and the CLI output
only gains fields. Nothing here breaks a running cluster on upgrade.

## Testing

Verified on a live single-node cluster (`spurctld` + `spurd` built from this branch,
accounting against a real PostgreSQL 16, `auth.plugin = "none"`), by setting each cap
through `sacctmgr` and reading it back through `scontrol show assoc_mgr`.

Setting the three caps this PR surfaces:

```
$ spur sacctmgr -i add qos name=verifycaps
 QOS added.
$ spur sacctmgr -i modify qos name=verifycaps set maxtresperjob=cpu=4
 Modified QOS 'verifycaps'.
$ spur sacctmgr -i modify qos name=verifycaps set maxsubmitjobsperaccount=7
 Modified QOS 'verifycaps'.
$ spur sacctmgr -i modify qos name=verifycaps set grpwall=90
 Modified QOS 'verifycaps'.
```

They read back with real values rather than the `N` they showed before, and the 90 is
rendered as a wall-clock duration:

```
QOS=verifycaps MaxWall=N MaxTRESPJ=cpu=4 MaxJobsPU=N MaxSubmitJobsPU=N MaxTRESPU=N MaxSubmitJobsPA=7
   GrpJobs=N(0) GrpSubmitJobs=N(0) GrpTRES= GrpWall=01:30:00(00:00:00)
```

A pre-existing QOS on the same cluster exercises the consumption side and the breach
flag, which is the `>=` boundary the unit tests also cover:

```
QOS=yanbudget MaxWall=N ... MaxSubmitJobsPA=N
   GrpJobs=N(0) GrpSubmitJobs=N(0) GrpTRES= GrpWall=00:01:00(00:01:00) OverLimit=GrpWall
```

The association path, with a per-job cap set on two separate associations:

```
Association Records
Account=refreshtest MaxWall=N MaxTRESPJ=cpu=6
   GrpJobs=N(0) GrpSubmitJobs=N(0) GrpTRES=
   User=nikhil MaxJobs=N(0) MaxSubmitJobs=N(0) MaxTRES=

Account=verifyacct MaxWall=N MaxTRESPJ=cpu=2
   GrpJobs=N(0) GrpSubmitJobs=N(0) GrpTRES=
   User=nikhil MaxJobs=N(0) MaxSubmitJobs=N(0) MaxTRES=
```

Both match the rows actually in the database (`cpu=2` for `verifyacct`, `cpu=6` for
`refreshtest`), confirming the values are read rather than defaulted.

Two gaps worth stating rather than glossing over. The `INFINITE`-when-no-snapshot
reading of `grp_wall_consumed_minutes` was not reproduced live — every QOS observed
had a populated `GrpWallCache`, so only the unit test covers the cold-cache path. And
the association records show the QOS-only caps as absent rather than as a visible
`INFINITE` sentinel, so that rendering is covered by unit tests only.

Alongside that, `cargo clippy --workspace --exclude spur-ffi --all-targets --locked`
is warning-free and `cargo test --locked` passes. New tests cover the gate boundary
(600 against a 600 budget is exceeded, 599 is not), an unknown spend never counting as
a breach, the controller populating all three caps from a real `QosCache` and seeded
`GrpWallCache`, the all-unset and cold-cache cases, an association carrying only the
per-job cap, and the rendered output for each shape.

`docs/user-guide/monitoring-jobs.rst` gains the new fields with the `PJ`/`PU`
distinction spelled out.


crates/spurctld/src/cluster.rs
```diff
@@ -4863,6 +4869,7 @@ impl ClusterManager {
                 // rather than inventing a merge across rows that disagree.
                 if let Some((_, limits)) = rows.and_then(|rows| rows.first()) {
                     record.max_wall_minutes = limits.max_wall_minutes;
+                    record.max_tres_per_job = limits.max_tres_per_job.clone();
```
shiv-tyagi, line 4872: Association `MaxTRESPJ` is taken from `rows.first()`, same as `MaxWall`. Enforcement uses the `(user, account)` row. Two users on one account with different per-job caps will show one of them on `Account=` and hide the other.

Put `MaxTRESPJ` on each `User=` line (next to `MaxTRES`), or document that the account line is first-row-only.

crates/spurctld/src/cluster.rs
```diff
@@ -11430,6 +11437,100 @@ mod tests {
         assert!(bob.exceeded_caps().is_empty());
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn assoc_mgr_info_surfaces_the_new_qos_caps_and_grp_wall_spend() {
+        // The three caps the record could not previously explain: the per-job TRES
+        // cap, the per-account submit cap, and the group wall budget with its spend
+        // read from the GrpWall cache. Spend has reached the budget, so the scope
+        // record must also name GrpWall as a live breach.
+        let dir = TempDir::new().unwrap();
```
shiv-tyagi, line 11446: Nit: these comments retell the PR. The assertions already name the caps and the GrpWall breach.
```suggestion
        let dir = TempDir::new().unwrap();
```

proto/slurm.proto
```diff
@@ -811,6 +811,21 @@ message AssocMgrRecord {
...
+  // (`grp_tres`). A QOS and an association both enforce it. Empty when unset,
+  // matching the other TRES fields; `<name>=<value>` otherwise.
+  string max_tres_per_job = 11;
+  // Caps a QOS holds that an association has no equivalent for, so they carry the
+  // INFINITE sentinel in association records. `max_submit_jobs_per_account` bounds
+  // pending+running jobs per account within the QOS; `grp_wall_minutes` is the
+  // group wall-clock budget, with `grp_wall_consumed_minutes` its consumption from
+  // the controller's GrpWall cache. Consumption is INFINITE when that cache holds
+  // no snapshot yet — distinct from a real zero — so a consumer does not read
+  // "unknown" as "none spent".
```
shiv-tyagi, line 825: Nit: six lines covering three fields. Wire shape (empty vs `INFINITE`) is the non-obvious part. The rest restates names the CLI already documents. Keep the empty-vs-INFINITE distinction, drop the rest.

crates/spur-cli/src/scontrol.rs
```diff
@@ -995,8 +1015,11 @@ fn render_assoc_mgr(
...
         };
         out.push_str(&format!(
-            "{}={} MaxWall={}",
-            section.scope, r.scope, max_wall
+            "{}={} MaxWall={} MaxTRESPJ={}",
+            section.scope,
+            r.scope,
+            max_wall,
+            tres_cap_or_n(&r.max_tres_per_job),
         ));
```
shiv-tyagi, line 1023: Nit: `MaxWall` just above and `GrpWall` both map `INFINITE` to `N` and otherwise call `format_time`. Pull that into one helper so the two slots cannot drift.

shiv-tyagi (commented):
Some comments. PTAL

PR #807, opened 2026-09-02
fix(spur-k8s): give SPUR_PEER_NODES resolvable pod names

pre:
Related:
* https://github.com/ROCm/spur/pull/808
* https://github.com/ROCm/spur/pull/809
* https://github.com/ROCm/spur/issues/873 (a Pod name conflict is reported as a successful launch, found during review)

## Motivation

In Pod mode `SPUR_PEER_NODES` is unusable, so no distributed workload can start.

Every Kubernetes node registers the single operator address as its agent address.
The peer list the controller dispatches therefore holds that one address repeated
once per node, and the variable reads
`"<operator-ip>:6818,<operator-ip>:6818,<operator-ip>:6818"`. An MPI or torchrun
launcher that reads it addresses the operator three times and never reaches a
peer. A single node job shows the same value, so it was never usable in Pod mode.

## Technical Details

The Pods of a multi-node job already resolve each other. `launch_job` sets the Pod
hostname to the target node name and the subdomain to the job's headless Service,
so `<node>.<job-service>.<namespace>.svc.cluster.local` is a working name for each
peer.

`SPUR_PEER_NODES` now carries those names, in nodelist order, so index N is the
peer whose `SPUR_NODE_RANK` is N. A small `headless_peer_dns` helper builds them.

A single node job has no headless Service and therefore no name to publish. The
variable is left unset rather than holding an address that resolves to nothing,
so a launcher fails loudly instead of hanging on a dead address.

This also corrects the message the operator returns when no `SpurJob` carries the
requested job id. It said "agent unreachable", although the operator is running and
answering. It now says that Pod mode launches only a `SpurJob`, and that a job from
`sbatch` or `spur submit` has nothing for the operator to launch.

Related:
* https://github.com/ROCm/spur/pull/808
* https://github.com/ROCm/spur/pull/809

## Test Plan

1. Run a three node job in Pod mode and read `SPUR_PEER_NODES` inside every Pod.
2. Resolve each name from inside a Pod and compare against the real Pod IPs.
3. Run a single node job and confirm the variable is absent.
4. Unit tests for the name builder, including the nodelist ordering.

## Test Result

Environment: three node RKE2 cluster on cloud VMs, 8 vCPU and 96 GiB each, no GPU.

- All three Pods of a three node job read the same list of three names, in nodelist
  order. Each name resolved from inside a Pod to the Pod IP of the matching node.
- Before the change the same job read one address three times.
- A single node job left `SPUR_PEER_NODES` unset.
  nothing; `cargo test --locked` passes 3528 tests, including three new ones for
  the name builder.

## Submission Checklist


crates/spur-k8s/src/agent.rs
```diff
@@ -955,6 +965,37 @@ fn sanitize_k8s_name(s: &str) -> String {
...
+        .filter(|n| !n.is_empty())
+        .collect();
+    if nodes.len() < 2 {
+        return Vec::new();
+    }
+    nodes
+        .iter()
+        .map(|n| {
+            format!(
+                "{}.spur-job-{}.{}.svc.cluster.local",
```
shiv-tyagi, line 990: The `spur-job-<id>` name is spelled out in 7 spots now (176, 452, 578, 622, 716, 860, and here). The DNS built here only resolves as long as this matches the Service name at 860 and the subdomain at 452. That is easy to break later.

Pull the service name into one helper and build off it:

```suggestion
                "{}.{}.{}.svc.cluster.local",
                sanitize_k8s_name(n),
                job_service_name(job_id),
                namespace
```

Where `job_service_name(job_id)` returns `format!("spur-job-{job_id}")`, reused at 452, 578, 860 too.

crates/spur-k8s/src/agent.rs
```diff
@@ -1283,3 +1324,33 @@ mod tests {
...
+    fn single_node_job_gets_nothing_because_it_has_no_headless_service() {
+        assert!(headless_peer_dns("node-a", 7, "spur").is_empty());
+        assert!(headless_peer_dns("", 7, "spur").is_empty());
+    }
+
+    #[test]
+    fn node_names_are_sanitized_the_same_way_as_the_pod_hostname() {
+        let out = headless_peer_dns("Node_A.example,node-b", 3, "ns");
+        assert_eq!(out[0], "node-a-example.spur-job-3.ns.svc.cluster.local");
+    }
```
shiv-tyagi, line 1355: None of these cover a trailing comma or padding. That path matters. `"node-a,"` splits into two parts and the empty one only gets dropped by the filter. Without it you would emit a junk FQDN and still pass the len check.

Add a case that locks it in:

```rust
#[test]
fn segments_are_trimmed_and_empty_ones_dropped() {
    let out = headless_peer_dns(" node-a , node-b, ", 7, "spur");
    assert_eq!(
        out,
        vec![
            "node-a.spur-job-7.spur.svc.cluster.local".to_string(),
            "node-b.spur-job-7.spur.svc.cluster.local".to_string(),
        ]
    );
}
```

crates/spur-k8s/src/agent.rs
```diff
@@ -955,6 +965,37 @@ fn sanitize_k8s_name(s: &str) -> String {
...
+        .collect();
+    if nodes.len() < 2 {
+        return Vec::new();
+    }
+    nodes
+        .iter()
+        .map(|n| {
+            format!(
+                "{}.spur-job-{}.{}.svc.cluster.local",
+                sanitize_k8s_name(n),
```
shiv-tyagi, line 991: Q: two node names that sanitize to the same string (say `node_a` and `node-a`) now collide here and on the pod name at 180. The pod one comes back as a 409 which launch_job counts as success, so a pod just never shows up and the peer DNS points at nothing. This PR leans on the sanitize for DNS now, so the collision surface is wider. Is that worth guarding, or are node names already constrained upstream?

shiv-tyagi (commented):
The DNS approach is right and the tests are solid. A few small things: one duplication nit around the service name, one missing test for the trim path, and a question about sanitize collisions.

crates/spur-k8s/src/agent.rs
pre: Done in 3bc26d5: `job_service_name` builds the Service name, and the Pod subdomain, the Service create and delete, and the peer DNS all use it. Pod names stay literal, they are not the Service name.

crates/spur-k8s/src/agent.rs
pre: Added as `segments_are_trimmed_and_empty_ones_dropped` in da90773.

crates/spur-k8s/src/agent.rs
pre: Not constrained upstream, `register_agent` accepts any hostname. K8s node names are DNS-1123 already, so only a dotted and a dashed variant (`a.b`, `a-b`) can collide. Since this PR widens the surface, 6accf17 refuses a nodelist in which two names sanitize alike, before any Pod is created, with an `invalid_argument` that names both. The 409-as-success path is a separate bug, filed as #873.

shiv-tyagi:

shiv-tyagi (approved):
Thanks for addressing my comments. @sgopinath1 @sajmera-pensando please merge if all looks good to you

PR #808, opened 2026-09-02
fix(k8s): keep the cancel of a deleted SpurJob

pre:
Related:
* https://github.com/ROCm/spur/pull/807
* https://github.com/ROCm/spur/pull/809

## Motivation

When a `SpurJob` is deleted, `handle_deletion` cancels the Spur job and then
removes the finalizer whatever the cancel returned. The result of `cancel_job`
is discarded. A cancel that fails at that moment, because the controller replica
that answered is dead or a connection error is transient, is therefore silent
and unrecoverable: the finalizer is gone, the operator never sees the object
again, and the job stays in the controller queue with nothing pointing at it.

With a healthy controller the cancel succeeds and nothing is lost. The failure
was seen on a cluster where a controller replica served with a dead Raft core;
that root cause is fixed in #810. This PR closes the path that turned it into
an orphaned job.

## Technical Details

The operator now inspects the result of the cancel. A failure keeps the
finalizer and returns `ReconcileError::Grpc`, so the delete is retried on the
next reconcile until the cancel is confirmed.

Two answers mean that nothing is left to cancel and must not block the delete:
the controller has no such job, or the job is already terminal. The controller
could not express either, because the `cancel_job` RPC mapped every error to
`Internal`. `ClusterManager::cancel_job` now returns a typed `CancelError`, and
the RPC maps it to a precise gRPC code:

- `NotFound` when the controller has no such job,
- `FailedPrecondition` when the job is already terminal,
- `PermissionDenied` when the caller does not own the job,
- `Internal` when the Raft proposal fails.

The operator treats `NotFound` and `FailedPrecondition` as settled
(`cancel_is_settled`) and lets the finalizer go.

The CLI, REST API, and FFI keep their behaviour. The error messages are
byte-identical, `scancel` prints only the message, the REST handler wraps any
error the same way as before, and the FFI only checks for success. `scancel` on
a finished job still reports an error, as Slurm does.

## Test Plan

1. Unit tests for the `CancelError` variants and their messages, the gRPC code
   of each variant, the `cancel_job` RPC end to end (unknown, foreign, and
   terminal job), and the operator deletion path against a stub controller for
   every outcome.
2. Cluster runs on a three-node Kaytoo cluster with k0s from `spur k8s up`:
   `main`, the first head of this branch, and the current head, with a
   Completed, a Cancelled, a Running, a Pending, and a forgotten job, and a
   delete inside the status-lag window. Results are in the PR comments.

## Test Result

  nothing; `cargo test --locked -p spurctld -p spur-k8s` passes.
- Cluster results: see the comments.

## Submission Checklist


crates/spur-k8s/src/job_controller.rs
```diff
@@ -347,13 +347,30 @@ async fn handle_deletion(job: &SpurJob, ctx: &JobControllerCtx) -> Result<Action
...
+                if e.code() != tonic::Code::NotFound {
+                    warn!(
+                        spurjob = %name, job_id, error = %e,
+                        "failed to cancel the Spur job; retrying before the SpurJob is removed"
+                    );
+                    return Err(ReconcileError::Other(format!(
+                        "cancel of Spur job {job_id} failed: {e}"
+                    )));
+                }
+            }
```
shiv-tyagi, line 373: What does `cancel_job` return when the job is already terminal? Say someone cancels it, or it finishes, right before this delete. `status.state` is still cached as `Pending`, so we hit this branch. `cluster.cancel_job` bails with `"job N is already Cancelled"`. Does that reach us as `NotFound`?

Same for a job the controller has actually dropped. Does `"job N not found"` come back as `NotFound`, or as `Internal`?

If it's `Internal`, this returns `Err`, the finalizer stays, and `kubectl delete` hangs until someone clears it by hand. Did you test the failing cancel, or only the one that succeeds first try?

crates/spur-k8s/src/job_controller.rs
```diff
@@ -347,13 +347,30 @@ async fn handle_deletion(job: &SpurJob, ctx: &JobControllerCtx) -> Result<Action
...
             let mut ctrl = ctx.ctrl_client.lock().await;
-            let _ = ctrl
+            let result = ctrl
                 .cancel_job(CancelJobRequest {
                     job_id,
                     signal: 0,
                     user: String::new(),
                 })
                 .await;
+            drop(ctrl);
```
shiv-tyagi, line 357: Is `NotFound` the only case worth passing here? Re-cancelling a terminal job already gets you what you want. It's not running, not queued. Erroring and retrying forever seems wrong. Would it be cleaner to make the terminal case idempotent on the controller side?

shiv-tyagi (commented):
Ran this on a 4-node cluster. Happy path works: unschedulable job, delete while Pending, gone.

One thing I can't square. The fix relies on `cancel_job` returning `NotFound`. Where does that come from? On the server side, `cancel_job` runs everything through `cluster_err_to_status`. That only handles `AuthError` and maps the rest to `Internal`. So when would we ever actually see `NotFound` here?

crates/spur-k8s/src/job_controller.rs
pre: You are right, both cases came back as `Internal`, so the finalizer stayed. eb0ddc2 gives `cluster.cancel_job` a typed `CancelError`: unknown job -> `NotFound`, terminal job -> `FailedPrecondition`, owner mismatch -> `PermissionDenied`. b20afa7 makes the operator treat the first two as settled. `cancel_job_rpc_maps_errors` pins the codes end to end.

crates/spur-k8s/src/job_controller.rs
pre: Kept the controller side as is, on purpose. Slurm returns `ESLURM_ALREADY_DONE` for `scancel` on a finished job, and `scancel`, `slurm_kill_job` in the FFI and the REST route all follow that today. An idempotent `Ok` would silently change all three. The one caller that wants idempotence is the operator, so it lives there: `cancel_is_settled` accepts `NotFound` and `FailedPrecondition`. A before/after run on a three-node cluster follows as a PR comment.

shiv-tyagi:

shiv-tyagi:

shiv-tyagi (approved):
Thanks for addressing my comments since you said you have tested it thoroughly.

crates/spur-k8s/src/job_controller.rs
shiv-tyagi: okay

PR #809, opened 2026-09-02
fix(k8s): recover a node the controller has forgotten

pre:
Related:
* https://github.com/ROCm/spur/pull/807
* https://github.com/ROCm/spur/pull/808

## Motivation

Two faults that both come from a controller answering `NotFound`.

**A node the controller has forgotten never comes back.** The operator registers
each Kubernetes node once, at startup, then only sends heartbeats. If the
controller loses its record of a node, through `spur node remove` or through a
rebuild from an older state, every later heartbeat fails with `NotFound` and the
operator merely logs it. The node stays missing from `sinfo` until somebody
restarts the operator, and the cluster quietly runs at reduced capacity.

**A refused launch is reported as an unreachable agent.** When an agent answers
`NotFound` or `FailedPrecondition`, the controller maps it to `Unreachable`. The
user sees "agent unreachable" although the agent answered, and the real reason is
lost. That sends anyone debugging it towards the network instead of the job.

## Technical Details

The operator keeps the `RegisterAgentRequest` it built at startup. When a heartbeat
comes back `NotFound`, it registers again from that stored request and logs that it
did so. Nothing else changes: any other error keeps the existing behaviour.

On the controller side, `tonic::Code::NotFound` and `tonic::Code::FailedPrecondition`
now map to `DispatchError::AgentRejected` and carry the agent's own message, rather
than falling through to `Unreachable`. The two codes mean the agent answered and
declined, which is the definition of a rejection.

The operator's "no `SpurJob` carries this job id" path is also made explicit, so
the rejection that reaches the user says what actually happened instead of
`deadline_exceeded`, which the controller used to read as unreachable.

Related:
* https://github.com/ROCm/spur/pull/807
* https://github.com/ROCm/spur/pull/808

## Test Plan

1. With the operator running, remove a node from the controller with
   `spur node remove`, then wait for one heartbeat. Do not restart the operator.
2. Check `sinfo`.
3. Submit a job through the CLI that the agent will refuse, and read the message
   the user gets and the reason in the controller log.

## Test Result

Environment: three node RKE2 cluster on cloud VMs, 8 vCPU and 96 GiB each, no GPU.

- `spur node remove` took a node away and `sinfo` showed 2. One heartbeat later the
  operator logged that the controller did not know the node and that it registered
  it again, and `sinfo` showed 3. **The operator was not restarted.**
- Before the change the node stayed missing and the operator logged the failed
  heartbeat on every cycle.
- A CLI job the agent refused now reads `1 agent rejected launch`, and the
  controller log holds the agent's reason. Before the change it read
  "agent unreachable".
  nothing; `cargo test --locked` passes 3253 tests.

## Submission Checklist


crates/spur-k8s/src/heartbeat.rs
```diff
@@ -67,6 +67,28 @@ impl HeartbeatManager {
...
+                                            "spurctld did not know this node; registered it again"
+                                        ),
+                                        Err(e) => warn!(
+                                            node = %name, error = %e,
+                                            "failed to register the node again"
+                                        ),
+                                    },
+                                    None => warn!(node = %name, "node no longer tracked"),
+                                }
+                            }
```
shiv-tyagi: This re-registration arm is the core of the fix and has no test. `spurd` already handles the identical NOT_FOUND case as a pure, unit-tested predicate (`should_reregister` in `crates/spurd/src/reporter.rs`, covered by `should_reregister_on_not_found` and `should_not_reregister_on_other_errors`). Mirror that here so the "NotFound re-registers, everything else does not" decision is pinned without standing up a live controller:

```rust
fn should_reregister(status: &tonic::Status) -> bool {
    status.code() == tonic::Code::NotFound
}
```

Then gate the arm on `Err(e) if should_reregister(&e)` and test the predicate directly, matching the `spurd` pattern.

crates/spurctld/src/scheduler_loop.rs
```diff
@@ -1224,6 +1224,11 @@ async fn dispatch_to_agent(
             tonic::Code::Unavailable | tonic::Code::DeadlineExceeded => {
                 DispatchError::Unreachable(s.into())
             }
+            // The agent answered and refused. Reporting this as unreachable sent
+            // the operator looking for a network fault that does not exist.
+            tonic::Code::NotFound | tonic::Code::FailedPrecondition => {
+                DispatchError::AgentRejected(s.message().to_string())
+            }
```
shiv-tyagi, line 1231: This NotFound/FailedPrecondition to `AgentRejected` mapping isn't covered, and the harness to test it already lives in this file. `dispatch_trigger_tests` has `MockAgent { reject_resources }` and `resources_unavailable_reject_cools_down_the_node` exercising the `ResourceExhausted` to `ResourcesUnavailable` arm. Add a sibling: a mock agent that returns `Status::not_found`, asserting the job is classified `AgentRejected` and, unlike `Unreachable`/`ResourcesUnavailable`, the node is **not** cooled down. That no-cooldown behavior is the contract that makes this change worth it, and nothing pins it today.

crates/spur-k8s/src/agent.rs
```diff
@@ -115,9 +115,17 @@ impl VirtualAgent {
...
             Ok(Err(status)) => return Err(status),
+            // The retry above swallows NotFound while it waits for a label that
+            // may still be propagating. When the budget runs out the answer is
+            // that no SpurJob exists, which is an explicit rejection and not a
+            // transport failure: deadline_exceeded here made the controller
+            // report "agent unreachable" for a running, reachable operator.
             Err(_elapsed) => {
-                return Err(Status::deadline_exceeded(format!(
-                    "namespace lookup for spur.amd.com/job-id={job_id} timed out after {}s",
+                return Err(Status::not_found(format!(
```
shiv-tyagi, line 124: Nit: the `not_found` budget-exhausted path has no test, so the code and the operator-facing message are unverified. Lower priority than the other two: there's no `kube::Client` mock precedent in this crate, so covering it means introducing one. Fine as a follow-up, flagging it so leaving it uncovered is a conscious call.

shiv-tyagi (commented):
Some comments. PTAL.

crates/spur-k8s/src/heartbeat.rs
pre: Done in f906456: `should_reregister` predicate with `should_reregister_on_not_found` and `should_not_reregister_on_other_errors`, mirroring spurd.

crates/spurctld/src/scheduler_loop.rs
pre: Done in f906456: `not_found_reject_is_agent_rejected_and_does_not_cool_down_the_node` and the `FailedPrecondition` sibling, with a mock agent that rejects with a given status.

crates/spur-k8s/src/agent.rs
pre: Covered in 88827fc with a fake API server (`test_support.rs`): `no_spurjob_within_the_budget_is_a_not_found_rejection`, plus three more lookup cases.

crates/spur-k8s/src/heartbeat.rs
shiv-tyagi: okay thanks

shiv-tyagi (dismissed):
Thanks for addressing my comments and testing Petrus. 

@sgopinath1 @sajmera-pensando Please merge if this looks okay to you.

sgopinath1 (approved):
LGTM

PR #814, opened 2026-09-03
fix(spurd): kernel-enforced GPU device isolation across every launch path

biluriuday:
## What this fixes

A job could open any GPU device node on its node, including one allocated to
another user's job. The existing `*_VISIBLE_DEVICES` deny was advisory — a job
could re-export the variable and regain access.

Closing it took two things: the kernel has to do the denying, and every process
a job runs has to be subject to it. Only the batch payload was in the job's
cgroup at all — `srun` steps, `spur exec`, and interactive attach ran in
spurd's own cgroup, so a filter attached to the job would have missed them.

## Approach

`spurd` attaches a default-deny `BPF_PROG_TYPE_CGROUP_DEVICE` program to each
job's cgroup. The instructions are generated in-process and loaded through the
raw `bpf()` syscall — the technique systemd uses — so there is no eBPF build
step and no nightly toolchain. The allow-list is the job's allocated device
paths, plus standard pseudo-devices, plus host-infrastructure nodes that are
never per-job allocatable.

Containment was then extended so the filter actually covers everything.
Interactive allocations create the cgroup at `RegisterJobAllocation`, and every
process the agent starts joins it from `pre_exec` while still privileged, since
an unprivileged process cannot write another cgroup's `cgroup.procs`. That
includes both spawn shapes of `spur exec`, one of which had no `pre_exec` hook
at all because it drops privilege inside the namespace via `setpriv`.

Container jobs come along for free — they already live in the same `job_N`
cgroup, so there is no container-specific device code.

## Design choices

- **Per-job program load, no BPF map.** Matches Slurm's `cgroup/v2` plugin. A
  map-based design would trade one `BPF_PROG_LOAD` for `MAP_CREATE` plus
  `MAP_UPDATE` and a hand-encoded lookup: more `unsafe` for no measured win on
  a path dwarfed by fork/exec.
- **No detach call.** `BPF_PROG_ATTACH` takes its own kernel reference, so
  removing the cgroup at teardown frees the program. Both descriptors drop
  before install returns; holding one would pin a program per job for spurd's
  lifetime.
- **Access is a subset test** (`requested & !granted == 0`), with the complement
  spanning the full 16-bit access field so an access bit a future kernel adds
  fails closed rather than being ignored.
- **Per-GPU compute nodes are deliberately absent from the host-infrastructure
  list** — gating those on the allocation is the security property. What is on
  it (InfiniBand verbs, NVIDIA control/UVM/MIG, `/dev/fuse`) is shared by every
  workload and owned by no allocation, so denying it breaks jobs that were never
  reaching for a GPU.
- **Steps share the job cgroup rather than getting nested ones.** Correct for
  device isolation and much simpler; the cost is no per-step limits or
  accounting.

## Behavior changes

[1] `constrain_devices` defaults on, so a job can no longer open a device node
its allocation does not cover. A site needing an extra node can list it in
`[cgroup] extra_device_paths` instead of disabling the filter.

[2] A step now counts against the job's `memory.max` and cpuset, where it
previously ran unbounded. A site whose `srun` steps routinely overrun what the
job asked for will start seeing OOM kills.

[3] `[cgroup] required` now also gates allocation registration, not just launch.

[4] `spurd` raises its own `RLIMIT_MEMLOCK` at startup, since kernels before
5.11 charge BPF program memory against it. A job configured
`rlimits.memlock = "inherit"` therefore inherits the raised value; the
`"unlimited"` default is unaffected.

## Testing

Unit tests execute the real generated instruction stream through a minimal BPF
interpreter rather than asserting structure, because the kernel verifier is not
reachable from CI. That caught a jump-overshoot bug where a mismatched rule
skipped the next rule's device-type check — a silent isolation hole that
structural assertions would have passed.

Validated on a bare-metal node using synthetic character devices in the local
major range declared as GRES, which separates the filter from a missing driver:
an allowed open fails `ENXIO`, a filtered one fails `EPERM`. A zero-GPU job was
denied both devices while `/dev/urandom` still worked; a one-GPU job opened its
own and was denied the other (same major, different minor); re-exporting
`ROCR_VISIBLE_DEVICES` did not restore access; the program was freed when the
cgroup went away, with no leak across twelve back-to-back jobs; and
`required = true` refused a launch when the filter could not be installed.

Workspace suite is at 3558 tests. The e2e suite adds 11 device-isolation cases,
which skip without GPU hardware.

## Follow-ups

- Nested per-step cgroups, for per-step limits and accounting. A filter attached
  at `job_N` is inherited by descendants, so it keeps working unchanged.
- seccomp and Landlock still reach only the batch path.
- `cleanup_cgroup` calls `remove_dir` straight after SIGKILL, which returns
  `EBUSY` until processes are reaped, so cancelling an allocation with live
  steps can strand the directory and its program.
- `required` is not verified on `spur exec` and attach; making the join fatal in
  the child would close that race-free.

### DIFF
--- a/crates/spurd/src/agent_server.rs
+++ b/crates/spurd/src/agent_server.rs
@@ -149,7 +149,10 @@ async fn teardown_completed_job(
         return;
     }
 
-    crate::container::cleanup_rootfs(job_id, &completed.rootfs_mode);
+    crate::container::cleanup_rootfs(
+        &crate::container::job_rootfs_base(job_id),
+        &completed.rootfs_mode,
+    );
     crate::executor::cleanup_job_spool(job_id);
     if let Some(ref cgroup) = completed.cgroup {
         crate::executor::cleanup_cgroup(cgroup);
@@ -399,6 +402,23 @@ fn signal_step_tree(pid: u32, signal: i32) {
                     "failed to signal step (it may already have exited)"
                 );
             }
+        }
+    }
+}
+
+fn signal_step_process_group(pid: u32, signal: i32) {
+    let sig =
+        nix::sys::signal::Signal::try_from(signal).unwrap_or(nix::sys::signal::Signal::SIGTERM);
+    let leader = nix::unistd::Pid::from_raw(pid as i32);
+    if let Err(e) = nix::sys::signal::killpg(leader, sig) {
+        if let Err(kill_err) = nix::sys::signal::kill(leader, Some(sig)) {
+            warn!(
+                pid,
+                signal,
+                killpg = %e,
+                kill = %kill_err,
+                "step process group signal failed (step may already have exited)"
+            );
         }
     }
 }
@@ -718,6 +738,8 @@ async fn run_tokio_step_to_spool(
     step_files: crate::executor::StepOutputFiles,
     active_steps: &Arc<Mutex<HashMap<(u32, u32), ActiveStep>>>,
     step_key: (u32, u32),
+    cgroup: Option<&std::path::Path>,
+    cgroup_required: bool,
 ) -> Result<Option<std::process::ExitStatus>, Status> {
     use std::process::Stdio;
     let mut child = cmd
@@ -725,6 +747,16 @@ async fn run_tokio_step_to_spool(
         .stderr(Stdio::from(step_files.stderr))
         .spawn()
         .map_err(|e| Status::internal(format!("step command failed to spawn: {e}")))?;
+
+    // The pre-exec join reports whether it wrote `cgroup.procs`; this verifies the
+    // step actually landed there. Under `[cgroup] required` a step that did not is
+    // refused and its group killed rather than run outside the limits and filter.
+    if escaped_job_cgroup(cgroup_required, cgroup, child.id()) {
+        kill_step_process_group(&mut child).await;
+        return Err(Status::failed_precondition(
+            "[cgroup] required but the step did not join its cgroup",
+        ));
+    }
 
     if let Some(pid) = child.id() {
         let cancel_now = {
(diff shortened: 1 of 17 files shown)


crates/spurd/src/agent_server.rs
```diff
@@ -143,6 +154,93 @@ fn resolve_cgroup_budget(
...
+
+/// Drop a job from the running set and release the cgroup it still owns.
+///
+/// The single exit from `running` bar the monitor loop, which takes the cgroup
+/// itself to read `memory.events` first. Taking it here keeps the release single.
+fn remove_tracked_job(running: &mut HashMap<u32, TrackedJob>, job_id: u32) -> Option<TrackedJob> {
+    let mut tracked = running.remove(&job_id)?;
+    if let Some(cgroup) = tracked.take_cgroup() {
+        executor::cleanup_cgroup(&cgroup);
+    }
```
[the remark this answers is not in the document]
biluriuday: addressed the review comment.

crates/spurd/src/device_cgroup.rs
```diff
@@ -0,0 +1,1166 @@
...
+/// Shared by every workload and owned by no allocation, so nothing puts them in a
+/// job's `device_paths`; denying them breaks jobs not reaching for a GPU at all.
+const HOST_INFRA_DEVICE_NODES: &[&str] = &[
+    "/dev/fuse", // Apptainer/Singularity run inside the batch job
+    // CUDA initialization; enumerating a GPU still needs /dev/nvidia<N>.
+    "/dev/nvidiactl",
+    "/dev/nvidia-uvm",
+    "/dev/nvidia-uvm-tools",
+    "/dev/nvidia-modeset",
+];
```
shiv-tyagi, line 67: Is this right? Should we carry nvidia consts in our codebase?

nvidia injection should come through CDI or gres. We should not need to hardcode like this.

shiv-tyagi (commented):
Tested this PR. Looks mostly fine.  

Just one comment.

crates/spurd/src/device_cgroup.rs
biluriuday: agreed. this was added as a fallback in case user mis-configures. I have updated the docs to specify these explicitly now.

shiv-tyagi (approved):

PR #815, opened 2026-09-03
fix(spur-cli): render squeue %L (TIME_LEFT) instead of ?

shiv-tyagi:
## What

`squeue -o "%L"` (the `TIME_LEFT` column) printed `?` for every job. The header rendered `TIME_LEFT` and `-S L` sorting worked, but the value was never displayed, so a user could sort by remaining time without being able to see it.

## Root cause

`squeue` keys three tables by format letter: the header (`squeue_header`), the sort comparator (`compare_field`), and the value renderer (`resolve_job_field`). `'L'` was present in the first two but missing from `resolve_job_field`, so it fell through to the `_ => "?"` catch-all.

## Fix

Add the `'L'` arm to `resolve_job_field`, reusing the existing `time_left_secs` and `format_duration_hms` helpers. No-limit jobs render `UNLIMITED` (consistent with `%l`); overrun (used > limit) clamps to `0:00`. Sorting is left unchanged.

Note: Slurm distinguishes `NOT_SET` from `UNLIMITED` for `%l`/`%L`, but Spur's model has no such distinction (a missing time limit means no limit), so `%L` follows the existing `%l` behavior.

## Tests

- Render tests for `%L`: remaining duration, `UNLIMITED`, and overrun clamp.
- Regression guard `every_header_letter_resolves`: asserts no `squeue_header` letter falls through to `?`, catching this whole class of header/render drift.

## Docs

- Add `%L` (TIME_LEFT) and `%b` (GRES) to the format-letter table in `monitoring-jobs.rst`.
- Drop the now-stale migration note claiming `%b`/`%L` are unresolved.

## Validation

- `cargo clippy -p spur-cli --all-targets --locked` clean; `cargo test -p spur-cli --locked` green.
- Smoke-tested on the 4-node bare-metal cluster: `%L` renders real values, matches `%l - %M`, and `-S L` / `-S -L` ordering is consistent with the displayed values.

### DIFF
--- a/crates/spur-cli/src/squeue.rs
+++ b/crates/spur-cli/src/squeue.rs
@@ -935,7 +935,7 @@ mod tests {
 
     /// Every documented squeue format letter must resolve to a real value, not
     /// the `?` fallback. Guards against a header/sort entry without a matching
-    /// render arm (the SPUR-245 %L regression).
+    /// render arm (for example, a %L time-left regression).
     #[test]
     fn every_header_letter_resolves() {
         let mut j = job(1, "default", P::JobRunning, 100);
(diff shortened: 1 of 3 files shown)


biluriuday (approved):
LGTM

Issue #781, opened 2026-08-31
srun step output is withheld until exit; blocking steps produce no output at all


sunxxuns:
## Summary

**A blocking step produces no output at all, ever.** Output is withheld until the process exits, so `ray start --block` and every other long-running service step is silent for its entire lifetime.

The same command run over SSH or under `docker run` on that node streams immediately. Only Spur's step path withholds it.

On bare-metal compute nodes, `srun` step execution buffers each node's entire stdout/stderr in memory on that node and returns it only after the process exits. Long-running service steps such as `ray start --block` are therefore silent for their whole lifetime, and step output is bounded by the node daemon's memory and the unary gRPC response limit.

Spur's own standalone `srun` streams output live, so this is a divergence between two paths of the same command rather than a capability Spur lacks. The fix is not client-side, though: `spurd` drains the pipe only at process exit, and `RunStepResponse` is a unary message with no way to carry incremental output, so both have to change before the client can print anything early.

## Code evidence (`b5c6ece`)

The per-node daemon `spurd` (Spur's `slurmd`) launches the process with piped output and waits for completion, in `crates/spurd/src/agent_server.rs::run_command`:

```rust
let mut child = cmd
    .stdout(Stdio::piped())
    .stderr(Stdio::piped())
    .spawn()?;

let output = child.wait_with_output().await?;
```

The full strings are returned in `RunCommandResponse`. `crates/spurctld/src/server.rs::run_step` waits for every node and concatenates all returned strings into a single `RunStepResponse`; `crates/spur-cli/src/srun.rs::dispatch_step` prints them only after that response arrives.

By contrast, the standalone `srun` path calls `try_stream_output` (`crates/spur-cli/src/srun.rs:930`) and streams live. Step mode — the path taken whenever `srun` runs inside an existing allocation via inherited `SPUR_JOB_ID`/`SLURM_JOB_ID` — never calls it.

Because `RunStepResponse` is a unary message subject to the configured max gRPC message size, a sufficiently verbose step can also fail at completion even though the process itself succeeded.

## This is not application-side stdio buffering

Slurm's `srun` forwards task stdout/stderr from `slurmstepd` over its I/O connection while the step runs, so output appears as the application flushes it. The familiar delay there is libc block buffering inside the application when stdout is not a tty, which `srun -u/--unbuffered`, `stdbuf`, or `python -u` resolves.

[...]

Issue #780, opened 2026-08-31
`srun --pty` is silently ignored inside allocations


sunxxuns:
## Summary

**`--pty` is silently discarded in step mode.** `dispatch_step` hardcodes `pty: false` and never reads `args.pty`, so the flag is accepted and the user gets a non-interactive buffered step instead of an error.

On bare metal you keep a shell on the node while the run is live. Inside a Spur allocation there is no equivalent, so inspecting a rank means giving up the allocation holding the state you wanted to inspect.

Inside an existing `salloc` or `sbatch` allocation, `srun --pty ...` does not create an interactive PTY. The step-dispatch path hardcodes `pty: false`, so the command follows the buffered `RunStep`/`RunCommand` path instead.

This prevents the standard interactive development workflow of retaining a multi-node allocation and opening a debugger or shell on a selected compute node.

## Code evidence (`b5c6ece`)

`crates/spur-cli/src/srun.rs::dispatch_step` creates the step with:

```rust
.create_job_step(CreateJobStepRequest {
    ...
    overlap: false,
    pty: false,
    winsize: None,
    node: String::new(),
    ...
})
```

The function does not inspect `args.pty`. It then calls synchronous `RunStep` and emits output only after the RPC returns.

The separate standalone/`--jobid --overlap` interactive path uses `run_interactive_pty`, but ordinary step mode selected through the inherited `SPUR_JOB_ID`/`SLURM_JOB_ID` does not.

## Reproduction

```bash
salloc -N2 --gpus-per-node=8
srun --pty bash
```

Expected: an interactive raw-mode terminal on an allocated node, with terminal resize and signal forwarding.

Actual: no interactive PTY is created; the command uses the synchronous buffered step path.

A realistic development command is:

```bash
srun --overlap -N1 -w node2 --pty bash
```

which should allow inspecting a specific SGLang/Ray worker while the rest of the allocation remains active.

Node targeting has additional existing gaps tracked by #597, and the current overlap/nsenter GPU-group problem is tracked by #596. This issue is specifically that `--pty` itself is discarded in ordinary step mode.

## Development impact

SGLang/Miles optimization requires keeping Ray, rollout engines, and trainers alive while opening per-node shells for `rocprofv3`, `py-spy`, `amd-smi`, RCCL diagnostics, core inspection, and source edits. Releasing and reallocating nodes for each shell destroys the state being debugged.

## Expected behavior

[...]

PR #818, opened 2026-09-05
feat(spurd): implement containerized srun job steps on bare-metal nodes

### LINKED ISSUE #777: `srun --container-image` is silently ignored on bare-metal job steps
## Summary

**This is a silent no-op, not a missing feature.** `srun` accepts `--container-image`, runs the command somewhere else entirely, and exits 0. Nothing warns the user, so the failure surfaces later as a missing binary or as the host's unpatched Python packages being imported instead of the image's.

Running that same image by hand on the node with `docker run` gives you the image's environment. Asking Spur for it gives you the host's.

On bare-metal compute nodes (the non-Kubernetes mode Spur calls "native host"), `srun --container-image=...` accepts the container options but runs the step directly on the host. The image, mounts, working directory, named-container state, and container environment are lost before they ever reach the compute node: the RPC that the controller (`spurctld`) sends to the per-node daemon (`spurd`, Spur's `slurmd`) to run a step has no fields to carry them.

This blocks the standard Slurm + Pyxis/Enroot pattern used by distributed AI launchers, including Miles' checked-in multi-node launcher, which starts one containerized Ray head/worker step per allocated node with `srun --overlap -w <node> --container-image=...`.

## Code path (`b5c6ece`)

`crates/spur-cli/src/srun.rs` builds a `JobSpec` containing the container fields and sets `srun_job: true`.

For native nodes, `crates/spurctld/src/scheduler_loop.rs` sets `srun_step_dispatch = true`, calls `register_allocation_on_nodes`, and deliberately sets `dispatch_spec` to `None`. The command is subsequently sent through `CreateJobStep` / `RunStep`.

`RunStepRequest` (`proto/slurm.proto:881`) and `RunCommandRequest` (`proto/slurm.proto:1099`) have no container fields at all — no image, mounts, workdir, name, or save/remap options. In `spurd`'s step handler, `crates/spurd/src/agent_server.rs::run_command`, the command is launched as a plain host process:

```rust
let mut cmd = tokio::process::Command::new(&program);
cmd.args(&program_args)
    .current_dir(&work_dir)
    .process_group(0);
```

This path does not call `setup_rootfs`, `container_init`, `pivot_root`, or `build_launch_plan`, and it does not enter an existing job container.

## Reproduction

Use an image containing a marker that does not exist on the host:

```bash
srun --container-image=marker-image.sqsh \
     sh -c 'test -e /inside-image-marker'
```

Expected: success inside the requested image.

Actual on a bare-metal node: the command is executed by the host `sh`; the marker is absent. If the executable exists only in the image, the step fails with `command not found`.

The same applies inside an allocation:

```bash
salloc -N2 --gpus-per-node=8
srun --overlap -N1 -w node1 --container-image=miles.sqsh ray start --head ...
```

`ray` is resolved on the host, not in `miles.sqsh`.

## Workaround, and why it does not close the gap

Multi-node containerized execution does work through the batch path: `sbatch` dispatches `LaunchJob` to every allocated node, and each node's script runs inside the image with `SPUR_NODE_RANK`, `SPUR_NNODES`, `RANK`, `WORLD_SIZE`, and `MASTER_ADDR` set. This is covered by `tests/native_host/e2e/test_container.py::TestContainerMultiNode`. A Ray cluster can therefore be brought up by branching on rank inside one script:

```bash
#!/bin/bash
#SBATCH -N2 --gpus-per-node=8 --container-image=miles.sqsh
if [ "$SPUR_NODE_RANK" = 0 ]; then
  ray start --head --block ...
else
  ray start --address="$MASTER_ADDR:6379" --block ...
fi
```

That workaround requires replacing the launcher of every framework being brought up on Spur, and it does not restore what steps provide:

- no way to restart one node's rollout engine or trainer inside an existing allocation — the unit of restart becomes the whole job;
- no containerized interactive iteration under `salloc`;
- upstream launchers (Miles, and any Pyxis-based launcher) cannot run unmodified;
- per-step images, mounts, and named containers are unavailable, so rollout and training steps cannot use different images in one allocation.

Submitting the parent script with `sbatch --container-image` does not help either: nested `srun` steps still go through `RunCommand` and execute on the host rather than entering the parent container.

## Development impact

Miles' Slurm adapter uses the step pattern directly:

It starts the Ray head and every Ray worker as concurrent containerized `srun --overlap` steps. Under Spur, those steps lose Miles' patched SGLang/Megatron/Ray stack and all requested mounts. The steps do not fail loudly at submission — the container flags are accepted and silently dropped, so the failure surfaces later as a missing binary or an import of the host's unpatched packages.

## Expected behavior

For bare-metal allocations, container options on `srun` should either:

1. create/reuse the requested step container and execute the step inside it, or
2. enter the parent allocation's container when appropriate.

The step must retain normal job-step semantics: selected node/GPU subset, cgroup membership, signal propagation, output handling, and cleanup on allocation termination.

If containerized steps are out of scope for bare-metal nodes, `srun` should reject the container flags there instead of accepting and ignoring them.

## Suggested coverage

Bare-metal E2E tests for: standalone `srun --container-image`; `salloc` followed by a containerized `srun`; `sbatch --container-image` followed by a nested containerized `srun`; simultaneous `srun --overlap` container steps on different nodes. Each should assert an image-only executable runs, requested mounts are present, GPUs are visible, and cancellation cleans up the step container.

## Decision requested

Pick one and say which:

1. implement containerized steps on bare-metal nodes, which requires container fields on `RunStepRequest`/`RunCommandRequest` plus the container path in the step handler; or
2. reject `--container-image` and the related container flags at submission on that path, with an error naming the batch workaround.

Continuing to accept the flags and drop them is the only outcome that leaves users unable to tell whether their job ran in the requested image. Frameworks targeting Spur need to know which of the two to plan against.

sajmera-pensando:
## Summary

Fixes rocm/spur#777 — `srun --container-image` was silently accepted and dropped on bare-metal ("native host") job steps, running the command on the host instead of in the requested image and exiting 0 with no warning.

The root cause was that `RunStepRequest` / `RunCommandRequest` carried no container fields, so `spurd`'s `run_command` always spawned a plain host process. This PR carries the container intent across the RPC boundary and wires the existing container runtime into the step-dispatch path.

## Approach

- **Proto:** add container fields to `RunStepRequest` (tags 11–19) and `RunCommandRequest` (tags 15–23), mirroring the existing `JobSpec` fields 70–78. Append-only — no renumbering or retyping, so wire/FFI/REST compatibility is preserved.
- **CLI (`srun`):** populate the container fields on `RunStepRequest`, and add `--container-name`, `--container-readonly`, `--container-entrypoint` for parity with `sbatch`.
- **Controller (`spurctld`):** `run_step` resolves the effective container config per step (step-level fields take precedence over the parent job's) and forwards it to each agent.
- **Agent (`spurd`):** `run_command` now dispatches across three paths:
  1. **Parent job has live namespaces** → enter them via `nsenter` (`build_launch_plan`). Covers a nested `srun` inside a containerized `sbatch` allocation, sharing the job's image, mounts, and GPU devices.
  2. **Step carries its own image, no parent container** → build a fresh rootfs (`setup_rootfs` + new `run_containerized_step`), matching the batch launch path: namespace unshare, mount setup, GPU device injection, shadow/user, bind mounts, DNS, home, hooks, `pivot_root`, privilege drop; stdout/stderr collected inline. Reached by standalone `srun` and by `salloc` + `srun` (the primary Miles pattern).
  3. **No container** → existing host spawn, unchanged.

## Notable fixes found along the way

Getting the bare-metal E2E suite green surfaced several real defects, all fixed here:

- **GPU device injection for step containers.** The container injection plan (`/dev/kfd`, `/dev/dri/renderD*`, ROCm library bind-mounts, render-group membership) is captured from `build_job_injection_plans` and wired into `ContainerConfig.device_plan` for standalone step containers.
- **GPU visibility deny.** `maybe_deny_gpu_env` is applied to `container_env` so a zero-GPU step cannot re-enable GPU visibility via `--container-env` — matching the batch path.
- **`--container-entrypoint`** was plumbed but ignored by the step path; it now runs before the step command, matching `sbatch`.
- **Cancellation cleanup.** Cancelling an allocation-only job (standalone `srun` / `salloc`) now signals its in-flight steps instead of only dropping the tracked allocation. Additionally, a containerized step's real workload is a grandchild that is PID 1 of its own PID namespace (and ignores SIGTERM from an ancestor), so signalling only the tracked pid left it orphaned; step cancellation now walks the `/proc` process tree (`kill_process_tree`, the same approach the batch container path uses). This applies to both `scancel` and `srun`'s Ctrl-C.
- **Rootless nsenter into a container (pre-existing bug, also affects `spur exec`).** Entering a rootless (user-namespace) container failed with `nsenter: setgroups failed: Operation not permitted`: the kernel forbids `setgroups()` in an unprivileged user namespace, and both `nsenter`'s default credential reset and `setpriv --init-groups` call it. For user-namespace targets, entry now uses `nsenter --preserve-credentials` and skips the `setpriv` group init — the job user is already mapped inside the namespace, so no in-namespace drop is needed, and there are no supplementary groups to preserve there anyway. This is **gated on `has_user_namespace`, which is false for every case where spurd is root**, so the root paths (including GPU-group preservation and its `@rootful` test) emit the identical command as before and are unaffected. This unblocks Case 1 (nested `srun` / `spur exec`) under rootless spurd.

## `--container-readonly` is rejected at submission (follow-up)

`--container-readonly` is a no-op on **every** backend today — `ContainerConfig.readonly` is populated but never consumed (no read-only remount on native host, no `readOnlyRootFilesystem` on k8s). Rather than continue silently accepting it (the exact failure mode #777 is about), both `srun` and `sbatch` now reject it at submission with a clear message.

> **Follow-up:** implement `--container-readonly`. This needs a read-only remount of the merged rootfs in `container_init` (native host) and `readOnlyRootFilesystem` in the pod SecurityContext (k8s); once done, remove the submission-time rejection in `srun`/`sbatch`. This is a pre-existing gap surfaced by #777, not new to this change.

**Behavior-change note:** `sbatch --container-readonly` was previously accepted (as a no-op) and is now rejected. Existing batch scripts passing this flag will error instead of silently ignoring it.

## Known limitations / follow-ups

- **File bind mounts** (`--container-mounts host_file:/dst`) are an untested edge case for containers (the existing batch coverage uses directory mounts). The E2E here mounts directories; a file-target bind hit an `ENOENT` and is left for a separate change (it would affect the batch path too).
- **`--container-mount-home`** end-to-end host-visibility under rootless is smoke-tested (home is present and writable inside the step) rather than asserting a host-written file round-trips, pending a separate investigation.

## Related issues

#777, #780 (`--pty`) and #781 (output streaming) share the same bare-metal step-dispatch root. This PR addresses #777; the other two are separate follow-ups.

## Test plan

- [ ] `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` — clean
- [ ] `cargo test --locked` — passes (spurd unit tests cover: container path taken vs silent host fallthrough, nsenter precedence over new container, rootless nsenter preserves credentials / skips setgroups, host fallthrough, cancellation signals in-flight steps, `--container-readonly` rejection on `srun` + `sbatch`)
- [ ] Bare-metal E2E (`tests/native_host/e2e/test_srun_container.py`): standalone `srun`; `salloc` + `srun`; `sbatch` + nested `srun` (nsenter) and `spur exec` into a running container; concurrent `--overlap` steps; a single multi-node `srun -N2` step; bind mounts / workdir / home / user identity / DNS; GPU device injection for both new-container and nsenter paths (auto-skips without GPUs); `--container-name`, `--container-entrypoint`; cancellation via `scancel` and Ctrl-C; `--container-readonly` rejection

### DIFF
--- a/crates/spurctld/src/server.rs
+++ b/crates/spurctld/src/server.rs
@@ -2768,6 +2768,41 @@ impl SlurmController for ControllerService {
         let label = req.label;
         let job_mpi = job.spec.mpi.as_deref().unwrap_or(spur_core::mpi::MPI_NONE);
         let mpi = spur_core::mpi::resolve_step_mpi(req.mpi.as_str(), job_mpi).to_string();
+        // Step-level container override from srun --container-image; falls back
+        // to the parent job's container config (e.g. sbatch --container-image).
+        let step_container_image = if !req.container_image.is_empty() {
+            req.container_image.clone()
+        } else {
+            job.spec.container_image.clone().unwrap_or_default()
+        };
+        let step_container_mounts = if !req.container_mounts.is_empty() {
+            req.container_mounts.clone()
+        } else {
+            job.spec.container_mounts.clone()
+        };
+        let step_container_workdir = if !req.container_workdir.is_empty() {
+            req.container_workdir.clone()
+        } else {
+            job.spec.container_workdir.clone().unwrap_or_default()
+        };
+        let step_container_name = if !req.container_name.is_empty() {
+            req.container_name.clone()
+        } else {
+            job.spec.container_name.clone().unwrap_or_default()
+        };
+        let step_container_readonly = req.container_readonly || job.spec.container_readonly;
+        let step_container_mount_home = req.container_mount_home || job.spec.container_mount_home;
+        let step_container_env = if !req.container_env.is_empty() {
+            req.container_env.clone()
+        } else {
+            job.spec.container_env.clone()
+        };
+        let step_container_entrypoint = if !req.container_entrypoint.is_empty() {
+            req.container_entrypoint.clone()
+        } else {
+            job.spec.container_entrypoint.clone().unwrap_or_default()
+        };
+        let step_container_remap_root = req.container_remap_root || job.spec.container_remap_root;
         let pmix_tmpdir = self.cluster.config().mpi.pmix_tmpdir.clone();
         let modex_connect_timeout_secs = self.cluster.config().mpi.modex_connect_timeout_secs;
         let modex_fence_timeout_secs = self.cluster.config().mpi.modex_fence_timeout_secs;
@@ -2937,6 +2972,12 @@ impl SlurmController for ControllerService {
             let work_dir = work_dir.clone();
             let environment = environment.clone();
             let step_mpi = mpi.clone();
+            let container_image = step_container_image.clone();
+            let container_mounts = step_container_mounts.clone();
+            let container_workdir = step_container_workdir.clone();
+            let container_name = step_container_name.clone();
+            let container_env = step_container_env.clone();
+            let container_entrypoint = step_container_entrypoint.clone();
             set.spawn(async move {
                 let mut agent = crate::agent_client::connect(agent_addr.clone())
                     .await
@@ -2962,6 +3003,15 @@ impl SlurmController for ControllerService {
                         pmix_plan,
                         mpi: step_mpi.clone(),
                         pmix_prepared: needs_pmix_prepare,
+                        container_image,
+                        container_mounts,
+                        container_workdir,
+                        container_name,
+                        container_readonly: step_container_readonly,
+                        container_mount_home: step_container_mount_home,
+                        container_env,
+                        container_entrypoint,
+                        container_remap_root: step_container_remap_root,
                     })
                     .await
                     .map_err(|e| {
--- a/crates/spurd/src/agent_server.rs
+++ b/crates/spurd/src/agent_server.rs
@@ -223,21 +223,16 @@ fn cancelled_step_response() -> RunCommandResponse {
     }
 }
 
+/// Signal a step's whole process subtree. The tracked pid may be an
+/// intermediate parent (a containerized step runs the workload as a grandchild
+/// that is PID 1 of its own PID namespace), so a single `killpg`/`kill` on it
+/// never reaches the real process. Walking the `/proc` child tree — the same
+/// approach the batch container path uses — reaches the workload directly, so a
+/// plain `SIGTERM` lands on it (a namespace init would otherwise discard it).
 fn signal_step_process_group(pid: u32, signal: i32) {
     let sig =
         nix::sys::signal::Signal::try_from(signal).unwrap_or(nix::sys::signal::Signal::SIGTERM);
-    let leader = nix::unistd::Pid::from_raw(pid as i32);
-    if let Err(e) = nix::sys::signal::killpg(leader, sig) {
-        if let Err(kill_err) = nix::sys::signal::kill(leader, Some(sig)) {
-            warn!(
-                pid,
-                signal,
-                killpg = %e,
-                kill = %kill_err,
-                "step process group signal failed (step may already have exited)"
-            );
-        }
-    }
+    crate::executor::kill_process_tree(pid as i32, sig);
 }
 
 #[cfg(test)]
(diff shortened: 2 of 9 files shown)


crates/spurd/src/agent_server.rs
```diff
@@ -2255,64 +2525,193 @@ impl SlurmAgent for AgentService {
...
-                let _ = child.wait_with_output().await;
-                return Ok(Response::new(cancelled_step_response()));
+
+            let image_path =
+                crate::container::resolve_image(&req.container_image, None, Some(step_uid))
+                    .map_err(|e| Status::failed_precondition(e.to_string()))?;
+
+            // Use a unique rootfs ID per step to avoid collisions with the parent
+            // job's rootfs (which uses `job_id` alone).
+            let step_rootfs_id = job_id * 10000 + step_id;
```
yansun1996, line 2644: This shares an id space with the batch path. `setup_rootfs(.., id, None)` resolves to `container_dir()/job_{id}` and `cleanup_rootfs` deletes that same path — so job 1 / step 0 gives `job_10000`, which is batch job 10000's live rootfs on that node. The step extracts into it, and `cleanup_rootfs` at line 2692 then removes it. Steps also collide with each other once `step_id >= 10000`, and `job_id * 10000` overflows `u32` above ~429k (debug panic inside the handler, silent wrap in release).

Could `setup_rootfs`/`cleanup_rootfs` take an explicit base name (e.g. `step_{job_id}_{step_id}`) so the two namespaces can't overlap?

crates/spurd/src/agent_server.rs
```diff
@@ -312,6 +307,249 @@ async fn step_cancel_requested(
...
+                    child_pid.as_raw(),
+                    nix::sys::signal::Signal::SIGKILL,
+                );
+                let _ = nix::sys::wait::waitpid(child_pid, None);
+                return Err(Status::cancelled("step cancelled"));
+            }
+
+            // Read stdout and stderr concurrently using blocking tasks so we
+            // don't deadlock waiting for one pipe while the child fills the other.
+            let stdout_file = unsafe { std::fs::File::from_raw_fd(stdout_r_fd) };
```
yansun1996, line 507: `nix::unistd::pipe()` returns `OwnedFd`s, and in this arm `stdout_r`/`stderr_r` are still alive — so `File::from_raw_fd` takes a second ownership and the fd is closed twice: once when the reader task finishes, once at scope exit. In a multi-threaded runtime that second `close()` can land on an fd another worker has since opened.

`std::fs::File::from(stdout_r)` moves the `OwnedFd` and avoids the raw-fd copy entirely (same for stderr).

crates/spurd/src/agent_server.rs
```diff
@@ -2255,64 +2525,193 @@ impl SlurmAgent for AgentService {
...
             task_offset = req.task_offset,
             uid = req.uid,
             work_dir = %work_dir,
+            container_image = %req.container_image,
+            parent_namespaces = job_entry.has_namespaces(),
             "RunCommand: executing step"
         );
 
-        use std::process::Stdio;
+        let output = if job_entry.has_namespaces() && job_entry.pid > 0 {
```
yansun1996, line 2541: This branch never consults `req.container_image`, so `sbatch --container-image A` + `srun --container-image B` runs in A with no indication that B was dropped — even though the controller resolved the override at `server.rs:2773`. That's the same accept-and-silently-ignore shape as #777.

Could we either build a fresh container when the step names a different image, or return `failed_precondition` if that's out of scope for this PR? A `warn!` would at least leave a trace — the `info!` above logs the image but not that it was ignored.

crates/spurd/src/agent_server.rs
```diff
@@ -2255,64 +2525,193 @@ impl SlurmAgent for AgentService {
...
+            // "srun inside salloc --container-image" without setting up a new rootfs.
+            let full_command = [vec![program], program_args].concat();
+            let priv_drop = crate::privdrop::PrivDrop::resolve_if_needed(req.uid, req.gid);
+            let plan = build_launch_plan(&job_entry, priv_drop.as_ref(), &full_command);
+            let mut cmd = tokio::process::Command::new(&plan.program);
+            cmd.args(&plan.args).process_group(0);
+            for (k, v) in env {
+                cmd.env(k, v);
+            }
+            if plan.apply_priv_in_child {
```
yansun1996, line 2553: `current_dir(&work_dir)` only runs when privilege is dropped in the child; the nsenter arm skips it, and `nsenter_args()` emits no `--wd`, so the step inherits spurd's cwd. Pre-change, `run_command` set `.current_dir(&work_dir)` unconditionally, so a nested `srun pwd` regresses here.

Setting `cmd.current_dir(&work_dir)` before the branch would restore it. Worth noting `test_nested_srun_enters_parent_container` uses an absolute path, so it wouldn't catch this.

crates/spurd/src/agent_server.rs
```diff
@@ -312,6 +307,249 @@ async fn step_cancel_requested(
...
+                step.cancel_requested
+            } else {
+                false
+            }
+        };
+        if cancel_now {
+            signal_step_process_group(pid, nix::sys::signal::Signal::SIGTERM as i32);
+            let _ = child.kill().await;
+            let _ = child.wait_with_output().await;
+            return Err(Status::cancelled("step cancelled"));
```
yansun1996, line 338: This returns `Err(Status::cancelled(..))` where the pre-change code returned `Ok(cancelled_step_response())` (exit 143 + `"step cancelled"` on stderr). In `spurctld/src/server.rs` an `Err` from a node lands in the `Ok(Err(e))` arm, which pushes `run_command on <node> failed: status: Cancelled...` into the step's stderr, sets the exit code to 1 instead of 143, and fires `cancel_step_on_nodes` at every node of the step. So a plain `scancel` produces a spurious error dump plus a cancel fan-out.

The check at line 2525 still returns `Ok(..)`, so the two paths also disagree. Could both return `Ok(cancelled_step_response())`? Same applies at line 502.

crates/spurd/src/agent_server.rs
```diff
@@ -2255,64 +2525,193 @@ impl SlurmAgent for AgentService {
...
+                container_cfg,
+                rootfs,
+                script_in_container,
+                env,
+                &self.active_steps,
+                step_key,
+                memlock,
+            )
+            .await;
+            crate::container::cleanup_rootfs(step_rootfs_id, &rootfs_mode);
```
yansun1996, line 2692: Nothing guarantees this runs. The `fs::write` at 2669 returns early between `setup_rootfs` and here, and if the tonic future is dropped (srun Ctrl-C, client disconnect, controller RPC timeout) the `.await` never resumes — leaving an extracted/mounted rootfs behind per attempt and the forked child with nobody to signal it.

A small `Drop` guard created right after `setup_rootfs`, disarmed on the normal return, would make this cancel-safe.

crates/spurd/src/agent_server.rs
```diff
@@ -223,21 +223,16 @@ fn cancelled_step_response() -> RunCommandResponse {
...
-            warn!(
-                pid,
-                signal,
-                killpg = %e,
-                kill = %kill_err,
-                "step process group signal failed (step may already have exited)"
-            );
-        }
-    }
+    crate::executor::kill_process_tree(pid as i32, sig);
```
yansun1996, line 235: Two side effects worth weighing here. `kill_process_tree` walks `/proc/<pid>/task/<pid>/children`, so a double-forked descendant reparented away from the step is no longer reached, whereas `killpg` covered it — and Case 3 host steps are still spawned with `.process_group(0)`. The `warn!` on failure is also gone, which was the only signal when a step won't die.

Would it work to keep the `killpg` and *add* the tree walk, rather than swapping one for the other? Also, the function no longer signals a process group — `signal_step_tree` might read truer.

crates/spurd/src/agent_server.rs
```diff
@@ -312,6 +307,249 @@ async fn step_cancel_requested(
...
+
+            let shell = std::ffi::CString::new("/bin/bash").unwrap();
+            // Honor --container-entrypoint, matching the batch path: run the
+            // entrypoint, then the step script.
+            let argv: Vec<std::ffi::CString> = if let Some(ref ep) = entrypoint {
+                let cmd = format!("{ep} && /bin/bash {script_path_clone}");
+                vec![
+                    std::ffi::CString::new("bash").unwrap(),
+                    std::ffi::CString::new("-c").unwrap(),
+                    std::ffi::CString::new(cmd).unwrap_or_default(),
```
yansun1996, line 443: If the entrypoint contains a NUL, `unwrap_or_default()` turns this into `bash -c ""`, which exits 0 — a step that never ran reports success. Building the argv fallibly before `fork()` and failing the RPC would avoid that, and would let the `.unwrap()`s at 435/441/442/447 go too (AGENTS.md discourages those in library code).

Two related gaps in this block: the shell is hardcoded `/bin/bash` where `executor.rs:1720` probes `/bin/bash` then `/bin/sh`, so a busybox/alpine image gets a bare 127; and `let _ = execve(..)` swallows the error where the batch path prints `spur: execve failed: ...`.

crates/spurd/src/agent_server.rs
```diff
@@ -2255,64 +2525,193 @@ impl SlurmAgent for AgentService {
...
+                &image_path,
+                step_rootfs_id,
+                container_cfg.name.as_deref(),
+            )
+            .map_err(|e| Status::internal(format!("step container setup failed: {e}")))?;
+
+            // Write the step command as a script inside the rootfs so it's
+            // accessible after pivot_root hides the host filesystem.
+            let step_script_content = {
+                let joined = shlex::try_join(
```
yansun1996, line 2655: `unwrap_or_else(|_| program.clone())` drops every argument on failure, so a command that isn't shell-safe silently runs as the bare program. The host path handles this in `build_one_shot_command_script` by returning `invalid_argument("command is not shell-safe")` — could this do the same?

crates/spurd/src/agent_server.rs
```diff
@@ -2887,14 +3287,61 @@ impl AgentService {
...
+    ///
+    /// A containerized step runs as PID 1 of its own PID namespace. Signals from
+    /// an ancestor namespace to a namespace's init are *discarded* unless init
+    /// installed a handler for them — SIGTERM/SIGINT to a `bash`/`sleep` init are
+    /// dropped. Only SIGKILL and SIGSTOP are force-delivered. So the requested
+    /// signal is sent first (graceful for host steps), then SIGKILL after a short
+    /// grace period guarantees a container init dies.
+    async fn cancel_active_steps_for_job(&self, job_id: u32, signal: i32) {
+        let keyed: Vec<((u32, u32), u32)> = {
+            let mut steps = self.active_steps.lock().await;
```
yansun1996, line 3305: The `/proc` tree walk inside `signal_step_process_group` runs while holding the `active_steps` lock, so every concurrent `run_command` blocks on it. That matters because `ActiveStepGuard::drop` uses `try_lock` and silently skips removal on contention — this is exactly the window where it loses — leaving a stale entry whose pid the 2s escalation at 3329 may then `SIGKILL` after the host has recycled it.

Snapshotting the pids under the lock and signalling outside it would close most of this. A monotonic `epoch` on `ActiveStep`, compared by the escalation instead of pid equality, would make it robust against recycling.

crates/spurd/src/agent_server.rs
```diff
@@ -4302,6 +4777,144 @@ mod tests {
...
+            work_dir: String::new(),
+            environment: HashMap::new(),
+            job_id,
+            container_image: "/nonexistent/bogus-step-image.sqsh".into(),
+            ..Default::default()
+        });
+        // The nsenter path may or may not succeed depending on the runner's
+        // privileges, but it must never report the bogus image as not found —
+        // that would mean Case 2 was wrongly entered ahead of Case 1.
+        if let Err(err) = svc.run_command(req).await {
```
yansun1996, line 4836: On the happy path this asserts nothing: the test job spawns a real `sleep`, so `pid > 0` and Case 1 runs a real `nsenter` against a process with no namespaces, which exits non-zero and yields `Ok(output)` — the `if let Err` body never executes. It isn't dead (it still fails if Case 2 is wrongly entered), but it would be stronger asserting on the observable branch, e.g. the nsenter output in stderr, rather than the absence of a string.

crates/spurctld/src/server.rs
```diff
@@ -2768,6 +2768,41 @@ impl SlurmController for ControllerService {
...
+            req.container_workdir.clone()
+        } else {
+            job.spec.container_workdir.clone().unwrap_or_default()
+        };
+        let step_container_name = if !req.container_name.is_empty() {
+            req.container_name.clone()
+        } else {
+            job.spec.container_name.clone().unwrap_or_default()
+        };
+        let step_container_readonly = req.container_readonly || job.spec.container_readonly;
```
yansun1996, line 2793: Two semantics here worth a second look.

The booleans are OR'd with the job's, so a step can never opt *out* of a job-inherited `--container-mount-home` or `--container-remap-root` — Slurm/Pyxis treat step flags as an independent override. `Option<bool>` on the proto (absent = inherit, `false` = off) would make that expressible.

And `step_container_env` at 2795 *replaces* the job's map wholesale when non-empty, so adding a single step variable silently drops every job-level container env var. Merging with step-wins is probably what users expect.

docs/user-guide/running-containers.rst
```diff
@@ -59,13 +59,15 @@ runs inside the container.
...
+     - Not yet implemented — rejected at submission. (A read-only container
+       root is not enforced yet; the flag is refused rather than silently
+       ignored.)
    * - ``--container-mount-home``
      - Mount your home directory into the container.
    * - ``--container-env KEY=VAL``
      - Set an environment variable inside the container. Repeatable.
    * - ``--container-entrypoint``
-     - Use the image's entrypoint (``sbatch``).
+     - Use the image's entrypoint.
```
yansun1996, line 70: This doesn't match the code: `agent_server.rs:445` and `executor.rs:1728` build `"{ep} && <script>"`, so it's a user-supplied shell command run *before* the job script, not the image's entrypoint. The table also shows no value placeholder, which reads as a boolean while the flag is `Option<String>`.

Something like ``--container-entrypoint <cmd>`` — "Shell command run inside the container before the job script" would line up. The `srun`/`sbatch` help strings ("Override the container entrypoint") imply replacement too, so all three could move together.

docs/user-guide/running-containers.rst
```diff
@@ -83,6 +85,36 @@ An interactive shell in a container with your home directory mounted:
...
+- **Step with its own image.** A standalone ``srun --container-image`` (or an
+  ``salloc`` allocation followed by ``srun --container-image``) creates a fresh
+  container for that step. Different steps in one allocation can use different
+  images and mounts.
+
+Concurrent steps on separate nodes each run their own container:
+
+.. code-block:: bash
+
+   salloc -N2 --gpus-per-node=8
```
yansun1996, line 109: This example won't run as written. `salloc` spawns an interactive shell and blocks, so the three following lines never execute in that block (the E2E wraps them via `salloc_run` for exactly this reason). `--overlap` is also inert without `--jobid`, since step mode keys off `SPUR_JOB_ID`. And `RunStepRequest` carries no nodelist/num_nodes — the controller fans the step over every entry in `job.allocated_nodes` — so `-N1 -w node1` is dropped and both `srun`s land on both nodes.

Could this be split into "allocate, then inside the allocation shell:", with the per-node targeting either removed or called out as not yet supported?

yansun1996 (changes requested):
Nice piece of work — this closes a genuinely confusing gap (#777), and the three-case dispatch reads clearly. The E2E suite is broad, and the "notable fixes found along the way" section is the right level of honesty about what had to be repaired. A few things I think need to change before merge, then some smaller notes.

**Blocking**

1. `step_rootfs_id = job_id * 10000 + step_id` shares an id space with the batch path, so a step can extract into — and later delete — a live batch job's rootfs.
2. The pipe read-ends are owned by `OwnedFd` **and** re-wrapped with `File::from_raw_fd`, giving a double `close()`.
3. Case 1 wins unconditionally, so a step's own `--container-image` is silently discarded inside a containerized job — the same silent-drop shape this PR is fixing.
4. The nsenter step path no longer sets `current_dir(&work_dir)`, which the pre-change `run_command` always did.
5. The cancel paths now return `Err(Status::cancelled)` where they previously returned `Ok(cancelled_step_response())`; the controller treats that as a dispatch failure.

Details inline.

**PR title / breaking change.** `sbatch --container-readonly` was previously accepted (as a no-op) and now hard-errors at submission. Per AGENTS.md that's a user-facing CLI break, so the title likely wants the marker: `feat(spurd)!: ...`.

**`--container-readonly` enforcement.** Rejecting in the CLI is a good call, but `ContainerConfig.readonly` is still never consumed and the flag is only refused in `srun`/`sbatch` — direct gRPC/REST/FFI and the k8s path still accept-and-ignore it. Worth a server-side rejection in `submit_job`/`run_step` so the guarantee doesn't depend on which client you use. Happy for that to be part of the follow-up you already flagged.

**Comment length.** The repo convention is that comments stay short (1–3 lines, explaining *why*). This diff adds 12 comment blocks of 3+ consecutive lines in `agent_server.rs` alone, including two 13-line blocks (the rootless-nsenter rationale in `build_nsenter_argv`, and the `cancel_active_steps_for_job` doc comment) and a 6-line one on `signal_step_process_group`, plus several multi-line docstrings in the E2E file (the 23-line module docstring especially). Most of that reasoning would land better in the commit message / PR description — could these be collapsed to a line or two each?

**Test coverage gaps.** The E2E breadth is good; the gaps are on the seams this PR introduces.

- The controller's step-vs-job container resolution (`server.rs:2773-2805`, nine stanzas) has no unit test at all. Something covering "step image overrides job image", "step env merges with job env", and "step can turn `mount_home` off" would pin the semantics.
- Nothing covers a step image that *differs* from the parent job's — the case in [3] above.
- No test asserts a cancelled step surfaces exit 143 rather than an RPC error.
- `run_command_uses_provided_work_dir` only exercises the host path; the nsenter path — where the regression in [4] is — is uncovered.
- `test_concurrent_overlap_steps` asserts `MARKER_CONTENT in out1_c`, which passes whether the step landed on one node or both. `out1_c.count(MARKER_CONTENT) == 1` would make it a real oracle for per-step node targeting.
- `test_named_container` runs a single step, so persistence — the whole point of `--container-name` — isn't exercised.

**Smaller notes**

- `proto/slurm.proto:1243-1245` says a non-empty image creates a container, but the server ignores it whenever the parent has namespaces.
- `agent_server.rs:2542` says Case 1 covers "`srun` inside `salloc --container-image`", but `register_job_allocation` sets all three namespace flags false, so salloc always reaches Case 2.
- `running-containers.rst:62` — "rejected at submission" holds for `sbatch`/`srun` but not for the other submission paths; maybe scope the wording.
- `running-containers.rst:114-116` — the cleanup claim doesn't hold for `--container-name` (`cleanup_rootfs` looks for `job_<id>`, named rootfs live at `<name>`) or when the client disconnects.
- The flag table's only per-command annotation ("(`sbatch`)") was removed without a replacement, so nothing now says which flags apply to `srun`; and `--container-name` is documented as "persist across jobs" while `srun.rs:161` says "persists across steps in an allocation".

A couple of things I looked at and don't think need action here: `parse_mount(..).ok()` silently dropping malformed mounts, and the root-owned script write into the rootfs, are both copied verbatim from the existing batch path, so they read as pre-existing rather than new; and `resolve_image(.., None, Some(uid))` is fine because `resolve_job_user_home` tries the uid first.

PR #819, opened 2026-09-05
feat(spurd): file-back and live-stream srun step output

### LINKED ISSUE #781: srun step output is withheld until exit; blocking steps produce no output at all

powderluv:
## Summary
Back srun **step** output with per-step spool files and stream it to the client live, instead of buffering the whole thing in the `RunStep` response until the step exits. Part of the bare-metal step-dispatch convergence (#777/#780/#781).

Two commits:
- **file-backed step output** — `spurd::run_command` now redirects each step's stdout/stderr to per-step spool files under the job spool dir (keyed on `(job_id, step_id)`) instead of `Stdio::piped()` + `wait_with_output()`. The child inherits the write fds (writes fine after privilege-drop); paths are recorded in `active_steps`. Response stays backward-compatible via a read-back for now.
- **live streaming** — `StreamJobOutputRequest` gains `step_id`; `stream_job_output` tails the step's spool file and ends when the step leaves `active_steps` (with a deterministic-path fallback for a step that finished before it was observed). The client `dispatch_step` runs a concurrent tail of the step's stdout+stderr while `RunStep` is in flight (single-node), suppressing the buffered print when streaming covered it. A client disconnect does not kill the step.

## Scope
Delivers **live delivery for single-node steps** — output appears before the step exits. It does not yet remove the response read-back for the multi-node / `-o`-file / streaming-failure fallbacks; that needs multi-node stream muxing and streaming into a client `-o` file (follow-ups).

## Testing
- `cargo test -p spurd` / `-p spur-cli` (new unit tests: on-disk spool file, step tail)
- clippy clean
- E2E on a single-node cluster: `test_srun_step_output.py` (stdout, stderr, exit-code + output) — green on shark-a

Closes #781.

### DIFF
--- a/crates/spur-cli/src/sattach.rs
+++ b/crates/spur-cli/src/sattach.rs
@@ -104,6 +104,7 @@ async fn stream_output_only(
             job_id,
             stream: stream_name.to_string(),
             user: user.to_string(),
+            ..Default::default()
         })
         .await
         .context("failed to start output stream")?
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -10,7 +10,8 @@ use spur_core::config::HooksConfig;
 use spur_proto::proto::slurm_controller_client::SlurmControllerClient;
 use spur_proto::proto::{
     CancelJobRequest, CompleteJobRequest, CreateJobStepRequest, GetJobRequest, GetNodeRequest,
-    JobSpec, JobState, RunStepRequest, StreamJobOutputRequest, SubmitJobRequest,
+    JobSpec, JobState, RunStepRequest, StreamJobOutputChunk, StreamJobOutputRequest,
+    SubmitJobRequest,
 };
 use std::collections::HashMap;
 use std::io::Write as _;
@@ -774,6 +775,22 @@ async fn dispatch_step(
     {
         anyhow::bail!("{err}");
     }
+    // Live-stream the step's output when it lands on a single node and the user
+    // has not redirected to a file. The tail runs concurrently with the blocking
+    // RunStep and ends when the step leaves the agent's active_steps (#781).
+    let live_node = if io.stdout.is_empty() && io.stderr.is_empty() {
+        single_step_node(client, job_id).await
+    } else {
+        None
+    };
+    let mut stream_handle = live_node.map(|node| {
+        let mut ctrl = client.clone();
+        let user = params.user.to_string();
+        tokio::spawn(async move {
+            stream_step_output_live(&mut ctrl, &node, job_id, step_id, &user).await
+        })
+    });
+
     let resp = client
         .run_step(RunStepRequest {
             job_id,
@@ -795,7 +812,24 @@ async fn dispatch_step(
         eprintln!("srun: dispatched to node {}", resp.node);
     }
 
-    emit_step_output(io, job_id, work_dir, &resp.stdout, &resp.stderr).await;
+    // The step has left active_steps, so a live tail eofs promptly; bound the
+    // join so a tail that never resolved (step failed to start) can't hang, and
+    // fall back to the buffered output whenever streaming didn't cover it.
+    let streamed = if let Some(handle) = stream_handle.as_mut() {
+        match tokio::time::timeout(std::time::Duration::from_secs(3), &mut *handle).await {
+            Ok(joined) => joined.unwrap_or(false),
+            Err(_) => {
+                handle.abort();
+                false
+            }
+        }
+    } else {
+        false
+    };
+
+    if !streamed {
+        emit_step_output(io, job_id, work_dir, &resp.stdout, &resp.stderr).await;
+    }
 
     Ok(StepDispatchResult {
         exit_code: resp.exit_code,
@@ -1060,6 +1094,7 @@ async fn try_stream_output(
             job_id,
             stream: "stdout".into(),
             user: user.to_string(),
+            ..Default::default()
         })
         .await
     {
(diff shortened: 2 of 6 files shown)


sajmera-pensando (commented):
## Review

**#781** is addressed cleanly — the spool-file + live-tail mechanism is sound, the backward-compatible read-back is a safe bridge, and the unit + E2E tests cover the key behaviors.

**#777** ( silently ignored) and **#780** ( silently discarded) are not addressed here and should be tracked in follow-up PRs.

### Non-blocking notes

- **200ms poll granularity** ([agent_server.rs, step tail loop](crates/spurd/src/agent_server.rs)): chunks arrive up to 200ms late. Fine for long-running steps; noticeable for short ones. inotify/kqueue would help but polling is simpler — worth a follow-up if latency becomes an issue.

- **Hardcoded 3s join timeout** ([srun.rs ~L818](crates/spur-cli/src/srun.rs)): after `RunStep` returns, the client waits up to 3s for the stream to eof. Under network jitter this adds latency to every step. Consider making it configurable.

- **TOCTOU on spool path** ([agent_server.rs ~L2295](crates/spurd/src/agent_server.rs)): paths are written into `active_steps` before `cmd.spawn()`. If spawn fails, a concurrent tail would open the empty file and wait the full 30s grace before giving up.

- **Test uses hardcoded job_id 77** ([agent_server.rs ~L4121](crates/spurd/src/agent_server.rs)): `stream_job_output_tails_step_spool_file` hard-codes job id 77, which could theoretically collide with the `NEXT_JOB_ID` counter added in the same PR. Should use the same counter pattern as `run_command_test_setup`.

- **Two agent connections for stdout/stderr** ([srun.rs ~L1162](crates/spur-cli/src/srun.rs)): each step opens two separate connections to the agent. Functional, but worth consolidating in a follow-up.

None of these affect the fix's correctness for the single-node streaming case.

sajmera-pensando (commented):
## Review

**#781** is addressed cleanly — the spool-file + live-tail mechanism is sound, the backward-compatible read-back is a safe bridge, and the unit + E2E tests cover the key behaviors.

**#777** (`srun --container-image` silently ignored) and **#780** (`srun --pty` silently discarded) are not addressed here and should be tracked in follow-up PRs.

### Non-blocking notes

- **200ms poll granularity** (`agent_server.rs`, step tail loop): chunks arrive up to 200ms late. Fine for long-running steps; noticeable for short ones. inotify/kqueue would help but polling is simpler — worth a follow-up if latency becomes an issue.

- **Hardcoded 3s join timeout** (`srun.rs` ~L818): after `RunStep` returns, the client waits up to 3s for the stream to eof. Under network jitter this adds latency to every step. Consider making it configurable.

- **TOCTOU on spool path** (`agent_server.rs` ~L2295): paths are written into `active_steps` before `cmd.spawn()`. If spawn fails, a concurrent tail would open the empty file and wait the full 30s grace before giving up.

- **Test uses hardcoded job_id 77** (`agent_server.rs` ~L4121): `stream_job_output_tails_step_spool_file` hard-codes job id 77, which could collide with the `NEXT_JOB_ID` counter added in the same PR. Should use the same counter pattern as `run_command_test_setup`.

- **Two agent connections for stdout/stderr** (`srun.rs` ~L1162): each step opens two separate connections to the agent. Functional, but worth consolidating in a follow-up.

None of these affect the fix's correctness for the single-node streaming case.

sajmera-pensando (approved):

PR #820, opened 2026-09-05
feat(spur-cli): honor --pty for srun steps in an allocation

### LINKED ISSUE #780: `srun --pty` is silently ignored inside allocations

powderluv:
## Summary
Honor `srun --pty` for **steps** inside an allocation, and warn about container/pty flags the step path silently dropped. Part of the bare-metal step-dispatch convergence (#777/#780/#781).

Two commits:
- **warn on dropped step-mode flags** — a step run inside an allocation dispatches via `run_as_step` → `RunStep`, which did not carry container or pty intent. Warn (rather than silently drop) when `--container-image` or `--pty` is set, via a pure testable helper.
- **honor `--pty` in step mode** — route a `--pty` step to `run_interactive_pty` (client → `InteractiveSession`/`run_pty_bridge` on the agent), the same machinery `srun --jobid --overlap --pty` already uses; step mode just never dispatched to it. Removes the now-obsolete `--pty` warning (`--container-image` still warns until #777 lands).

## Testing
- `cargo test -p spur-cli step_mode` (updated unit tests) + clippy clean
- E2E on a cluster: `test_srun_pty.py` — `srun --pty echo` runs through the interactive session; exit code propagates. Green on shark-a.

Closes #780.

### DIFF
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -1572,7 +1572,8 @@ async fn run_as_step(
     if args.pty {
         let node = first_node(args.nodelist.as_deref().unwrap_or_default());
         let exit_code =
-            run_interactive_pty(&args.controller, job_id, args.command.clone(), node, &user).await?;
+            run_interactive_pty(&args.controller, job_id, args.command.clone(), node, &user)
+                .await?;
         std::process::exit(exit_code);
     }
 
@@ -1665,14 +1666,9 @@ mod tests {
 
     #[test]
     fn step_mode_warns_only_on_container_image_when_pty_also_set() {
-        let args = SrunArgs::try_parse_from([
-            "srun",
-            "--container-image",
-            "img.sqsh",
-            "--pty",
-            "bash",
-        ])
-        .expect("parse failed");
+        let args =
+            SrunArgs::try_parse_from(["srun", "--container-image", "img.sqsh", "--pty", "bash"])
+                .expect("parse failed");
         let warnings = step_mode_unsupported_warnings(&args);
         assert_eq!(warnings.len(), 1);
         assert!(warnings[0].contains("--container-image"));
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -1524,6 +1524,30 @@ fn is_retryable_status(status: &tonic::Status) -> bool {
     )
 }
 
+/// Flags a step run inside an allocation accepts on the command line but does
+/// not yet honor. The batch and standalone-srun paths apply these; the step
+/// path (`dispatch_step` → `RunStep`) silently drops them, so warn rather than
+/// let a `--container-image`/`--pty` request quietly become a plain host run.
+/// Convergence tracked in spur#777 (container) and spur#780 (pty).
+fn step_mode_unsupported_warnings(args: &SrunArgs) -> Vec<String> {
+    let mut warnings = Vec::new();
+    if args.container_image.is_some() {
+        warnings.push(
+            "srun: warning: --container-image is not yet honored in step mode; \
+             the command runs on the host, not inside the image (spur#777)"
+                .to_string(),
+        );
+    }
+    if args.pty {
+        warnings.push(
+            "srun: warning: --pty is not yet honored in step mode; \
+             the command runs non-interactively without a terminal (spur#780)"
+                .to_string(),
+        );
+    }
+    warnings
+}
+
 async fn run_as_step(
     args: &SrunArgs,
     job_id: u32,
@@ -1538,6 +1562,9 @@ async fn run_as_step(
 
     if args.input.is_some() {
         eprintln!("srun: warning: --input is not supported in step mode, ignoring");
+    }
+    for warning in step_mode_unsupported_warnings(args) {
+        eprintln!("{warning}");
     }
 
     let io = resolve_io_paths(args);
@@ -1611,6 +1638,45 @@ fn srun_hook_context(script_context: &str, work_dir: &str) -> spur_core::hooks::
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn step_mode_warns_on_container_image() {
+        let args = SrunArgs::try_parse_from(["srun", "--container-image", "img.sqsh", "hostname"])
+            .expect("parse failed");
+        let warnings = step_mode_unsupported_warnings(&args);
+        assert_eq!(warnings.len(), 1);
+        assert!(warnings[0].contains("--container-image"));
+        assert!(warnings[0].contains("spur#777"));
+    }
+
+    #[test]
+    fn step_mode_warns_on_pty() {
+        let args =
+            SrunArgs::try_parse_from(["srun", "--pty", "bash"]).expect("parse failed");
+        let warnings = step_mode_unsupported_warnings(&args);
+        assert_eq!(warnings.len(), 1);
+        assert!(warnings[0].contains("--pty"));
+        assert!(warnings[0].contains("spur#780"));
+    }
+
+    #[test]
+    fn step_mode_warns_on_both_flags_together() {
+        let args = SrunArgs::try_parse_from([
+            "srun",
+            "--container-image",
+            "img.sqsh",
+            "--pty",
+            "bash",
+        ])
+        .expect("parse failed");
+        assert_eq!(step_mode_unsupported_warnings(&args).len(), 2);
+    }
+
+    #[test]
+    fn step_mode_silent_without_unsupported_flags() {
+        let args = SrunArgs::try_parse_from(["srun", "hostname"]).expect("parse failed");
+        assert!(step_mode_unsupported_warnings(&args).is_empty());
+    }
 
     #[test]
     fn first_node_reduces_comma_list_to_head() {
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -9,9 +9,9 @@ use clap::{ArgMatches, CommandFactory, FromArgMatches, Parser};
 use spur_core::config::HooksConfig;
 use spur_proto::proto::slurm_controller_client::SlurmControllerClient;
 use spur_proto::proto::{
-    CancelJobRequest, CompleteJobRequest, ContainerSpec, CreateJobStepRequest, GetJobRequest,
-    GetNodeRequest, JobSpec, JobState, RunStepRequest, StreamJobOutputChunk,
-    StreamJobOutputRequest, SubmitJobRequest,
+    CancelJobRequest, CompleteJobRequest, CompleteJobStepRequest, ContainerSpec,
+    CreateJobStepRequest, GetJobRequest, GetNodeRequest, JobSpec, JobState, RunStepRequest,
+    StreamJobOutputChunk, StreamJobOutputRequest, SubmitJobRequest,
 };
 use std::collections::HashMap;
 use std::io::Write as _;
@@ -241,6 +241,12 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
 
     // --jobid --overlap: exec into a running job (interactive PTY session)
     if let Some(job_id) = args.jobid {
+        let warnings = step_unsupported_warnings(&args)
+            .into_iter()
+            .chain(pty_step_unsupported_warnings(&args, &matches));
+        for warning in warnings {
+            eprintln!("{warning}");
+        }
         let node = first_node(args.nodelist.as_deref().unwrap_or_default());
         let channel = crate::authclient::connect(&args.controller)
             .await
@@ -254,8 +260,7 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
         let known_owner = (!job.user.is_empty()).then_some(job.user.as_str());
         let user = crate::interactive::job_caller_user(&mut ctrl, job_id, known_owner).await?;
         let exit_code =
-            run_interactive_pty(&args.controller, job_id, args.command.clone(), node, &user)
-                .await?;
+            run_interactive_pty(&mut ctrl, job_id, args.command.clone(), node, &user).await?;
         std::process::exit(exit_code);
     }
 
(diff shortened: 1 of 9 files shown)


crates/spur-cli/src/srun.rs
```diff
@@ -1353,13 +1382,29 @@ async fn run_as_step(
...
+    // same machinery `srun --jobid --overlap --pty` already uses. It runs its
+    // own step and owns the exit code, so return here rather than dispatching a
+    // second, non-interactive step.
+    if args.pty {
+        for warning in step_mode_unsupported_warnings(args) {
+            eprintln!("{warning}");
+        }
+        let node = first_node(args.nodelist.as_deref().unwrap_or_default());
+        let exit_code =
+            run_interactive_pty(&args.controller, job_id, args.command.clone(), node, &user)
```
yansun1996: The step created here never reaches a terminal state. `record_step_complete` (→ `WalOperation::JobStepComplete`) has exactly one caller — the `RunStep` handler in `spurctld/src/server.rs` — so an `InteractiveSession` step stays `Running` until `complete_job_steps` sweeps it when the whole job ends.

Two consequences for the in-allocation case this newly routes here:

- a long-lived `salloc` accumulates phantom `Running` steps in `squeue -s`
- the step's exit code never feeds the job's `DerivedExitCode`, which is accumulated only on `JobStepComplete`. Previously `srun --pty bash -c 'exit 9'` inside an allocation contributed 9; now it contributes nothing.

The leak already exists for `srun --jobid --overlap --pty`, but that's a rare path — this makes it the common one. Would it be worth recording completion after `drive_interactive_session` returns, and on the error paths?

crates/spur-cli/src/srun.rs
```diff
@@ -1353,13 +1382,29 @@ async fn run_as_step(
...
+                .await?;
+        std::process::exit(exit_code);
```
yansun1996: `std::process::exit` here skips `handle_terminal_state`, which is the only place `hooks.srun_epilog` runs. `SrunProlog` still runs for a `--pty` step (in `main_with_args`), so the two become unpaired — and this is a change in behavior: before this PR a `--pty` step fell through to the buffered path and did run the epilog.

Since `handle_terminal_state` is `-> !` and `output_streamed: true` suppresses re-printing, routing through it looks like a drop-in replacement for the `exit`. The standalone `--pty` path has the same gap today, so a small shared helper would cover both.

crates/spur-cli/src/srun.rs
```diff
@@ -1353,13 +1382,29 @@ async fn run_as_step(
...
+    // that path here (--pty itself is honored, so it needs no warning).
     let io = resolve_io_paths(args);
     let user = crate::interactive::job_caller_user(&mut client, job_id, None).await?;
 
+    // An interactive step (`srun --pty` inside an allocation) drives a PTY via
+    // InteractiveSession instead of the buffered RunStep path (spur#780), the
+    // same machinery `srun --jobid --overlap --pty` already uses. It runs its
+    // own step and owns the exit code, so return here rather than dispatching a
+    // second, non-interactive step.
+    if args.pty {
```
yansun1996: Three things `dispatch_step` does that a `--pty` step got before this change are now skipped:

- the `--input` warning — it moved into `dispatch_step` and `step_mode_unsupported_warnings` doesn't cover it, so `srun --pty --input=f` is now silent where it previously warned
- `--output`/`--error` — `io` is computed just above and unused here; `emit_step_output` honored them
- `warn_unsupported_cpu_bind` plus the `map_cpu_bind_error`/`mask_cpu_bind_error` `bail!`

The last is the concerning direction: `srun --pty --cpu-bind=map_cpu:0,1` used to fail loudly on a malformed spec and now succeeds silently unbound. Could the cpu-bind validation be hoisted above the branch, and `--input`/`-o`/`-e` folded into `step_mode_unsupported_warnings` so both paths share one source of truth?

crates/spur-cli/src/srun.rs
```diff
@@ -1353,13 +1382,29 @@ async fn run_as_step(
         .context("failed to connect to spurctld")?;
     let mut client = spur_proto::controller_client(channel);
 
-    if args.input.is_some() {
-        eprintln!("srun: warning: --input is not supported in step mode, ignoring");
-    }
-
+    // Step-mode warnings (--input, dropped container options) are emitted in
```
yansun1996: `dispatch_step` isn't only reached from step mode — `run_standalone_srun` calls `dispatch_step_cancellable` for `srun_step_dispatch` jobs, which is the normal path for a standalone `srun` on native-host nodes. So this warning and the pre-existing `--input` one now print "in step mode" during a plain standalone `srun`. The substance is still correct there (the container fields really are dropped), but the wording will confuse — maybe drop "in step mode" from the message?

Also, the block sits above `resolve_io_paths`/`job_caller_user`, which it doesn't describe; the code it's about is roughly ten lines below.

crates/spur-cli/src/srun.rs
```diff
@@ -1341,6 +1344,32 @@ fn is_retryable_status(status: &tonic::Status) -> bool {
     )
 }
 
+/// Flags a step run inside an allocation accepts on the command line but does
```
yansun1996: AGENTS.md: "Do not reference issue numbers, PR numbers, or task IDs in code comments." The `spur#777`/`spur#780` refs appear here and at 1393, 1498, and in the new test's docstring — worth moving to the commit message.

Separately, `(spur#777)` also lands in the user-facing warning at 1366 and is asserted on at 1484. There's no precedent for issue refs in `eprintln!` output elsewhere in the crate — `warn_unsupported_cpu_bind` names the supported values instead — and the ref isn't actionable for an end user. Dropping it from the message (and the assertion with it) would keep it consistent.

tests/native_host/e2e/test_srun_pty.py
```diff
@@ -0,0 +1,26 @@
...
+"""
+
+
+class TestSrunPtyStep:
+    def test_pty_step_runs_and_returns_output(self, cluster):
+        code, out = cluster.salloc_run("srun --pty echo PTY-MARKER-BRAVO\n")
+        assert code == 0, out
+        assert "PTY-MARKER-BRAVO" in out, out
+        # --pty is honored now, so the "not yet honored" warning must be gone.
+        assert "not yet honored" not in out, out
```
yansun1996: This assertion can't fail: the string only exists in the container warning added by this PR, and this test sets no container flags. It never appeared on the old code either — `--pty` was dropped silently rather than warned about.

Suggest dropping it, or moving it to a test that does set `--container-image` and asserts the warning *is* present (which would also be the only coverage of the new warning reaching the user through the pty path).

tests/native_host/e2e/test_srun_pty.py
```diff
@@ -0,0 +1,26 @@
...
+
+Inside an allocation, `srun --pty <cmd>` must run the command through an
+interactive PTY session (the InteractiveSession path) rather than the buffered
+RunStep path — the same machinery `srun --jobid --overlap --pty` uses — instead
+of silently dropping the flag and running non-interactively.
+"""
+
+
+class TestSrunPtyStep:
+    def test_pty_step_runs_and_returns_output(self, cluster):
```
yansun1996: Both tests pass unchanged against the pre-fix code: the old buffered `RunStep` path also printed `PTY-MARKER-BRAVO` and also exited 9, so neither would go red if the routing were reverted. The five unit tests only exercise the pure `step_mode_unsupported_warnings` helper, which the `if args.pty` routing never calls — so nothing currently guards the change itself.

A cheap oracle: the agent allocates a real PTY server-side (`spawn_pty_in_job` → `openpty_with_winsize`) regardless of the client's TTY, so

```python
code, out = cluster.salloc_run("srun --pty bash -c 'test -t 1 && echo IS-TTY'\n")
assert "IS-TTY" in out, out
```

is false on the old path and true on the new one. A unit-testable routing predicate would also give CI something to hold, since the e2e needs a live cluster.

yansun1996 (commented):
Nice change — routing a `--pty` step into the existing `InteractiveSession` machinery is the right shape, and reusing `run_interactive_pty` rather than growing `RunStep` keeps it small. A few things worth a look before this lands; most are about what the new early return in `run_as_step` skips.

The main one: the interactive step is created but never completed, so it stays `Running` and its exit code no longer reaches the job's `DerivedExitCode` (details inline). That's a behavior regression for the in-allocation case this PR newly routes here.

Also worth a look — `SrunEpilog` no longer runs for a `--pty` step, and the `--input` warning, `-o`/`-e` handling, and `--cpu-bind` validation are all skipped by the early return.

On tests: neither new e2e test would go red against the unfixed code, and the unit tests only cover the pure warning helper — so nothing currently guards the routing itself. There's a one-line oracle suggested inline.

Two that don't anchor to the diff:

- `docs/deployment/native-host.rst` (cgroup containment) says an interactive attach is "distinct from a `--pty` job, which *is* contained because it launches as a batch payload." After this change `srun --pty` inside an allocation goes through the agent's PTY spawn path — i.e. the "`srun` steps inside a job" escape that the same table already lists — so that line needs qualifying. The section closes by warning against relying on these limits as a security boundary, so it seems worth two sentences.
- `docs/user-guide/interactive.rst`'s `--pty` row could mention the in-allocation behavior (optional).

Minor / take or leave:

- `run_interactive_pty` hardcodes `num_tasks: 1, cpus_per_task: 1`, so `srun --pty -n8 -c4` silently collapses to one task. Slurm's `--pty` is single-task by design so this is probably fine, but by this PR's own "warn rather than silently drop" standard it may deserve a line in the helper.
- It also hardcodes `overlap: true` where every other step passes `false`. Inert today (`create_job_step` never reads the field), so no behavior impact — just an inconsistency a future reader could misread.
- `step_mode_no_longer_warns_on_pty` can't fail for the reason its comment gives: the helper never inspects `args.pty`, so it is equivalent to `step_mode_silent_without_unsupported_flags`.
- A few of the new comment blocks run 4–6 lines (1347–1352, 1385–1388, 1392–1396); AGENTS.md asks for 1–3, and most of it reads as commit-message material.

proto/slurm.proto
```diff
@@ -394,6 +394,9 @@ service SlurmController {
   // Job steps
   rpc GetJobSteps(GetJobStepsRequest) returns (GetJobStepsResponse);
   rpc CreateJobStep(CreateJobStepRequest) returns (CreateJobStepResponse);
+  // RunStep records its own steps; a PTY step streams client-to-agent, so its
```
sajmera-pensando, line 397: The proto comment here explains *what* the new RPC does rather than *why* it exists. Per project conventions, comments should capture non-obvious intent. A brief *why* — that RunStep owns its step lifecycle automatically but a PTY step runs client-side so nothing else can close it — would make this self-contained for a future reader. Fine to address in a follow-up.

crates/spur-cli/src/srun.rs
```diff
@@ -1417,104 +1424,144 @@ fn warn_unsupported_cpu_bind(environment: &HashMap<String, String>) {
     }
 }
 
-/// Create an interactive PTY step on a running job and attach to it.
-///
-/// Retries transient failures (job not yet visible on agent after controller
-/// reports it as Running) up to a few seconds before giving up.
+/// A session that never ran still closes its step out, reporting 1 so the step
```
sajmera-pensando, line 1427: Nit: `interactive_exit_code` is a one-liner that wraps a single expression. Its name is clear, but given the project convention ("three similar lines is better than a premature abstraction"), inlining `outcome.as_ref().copied().unwrap_or(1)` at the call site with a short comment explaining the default-to-1 policy would be simpler. Fine to address in a follow-up.

sajmera-pensando (approved):
Good PR. The design is sound: routing PTY steps through the existing interactive-session machinery is the right call, and the new `CompleteJobStep` RPC is necessary and correctly guarded (reserved IDs rejected, idempotency handled, 5-second timeout on the report). After this, Spur matches or beats Slurm on every `--pty` step dimension — including the epilog-hook fix for standalone srun that Slurm was missing. Test coverage is thorough end-to-end. Two minor nits left as inline comments, fine to address in a follow-up.

PR #829, opened 2026-09-05
fix(spur-core): parse --time minutes:seconds and days-hours correctly

### LINKED ISSUE #804: `-t MM:SS` is parsed as `HH:MM` (60x too long) and bare `-t days-hours` silently yields no time limit
## What happened

Two of the six time formats Slurm documents for `--time` are parsed incorrectly. `minutes:seconds` is read as `hours:minutes`, making every such limit 60x longer than requested, and bare `days-hours` fails to parse at all and is silently treated as no limit.

Running Spur's own parser (`parse_time_seconds`, extracted verbatim from `crates/spur-core/src/config.rs` at `fba350d`) against each documented form:

```
-t INPUT     SPUR COMPUTES          SLURM MEANS              RATIO  FORM
------------------------------------------------------------------------------------------------
10           600s (10m0s)           600s (10m0s)                1x  minutes
4:00         14400s (4h0m)          240s (4m0s)                60x  minutes:seconds  <== MISMATCH
30:00        108000s (1d6h0m)       1800s (30m0s)              60x  minutes:seconds  <== MISMATCH
1:30         5400s (1h30m)          90s (1m30s)                60x  minutes:seconds  <== MISMATCH
0:30         1800s (30m0s)          30s (30s)                  60x  minutes:seconds  <== MISMATCH
1:00:00      3600s (1h0m)           3600s (1h0m)                1x  hours:minutes:seconds
0:00:10      10s (10s)              10s (10s)                   1x  hours:minutes:seconds
2-12         NO LIMIT               216000s (2d12h0m)          INF  days-hours       <== MISMATCH
1-0          NO LIMIT               86400s (1d0h0m)            INF  days-hours       <== MISMATCH
2-12:30      217800s (2d12h30m)     217800s (2d12h30m)          1x  days-hours:minutes
2-12:30:15   217815s (2d12h30m)     217815s (2d12h30m)          1x  days-hours:minutes:seconds
```

Note that `days-hours:minutes` and `days-hours:minutes:seconds` are correct. Only the colon-less `days-hours` form falls through, because once the remainder contains a colon `parse_hms_seconds` accepts it.

## The reference grammar

From the Slurm `sbatch` documentation for `-t, --time`:

> Acceptable time formats include "minutes", "minutes:seconds", "hours:minutes:seconds", "days-hours", "days-hours:minutes" and "days-hours:minutes:seconds".

The two-field colon form is `minutes:seconds`. Spur reads it as `hours:minutes`.

## Observed on a deployed cluster

A job submitted with `-t 4:00`, which Slurm defines as four minutes:

```
JobId=<id> JobName=wrap
   RunTime=00:22:29 TimeLimit=04:00:00
   StartTime=2026-09-02T08:19:12 EndTime=2026-09-02T12:19:12
   SubmitLine=sbatch --wrap 'sleep 1800' -p <partition> -A <account> -q <qos> -N 1 -G 1 -t 4:00
```

`TimeLimit=04:00:00`. The job was granted a four-hour hold on a GPU for a request of four minutes, and `scontrol` reports the inflated value back in a plausible-looking form, so nothing about the output suggests a misparse.

## Code trace

Refs at `fba350d`.

The two-field branch, in the seconds variant:

```
// crates/spur-core/src/config.rs:2036-2041, in parse_slurm_time_seconds()
        2 => {
            // HH:MM → hours and minutes (no seconds)
            let h: u64 = parts[0].parse().ok()?;
            let m: u64 = parts[1].parse().ok()?;
            h.checked_mul(3600)?.checked_add(m.checked_mul(60)?)
        }
```

and identically in the minutes variant:

```
// crates/spur-core/src/config.rs:1996-2000, in parse_slurm_time_minutes()
        2 => {
            let h: u32 = parts[0].parse().ok()?;
            let m: u32 = parts[1].parse().ok()?;
            h.checked_mul(60)?.checked_add(m)
        }
```

The intent is stated in the doc comment on `parse_time_seconds` (`config.rs:2009-2011`): `"N" → N minutes, "H:MM" → hours+minutes, "H:MM:SS" → exact`. So this is a deliberate reading of the grammar, not an oversight in one branch — which is why the fix needs the documentation quoted above rather than a one-line patch.

For bare `days-hours`, the failure is a fallthrough. `parse_slurm_time_seconds` splits on `-` and hands the remainder to `parse_hms_seconds`, which only accepts 2 or 3 colon-separated fields (`config.rs:2089-2107`), so a bare `12` returns `None` and the `?` propagates. The fallback `parse_suffix_duration_seconds` then rejects the input on reaching the `-` character (`config.rs:2071-2076`, `_ => return None`). `parse_time_seconds` therefore returns `None`, which is the same value it returns for `INFINITE` and `UNLIMITED` (`config.rs:2014-2016`), and `sbatch` maps that to an absent `time_limit` field:

```
// crates/spur-cli/src/sbatch.rs:837-845
    let time_limit = args
        .time
        .as_ref()
        .and_then(|t| spur_core::config::parse_time_seconds(t))
        .map(|secs| prost_types::Duration { ... });
```

A user asking for `-t 2-12` gets an unlimited job.

## Blast radius

Both parsers carry the two-field branch, so the 60x inflation reaches every submission path:

- `sbatch -t` — `crates/spur-cli/src/sbatch.rs:838` (seconds variant)
- `srun -t` — `crates/spur-cli/src/srun.rs:557` (minutes variant)
- `salloc -t` — `crates/spur-cli/src/salloc.rs:301` (minutes variant). Note `time` is a plain `String` with `default_value = "1:00:00"` (`salloc.rs:48`), so this path always parses a value.
- partition `MaxTime` / `DefaultTime` — `crates/spur-core/src/config.rs:1930-1931` (config file), `crates/spurctld/src/cluster.rs:3848`, `3860`, and `crates/spurctld/src/server.rs:2168`, `2178` (RPC)

The `days-hours` → no-limit failure has a narrower reach, because the partition paths validate strictly. `parse_partition_time` (`config.rs:1957-1969`) returns `Err` where `parse_time_minutes` returns `None`, and `validate()` runs it over every partition's `max_time` and `default_time` (`config.rs:1886-1909`) from both loaders (`load_from_file:1599`, `load_from_str:1606`); the RPC paths reject with `InvalidArgument` likewise. So `MaxTime = "2-12"` is a hard failure, not a silent unlimited. **The unlimited fallthrough is confined to job submission** — `sbatch`, `srun`, and `salloc`.

Two paths are worth naming explicitly so they aren't mistaken for live vectors:

- `scontrol update job TimeLimit=` (`scontrol.rs:1241`) yields `None` on an unparseable value, and `None` in `UpdateJobRequest` means "not updating". A `days-hours` value is therefore a silent no-op rather than an unlimited job — still wrong, but a different failure.
- REST submission (`rest/handlers.rs:96`) parses identically but the endpoint refuses every request at `handlers.rs:126` (`spec.uid == 0`, as it carries no authenticated caller), so it is unreachable today.

The consequential case is the partition ceiling. An operator who configures `MaxTime = "30:00"` intending a thirty-minute cap has configured thirty hours, and because `"30:00"` is a *valid* parse it sails through validation untouched. The bound meant to limit how long a job holds a node is inflated by the same factor as the requests it is supposed to bound, so it stops constraining anything, and both the request and the limit read back in a form that looks deliberate.

## The quirk is already documented in one place

`parse_reservation_duration` (`crates/spur-cli/src/scontrol.rs:1583-1589`) carries this doc comment:

> Quirk of that grammar: a bare `MM:SS` is read as `HH:MM` and bare `days-hours` is rejected, so prefer the unambiguous colon forms.

So the behaviour is known and was worked around locally for reservation durations rather than fixed in the parser. That function is not part of the job time-limit path and errors out loudly, so it is not itself affected.

## Tests currently assert the incorrect reading

Any fix must update three assertions across two files:

- `crates/spur-core/src/config.rs:2606` — `assert_eq!(parse_time_minutes("1:30"), Some(90));`. Ninety minutes, where Slurm means ninety seconds.
- `crates/spur-core/src/config.rs:2618` — `assert_eq!(parse_time_seconds("1:30"), Some(5400)); // 1h30m`. Slurm means 90 seconds.
- `crates/spur-tests/src/t52_config.rs:22` — `assert_eq!(parse_time_minutes("1:30"), Some(90));`. Same as the first.

## Reproduction

```rust
#[test]
fn two_field_form_is_minutes_seconds_and_days_hours_parses() {
    // Slurm: "minutes:seconds"
    assert_eq!(parse_time_seconds("4:00"), Some(240));     // actual: Some(14400)
    assert_eq!(parse_time_seconds("1:30"), Some(90));      // actual: Some(5400)
    // Slurm: "days-hours"
    assert_eq!(parse_time_seconds("2-12"), Some(216000));  // actual: None
    assert_eq!(parse_time_seconds("1-0"), Some(86400));    // actual: None
}
```

Or end to end: `sbatch --wrap 'sleep 60' -t 4:00` and observe `TimeLimit=04:00:00` in `scontrol show job`.

## Expected behaviour

`H:MM` should become `MM:SS`, and bare `days-hours` should parse as days plus hours. Single-field minutes, `hours:minutes:seconds`, `days-hours:minutes`, and `days-hours:minutes:seconds` are already correct and should not change. The suffixed extension (`90m`, `1h30m`) is a Spur addition outside the Slurm grammar and is unaffected.

## Migration note

This is a behaviour break. Every job script and partition config written against the current reading will get 1/60th of the wall time its author has been observing, and sites relying on `MaxTime = "H:MM"` will see their effective ceiling drop sharply. It warrants a release note and probably a startup warning when a config value uses the two-field form, rather than a silent correction.

powderluv:
## Summary
Two of the six documented Slurm `--time` forms were parsed incorrectly (issue #804):

- **`minutes:seconds` read as `hours:minutes`** — a two-field value with no days was treated as `HH:MM`, so every such limit was 60× too long: `-t 4:00` became 4 hours instead of 4 minutes.
- **bare `days-hours` yielded no limit** — `-t 2-12` failed to parse (the days-remainder parser required at least `hours:minutes`), and a failed parse is silently treated as unlimited.

## Fix
- `parse_slurm_time_seconds` / `parse_slurm_time_minutes`: the no-days two-field arm is `minutes:seconds`, not `hours:minutes` (Slurm has no bare `HH:MM` form — hours only appear with days or the three-field `H:M:S`).
- `parse_hms_seconds` / `parse_hms`: accept a bare hours field after `days-`.

## Testing
- New table-driven test over all six documented forms (`minutes`, `minutes:seconds`, `hours:minutes:seconds`, `days-hours`, `days-hours:minutes`, `days-hours:minutes:seconds`).
- Corrected two existing tests that encoded the old (buggy) `HH:MM` behavior; overflow tests updated for the new arm.
- `cargo test -p spur-core parse_time` green.

Closes #804.

### DIFF
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -1995,9 +1995,10 @@ fn parse_slurm_time_minutes(s: &str) -> Option<u32> {
     match parts.len() {
         1 => parts[0].parse().ok(),
         2 => {
-            let h: u32 = parts[0].parse().ok()?;
-            let m: u32 = parts[1].parse().ok()?;
-            h.checked_mul(60)?.checked_add(m)
+            // minutes:seconds → round a sub-minute remainder up, matching parse_hms
+            let m: u32 = parts[0].parse().ok()?;
+            let sec: u32 = parts[1].parse().ok()?;
+            m.checked_add(if sec > 0 { 1 } else { 0 })
         }
         3 => parse_hms(s),
         _ => None,
@@ -2034,13 +2035,13 @@ fn parse_slurm_time_seconds(s: &str) -> Option<u64> {
             mins.checked_mul(60)
         }
         2 => {
-            // HH:MM → hours and minutes (no seconds)
-            let h: u64 = parts[0].parse().ok()?;
-            let m: u64 = parts[1].parse().ok()?;
-            h.checked_mul(3600)?.checked_add(m.checked_mul(60)?)
+            // minutes:seconds (Slurm has no bare hours:minutes form without days)
+            let m: u64 = parts[0].parse().ok()?;
+            let sec: u64 = parts[1].parse().ok()?;
+            m.checked_mul(60)?.checked_add(sec)
         }
         3 => {
-            // HH:MM:SS
+            // hours:minutes:seconds
             let h: u64 = parts[0].parse().ok()?;
             let m: u64 = parts[1].parse().ok()?;
             let sec: u64 = parts[2].parse().ok()?;
@@ -2087,8 +2088,13 @@ fn parse_suffix_duration_seconds(s: &str) -> Option<u64> {
 }
 
 fn parse_hms_seconds(s: &str) -> Option<u64> {
+    // The part after `days-`: hours, hours:minutes, or hours:minutes:seconds.
     let parts: Vec<&str> = s.split(':').collect();
     match parts.len() {
+        1 => {
+            let h: u64 = parts[0].parse().ok()?;
+            h.checked_mul(3600)
+        }
         2 => {
             let h: u64 = parts[0].parse().ok()?;
             let m: u64 = parts[1].parse().ok()?;
@@ -2107,12 +2113,17 @@ fn parse_hms_seconds(s: &str) -> Option<u64> {
 }
 
 fn parse_hms(s: &str) -> Option<u32> {
+    // The part after `days-`: hours, hours:minutes, or hours:minutes:seconds.
     let parts: Vec<&str> = s.split(':').collect();
-    if parts.len() != 3 && parts.len() != 2 {
+    if parts.len() > 3 || parts.is_empty() {
         return None;
     }
     let h: u32 = parts[0].parse().ok()?;
-    let m: u32 = parts[1].parse().ok()?;
+    let m: u32 = if parts.len() >= 2 {
+        parts[1].parse().ok()?
+    } else {
+        0
+    };
     let s: u32 = if parts.len() == 3 {
         parts[2].parse().ok()?
     } else {
@@ -2603,7 +2614,8 @@ cni_mtu = 1400
     #[test]
     fn test_parse_time() {
         assert_eq!(parse_time_minutes("60"), Some(60));
-        assert_eq!(parse_time_minutes("1:30"), Some(90));
+        // minutes:seconds — 1m30s rounds up to 2 minutes
+        assert_eq!(parse_time_minutes("1:30"), Some(2));
         assert_eq!(parse_time_minutes("72:00:00"), Some(4320));
         assert_eq!(parse_time_minutes("1-00:00:00"), Some(1440));
         assert_eq!(parse_time_minutes("INFINITE"), None);
@@ -2614,9 +2626,9 @@ cni_mtu = 1400
         // "N" → N minutes in seconds
         assert_eq!(parse_time_seconds("1"), Some(60));
         assert_eq!(parse_time_seconds("60"), Some(3600));
-        // "H:MM" → exact seconds
-        assert_eq!(parse_time_seconds("1:30"), Some(5400)); // 1h30m
-                                                            // "H:MM:SS" → exact seconds (the key case)
+        // "MM:SS" → exact seconds
+        assert_eq!(parse_time_seconds("1:30"), Some(90)); // 1m30s
+                                                          // "H:MM:SS" → exact seconds (the key case)
         assert_eq!(parse_time_seconds("0:00:10"), Some(10));
         assert_eq!(parse_time_seconds("0:01:30"), Some(90));
         assert_eq!(parse_time_seconds("1:00:00"), Some(3600));
@@ -2626,6 +2638,34 @@ cni_mtu = 1400
         // limits
         assert_eq!(parse_time_seconds("INFINITE"), None);
         assert_eq!(parse_time_seconds("UNLIMITED"), None);
+    }
+
+    #[test]
+    fn parse_time_covers_all_six_slurm_forms() {
+        // The six documented `--time` forms, in seconds.
+        let cases = [
+            ("10", 10 * 60),                                      // minutes
+            ("4:00", 4 * 60),                                     // minutes:seconds
+            ("30:00", 30 * 60),                                   // minutes:seconds
+            ("1:30", 90),                                         // minutes:seconds
+            ("0:30", 30),                                         // minutes:seconds
+            ("1:00:00", 3600),                                    // hours:minutes:seconds
+            ("2-12", 2 * 86400 + 12 * 3600),                      // days-hours
+            ("2-12:30", 2 * 86400 + 12 * 3600 + 30 * 60),         // days-hours:minutes
+            ("2-12:30:45", 2 * 86400 + 12 * 3600 + 30 * 60 + 45), // days-hours:minutes:seconds
+            ("1-0", 86400),                                       // days-hours (bare hours = 0)
+        ];
+        for (input, want) in cases {
+            assert_eq!(parse_time_seconds(input), Some(want), "seconds for {input}");
+        }
+
+        // minutes view: sub-minute remainders round up.
+        assert_eq!(parse_time_minutes("10"), Some(10));
+        assert_eq!(parse_time_minutes("4:00"), Some(4));
+        assert_eq!(parse_time_minutes("1:30"), Some(2)); // 1m30s → 2m
+        assert_eq!(parse_time_minutes("0:30"), Some(1)); // 30s → 1m
+        assert_eq!(parse_time_minutes("2-12"), Some(2 * 1440 + 12 * 60));
+        assert_eq!(parse_time_minutes("2-12:30"), Some(2 * 1440 + 12 * 60 + 30));
     }
 
     #[test]
@@ -2976,8 +3016,8 @@ memory_mb = 1024000
         // Overflow on the `+ hms`, not the multiply: 2982616 days leaves only
         // 255 minutes of headroom and "05:00:00" is 300.
         assert_eq!(parse_time_minutes("2982616-05:00:00"), None);
-        // HH:MM arm.
-        assert_eq!(parse_time_minutes("71582789:00"), None);
+        // minutes:seconds arm — a minutes field past u32::MAX fails to parse.
+        assert_eq!(parse_time_minutes("4294967296:00"), None);
         // u64 seconds overflow.
         assert_eq!(parse_time_seconds("213503982334602-00:00:00"), None);
         // The largest non-overflowing value is still accepted unchanged.
--- a/crates/spur-tests/src/t52_config.rs
+++ b/crates/spur-tests/src/t52_config.rs
@@ -151,7 +151,7 @@ priority_tier = 1
 [[partitions]]
 name = "debug"
 nodes = "node[001-002]"
-max_time = "1:00"
+max_time = "1:00:00"
 "#,
         )
         .unwrap();


crates/spur-tests/src/t52_config.rs
```diff
@@ -151,7 +151,7 @@ priority_tier = 1
 [[partitions]]
 name = "debug"
 nodes = "node[001-002]"
-max_time = "1:00"
+max_time = "1:00:00"
```
yansun1996, line 154: This fixture edit is right, but the assertion 130 lines above it in the same file was missed — `t52_config.rs:22` still has:

```rust
assert_eq!(parse_time_minutes("1:30"), Some(90));
```

which now returns `Some(2)`, so `cargo test` fails (the module is wired in at `crates/spur-tests/src/lib.rs:50`). The enclosing test name `t52_2_parse_hours_minutes` is worth renaming too — it no longer describes what it checks.

Worth noting that needing this fixture edit at all is the clearest evidence for the `!` request in the summary: a real config carrying `max_time = "1:00"` silently means one minute after this lands.

crates/spur-core/src/config.rs
```diff
@@ -1995,9 +1995,10 @@ fn parse_slurm_time_minutes(s: &str) -> Option<u32> {
     match parts.len() {
         1 => parts[0].parse().ok(),
         2 => {
-            let h: u32 = parts[0].parse().ok()?;
-            let m: u32 = parts[1].parse().ok()?;
-            h.checked_mul(60)?.checked_add(m)
+            // minutes:seconds → round a sub-minute remainder up, matching parse_hms
+            let m: u32 = parts[0].parse().ok()?;
```
yansun1996, line 1999: The round-up assumes the seconds field is `< 60`, so it undercounts once it isn't: `parse_time_minutes("0:90")` gives 1 minute, while `parse_time_seconds("0:90")` gives 90 seconds. Since `sbatch` uses the seconds parser (`sbatch.rs:879`) and `srun`/`salloc` use the minutes one (`srun.rs:582`, `salloc.rs:301`), the same `-t` string means different things depending on the command — and with this change `-t 0:05` is 5 seconds under `sbatch` versus 1 minute under `srun`.

Would it be worth deriving the minutes view from the seconds view rather than maintaining a parallel grammar?

```rust
pub fn parse_time_minutes(s: &str) -> Option<u32> {
    let secs = parse_time_seconds(s)?;
    u32::try_from(secs.div_ceil(60)).ok()
}
```

That lets `parse_slurm_time_minutes` and `parse_hms` go away entirely, and it also fixes the same `sec >= 60` rounding bug in the three-field arm (`"0:0:90"` → 1 minute today), which is pre-existing but sits right next door.

The bot review raised the same rounding issue independently and suggested `div_ceil(60)` — that lands in the same place.

crates/spur-core/src/config.rs
```diff
@@ -2107,12 +2113,17 @@ fn parse_hms_seconds(s: &str) -> Option<u64> {
 }
 
 fn parse_hms(s: &str) -> Option<u32> {
+    // The part after `days-`: hours, hours:minutes, or hours:minutes:seconds.
```
yansun1996, line 2116: `parse_hms` is reached from two places — the days branch at `config.rs:1989` *and* the non-days three-field arm at `config.rs:2003` — so "the part after `days-`" understates its scope. The identical comment on `parse_hms_seconds` (line 2091) is accurate, since that one really does have a single call site at line 2027.

Simplest fix is to drop it from `parse_hms`; per `AGENTS.md` it's describing *what* rather than *why*, and the body is already clear. (Moot if you take the consolidation suggested on line 1999.)

crates/spur-core/src/config.rs
```diff
@@ -2107,12 +2113,17 @@ fn parse_hms_seconds(s: &str) -> Option<u64> {
...
     let parts: Vec<&str> = s.split(':').collect();
-    if parts.len() != 3 && parts.len() != 2 {
+    if parts.len() > 3 || parts.is_empty() {
```
yansun1996, line 2118: `parts.is_empty()` can't be true — `str::split(':')` always yields at least one element (`"".split(':')` yields `[""]`), and the empty case is already caught by the `parse()` two lines down.

`if parts.len() > 3` is enough, or `if !(1..=3).contains(&parts.len())` if you'd rather state the accepted range explicitly.

crates/spur-core/src/config.rs
```diff
@@ -2976,8 +3016,8 @@ memory_mb = 1024000
         // Overflow on the `+ hms`, not the multiply: 2982616 days leaves only
         // 255 minutes of headroom and "05:00:00" is 300.
         assert_eq!(parse_time_minutes("2982616-05:00:00"), None);
-        // HH:MM arm.
-        assert_eq!(parse_time_minutes("71582789:00"), None);
+        // minutes:seconds arm — a minutes field past u32::MAX fails to parse.
+        assert_eq!(parse_time_minutes("4294967296:00"), None);
```
yansun1996, line 3020: This one no longer reaches the overflow path: `u32::MAX` is `4294967295`, so `4294967296` fails at `str::parse` before `checked_add` is ever called. The assertion passes for the wrong reason, and it replaced a case (`"71582789:00"`) that did genuinely hit `checked_mul(60)`.

A differential pair would pin it properly:

```rust
assert_eq!(parse_time_minutes("4294967295:00"), Some(4294967295));
assert_eq!(parse_time_minutes("4294967295:01"), None); // checked_add overflow
```

crates/spur-core/src/config.rs
```diff
@@ -2628,6 +2640,34 @@ cni_mtu = 1400
...
+            ("2-12:30", 2 * 86400 + 12 * 3600 + 30 * 60),         // days-hours:minutes
+            ("2-12:30:45", 2 * 86400 + 12 * 3600 + 30 * 60 + 45), // days-hours:minutes:seconds
+            ("1-0", 86400),                                       // days-hours (bare hours = 0)
+        ];
+        for (input, want) in cases {
+            assert_eq!(parse_time_seconds(input), Some(want), "seconds for {input}");
+        }
+
+        // minutes view: sub-minute remainders round up.
+        assert_eq!(parse_time_minutes("10"), Some(10));
```
yansun1996, line 2663: Given the test's name, worth closing the gap: the seconds table above does cover all six forms, but these minutes assertions cover only four — `hours:minutes:seconds` and `days-hours:minutes:seconds` are missing.

Rather than adding two more ad-hoc lines, could the minutes half loop over the same `cases` table and assert the invariant?

```rust
for (input, want) in cases {
    assert_eq!(parse_time_minutes(input), u32::try_from(want.div_ceil(60)).ok(),
               "minutes for {input}");
}
```

Then adding `("0:90", 90)` to `cases` also gives you regression coverage for the divergence noted on line 1999. As written the table passes only because every case happens to have a seconds field under 60.

crates/spur-core/src/config.rs
```diff
@@ -2614,9 +2626,9 @@ cni_mtu = 1400
         // "N" → N minutes in seconds
         assert_eq!(parse_time_seconds("1"), Some(60));
         assert_eq!(parse_time_seconds("60"), Some(3600));
-        // "H:MM" → exact seconds
-        assert_eq!(parse_time_seconds("1:30"), Some(5400)); // 1h30m
-                                                            // "H:MM:SS" → exact seconds (the key case)
+        // "MM:SS" → exact seconds
+        assert_eq!(parse_time_seconds("1:30"), Some(90)); // 1m30s
+                                                          // "H:MM:SS" → exact seconds (the key case)
```
yansun1996, line 2631: rustfmt has pushed this comment out to column 59 where it hangs off the end of the previous assertion. Moving it onto its own line above the `assert_eq!` it describes reads better.

yansun1996 (changes requested):
Thanks for picking this up — the grammar analysis is right (Slurm genuinely has no bare `HH:MM` form), and `2-12` silently meaning "unlimited" was a nasty one to have in there. I'm requesting changes rather than nitting, because the reinterpretation is a silent 60x change to values that appear in deployed configs and live scripts, and the rollout of it is incomplete in a few places.

### Blockers

**CI is red.** `crates/spur-tests/src/t52_config.rs:22` still asserts `parse_time_minutes("1:30") == Some(90)`; it now returns `Some(2)`. Noted inline on the fixture change in the same file.

**This is a breaking change and should carry `!`.** Two-field values reach a lot more than `-t`:

- partition `max_time` / `default_time` in `/etc/spur/spur.conf` (`config.rs:1888-1896` → `build_partitions` at `config.rs:1930-1931`, enforced at `cluster.rs:7106` `partition_time_allows` and applied at `cluster.rs:7753`)
- REST job submit `time_limit` (`spurctld/src/rest/handlers.rs:96`)
- runtime partition CRUD (`spurctld/src/server.rs:2263,2273`; `cluster.rs:3868,3880`)
- `scontrol update Job TimeLimit=` (`spur-cli/src/scontrol.rs:1431`)

A deployed cluster with `max_time = "4:00"` goes from 4 hours to 4 minutes on restart with no error — every job over 4 minutes just pends on `PartitionTimeLimit`. `default_time = "1:00"` is worse, since it silently caps jobs that never asked for a limit. Per the Breaking Changes section of `AGENTS.md` this wants `!` on the type/scope, plus a release note telling operators to audit two-field time values. A startup WARN when a partition time parses through the two-field arm would make it self-diagnosing.

**The e2e suite encodes the old meaning in ~27 places.** Most importantly `tests/native_host/e2e/cluster.py:518` (the default `salloc_args` used by every `salloc_run` caller, 5 min → 1 min) and `tests/native_host/e2e/test_srun_container.py:439`, which does `sbatch -t 0:05` over a `sleep 300` payload — that becomes a **5-second** limit, and the `wait_job_state(..., "R", timeout=60)` on the next line won't reliably see it running. `test_srun_pty.py:62,113` and ~20 `srun_with_exit([..., "-t", "0:02"])` sites drop to 1 minute. These need to move to `-t 5` / `-t 0:05:00` in this PR, or the suite starts flaking on merge.

**`minutes` and `seconds` now disagree, and the gap widened.** Covered inline, but the headline is that `-t 0:05` is now 5 *seconds* through `sbatch` (`parse_time_seconds`) and 1 *minute* through `srun`/`salloc` (`parse_time_minutes`).

### Also worth addressing

**`sacctmgr` still speaks the old grammar.** `spur-cli/src/sacctmgr.rs:1344` `parse_wall_time` is a second copy of this grammar, still reading two fields as `hh:mm` (`:1365-1369`) and still rejecting bare `days-hours`. After this merges, `sbatch -t 4:00` is 4 minutes while `sacctmgr ... GrpWall=4:00` is 4 hours, from the same binary. The duplication predates this PR but the 60x divergence is new here. Routing `parse_wall_limit` through `spur_core::config::parse_time_minutes` would fix it and pick up the `checked_*` arithmetic the core parser already has (`sacctmgr.rs:1359` is unchecked today, and `parse_wall_limit_accepts_duration_syntax` at `:1680` locks in the old behavior). A follow-up issue is fine if you'd rather keep scope tight — but note `docs/admin-guide/accounting.rst:515-522` documents `grpwall=10:00` as ten hours, which stays correct only for as long as the two parsers disagree.

**Stale docs left behind by the change** (all outside the diff, so noting here):

- `config.rs:1971-1973` — `parse_time_minutes`'s doc still lists `H:MM` as accepted and omits the newly accepted `D-HH` / `D-HH:MM`.
- `config.rs:2011-2012` — `parse_time_seconds`'s doc still says `"H:MM" → hours+minutes`.
- `scontrol.rs:258` — the `--duration` clap help advertises `H:MM` to users; this is printed output, not a comment.
- `scontrol.rs:1772-1775` — documents the old quirk verbatim ("a bare `MM:SS` is read as `HH:MM` and bare `days-hours` is rejected"). Both halves are now inverted; the caveat can just be deleted.
- `docs/admin-guide/configuration.rst:713-722` — nothing there is wrong, but the `max_time` examples don't mention `days-hours` or that `"4:00"` now means four minutes. Worth a line given `AGENTS.md` asks user-facing changes to ship docs.

**Test coverage.** Beyond the inline notes: the seconds two-field arm's `checked_mul(60)`/`checked_add` (`config.rs:2041`) has no overflow coverage at all — the only `parse_time_seconds` overflow assertion exercises the days branch. And negative cases (`"1:"`, `":30"`, `"1:2:3:4"`, `"5-"`) aren't pinned anywhere, so a future refactor could loosen the grammar unnoticed.

One thing I checked that's fine, for the record: format/parse round-trips are unaffected, and one case actually improves. `format_time`, `format_time_seconds`, and `timefmt::format_duration_dhms` all emit three colon-fields or `D-HH:MM:SS`, so nothing re-parses through the changed arm. `squeue`'s `format_duration_hms` (`squeue.rs:477`) does emit `M:SS` for sub-hour durations — under the old parser, pasting that back into `-t` inflated it 60x, and it now round-trips correctly. Good supporting evidence for the change.

PR #833, opened 2026-09-05
feat(spurd): containerized srun steps on bare-metal nodes

### LINKED ISSUE #777: `srun --container-image` is silently ignored on bare-metal job steps

sajmera-pensando:
## Summary

Fixes rocm/spur#777 — `srun --container-image` was silently accepted and dropped on bare-metal ("native host") job steps, running the command on the host instead of in the requested image. This carries the container intent across the step RPC and wires the existing container runtime into the step-dispatch path.

**This supersedes #818.** It is #818 rebased onto the merged output-streaming work (#819) and reworked to adopt the two architectural wins from @powderluv's #821, plus every review comment from #818 addressed. See "Relationship to #818/#819/#821" below.

## Approach

- **Proto:** a single `ContainerSpec` message on `RunStepRequest` (tag 11) and `RunCommandRequest` (tag 15) — append-only, no renumber/retype. `JobSpec`'s existing container fields (70–78) are unchanged.
- **CLI (`srun`):** fills `ContainerSpec` from the container flags (parity with `sbatch`, incl. `--container-name/--container-readonly/--container-entrypoint`).
- **Controller (`spurctld`):** `run_step` resolves the effective container per step (step override, else inherit the parent job; the step's env is **merged over** the job's, step-wins) and forwards it to each agent.
- **Agent (`spurd`):** `run_command` dispatches across three paths, each wiring the step's stdout/stderr to the per-step spool files (so `stream_job_output` tails them live — the #819 model):
  1. **Parent job has live namespaces** → enter them via `nsenter` (nested `srun`/`spur exec` into a running container). Includes the rootless `setgroups` fix (see below).
  2. **Step carries its own image, no parent container** → build a fresh rootfs (`setup_rootfs` + `run_containerized_step`): namespaces, mounts, GPU device injection, shadow/user, DNS, home, hooks, `pivot_root`, priv-drop. Reached by standalone `srun` and by `salloc` + `srun` (the primary Miles pattern).
  3. **No container** → host spawn.

## Notable correctness work

- **Rootless nsenter into a container (pre-existing bug, also fixed `spur exec`).** Entering a rootless (user-namespace) container failed with `nsenter: setgroups failed: Operation not permitted` (the kernel forbids `setgroups()` in an unprivileged user namespace). For user-namespace targets, entry now uses `nsenter --preserve-credentials` and skips `setpriv --init-groups`. **Gated on `has_user_namespace`, which is false whenever spurd is root**, so the root paths (and their `@rootful` groups test) emit the identical command as before.
- **Cancellation.** A containerized step's real workload is a grandchild that is PID 1 of its own namespace (ignores SIGTERM from an ancestor). Cancellation walks the `/proc` tree **and** signals the process group, snapshots pids under the `active_steps` lock and signals outside it, and escalates to SIGKILL guarded by a monotonic epoch (never a recycled pid). Applies to `scancel` and `srun` Ctrl-C.
- **GPU:** device-plan injection for step containers; `maybe_deny_gpu_env` on `container_env` so a zero-GPU step can't re-enable GPU visibility via `--container-env`.
- **`--container-readonly`** is a no-op on every backend today, so it is **rejected at submission** (with a follow-up to implement it) rather than silently accepted.

## #818 review comments (from @yansun1996) — all addressed here

- rootfs id collision / `u32` overflow → `setup_rootfs`/`cleanup_rootfs` take an explicit base; steps use a `step_<job>_<step>` namespace disjoint from `job_<id>` (unit test added).
- nsenter step inherited spurd's cwd → `current_dir(&work_dir)` set unconditionally (E2E added comparing batch vs step cwd).
- rootfs leaked when the future was dropped → `StepRootfsGuard` (Drop) makes cleanup cancel-safe.
- `kill_process_tree` lost `killpg` coverage and the `warn!` → `signal_step_tree` does both and warns.
- entrypoint NUL → `bash -c ""` false success; hardcoded `/bin/bash`; swallowed execve error → NUL rejected pre-fork, shell probed, failure surfaced.
- non-shell-safe step command silently dropped args → `invalid_argument`.
- `/proc` walk under the `active_steps` lock + pid recycling → snapshot-then-signal + epoch guard.
- nested `srun --container-image B` silently ignored under image A → `warn!`.
- step container env replaced the job's wholesale → merged, step-wins.
- docs: `--container-entrypoint` description + help strings corrected; the `salloc` example rewritten (per-node `-w`/`--overlap` caveats noted).
- weak nsenter-precedence unit test strengthened to assert on observable output.
- The fd double-close and the cancelled-step `Err`-vs-`Ok` comments are resolved by the spool-file refactor.

## Relationship to #818 / #819 / #821

- **#819** (per-step spool-file output streaming) is merged; this PR builds on it, so step output streams rather than being buffered inline.
- Adopts **#821**'s two architectural ideas: the `ContainerSpec` proto message and spool-file-native step output.
- Retains all of **#818**'s capability (Case 1 nsenter + rootless fix, flag parity, `--container-readonly` rejection, GPU injection, cancellation, the full E2E suite) — the test depth is unchanged from #818.
- **#818 should be closed in favor of this PR.**

## Related issues

#777, #780 (`--pty`) and #781 (output streaming, landed via #819) share the same bare-metal step-dispatch root. This addresses #777.

## Test plan

- [x] `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` — clean
- [x] `cargo test --locked` — passes (spurd/spurctld/spur-cli); unit coverage: container path taken vs silent host fallthrough, nsenter precedence (observable-branch assertion), rootless nsenter preserves credentials / skips setgroups, cancellation signals in-flight steps, rootfs base-name disjointness, `--container-readonly` rejection
- [x] Bare-metal E2E (`tests/native_host/e2e/test_srun_container.py`): standalone `srun`; `salloc` + `srun`; `sbatch` + nested `srun` (nsenter) and `spur exec`; nested-step cwd; concurrent `--overlap`; single multi-node `srun -N2`; bind mounts / workdir / home / user identity / DNS; GPU device injection (both paths, auto-skips without GPUs); `--container-name`, `--container-entrypoint`; cancellation via `scancel` and Ctrl-C; `--container-readonly` rejection

### DIFF
--- a/crates/spurd/src/agent_server.rs
+++ b/crates/spurd/src/agent_server.rs
@@ -196,10 +196,20 @@ fn start_pmix_launch(
     Ok((guard, plan, per_local_rank_env))
 }
 
+/// Monotonic, never-reused id stamped on each `ActiveStep`. The delayed SIGKILL
+/// escalation compares epochs (not pids) so it can never signal a recycled pid
+/// belonging to a different step that reused the same (job_id, step_id) key.
+static STEP_EPOCH: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(1);
+
+fn next_step_epoch() -> u64 {
+    STEP_EPOCH.fetch_add(1, std::sync::atomic::Ordering::Relaxed)
+}
+
 #[derive(Debug, Default)]
 struct ActiveStep {
     cancel_requested: bool,
     pid: Option<u32>,
+    epoch: u64,
     /// Spool files the step's stdout/stderr are redirected to, so
     /// `stream_job_output` can tail them live keyed on (job_id, step_id).
     stdout_path: String,
@@ -227,19 +237,32 @@ fn cancelled_step_response() -> RunCommandResponse {
     }
 }
 
-fn signal_step_process_group(pid: u32, signal: i32) {
+/// Signal a step's whole workload. The tracked pid may be an intermediate
+/// parent (a containerized step runs the workload as a grandchild that is PID 1
+/// of its own PID namespace), so a single `kill` on it never reaches the real
+/// process. It walks the `/proc` descendant tree first (before the group signal
+/// can reparent/reap children and empty `/proc/<pid>/children`), then signals
+/// the process group (host steps are group leaders, so this also covers a
+/// double-forked descendant that reparented out of the `/proc` tree).
+/// Best-effort; warns only when the tracked pid cannot be signalled at all.
+fn signal_step_tree(pid: u32, signal: i32) {
     let sig =
         nix::sys::signal::Signal::try_from(signal).unwrap_or(nix::sys::signal::Signal::SIGTERM);
-    let leader = nix::unistd::Pid::from_raw(pid as i32);
-    if let Err(e) = nix::sys::signal::killpg(leader, sig) {
-        if let Err(kill_err) = nix::sys::signal::kill(leader, Some(sig)) {
-            warn!(
-                pid,
-                signal,
-                killpg = %e,
-                kill = %kill_err,
-                "step process group signal failed (step may already have exited)"
-            );
+    let target = nix::unistd::Pid::from_raw(pid as i32);
+    crate::executor::kill_process_tree(pid as i32, sig);
+    if let Err(pg_err) = nix::sys::signal::killpg(target, sig) {
+        // Not a group leader (a container fork), or already gone. Fall back to
+        // the tracked pid so an unsignalable step still surfaces a warning.
+        if let Err(e) = nix::sys::signal::kill(target, Some(sig)) {
+            if e != nix::errno::Errno::ESRCH {
+                warn!(
+                    pid,
+                    signal,
+                    killpg = %pg_err,
+                    kill = %e,
+                    "failed to signal step (it may already have exited)"
+                );
+            }
         }
     }
 }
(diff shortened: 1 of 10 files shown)


crates/spurd/src/agent_server.rs
```diff
@@ -934,6 +1228,21 @@ impl Drop for StepScriptCleanup {
...
+struct StepRootfsGuard {
+    base: String,
+    mode: crate::container::RootfsMode,
+}
+
+impl Drop for StepRootfsGuard {
+    fn drop(&mut self) {
+        crate::container::cleanup_rootfs(&self.base, &self.mode);
+    }
+}
```
yansun1996, line 1244: `StepRootfsGuard::drop` tears down the rootfs unconditionally, but doesn't kill/reap the tracked child first. If `run_command`'s future is dropped mid-flight (client disconnect, RPC deadline) without going through `cancel_active_steps_for_job`, this guard fires immediately while the container process may still be running and pivoted into this rootfs — `cleanup_rootfs` can then race an `rm -rf`/umount against a live process. The doc comment above frames this guard as making cleanup fully cancel-safe, but that only covers the leak case, not this race. Would it be worth having the guard (or `run_containerized_step` itself) kill the tracked pid before removing the rootfs, so cleanup always happens after the process is gone?

crates/spurd/src/agent_server.rs
```diff
@@ -2307,32 +2616,238 @@ impl SlurmAgent for AgentService {
...
+                .unwrap_or_default();
+            let home_dir = env
+                .get("HOME")
+                .cloned()
+                .unwrap_or_else(|| format!("/home/{username}"));
+
+            let container_cfg = crate::container::ContainerConfig {
+                image: c.image.clone(),
+                mounts,
+                workdir: if c.workdir.is_empty() {
```
yansun1996, line 2720: This standalone-container step path only honors `--container-workdir` (`c.workdir`) — `req.work_dir` (i.e. `--chdir`) is never threaded into `ContainerConfig`, so `container_init` falls back to `/tmp` (`container.rs:1107`) when `--container-workdir` isn't set. That means `srun --chdir=/foo --container-image=X` silently lands in `/tmp` instead of `/foo`. The nsenter path (Case 1) explicitly `cd`s into `work_dir`, so this is an asymmetry — and it's the same silent-flag-drop class #777 was about, just on this path. Might be worth threading `req.work_dir` in as a fallback when `--container-workdir` isn't set.

yansun1996 (commented):
Went through this after it landed — solid fix overall for #777, and good consolidation of #818/#821. Found two things worth a follow-up, both around the new standalone-container step path (Case 2):

PR #834, opened 2026-09-06
fix(spurd): refresh node inventory and re-register when it changes

### LINKED ISSUE #800: Node resource inventory is captured once at spurd startup and never refreshed; heartbeats carry no resources
## What happened

Two nodes in the same partition on the same silicon (`0x75a3`) present different device counts, because they are in different compute-partition modes:

- `node-a` is in CPX and enumerates **64** KFD devices (8 packages x 8 XCDs).
- `node-b` is in SPX and enumerates **8**.

A node's partition mode was changed after `spurd` was already running. The controller continued to schedule against the inventory it learned at agent startup, and nothing in Spur noticed or reported the change.

## What was expected

When a node's device inventory changes, the controller converges on the new inventory without requiring someone to notice and restart the agent — or the node is marked unusable until it does. Allocating against an inventory that no longer describes the hardware should not be possible.

## Code trace

Refs at `fba350d`.

Resources are discovered exactly once, at agent startup:

```
// crates/spurd/src/main.rs:302-315
// Initialize device registry (CDI cache, GRES config, and discovery).
...
reporter::discover_resources(&reg)
```

`NodeReporter::resources` is a plain `ResourceSet` field populated at construction (`crates/spurd/src/reporter.rs:31-50`), and `register()` sends it once (`reporter.rs:104-114`).

The 30-second heartbeat carries no resource payload at all — only `hostname`, `cpu_load`, `free_memory_mb`, `running_jobs`, `node_token`, `wg_pubkey`, and `k0s_status` (`crates/spurd/src/reporter.rs:152-211`).

The only path that re-sends resources is an accidental one: a `NOT_FOUND` heartbeat means the controller lost the registration, which triggers `register()` again (`reporter.rs:196-204`, `should_reregister` at `reporter.rs:218-220`). Absent that, inventory is immutable for the lifetime of the `spurd` process.

Worth noting the asymmetry: the WireGuard public key *is* deliberately re-read live on every register and heartbeat, with a comment explaining that a key which "appears or changes after startup ... reaches the controller" (`reporter.rs:40-42`, `87-90`). Device inventory has no equivalent, though it is equally mutable in practice.

## Consequence

With stale inventory, the controller allocates device ids that no longer exist or no longer mean what it thinks. On dispatch, `spurd` rejects the launch and the job is reported as:

```
JobLaunchFailure (dispatch confirmation failed (0/1 confirmed): 1 gpu/resource allocation mismatch)
```

(the category is `DispatchError::ResourcesUnavailable`, `crates/spurctld/src/scheduler_loop.rs:1028`). The node is put on dispatch cooldown and the job is requeued with a backoff that counts against `max_batch_requeue`, so repeated attempts can exhaust the budget and park the job for an operator. I hit this on a submission that had no other reason to fail.

In the opposite direction (CPX to SPX, 64 down to 8) the controller believes in devices that are gone, and will keep handing out allocations that cannot be satisfied.

## Steps to reproduce

1. With `spurd` running on a GPU node, confirm the advertised count: `spur scontrol show node <node> | grep -i gres`
2. Change the node's compute-partition mode out of band (e.g. `rocm-smi --setcomputepartition cpx` on a drained node), without restarting `spurd`.
3. Query the node again. The advertised GRES count is unchanged and still reflects the pre-change mode.
4. Submit a GPU job that lands on that node and observe the dispatch outcome.

## Scope boundary with #801

This issue is the **data-path** gap: heartbeats carry no resource inventory, so the
controller has no mechanism to converge on a changed inventory even if it decided to. #801
is the **policy** gap: no health-check hook or gate exists anywhere in the node lifecycle.

Fixing #801 alone leaves the controller unable to learn a corrected inventory; fixing this
alone leaves other forms of node breakage undetected. Complementary, not duplicates.

powderluv:
## Summary
Node resources were discovered once at spurd startup and never refreshed, and the 30s heartbeat carried no resource payload — so a device count that changed out of band (a GPU partition-mode switch, a GPU dropping off the bus) never reached the controller, which kept scheduling against hardware that no longer existed (issue #800). The WireGuard key is already re-read live each heartbeat; inventory had no equivalent.

## Fix
A periodic (60s) re-discovery task: rebuild the device registry (re-running CDI / KFD discovery), recompute the `ResourceSet`, and when it differs from what was reported, **swap the shared registry** (so GPU injection uses the new set) and **re-register** with the controller so it converges on the new inventory.

- `resources` moves behind a `RwLock`; `update_resources()` returns whether it actually changed; `register()` snapshots the inventory before its await.

## Testing
- Unit test: `update_resources` reports a real change and no-ops an unchanged set.
- build + clippy + fmt green.

Closes #800.

### DIFF
--- a/crates/spur-sched/src/cons_tres.rs
+++ b/crates/spur-sched/src/cons_tres.rs
@@ -66,6 +66,29 @@ impl NodeAllocation {
         }
     }
 
+    /// Update total capacity from a re-discovered inventory, preserving current
+    /// allocations. CPUs/memory resize in place; GPUs are re-indexed by device
+    /// id so existing GPU allocations follow their device. A device that
+    /// vanished while allocated stops counting toward the total but its owning
+    /// job is untouched (it releases normally). Without this the agent's local
+    /// capacity stays frozen at startup and rejects launches for GPUs that only
+    /// appeared later (e.g. an SPX->CPX partition change), even after the
+    /// controller has learned about them via re-register.
+    pub fn update_capacity(&mut self, resources: &ResourceSet) {
+        self.total_cpus = resources.cpus;
+        self.allocated_cpus.resize(resources.cpus as usize, false);
+        self.total_memory_mb = resources.memory_mb;
+
+        let allocated_ids: std::collections::HashSet<u32> =
+            self.allocated_gpu_ids().into_iter().collect();
+        self.gpus = resources.gpus.clone();
+        self.gpu_allocated = self
+            .gpus
+            .iter()
+            .map(|g| allocated_ids.contains(&g.device_id))
+            .collect();
+    }
+
     /// Available (unallocated) CPU count.
     pub fn free_cpus(&self) -> u32 {
         self.allocated_cpus.iter().filter(|&&a| !a).count() as u32
@@ -324,6 +347,43 @@ mod tests {
         assert_eq!(node.free_cpus(), 64);
         assert_eq!(node.free_memory_mb(), 256_000);
         assert_eq!(node.free_gpus(None), 8);
+    }
+
+    #[test]
+    fn test_update_capacity_grows_and_preserves_allocations() {
+        // Start with 4 GPUs, allocate two of them, then an inventory refresh
+        // reveals 8 (e.g. SPX->CPX). The new GPUs become schedulable while the
+        // two already allocated stay allocated (by device id).
+        let mut node = make_node(32, 128_000, 4, "mi300x");
+        node.allocate_for_job(1, 8, 0, &[1, 2]).unwrap();
+        assert_eq!(node.free_gpus(None), 2);
+
+        let bigger = ResourceSet {
+            cpus: 64,
+            memory_mb: 256_000,
+            gpus: (0..8u32)
+                .map(|device_id| GpuResource {
+                    device_id,
+                    gpu_type: "mi300x".into(),
+                    memory_mb: 192_000,
+                    peer_gpus: vec![],
+                    link_type: GpuLinkType::XGMI,
+                })
+                .collect(),
+            ..Default::default()
+        };
+        node.update_capacity(&bigger);
+
+        assert_eq!(node.total_cpus, 64);
+        assert_eq!(node.free_memory_mb(), 256_000 - node.allocated_memory_mb);
+        assert_eq!(
+            node.free_gpus(None),
+            6,
+            "4 new GPUs added, 2 still allocated"
+        );
+        assert_eq!(node.allocated_gpu_ids(), vec![1, 2]);
+        // CPU allocation from before is preserved (8 cores still busy).
+        assert_eq!(node.free_cpus(), 64 - 8);
     }
 
     #[test]
--- a/crates/spurd/src/agent_server.rs
+++ b/crates/spurd/src/agent_server.rs
@@ -731,6 +731,14 @@ impl AgentService {
         self.k0s.clone()
     }
 
+    /// Shared handle to this node's local allocation, so the inventory-refresh
+    /// task can update its capacity (via `update_capacity`) when devices appear
+    /// or vanish. Without it the agent's capacity stays frozen at startup and
+    /// rejects launches for devices the controller has already re-learned.
+    pub fn allocation_handle(&self) -> Arc<Mutex<NodeAllocation>> {
+        self.allocation.clone()
+    }
+
     /// Spawn a background task to monitor running jobs and report completions.
     pub fn start_monitor(&self, controller_addr: String) {
         let running = self.running.clone();
--- a/crates/spurd/src/agent_server.rs
+++ b/crates/spurd/src/agent_server.rs
@@ -654,7 +654,7 @@ impl AgentService {
             hostname::get()
                 .map(|h| h.to_string_lossy().to_string())
                 .unwrap_or_else(|_| "unknown".into()),
-            &reporter.resources,
+            &reporter.resources.read().unwrap(),
         );
 
         // Load SPANK plugins from plugstack.conf if available
@@ -2124,9 +2124,9 @@ impl SlurmAgent for AgentService {
         &self,
         _request: Request<()>,
     ) -> Result<Response<NodeResourcesResponse>, Status> {
-        let resources = &self.reporter.resources;
+        let resources = self.reporter.resources.read().unwrap();
         Ok(Response::new(NodeResourcesResponse {
-            total: Some(crate::reporter::resource_to_proto(resources)),
+            total: Some(crate::reporter::resource_to_proto(&resources)),
             used: Some(crate::reporter::allocations_to_proto(
                 &spur_core::resource::ResourceAllocations::default(),
             )),
--- a/crates/spurd/src/main.rs
+++ b/crates/spurd/src/main.rs
@@ -353,36 +353,6 @@ async fn main() -> anyhow::Result<()> {
         hb_reporter.heartbeat_loop().await;
     });
 
-    // Periodically re-discover node inventory. Inventory is otherwise frozen at
-    // startup, so a device count that changes out of band (e.g. a GPU
-    // partition-mode switch, or a GPU dropping off the bus) never reaches the
-    // controller and it keeps scheduling against hardware that no longer exists.
-    // On a change, replace the shared registry (so injection uses the new set)
-    // and re-register with the controller.
-    {
-        let reporter = reporter.clone();
-        let registry = registry.clone();
-        let config = config.clone();
-        tokio::spawn(async move {
-            let mut ticker = tokio::time::interval(std::time::Duration::from_secs(60));
-            ticker.tick().await; // the first tick fires immediately; skip it
-            loop {
-                ticker.tick().await;
-                let rebuilt = init_device_registry(config.as_ref());
-                let fresh = reporter::discover_resources(&rebuilt);
-                if reporter.update_resources(fresh) {
-                    *registry.lock().await = rebuilt;
-                    match reporter.register().await {
-                        Ok(()) => {
-                            info!("node inventory changed; re-registered with the controller")
-                        }
-                        Err(e) => warn!(error = %e, "re-register after inventory change failed"),
-                    }
-                }
-            }
-        });
-    }
-
     // Start agent gRPC server (receives job launches + cluster-component RPCs from spurctld).
     // Pass the [cluster] config so the K0sAgent uses the operator's k0s version + install path.
     let limits = match config.as_ref() {
@@ -451,6 +421,51 @@ async fn main() -> anyhow::Result<()> {
     tokio::spawn(k0s.supervise());
 
     agent_service.start_monitor(args.controller.clone());
+
+    // Periodically re-discover node inventory. It is otherwise frozen at startup,
+    // so a device count that changes out of band (a GPU partition-mode switch, a
+    // GPU dropping off the bus) never reaches the controller and it keeps
+    // scheduling against hardware that no longer exists. On a change: swap the
+    // shared registry (so injection uses the new set), resize the local
+    // allocation capacity (so this node accepts launches for devices that
+    // appeared), and re-register with the controller.
+    {
+        let reporter = reporter.clone();
+        let registry = registry.clone();
+        let config = config.clone();
+        let allocation = agent_service.allocation_handle();
+        tokio::spawn(async move {
+            let mut ticker = tokio::time::interval(std::time::Duration::from_secs(60));
+            ticker.tick().await; // the first tick fires immediately; skip it
+                                 // `update_resources` commits the new inventory to the reporter as it
+                                 // detects the change, so a re-register that then fails would never be
+                                 // retried (the next tick sees no further change). Track that a
+                                 // re-register is still owed and keep trying until it lands.
+            let mut reregister_pending = false;
+            loop {
+                ticker.tick().await;
+                let rebuilt = init_device_registry(config.as_ref());
+                let fresh = reporter::discover_resources(&rebuilt);
+                let changed = reporter.update_resources(fresh.clone());
+                if changed {
+                    *registry.lock().await = rebuilt;
+                    allocation.lock().await.update_capacity(&fresh);
+                    reregister_pending = true;
+                }
+                if reregister_pending {
+                    match reporter.register().await {
+                        Ok(()) => {
+                            reregister_pending = false;
+                            info!("node inventory changed; re-registered with the controller")
+                        }
+                        Err(e) => {
+                            warn!(error = %e, "re-register after inventory change failed; will retry")
+                        }
+                    }
+                }
+            }
+        });
+    }
 
     let addr = args.listen.parse()?;
     info!(%addr, "agent gRPC server listening");
--- a/crates/spurd/src/main.rs
+++ b/crates/spurd/src/main.rs
@@ -353,6 +353,36 @@ async fn main() -> anyhow::Result<()> {
         hb_reporter.heartbeat_loop().await;
     });
 
+    // Periodically re-discover node inventory. Inventory is otherwise frozen at
+    // startup, so a device count that changes out of band (e.g. a GPU
+    // partition-mode switch, or a GPU dropping off the bus) never reaches the
+    // controller and it keeps scheduling against hardware that no longer exists.
+    // On a change, replace the shared registry (so injection uses the new set)
+    // and re-register with the controller.
+    {
+        let reporter = reporter.clone();
+        let registry = registry.clone();
+        let config = config.clone();
+        tokio::spawn(async move {
+            let mut ticker = tokio::time::interval(std::time::Duration::from_secs(60));
+            ticker.tick().await; // the first tick fires immediately; skip it
+            loop {
+                ticker.tick().await;
+                let rebuilt = init_device_registry(config.as_ref());
+                let fresh = reporter::discover_resources(&rebuilt);
+                if reporter.update_resources(fresh) {
+                    *registry.lock().await = rebuilt;
+                    match reporter.register().await {
+                        Ok(()) => {
+                            info!("node inventory changed; re-registered with the controller")
+                        }
+                        Err(e) => warn!(error = %e, "re-register after inventory change failed"),
+                    }
+                }
+            }
+        });
+    }
+
     // Start agent gRPC server (receives job launches + cluster-component RPCs from spurctld).
     // Pass the [cluster] config so the K0sAgent uses the operator's k0s version + install path.
     let limits = match config.as_ref() {
--- a/crates/spurd/src/reporter.rs
+++ b/crates/spurd/src/reporter.rs
@@ -31,7 +31,10 @@ impl<T: Send> HeldJobs for Mutex<HashMap<u32, T>> {
 pub struct NodeReporter {
     pub hostname: String,
     pub controller_addr: String,
-    pub resources: ResourceSet,
+    /// Node inventory. Behind a lock so a background refresh can update it live
+    /// (a device count that changes after startup, e.g. a GPU partition-mode
+    /// switch, must reach the controller rather than being frozen at boot).
+    pub resources: std::sync::RwLock<ResourceSet>,
     pub node_address: spur_net::NodeAddress,
     pub labels: HashMap<String, String>,
     pub free_memory_mb: AtomicU64,
@@ -64,7 +67,7 @@ impl NodeReporter {
         Self {
             hostname,
             controller_addr,
-            resources,
+            resources: std::sync::RwLock::new(resources),
             node_address,
             labels,
             free_memory_mb: AtomicU64::new(0),
@@ -94,6 +97,18 @@ impl NodeReporter {
         self.held_jobs.held_job_ids()
     }
 
+    /// Replace the reported inventory if it changed; returns whether it did, so
+    /// the caller can re-register the node with the controller.
+    pub fn update_resources(&self, fresh: ResourceSet) -> bool {
+        let mut cur = self.resources.write().unwrap();
+        if *cur != fresh {
+            *cur = fresh;
+            true
+        } else {
+            false
+        }
+    }
+
     /// Register with the controller.
     pub async fn register(&self) -> anyhow::Result<()> {
         let channel = spur_client::connect_channel(&self.controller_addr)
@@ -101,10 +116,13 @@ impl NodeReporter {
             .context("failed to connect to spurctld for registration")?;
         let mut client = spur_proto::controller_client(channel);
 
+        // Snapshot the inventory before the await so the lock guard is not held
+        // across it (the future must stay Send).
+        let resources = resource_to_proto(&self.resources.read().unwrap());
         let resp = client
             .register_agent(RegisterAgentRequest {
                 hostname: self.hostname.clone(),
-                resources: Some(resource_to_proto(&self.resources)),
+                resources: Some(resources),
                 version: env!("CARGO_PKG_VERSION").into(),
                 address: self.node_address.ip.clone(),
                 port: self.node_address.port as u32,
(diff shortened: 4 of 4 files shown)


crates/spurd/src/main.rs
```diff
@@ -422,6 +422,51 @@ async fn main() -> anyhow::Result<()> {
...
+            let mut ticker = tokio::time::interval(std::time::Duration::from_secs(60));
+            ticker.tick().await; // the first tick fires immediately; skip it
+                                 // `update_resources` commits the new inventory to the reporter as it
+                                 // detects the change, so a re-register that then fails would never be
+                                 // retried (the next tick sees no further change). Track that a
+                                 // re-register is still owed and keep trying until it lands.
+            let mut reregister_pending = false;
+            loop {
+                ticker.tick().await;
+                let rebuilt = init_device_registry(config.as_ref());
```
yansun1996, line 447: I'd treat this as blocking, since everything else depends on how far this signal can be trusted. `init_device_registry` swallows its errors — the CDI cache warns on parse failures and skips unreadable dirs, autodetect returns an empty vec — so "discovery failed" and "no GPUs" are the same value, and it's committed unconditionally. A driver reload rolling across the fleet would report every node as zero-GPU at once. It's also what makes the `cons_tres.rs` issues likely rather than rare.

Could discovery return a `Result` so the tick skips on error, and require two consecutive identical readings before acting? Also worth `spawn_blocking` here — both calls do synchronous sysfs/proc/CDI reads on a runtime worker, and the startup path already wraps comparable work.

crates/spurd/src/main.rs
```diff
@@ -422,6 +422,51 @@ async fn main() -> anyhow::Result<()> {
...
+                let fresh = reporter::discover_resources(&rebuilt);
+                let changed = reporter.update_resources(fresh.clone());
+                if changed {
+                    *registry.lock().await = rebuilt;
+                    allocation.lock().await.update_capacity(&fresh);
+                    reregister_pending = true;
+                }
+                if reregister_pending {
+                    match reporter.register().await {
```
yansun1996, line 456: Reporting a growth is always safe; reporting a shrink while a job still holds the vanished devices is what makes the controller's totals disagree with its own `alloc_resources`. Note the decision needn't live here — `register_node` already sees both the resource diff and `alloc_resources`, so the controller could drain instead, with no new plumbing. Longer term this is the seam where per-job evidence replaces the totals diff, so it's worth returning a classified outcome rather than deciding inline.

crates/spurd/src/main.rs
```diff
@@ -422,6 +422,51 @@ async fn main() -> anyhow::Result<()> {
...
+    // shared registry (so injection uses the new set), resize the local
+    // allocation capacity (so this node accepts launches for devices that
+    // appeared), and re-register with the controller.
+    {
+        let reporter = reporter.clone();
+        let registry = registry.clone();
+        let config = config.clone();
+        let allocation = agent_service.allocation_handle();
+        tokio::spawn(async move {
+            let mut ticker = tokio::time::interval(std::time::Duration::from_secs(60));
```
yansun1996, line 438: The 60s cadence is hardcoded while comparable timings (`heartbeat_timeout_secs`, `agent_keepalive_interval_secs`, ...) are config fields — worth an `inventory_refresh_secs` with `#[serde(default)]`, plus a short docs note since inventory changing on its own is new user-visible behaviour. Minor: `interval` defaults to `MissedTickBehavior::Burst`, so delayed ticks fire back-to-back.

crates/spur-sched/src/cons_tres.rs
```diff
@@ -66,6 +66,29 @@ impl NodeAllocation {
...
+    /// capacity stays frozen at startup and rejects launches for GPUs that only
+    /// appeared later (e.g. an SPX->CPX partition change), even after the
+    /// controller has learned about them via re-register.
+    pub fn update_capacity(&mut self, resources: &ResourceSet) {
+        self.total_cpus = resources.cpus;
+        self.allocated_cpus.resize(resources.cpus as usize, false);
+        self.total_memory_mb = resources.memory_mb;
+
+        let allocated_ids: std::collections::HashSet<u32> =
+            self.allocated_gpu_ids().into_iter().collect();
```
yansun1996, line 83: `allocated_gpu_ids()` is derived from `gpus` + `gpu_allocated`, so seeding from it loses any device not in the current list — a GPU that vanishes on one refresh and returns on the next comes back free while its job is running, and the first job's release then clears the second job's bit. Line 79 is the same shape: `resize` truncates allocated CPU entries on shrink, so `free_cpus()` over-reports. Both bite a healthy job on a busy node.

Gating on "nothing allocated" makes both impossible. Otherwise, `owners` is the documented source of truth:

```rust
let allocated_ids: HashSet<u32> =
    self.owners.values().flat_map(|a| a.gpu_ids.iter().copied()).collect();
```

crates/spur-sched/src/cons_tres.rs
```diff
@@ -66,6 +66,29 @@ impl NodeAllocation {
...
+        self.gpus = resources.gpus.clone();
+        self.gpu_allocated = self
+            .gpus
+            .iter()
+            .map(|g| allocated_ids.contains(&g.device_id))
```
yansun1996, line 88: `device_id` isn't stable identity — `assign_device_ids` hands out `0..N` positionally over the BDF-sorted registry, so a device added with a lower BDF renumbers everything above it and a partition switch re-enumerates wholesale. It isn't only removals; growth renumbers too. Steps compound it, re-resolving cached ids through the swapped registry into `build_job_injection_plans`, so a running job's `srun` step can bind to the wrong device. Only applying when nothing is allocated avoids needing a durable key here; PCI BDF would be the key if you want per-job precision later.

crates/spur-sched/src/cons_tres.rs
```diff
@@ -326,6 +349,43 @@ mod tests {
         assert_eq!(node.free_gpus(None), 8);
     }
 
+    #[test]
+    fn test_update_capacity_grows_and_preserves_allocations() {
+        // Start with 4 GPUs, allocate two of them, then an inventory refresh
+        // reveals 8 (e.g. SPX->CPX). The new GPUs become schedulable while the
+        // two already allocated stay allocated (by device id).
```
yansun1996, line 356: This covers growth on an idle node — the safe case. The gap is the allocated one: if the refresh becomes gated, assert a confirmed change is a local no-op while a job holds a device and applies once it releases. (Minor: `assert_eq!(node.free_memory_mb(), 256_000 - node.allocated_memory_mb)` reduces to asserting `total_memory_mb == 256_000`; asserting that directly reads more clearly.)

crates/spur-sched/src/cons_tres.rs
```diff
@@ -66,6 +66,29 @@ impl NodeAllocation {
...
     }
 
+    /// Update total capacity from a re-discovered inventory, preserving current
+    /// allocations. CPUs/memory resize in place; GPUs are re-indexed by device
+    /// id so existing GPU allocations follow their device. A device that
+    /// vanished while allocated stops counting toward the total but its owning
+    /// job is untouched (it releases normally). Without this the agent's local
+    /// capacity stays frozen at startup and rejects launches for GPUs that only
+    /// appeared later (e.g. an SPX->CPX partition change), even after the
+    /// controller has learned about them via re-register.
```
yansun1996, line 76: This runs to 8 lines where the repo keeps comments to 1-2 lines of *why*, and the last sentence narrates the fix — same text is repeated on `allocation_handle()`. Two claims are also inaccurate: the id follows the slot, not the device; and while the release itself is fine, the vanish-then-return path can hand the device to a second job first. Same length note for `main.rs` 425-431 and 440-443 (the latter reflowed into a trailing-comment continuation at column 34), `reporter.rs` 34-36, `agent_server.rs` 734-737, and the test comment at `cons_tres.rs` 354-356.

crates/spurd/src/reporter.rs
```diff
@@ -576,6 +594,44 @@ mod tests {
         assert_eq!(resources.generic.get("bandwidth:lustre"), Some(&4096));
     }
 
+    #[test]
+    fn update_resources_reports_only_real_changes() {
```
yansun1996, line 598: This only varies `cpus`. Since change detection now drives a controller-side write, cases varying `gpus` alone and `generic` alone would be worth adding — that's where an unstable discovery order would show up as a re-register every tick.

yansun1996 (commented):
Thanks — #800 is a real gap and the shape here is right. One suggestion plus a few notes.

**Could the apply be gated on whether anything is allocated?**

| change seen | action |
|---|---|
| discovery errored | skip the tick |
| seen once | wait for a second identical reading |
| confirmed, nothing allocated | swap registry, update capacity, re-register |
| confirmed, something allocated | drain; don't apply, don't report new totals |

Row 4 self-converges: draining stops new work, the jobs finish, the next tick takes row 3. So there's never an apply-while-running path — which removes most of the notes below instead of needing fixes for them, and collapses `update_capacity` to a rebuild from the fresh `ResourceSet`.

Why not apply under a live allocation: the controller can't reconcile it. `NodeUpdate` sets `total_resources` and stops — `alloc_resources` isn't pruned, so `can_satisfy_with_allocated` stops subtracting the vanished devices and over-reports free GPUs.

That's a missing evidence channel, not a missing calculation. The agent only sends `RunningJobStatus { job_id, state, exit_code }`, so the controller can only diff totals. The end state worth building toward is the agent re-collecting what each running job still holds and the controller reconciling per job — prune what's genuinely gone, keep the rest, leave healthy jobs alone. The agent already has the data (`NodeAllocation.owners` is `job_id -> {cpu_ids, gpu_ids, memory_mb}`), just no wire format for it. Until then drain is the placeholder — so it'd help to keep detection separate from policy here, e.g. `refresh_once()` returning a classified outcome and a small policy fn acting on it. That also makes both paths testable.

**Drain and auto-recovery need no new plumbing.** `register_node` already sees the resource diff and `alloc_resources`, and `cluster.drain_node()` returns `Draining` when jobs are running. Set it as a system drain (`admin_locked = false`) and lift it when the cause clears, the way `check_node_health` recovers a node from Down — an operator's own drain stays put. Idle nodes never drain; busy ones resume themselves.

**Labels (out of diff).** `register()` sends labels captured at startup. Under default permissive auth the agent is treated as privileged, so a re-register silently reverts an admin relabel; under `auth.mode = required` it fails `PermissionDenied`, and `reregister_pending` never clears — a warning every 60s indefinitely.

**Admission (out of diff).** `validate_admission` runs on every `register_agent`. Harmless under the default `open` mode, but on a token-mode cluster an expired join token means re-registration can never succeed. A bounded backoff would cover both this and the labels case.

**Upgrade compat is clean.** No proto change, no new persisted field or WAL variant; `NodeUpdate` preserves `Drain`/`Down`, `admin_locked`, `agent_start_time`. Raft growth is bounded — discovery ordering is deterministic. One edge: `discover_memory_mb` returns `0` on a `/proc/meminfo` read failure, which would flap and write every tick.

sajmera-pensando (dismissed):
_(withdrawn)_

crates/spur-sched/src/cons_tres.rs
```diff
@@ -66,6 +66,29 @@ impl NodeAllocation {
...
+    pub fn update_capacity(&mut self, resources: &ResourceSet) {
+        self.total_cpus = resources.cpus;
+        self.allocated_cpus.resize(resources.cpus as usize, false);
```
sajmera-pensando: **Critical: CPU shrink via `resize()` silently drops live allocation records.**

`Vec::resize` truncates the bitmap when `resources.cpus` is smaller than the current length. If any truncated slot holds `true` (a job is running on that core), its `AllocationResult` in `self.owners` still contains the now-out-of-range `cpu_ids`. A subsequent `release_job` calls `self.allocated_cpus.get_mut(cpu as usize)` on those indices, gets `None`, and silently skips them — the bitmap is never cleared and `allocated_memory_mb` is never decremented.

Before resizing, scan `self.owners` for entries with `cpu_ids >= resources.cpus` and either release them or emit a `warn!`. At minimum:

```rust
let new_cpu_count = resources.cpus as usize;
if new_cpu_count < self.allocated_cpus.len() {
    let affected: Vec<_> = self.owners.iter()
        .filter(|(_, a)| a.cpu_ids.iter().any(|&c| c as usize >= new_cpu_count))
        .map(|(id, _)| *id)
        .collect();
    for job_id in affected {
        warn!(job_id, "shrinking CPU capacity evicts allocated CPUs; releasing job");
        self.release_job(job_id);
    }
}
self.allocated_cpus.resize(new_cpu_count, false);
```

crates/spur-sched/src/cons_tres.rs
```diff
@@ -66,6 +66,29 @@ impl NodeAllocation {
...
+        self.total_memory_mb = resources.memory_mb;
```
sajmera-pensando: **Critical: `allocated_memory_mb` not clamped after `total_memory_mb` shrinks.**

After this line, `allocated_memory_mb` may exceed `total_memory_mb`. `free_memory_mb()` uses `saturating_sub` and silently returns 0, so the node reports no free memory — but `allocated_memory_mb` is permanently inflated, blocking future allocations on this node even after running jobs finish.

```suggestion
        self.total_memory_mb = resources.memory_mb;
        self.allocated_memory_mb = self.allocated_memory_mb.min(self.total_memory_mb);
```

crates/spur-sched/src/cons_tres.rs
```diff
@@ -66,6 +66,29 @@ impl NodeAllocation {
...
+
+        let allocated_ids: std::collections::HashSet<u32> =
+            self.allocated_gpu_ids().into_iter().collect();
+        self.gpus = resources.gpus.clone();
+        self.gpu_allocated = self
+            .gpus
+            .iter()
+            .map(|g| allocated_ids.contains(&g.device_id))
+            .collect();
```
sajmera-pensando: **Major: Vanished GPU device_ids preserved in `owners` with no log entry.**

When a GPU drops from the PCIe bus, `self.gpus` is replaced with the new list but `self.owners` still retains `AllocationResult` entries referencing the vanished `device_id`. The agent never kills or requeues the affected job, and no warning is emitted — operators have no signal that a live job's GPU assignment is now invalid.

After updating `self.gpu_allocated`, compute the vanished set and warn:

```rust
let new_device_ids: std::collections::HashSet<u32> =
    resources.gpus.iter().map(|g| g.device_id).collect();
let vanished: Vec<_> = self.owners.iter()
    .flat_map(|(&jid, a)| {
        a.gpu_ids.iter()
            .filter(|&&d| !new_device_ids.contains(&d))
            .map(move |&d| (jid, d))
            .collect::<Vec<_>>()
    })
    .collect();
if !vanished.is_empty() {
    warn!(jobs = ?vanished, "allocated GPU(s) vanished from inventory; manual intervention may be required");
}
```

crates/spur-sched/src/cons_tres.rs
```diff
@@ -326,6 +349,43 @@ mod tests {
         assert_eq!(node.free_gpus(None), 8);
     }
 
+    #[test]
+    fn test_update_capacity_grows_and_preserves_allocations() {
```
sajmera-pensando: **Minor: No test for the shrink path — the primary bug scenario has zero coverage.**

The only `update_capacity` test exercises the grow path (4 → 8 GPUs). Add a companion test that:
1. Allocates a job on CPUs [4, 5, 6, 7] of an 8-CPU node.
2. Calls `update_capacity` with `cpus=4`.
3. Asserts the node is in a consistent state (either the job was released, or its `cpu_ids` were clamped).
4. Verifies a subsequent `release_job` leaves `free_cpus() == 4` and does not panic.

A memory-shrink companion test (`allocated_memory_mb` non-zero, `total_memory_mb` reduced) should accompany it.

crates/spurd/src/main.rs
```diff
@@ -422,6 +422,51 @@ async fn main() -> anyhow::Result<()> {
...
+    // scheduling against hardware that no longer exists. On a change: swap the
+    // shared registry (so injection uses the new set), resize the local
+    // allocation capacity (so this node accepts launches for devices that
+    // appeared), and re-register with the controller.
+    {
+        let reporter = reporter.clone();
+        let registry = registry.clone();
+        let config = config.clone();
+        let allocation = agent_service.allocation_handle();
+        tokio::spawn(async move {
```
sajmera-pensando: **Major: `JoinHandle` is dropped immediately — a panicking refresh task is silently swallowed.**

If the task panics (poisoned lock, `unwrap` inside `update_capacity`, etc.), tokio logs to stderr and the task disappears. The node then stops refreshing inventory and re-registering forever with no recovery path short of restarting `spurd`.

At minimum, monitor the handle and propagate the panic:

```rust
let refresh_handle = tokio::spawn(async move { /* ... */ });
tokio::spawn(async move {
    if let Err(e) = refresh_handle.await {
        panic!("inventory refresh task exited unexpectedly: {e:?}");
    }
});
```

crates/spurd/src/main.rs
sajmera-pensando: **Major: `MissedTickBehavior::Burst` causes back-to-back re-registrations after a slow iteration.**

If one loop iteration takes longer than 60 s (controller outage, slow KFD scan), tokio queues all missed ticks and fires them back-to-back on recovery — causing a burst of consecutive `init_device_registry` + `register()` calls against a controller that is already under stress.

```suggestion
            let mut ticker = tokio::time::interval(std::time::Duration::from_secs(60));
            ticker.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
```

crates/spurd/src/main.rs
sajmera-pensando: **Major: Blocking I/O on a tokio worker thread.**

`init_device_registry` performs CDI spec directory I/O and KFD/ROCm enumeration synchronously inside `tokio::spawn`. On a GPU node with many devices this can take seconds, stalling every other task on the same worker thread.

Move the blocking work to the dedicated blocking thread pool:

```rust
let (rebuilt, fresh) = tokio::task::spawn_blocking({
    let config = config.clone();
    move || {
        let rebuilt = init_device_registry(config.as_ref());
        let fresh = reporter::discover_resources(&rebuilt);
        (rebuilt, fresh)
    }
}).await.expect("device discovery panicked");
```

crates/spurd/src/main.rs
```diff
@@ -422,6 +422,51 @@ async fn main() -> anyhow::Result<()> {
...
+                                 // `update_resources` commits the new inventory to the reporter as it
+                                 // detects the change, so a re-register that then fails would never be
+                                 // retried (the next tick sees no further change). Track that a
+                                 // re-register is still owed and keep trying until it lands.
+            let mut reregister_pending = false;
+            loop {
+                ticker.tick().await;
+                let rebuilt = init_device_registry(config.as_ref());
+                let fresh = reporter::discover_resources(&rebuilt);
+                let changed = reporter.update_resources(fresh.clone());
```
sajmera-pensando: **Major: `reporter.resources` updated before `allocation.update_capacity()` — brief window of inconsistency.**

`reporter.update_resources(fresh.clone())` (this line) commits the new `ResourceSet` into the `RwLock` visible to gRPC handlers. Only then does `allocation.lock().await.update_capacity(&fresh)` (line 452) update the local `NodeAllocation`. Between those two `await` points a concurrent `LaunchJob` RPC can check GPU availability against the stale local `NodeAllocation` while the heartbeat path is already advertising the new inventory to the controller.

Swap the commit order — update local enforcement first, then make the change externally visible:

```rust
                if changed {
                    // Update local enforcement capacity before advertising new
                    // inventory so LaunchJob RPCs never see a gap between the
                    // two views.
                    allocation.lock().await.update_capacity(&fresh);
                    *registry.lock().await = rebuilt;
                    reporter.update_resources(fresh);
                    reregister_pending = true;
                }
```

crates/spurd/src/main.rs
```diff
@@ -422,6 +422,51 @@ async fn main() -> anyhow::Result<()> {
...
     agent_service.start_monitor(args.controller.clone());
 
+    // Periodically re-discover node inventory. It is otherwise frozen at startup,
+    // so a device count that changes out of band (a GPU partition-mode switch, a
+    // GPU dropping off the bus) never reaches the controller and it keeps
+    // scheduling against hardware that no longer exists. On a change: swap the
+    // shared registry (so injection uses the new set), resize the local
+    // allocation capacity (so this node accepts launches for devices that
+    // appeared), and re-register with the controller.
+    {
```
sajmera-pensando: **Minor: 30-line anonymous `tokio::spawn` block should be a named function.**

Extracting the body into `fn spawn_inventory_refresh(reporter, registry, config, allocation)` gives the task a name visible in panic backtraces and `tokio-console`, makes parameter types explicit, and allows the loop body to be unit-tested without constructing the full service stack.

crates/spurd/src/reporter.rs
```diff
@@ -31,7 +31,10 @@ impl<T: Send> HeldJobs for Mutex<HashMap<u32, T>> {
 pub struct NodeReporter {
     pub hostname: String,
     pub controller_addr: String,
-    pub resources: ResourceSet,
+    /// Node inventory. Behind a lock so a background refresh can update it live
+    /// (a device count that changes after startup, e.g. a GPU partition-mode
+    /// switch, must reach the controller rather than being frozen at boot).
+    pub resources: std::sync::RwLock<ResourceSet>,
```
sajmera-pensando: **Minor: `std::sync::RwLock` in an async service — consider `tokio::sync::RwLock`.**

The existing comment in `register()` (`// Snapshot the inventory before the await so the lock guard is not held across it`) is evidence of the manual reasoning burden this imposes. `tokio::sync::RwLock` makes the snapshot pattern a compile-time guarantee (the guard is `!Send`, so holding it across `.await` is a compile error) and eliminates the poisoning concern raised below.

The same applies to `node_token: RwLock<String>`.

crates/spurd/src/reporter.rs
```diff
@@ -94,17 +97,32 @@ impl NodeReporter {
         self.held_jobs.held_job_ids()
     }
 
+    /// Replace the reported inventory if it changed; returns whether it did, so
+    /// the caller can re-register the node with the controller.
+    pub fn update_resources(&self, fresh: ResourceSet) -> bool {
```
sajmera-pensando: **Minor: Taking `ResourceSet` by value forces an unconditional clone at the call site.**

The caller in `main.rs` must write `reporter.update_resources(fresh.clone())` because `fresh` is still needed afterwards for `update_capacity`. Taking `&ResourceSet` and cloning only when the value actually changes eliminates the unconditional allocation:

```suggestion
    pub fn update_resources(&self, fresh: &ResourceSet) -> bool {
```

Inside the function, change the assignment to `*cur = fresh.clone();`. The call site becomes `reporter.update_resources(&fresh)`.

crates/spurd/src/reporter.rs
```diff
@@ -94,17 +97,32 @@ impl NodeReporter {
...
+        let mut cur = self.resources.write().unwrap();
```
sajmera-pensando: **Major: `std::sync::RwLock` poison propagates to gRPC handlers via `unwrap()`.**

A panic inside this write critical section poisons the lock. Every subsequent `.unwrap()` on `resources` — including in `register()` (line 121), `heartbeat()`, and `get_node_resources` in `agent_server.rs` — will itself panic, cascading the failure to all request handlers.

Either recover from poison at every lock site:
```rust
let mut cur = self.resources.write().unwrap_or_else(|e| e.into_inner());
```
or — better — switch to `tokio::sync::RwLock<ResourceSet>` throughout, which has no poisoning concept and is the idiomatic choice for state shared across async tasks.

crates/spurd/src/agent_server.rs
```diff
@@ -2124,9 +2132,9 @@ impl SlurmAgent for AgentService {
         &self,
         _request: Request<()>,
     ) -> Result<Response<NodeResourcesResponse>, Status> {
-        let resources = &self.reporter.resources;
+        let resources = self.reporter.resources.read().unwrap();
```
sajmera-pensando: **Minor: `std::sync::RwLockReadGuard` held across the full async fn body — fragile pattern.**

Currently safe because there is no `.await` between lock acquisition and return. However, `std::sync::RwLockReadGuard` is `!Send`, so any future change that adds an `.await` in this function body will either fail to compile or (with unsafe) introduce a data race. Scope the guard explicitly:

```rust
        let node_resources = {
            let r = self.reporter.resources.read().unwrap_or_else(|e| e.into_inner());
            crate::reporter::resource_to_proto(&r)
        };
        Ok(Response::new(NodeResourcesResponse {
            total: Some(node_resources),
            used: Some(crate::reporter::allocations_to_proto(
                &spur_core::resource::ResourceAllocations::default(),
            )),
        }))
```

sajmera-pensando (changes requested):
Requesting changes — details in the review comment above. The main blocker is the positional `device_id`: when the device set changes mid-run, surviving allocations on healthy GPUs can be mis-mapped (an in-use GPU shown free → double-schedulable). Also worth addressing before merge: no stable device identity, wholesale swap vs. diff-and-classify, and no explicit policy (warn/fail/requeue) for a job whose device vanishes.

shiv-tyagi:
> The main blocker is the positional device_id: when the device set changes mid-run, surviving allocations on healthy GPUs can be mis-mapped (an in-use GPU shown free → double-schedulable)

+1 on this

sajmera-pensando:
Closing this PR in favor of https://github.com/ROCm/spur/pull/898

PR #836, opened 2026-09-06
feat: node health checks as controller-submitted jobs

powderluv:
## Node health checks (spur#801)

Adds operator-authored node health checks. The **controller submits each check as a real job** that requests its node exclusively, so the scheduler runs it only when the node is idle — "probe only when idle" falls out of normal scheduling with no special-case code. Checks are visible in `squeue`, run in a cgroup, and get the usual job environment.

A **failing check drains the node** with the check as the reason — a sticky, operator-recoverable drain, like Slurm's `HealthCheckProgram`. If a check is starved because the node stays busy past `max_wait_secs`, the node is soft-drained so it empties, then resumed so the pending check lands ("drain, then resume-when-empty"). The two drain kinds are distinguished by the persisted `state_reason` prefix so they survive a leader failover; in-flight detection and the `max_wait` clock read from the durable job store, and only the per-node interval timestamp is leader-local.

> **Note on the design.** This replaces an earlier spurd-side approach (a raw process run in the completion path with a re-entry gate). Per the review discussion, that ran with no cgroup, was invisible to the controller, raced other free slots on a partially-idle node, and blocked the completion report. Making the check a scheduled job fixes those structurally, and the per-job re-entry gate is dropped entirely — per-job cleanup stays the epilog's job; health checks are node-level hardware verification.

### Configuration

```toml
[[health.checks]]
program = "/etc/spur/node-health.sh"   # runs as the job's command
nodes = 1                               # 1 = per-node (multi-node is a follow-up)
interval_secs = 300                     # re-run on each node about this often
timeout_secs = 60                       # kill + fail the check after this long
max_wait_secs = 600                     # soft-drain a busy node to run the check (0 = wait for idle)
user = "nobody"                         # identity the check runs as; a privileged probe needs
uid = 65534                             # root (uid/gid 0) AND allow_root_jobs on the agents
gid = 65534
```

Checks default to an **unprivileged** identity. Root probes need root `uid`/`gid` **and** `allow_root_jobs` on the agents — the agent doesn't authenticate its caller, so a name-based root exemption would be a spoofable escalation.

### Changes

- **spur-core config:** scalar `[health]` → `[[health.checks]]` array (per-check `program`/`nodes`/`interval_secs`/`timeout_secs`/`max_wait_secs` + `user`/`uid`/`gid`).
- **spurctld:** `run_node_health_pass` (scheduler-loop hook), `submit_health_job` (trusted internal submit, exclusive + node-pinned), and a finalization hook that drains on failure; the user submit path rejects the reserved job-name prefix so a job can't spoof a health check.
- **spurd:** removed `health.rs`, the monitor-loop re-entry gate, the periodic loop, `NodeReporter::drain_node`, and the agent `health` field/param.
- **tests:** config parse/default tests, 5 controller unit tests, and a rewritten E2E (failing check drains an idle node; passing check leaves it schedulable) — validated end-to-end on real spurctld+spurd.

### Testing

- `cargo build` / clippy / fmt clean.
- Full suites green: spurctld 1025, spurd 276; 5 new controller unit tests + 2 config tests.
- E2E (`test_health_check.py`) passes end-to-end on real spurctld+spurd.

### Follow-ups

- Multi-node (`nodes > 1`) fabric checks — topology-aware node pairing, gated on spur#800. The pass currently warns-and-skips `nodes != 1`.


sajmera-pensando (dismissed):
The redesign is solid — health checks as real controller-submitted exclusive jobs is the right model, and the implementation is faithful: probe-only-when-idle falls out of scheduling, visibility in squeue, sticky operator-recoverable drain, the concurrency cap bounding the wave on restart/first-enable, absolute-path validation, and 126/127 treated as deploy errors. Unit and e2e coverage is substantially complete. Two minor e2e gaps (cap end-to-end, cancelled-check) noted in comments for follow-up.

PR #839, opened 2026-09-07
feat(spurd)!: make the host-ROCm container library overlay opt-in

### LINKED ISSUE #779: AMD CDI injection silently overlays host ROCm libraries onto every GPU container
## Summary

**This is the default behavior on every AMD GPU container job, and it is silent.** The image's ROCm userspace is replaced by the host's with nothing at default log level, no warning, and no per-job opt-out.

The ROCm-documented way to run these images on bare metal — `docker run --device=/dev/kfd --device=/dev/dri --group-add video` — passes the device nodes and nothing else, leaving the image's userspace intact. Spur silently diverges from that on the same hardware with the same image.

When AMD GPUs are discovered by auto-detection, Spur's generated shared CDI edits bind-mount the host's `/opt/rocm/lib` and `/opt/rocm/lib64` over the same paths inside every GPU container. The image's own ROCm userspace is hidden at those paths, and there is no per-job or per-image way to keep it.

## Code evidence (`b5c6ece`)

`crates/spur-devices/src/cdi/discovery.rs::build_shared_edits`:

```rust
for lib_path in &["/opt/rocm/lib", "/opt/rocm/lib64"] {
    if Path::new(lib_path).is_dir() {
        edits.mounts.push(Mount {
            host_path: lib_path.to_string(),
            container_path: lib_path.to_string(),
            options: Some(vec!["ro", "nosuid", "nodev", "bind"]),
            ..
        });
    }
}
```

These edits are attached to the auto-generated AMD CDI spec (`discover_to_cdi`) and applied after the rootfs is prepared, so they land on top of the image's directories.

Note the contrast with the GRES-based path, `crates/spur-devices/src/gres/entry.rs::build_shared_edits`, which injects only `/dev/kfd` and no libraries. The library overlay is specific to CDI auto-detection, which is the default (`DevicesConfig::auto_detect = true` in `crates/spur-core/src/config.rs`).

## This is not a version swap, it is a split installation

Only `lib` and `lib64` are mounted, so on a standard ROCm layout the host supplies:

- the HIP and HSA runtimes, RCCL, and `comgr`;
- the math libraries **and their kernel/tuning databases** — `lib/rocblas/library`, `lib/hipblaslt/library`, `lib/MIOpen`;
- the profiler and SMI libraries, `librocprofiler64` and `librocm_smi64`.

while the image still supplies:

- `bin/` — `hipcc`, `rocprofv3`, `amd-smi`, `rocminfo`, `hipconfig`;
- `include/` — the headers anything built inside the job compiles against;
- `llvm/` and `amdgcn/bitcode` — the compiler and device bitcode;
- `.info/version`;
- Python packages, including the ROCm libraries that PyTorch wheels bundle under `site-packages/torch/lib`.

The compiler, the device bitcode, and the runtime can therefore be three different versions inside one process. `rocprofv3` from the image runs against the host's `librocprofiler64`. A torch wheel resolving through its own RPATH can keep image libraries while anything that `dlopen`s out of `/opt/rocm/lib` gets the host's. A HIP or torch extension built inside the job compiles against the image's headers and links against the host's libraries.

The performance consequence is worth stating separately, because it is silent: rocBLAS, hipBLASLt, and MIOpen tuning data comes from the host, so a benchmark run under Spur measures the host's tuning rather than the tuning the image pins. That is precisely what kernel and collective-communication optimization work is trying to measure.

## The mismatch is invisible to the obvious check

`/opt/rocm/.info/version` sits outside `lib/`, so it still reports the **image's** version while the libraries beneath it are the host's. The version file effectively lies. The only reliable in-job check is the mount table:

```bash
grep rocm /proc/self/mountinfo
```

The mount is also conditional on the host having the directory (`Path::new(lib_path).is_dir()`), so on a fleet where some nodes carry a host ROCm install and some do not, one job can run image libraries on some ranks and host libraries on others.

## Interaction with #649

Nothing in `crates/` sets `LD_LIBRARY_PATH`, `ROCM_PATH`, or `HIP_PATH`, and per #649 container jobs launch with the host environment instead of the image's `config.Env`, so the image's own resolution hints are discarded as well. Together these mean neither the libraries nor the search path are the image's, and a job cannot determine what it is actually running by inspection.

## Why this is not the same as device-node injection

The kernel driver and the `/dev/kfd` / `/dev/dri` device nodes must come from the host. It does not follow that the image's ROCm userspace libraries should be replaced. Standard ROCm containers deliberately pin a tested userspace stack against the host's kernel driver interface; that is the supported ROCm container model.

## The only current escape hatch is site-wide

Per `crates/spur-devices/src/cdi/cache.rs::load`, auto-detection runs only when no on-disk CDI specs are found:

```rust
if cache.is_empty() && auto_detect {
    cache.add_specs(&crate::cdi::discovery::discover_to_cdi());
} else if auto_detect {
    warn!("on-disk CDI specs found; auto_detect ignored");
}
```

So the overlay can be avoided only by an administrator hand-authoring CDI specs (or switching to GRES pools) for every GPU node, which disables AMD auto-detection entirely and must then track device topology by hand. There is no job-level or image-level opt-out, and a user on a shared cluster has none at all.

## Development impact

Miles' ROCm image pins patched versions of PyTorch, SGLang, Megatron-LM, and the ROCm userspace stack:

https://github.com/radixark/miles/blob/main/docker/Dockerfile.rocm

Loader failures and missing symbols are the good case, because they are loud. The bad cases are silent: different kernel selection, different collective algorithms, changed RCCL behavior, and throughput differences that read as model or kernel regressions. The full blast radius is not enumerable in advance, because it is a function of host version crossed with image version — which is the argument for not doing the substitution by default rather than for cataloguing its effects.

## Expected behavior

Separate AMD device injection from userspace-library injection:

- inject `/dev/kfd`, the allocated `/dev/dri/*` nodes, and the required supplementary groups, and leave the image's libraries alone. This should be the default, matching what `docker run` does on the same hardware;
- for sites that genuinely need the host stack, put it behind an explicit opt-in rather than applying it unconditionally;
- log the injected mounts in the job record, so a substitution is visible without inspecting the mount table at runtime.

This should not turn into a version-compatibility gate that can refuse to start a job. The ask is that Spur stop rewriting the image by default, not that it add a new check in front of it.

## Suggested coverage

Build an image with a marker library under `/opt/rocm/lib`, run it on a host carrying a different marker, and assert the selected policy deterministically exposes the intended version. Add a multi-node test asserting all ranks report the same ROCm/RCCL build identity, including on a fleet where only some hosts have `/opt/rocm` installed.

## Decision requested

Confirm whether overlaying the host's ROCm userspace onto every GPU container is intended.

If it is intended, it needs to be documented where users choosing an image will see it, and the injected mounts need to be logged per job so a version mismatch is diagnosable rather than invisible.

If it is not intended, the library mounts should be separated from device-node injection and put behind an explicit policy, with the image's libraries kept by default.

Either way, silently substituting a different ROCm userspace than the image pins, with no record in the job log, should not remain the default on AMD hardware.

powderluv:
## Summary

Fixes #779. Auto-detected AMD GPU containers bind-mounted the host's `/opt/rocm/lib{,64}` over the same paths inside **every** container, silently replacing the image's ROCm userspace with the host's — no warning, no per-job opt-out. `docker run` on the same hardware leaves the image's libraries in place. Overlaying a different runtime/math-library/tuning stack than the image pins is a silent version split (the compiler, device bitcode, and runtime can end up three different versions in one process; benchmark tuning data comes from the host, not the image).

The library overlay is now gated behind **`[devices] overlay_host_rocm_libs`, default off**. Device nodes (`/dev/kfd`, `/dev/dri/*`) and the `video`/`render` supplementary groups are still injected unconditionally — only the library bind mounts are gated. When mounts are injected, spurd logs them in the job's log so a substitution is visible without reading `/proc/self/mountinfo`.

## Changes

- `spur-core/config.rs`: `DevicesConfig.overlay_host_rocm_libs` (default false) + defaults/opt-in test.
- `spur-devices/cdi/{discovery,cache}.rs`: `discover_to_cdi`/`build_shared_edits`/`CdiCache::load` thread the flag; the library mounts are gated, device nodes stay independent (+ test).
- `spurd/main.rs`: passes the flag; `cluster.rs`: the k0s/containerd CDI path passes `false` (no overlay), matching the default.
- `spurd/container.rs`: logs injected library mounts per job.
- `examples/spur.conf`: documented `[devices]` section.

## Testing

- Unit: spur-devices 116, spur-core config 83 pass; clippy/fmt clean.
- **GPU-host E2E (`test_rocm_overlay.py`), validated on shark-a's gfx1201 (amdgpu + KFD):** a marker-library image run through a GPU container both ways. Overlay off (default) keeps the image's `/opt/rocm/lib` and mounts nothing over it; overlay on bind-mounts the host's over the image's (marker gone, mount visible). GPU-gated (skips without KFD + `/opt/rocm`).

### DIFF
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -1242,17 +1242,10 @@ pub struct DevicesConfig {
     #[serde(default)]
     pub gres: Vec<DevicesGresEntry>,
 
-    /// Bind-mount the host's `/opt/rocm/lib{,64}` over the same paths inside
-    /// every auto-detected AMD GPU container, replacing the image's ROCm
-    /// userspace with the host's. Off by default: `docker run` on the same
-    /// hardware leaves the image's libraries in place, and overlaying a
-    /// different runtime/math-library/tuning stack than the image pins is a
-    /// silent version split. Device nodes (`/dev/kfd`,
-    /// `/dev/dri/*`) and the GPU supplementary groups are injected regardless;
-    /// this flag governs only the library overlay. It applies only to the
-    /// auto-detected AMD spec, so it is inert on a node that has any on-disk CDI
-    /// spec (auto-detection is skipped there), and it does not affect k0s/pod
-    /// containers, whose CDI spec never carries the overlay.
+    /// Bind-mount the host's `/opt/rocm/lib{,64}` over the image's inside
+    /// auto-detected AMD GPU containers; off by default. Inert when on-disk CDI
+    /// specs exist, and never applied to managed-k0s pods. Device nodes and GPU
+    /// groups are injected regardless.
     #[serde(default)]
     pub overlay_host_rocm_libs: bool,
 }
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -1247,9 +1247,12 @@ pub struct DevicesConfig {
     /// userspace with the host's. Off by default: `docker run` on the same
     /// hardware leaves the image's libraries in place, and overlaying a
     /// different runtime/math-library/tuning stack than the image pins is a
-    /// silent version split (see spur#779). Device nodes (`/dev/kfd`,
+    /// silent version split. Device nodes (`/dev/kfd`,
     /// `/dev/dri/*`) and the GPU supplementary groups are injected regardless;
-    /// this flag governs only the library overlay.
+    /// this flag governs only the library overlay. It applies only to the
+    /// auto-detected AMD spec, so it is inert on a node that has any on-disk CDI
+    /// spec (auto-detection is skipped there), and it does not affect k0s/pod
+    /// containers, whose CDI spec never carries the overlay.
     #[serde(default)]
     pub overlay_host_rocm_libs: bool,
 }
@@ -2844,8 +2847,8 @@ job_submit_lua = "/etc/spur/job_submit.lua"
 
     #[test]
     fn test_devices_host_rocm_overlay_defaults_off() {
-        // The host-ROCm library overlay must be off by default (spur#779): a job
-        // gets the image's userspace unless the operator opts in.
+        // The host-ROCm library overlay must be off by default: a job gets the
+        // image's userspace unless the operator opts in.
         let config = SlurmConfig::load_from_str(r#"cluster_name = "x""#).unwrap();
         assert!(!config.devices.overlay_host_rocm_libs);
         assert!(
--- a/crates/spur-core/src/config.rs
+++ b/crates/spur-core/src/config.rs
@@ -1241,6 +1241,17 @@ pub struct DevicesConfig {
     /// File-based or countable GRES pools alongside CDI-discovered devices.
     #[serde(default)]
     pub gres: Vec<DevicesGresEntry>,
+
+    /// Bind-mount the host's `/opt/rocm/lib{,64}` over the same paths inside
+    /// every auto-detected AMD GPU container, replacing the image's ROCm
+    /// userspace with the host's. Off by default: `docker run` on the same
+    /// hardware leaves the image's libraries in place, and overlaying a
+    /// different runtime/math-library/tuning stack than the image pins is a
+    /// silent version split (see spur#779). Device nodes (`/dev/kfd`,
+    /// `/dev/dri/*`) and the GPU supplementary groups are injected regardless;
+    /// this flag governs only the library overlay.
+    #[serde(default)]
+    pub overlay_host_rocm_libs: bool,
 }
 
 impl Default for DevicesConfig {
@@ -1249,6 +1260,7 @@ impl Default for DevicesConfig {
             auto_detect: true,
             cdi_spec_dirs: Vec::new(),
             gres: Vec::new(),
+            overlay_host_rocm_libs: false,
         }
     }
 }
@@ -2828,6 +2840,24 @@ job_submit_lua = "/etc/spur/job_submit.lua"
         // hooks section omitted — metrics should keep defaults
         assert!(config.metrics.enabled);
         assert_eq!(config.metrics.listen_addr, "[::]:6822");
+    }
+
+    #[test]
+    fn test_devices_host_rocm_overlay_defaults_off() {
+        // The host-ROCm library overlay must be off by default (spur#779): a job
+        // gets the image's userspace unless the operator opts in.
+        let config = SlurmConfig::load_from_str(r#"cluster_name = "x""#).unwrap();
+        assert!(!config.devices.overlay_host_rocm_libs);
+        assert!(
+            config.devices.auto_detect,
+            "auto_detect stays on by default"
+        );
+
+        let opted_in = SlurmConfig::load_from_str(
+            "cluster_name = \"x\"\n[devices]\noverlay_host_rocm_libs = true\n",
+        )
+        .unwrap();
+        assert!(opted_in.devices.overlay_host_rocm_libs);
     }
 
     #[test]
--- a/crates/spur-devices/src/cdi/discovery.rs
+++ b/crates/spur-devices/src/cdi/discovery.rs
@@ -22,8 +22,8 @@ const IO_LINK_XGMI: u32 = 2;
 const GPU_SUPPLEMENTARY_GROUPS: [&str; 2] = ["video", "render"];
 
 /// Build the auto-detected AMD CDI spec. `overlay_host_rocm_libs` decides
-/// whether the host's `/opt/rocm/lib{,64}` is bind-mounted over the image's
-/// (spur#779); device nodes and GPU groups are injected regardless.
+/// whether the host's `/opt/rocm/lib{,64}` is bind-mounted over the image's;
+/// device nodes and GPU groups are injected regardless.
 pub fn discover_to_cdi(overlay_host_rocm_libs: bool) -> Vec<CdiSpec> {
     let gpus = discover_amd_gpus();
     if gpus.is_empty() {
@@ -361,30 +361,39 @@ fn build_shared_edits(overlay_host_rocm_libs: bool) -> ContainerEdits {
         });
     }
 
-    // The host-ROCm library overlay is opt-in (spur#779). Default off keeps the
-    // image's own userspace, matching `docker run --device=/dev/kfd ...`;
-    // overlaying the host's runtime/math/tuning libraries over the image's is a
-    // silent version split. Device nodes and GPU groups are injected either way.
-    if overlay_host_rocm_libs {
-        for lib_path in &["/opt/rocm/lib", "/opt/rocm/lib64"] {
-            if Path::new(lib_path).is_dir() {
-                edits.mounts.push(Mount {
-                    host_path: lib_path.to_string(),
-                    container_path: lib_path.to_string(),
-                    r#type: None,
-                    options: Some(vec![
-                        "ro".into(),
-                        "nosuid".into(),
-                        "nodev".into(),
-                        "bind".into(),
-                    ]),
-                });
-            }
-        }
-    }
-
+    edits.mounts = rocm_lib_overlay_mounts(overlay_host_rocm_libs, ROCM_LIB_DIRS);
     edits.additional_gids = gpu_supplementary_gids_from_group_file("/etc/group");
     edits
+}
+
+/// Host ROCm library directories overlaid into a GPU container, when enabled.
+const ROCM_LIB_DIRS: &[&str] = &["/opt/rocm/lib", "/opt/rocm/lib64"];
+
+/// Read-only bind mounts overlaying the host's ROCm libraries onto a container,
+/// or empty when the overlay is off. Opt-in: the default keeps the image's own
+/// userspace (matching `docker run --device=/dev/kfd ...`); overlaying the
+/// host's runtime/math/tuning libraries is a silent version split. Only
+/// directories that exist on the host are mounted. Split out from
+/// `build_shared_edits` so the gate is testable without a real `/opt/rocm`.
+fn rocm_lib_overlay_mounts(overlay_host_rocm_libs: bool, candidates: &[&str]) -> Vec<Mount> {
+    if !overlay_host_rocm_libs {
+        return Vec::new();
+    }
+    candidates
+        .iter()
+        .filter(|p| Path::new(p).is_dir())
+        .map(|lib_path| Mount {
+            host_path: lib_path.to_string(),
+            container_path: lib_path.to_string(),
+            r#type: None,
+            options: Some(vec![
+                "ro".into(),
+                "nosuid".into(),
+                "nodev".into(),
+                "bind".into(),
+            ]),
+        })
+        .collect()
 }
 
 fn lookup_group_gid_in(group_file: &str, name: &str) -> Option<u32> {
@@ -817,30 +826,28 @@ mod tests {
     }
 
     #[test]
-    fn test_shared_edits_library_overlay_is_gated() {
-        // The overlay flag governs only the /opt/rocm library bind mounts, and
-        // only when the host actually has those directories.
-        let host_has_rocm = std::path::Path::new("/opt/rocm/lib").is_dir()
-            || std::path::Path::new("/opt/rocm/lib64").is_dir();
+    fn test_rocm_lib_overlay_is_gated() {
+        // Inject a candidate directory that DOES exist (a tempdir) so the test is
+        // falsifiable regardless of whether the host has /opt/rocm: removing the
+        // gate would make the "off" case produce a mount and fail here.
+        let dir = tempfile::tempdir().unwrap();
+        let path = dir.path().to_str().unwrap();
+        let candidates = [path];
 
-        let off = build_shared_edits(false);
+        let off = rocm_lib_overlay_mounts(false, &candidates);
         assert!(
-            off.mounts.is_empty(),
-            "no library mounts must be injected when the overlay is off, got {:?}",
-            off.mounts
+            off.is_empty(),
+            "overlay off must inject no library mounts, got {off:?}"
         );
 
-        let on = build_shared_edits(true);
-        if host_has_rocm {
-            assert!(
-                on.mounts
-                    .iter()
-                    .any(|m| m.host_path.starts_with("/opt/rocm/lib")),
-                "overlay on + host ROCm present must inject the library mounts"
-            );
-        }
-        // Device-node injection is independent of the overlay flag.
-        assert_eq!(off.device_nodes.len(), on.device_nodes.len());
+        let on = rocm_lib_overlay_mounts(true, &candidates);
+        assert_eq!(on.len(), 1, "overlay on must mount the existing candidate");
+        assert_eq!(on[0].host_path, path);
+        assert_eq!(on[0].container_path, path);
+
+        // A candidate that does not exist is never mounted, even when opted in.
+        let missing = rocm_lib_overlay_mounts(true, &["/no/such/rocm/lib"]);
+        assert!(missing.is_empty());
     }
 
     #[test]
--- a/crates/spur-devices/src/cdi/discovery.rs
+++ b/crates/spur-devices/src/cdi/discovery.rs
@@ -361,7 +361,11 @@ fn build_shared_edits(overlay_host_rocm_libs: bool) -> ContainerEdits {
         });
     }
 
-    edits.mounts = rocm_lib_overlay_mounts(overlay_host_rocm_libs, ROCM_LIB_DIRS);
+    // `extend`, not assign, so a mount added above is never silently dropped.
+    edits.mounts.extend(rocm_lib_overlay_mounts(
+        overlay_host_rocm_libs,
+        ROCM_LIB_DIRS,
+    ));
     edits.additional_gids = gpu_supplementary_gids_from_group_file("/etc/group");
     edits
 }
@@ -371,10 +375,8 @@ const ROCM_LIB_DIRS: &[&str] = &["/opt/rocm/lib", "/opt/rocm/lib64"];
 
 /// Read-only bind mounts overlaying the host's ROCm libraries onto a container,
 /// or empty when the overlay is off. Opt-in: the default keeps the image's own
-/// userspace (matching `docker run --device=/dev/kfd ...`); overlaying the
-/// host's runtime/math/tuning libraries is a silent version split. Only
-/// directories that exist on the host are mounted. Split out from
-/// `build_shared_edits` so the gate is testable without a real `/opt/rocm`.
+/// userspace (`docker run --device=/dev/kfd ...`), since overlaying a different
+/// runtime/math/tuning stack is a silent version split. Only existing dirs mount.
 fn rocm_lib_overlay_mounts(overlay_host_rocm_libs: bool, candidates: &[&str]) -> Vec<Mount> {
     if !overlay_host_rocm_libs {
         return Vec::new();
@@ -901,8 +903,19 @@ mod tests {
     }
 
     #[test]
-    fn test_build_shared_edits_kfd() {
-        let _ = build_shared_edits(false);
+    fn test_build_shared_edits_gates_only_the_library_overlay() {
+        // The gate wires through build_shared_edits: off adds no library mounts,
+        // while device nodes and GPU groups are injected either way (both read
+        // the same host state, so the comparison is stable everywhere and only
+        // fails if someone later moves that injection inside the gate).
+        let off = build_shared_edits(false);
+        let on = build_shared_edits(true);
+        assert!(
+            off.mounts.is_empty(),
+            "overlay off must add no library mounts"
+        );
+        assert_eq!(off.device_nodes.len(), on.device_nodes.len());
+        assert_eq!(off.additional_gids, on.additional_gids);
     }
 
     #[test]
--- a/crates/spur-devices/src/cdi/discovery.rs
+++ b/crates/spur-devices/src/cdi/discovery.rs
@@ -21,7 +21,10 @@ const VRAM_HEAP_TYPE: u64 = 1;
 const IO_LINK_XGMI: u32 = 2;
 const GPU_SUPPLEMENTARY_GROUPS: [&str; 2] = ["video", "render"];
 
-pub fn discover_to_cdi() -> Vec<CdiSpec> {
+/// Build the auto-detected AMD CDI spec. `overlay_host_rocm_libs` decides
+/// whether the host's `/opt/rocm/lib{,64}` is bind-mounted over the image's
+/// (spur#779); device nodes and GPU groups are injected regardless.
+pub fn discover_to_cdi(overlay_host_rocm_libs: bool) -> Vec<CdiSpec> {
     let gpus = discover_amd_gpus();
     if gpus.is_empty() {
         return Vec::new();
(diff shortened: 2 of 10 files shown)


crates/spur-devices/src/cdi/discovery.rs
```diff
@@ -803,8 +812,35 @@ mod tests {
...
     #[test]
     fn test_discover_does_not_panic() {
-        let specs = discover_to_cdi();
-        let _ = specs;
+        let _ = discover_to_cdi(false);
+        let _ = discover_to_cdi(true);
+    }
+
+    #[test]
+    fn test_shared_edits_library_overlay_is_gated() {
```
yansun1996: I think this passes on CI even with the gate removed. Three things stack up:

- `build_shared_edits` has exactly one `edits.mounts.push` (line 371, inside the new `if`), so `off.mounts.is_empty()` at 828 holds on any host without `/opt/rocm/lib{,64}` regardless of whether the `if` is there.
- The only assertion that exercises the on-branch (835-840) sits inside `if host_has_rocm` (834), so it's skipped on a CI runner.
- `assert_eq!(off.device_nodes.len(), on.device_nodes.len())` at 843 can't fail — `device_nodes` is populated only from `/dev/kfd` existence at line 351 and never touches the flag.

`AGENTS.md` also asks tests not to depend on the runner's environment. Could the candidate paths be a parameter so both branches are deterministic everywhere?

```rust
const ROCM_LIB_CANDIDATES: [&str; 2] = ["/opt/rocm/lib", "/opt/rocm/lib64"];

fn build_shared_edits(overlay_host_rocm_libs: bool, lib_candidates: &[&str]) -> ContainerEdits
```

Then the test points at a `tempfile::TempDir` containing `lib/` and `lib64/` and asserts `off.mounts.len() == 0` **and** `on.mounts.len() == 2` unconditionally.

While you're there — the PR's other safety promise, that GPU groups are injected regardless of the flag, has no coverage at all. `additional_gids` is set at `discovery.rs:386`, outside the gate, so this one-liner is both true today and falsifiable if someone later moves that line inside the block:

```rust
assert_eq!(off.additional_gids, on.additional_gids);
```

crates/spurd/src/cluster.rs
```diff
@@ -514,7 +514,9 @@ const CDI_SPEC_PATH: &str = "/etc/cdi/amd.json";
 /// k8s-device-plugin; acceptable for Phase-2 containerd injection — a native spur-device-plugin is
 /// a later milestone.
 async fn write_cdi_spec() -> anyhow::Result<()> {
-    let specs = spur_devices::cdi::discovery::discover_to_cdi();
+    // No host-ROCm library overlay for k0s/containerd pods (spur#779): match the
+    // native-container default of leaving the image's userspace in place.
+    let specs = spur_devices::cdi::discovery::discover_to_cdi(false);
```
yansun1996, line 519: Worth being explicit that this makes the setting config-ignored for k0s: an operator who sets `overlay_host_rocm_libs = true` gets it for native containers and silently not for pods, with no warning.

I can see why it's hardcoded — `write_cdi_spec` is a free fn with no `SlurmConfig` in scope, so honoring it needs plumbing. If threading the value through is out of scope here, could the carve-out at least be stated in the field's doc comment and in `examples/spur.conf`? A `warn!` when the operator opted in but this path forces `false` would also close the surprise.

This is also one half of the stale-spec problem in the summary: the spec written here with `false` is what later shadows auto-detect on the native path.

crates/spurd/src/container.rs
```diff
@@ -610,6 +610,17 @@ fn apply_container_device_plan(rootfs: &Path, plan: &spur_devices::inject::Conta
...
     }
 
+    // Record the injected library mounts in the job's log (spur#779): overlaying
+    // host libraries over the image's is a substitution the job can otherwise
+    // only detect by reading /proc/self/mountinfo, so name it where it's visible.
+    if !plan.mounts.is_empty() {
+        let overlaid: Vec<&str> = plan.mounts.iter().map(|m| m.target.as_str()).collect();
+        info!(
+            mounts = ?overlaid,
+            "overlaying host library paths onto the container image (replacing the image's own)"
```
yansun1996: Two things here (the first was raised by the bot review as well):

- `plan.mounts` isn't ROCm-specific. It's `unique_mounts(&edits.device_edits.mounts, &edits.shared_edits.mounts)` (`spur-devices/src/inject.rs:216-232`), merged from *every* matched CDI spec including operator-supplied files in `/etc/cdi`. So "overlaying host library paths … (replacing the image's own)" can be wrong on both halves.
- It's emitted before the loop at 624-638, which `continue`s when the source doesn't exist (627-630) and only `warn!`s on bind failure (631-637) — so it announces mounts that may never be applied.

Moving it below the loop and collecting the targets that actually bound would fix both, with neutral wording like `"bind-mounted injected container paths"`.

For what it's worth, the "job's log" claim in the comment does check out — `JobIoRaw::wire` dup2s the job's stdout onto fd 1 at `executor.rs:1760` before `container_init` at 1779, so this lands in the user's `-o` file. It's incidental to fork+dup2 rather than a designed channel, but it's accurate.

crates/spur-core/src/config.rs
```diff
@@ -1241,6 +1241,17 @@ pub struct DevicesConfig {
     /// File-based or countable GRES pools alongside CDI-discovered devices.
     #[serde(default)]
     pub gres: Vec<DevicesGresEntry>,
+
+    /// Bind-mount the host's `/opt/rocm/lib{,64}` over the same paths inside
```
yansun1996: Two gaps worth closing, since this is the field's canonical documentation:

- It doesn't mention that the flag is inert whenever on-disk CDI specs are present (`cache.rs:75-81`) — which, per the summary, is every node that has run k0s.
- It doesn't mention the k0s/containerd carve-out (`cluster.rs:519`).

Also 8 lines where the sibling fields all use 1-2. Something like:

```rust
/// Bind-mount the host's `/opt/rocm/lib{,64}` over the image's inside
/// auto-detected AMD GPU containers; off by default. Ignored when on-disk CDI
/// specs exist or for managed-k0s pods. Device nodes and GPU groups are
/// injected regardless.
```

tests/native_host/e2e/test_rocm_overlay.py
```diff
@@ -0,0 +1,102 @@
...
+    cluster.container_preflight()
+
+    cluster.start(
+        config_overrides={"devices": {"overlay_host_rocm_libs": overlay_on}},
+        agent_as_root=True,
+    )
+    cluster.gpu_preflight(1)
+
+    img = cluster.build_container_image(tmp_path, rootfs_extra=ROOTFS_EXTRA)
+    probe = cluster.write_file("overlay-probe.sh", PROBE)
```
yansun1996, line 73: `write_file` defaults to node 0 only (`cluster.py:712`, `targets = self.nodes if all_nodes else self.nodes[:1]`), but the job isn't pinned — the args below are `["-J", "rocm-overlay", "-N", "1", "--gres=gpu:1", ...]` with no `-w`. On a multi-node cluster this can land on a node where the probe script doesn't exist. `build_container_image` does ship to all nodes (`cluster.py:1077-1078`), so it's just the probe that's node-local.

Either `write_file(..., all_nodes=True)` or add `f"--nodelist={cluster.node_names[0]}"` as `test_devices.py:320` does — the latter also matches the fact that the GPU/ROCm skip check only probes node 0.

tests/native_host/e2e/test_rocm_overlay.py
```diff
@@ -0,0 +1,102 @@
...
+  case "$line" in *"/opt/rocm/lib"*) OVERLAY=yes ;; esac
+done < /proc/self/mountinfo
+echo "OVERLAY=$OVERLAY"
+echo OVERLAY_PROBE_OK
+"""
+
+
+def _node_has_gpu_and_rocm(cluster) -> bool:
+    probe = cluster.nodes[0].exec_allow_fail(
+        "ls /dev/kfd /dev/dri/renderD* 2>/dev/null | head -1 && test -d /opt/rocm/lib "
```
yansun1996: The KFD half of this check never gates anything:

```sh
ls /dev/kfd /dev/dri/renderD* 2>/dev/null | head -1 && test -d /opt/rocm/lib && echo READY
```

`&&` binds to the pipeline's exit status, which is `head -1`'s, which is always 0 — so the effective condition is just `test -d /opt/rocm/lib`, and the function name and skip message overpromise. `gpu_preflight(1)` further down does catch it for real, but only after `cluster.start()` has already paid for a full cluster boot.

```sh
test -e /dev/kfd && test -d /opt/rocm/lib && echo READY
```

tests/native_host/e2e/test_rocm_overlay.py
```diff
@@ -0,0 +1,102 @@
...
+    )
+    return "READY" in probe
+
+
+def _run_probe(cluster, overlay_on: bool, tmp_path) -> tuple[dict, str]:
+    if not _node_has_gpu_and_rocm(cluster):
+        pytest.skip("node 0 lacks a GPU (KFD) and/or /opt/rocm/lib")
+    cluster.container_preflight()
+
+    cluster.start(
```
yansun1996, line 61: Missing the `cluster.root_agent_preflight()` call that precedes every other `agent_as_root=True` start (`conftest.py:256-261`). Without it a host without passwordless sudo gets a hard `RuntimeError` out of `start()` instead of a clean skip.

Minor, while here: `cluster.py:813-817` already has a `SpurCluster.devices_config(auto_detect=True, **extra)` helper that builds this override — using it keeps the two in sync.

examples/spur.conf
```diff
@@ -148,6 +148,18 @@ format = "text"
 # it to that many bytes (e.g. "1073741824" for 1 GiB).
 memlock = "unlimited"
 
+# GPU/device discovery and container injection. AMD GPUs are auto-detected from
+# KFD when no on-disk CDI specs are present, and their device nodes (/dev/kfd,
+# /dev/dri/*) plus the video/render groups are injected into GPU containers.
+[devices]
+# auto_detect = true            # discover AMD GPUs from KFD when the CDI cache is empty
+# Overlay the host's /opt/rocm/lib{,64} over the image's inside every GPU
```
yansun1996: "inside **every** GPU container" overstates the scope on two axes — the overlay lives only in the auto-detected spec (so it's a no-op once any on-disk CDI spec exists) and never applies to managed-k0s pods. Suggest "inside auto-detected AMD GPU containers (not on-disk CDI specs, not managed-k0s pods)".

Also, now that the section exists, worth documenting the other two `DevicesConfig` fields — `cdi_spec_dirs` and `[[devices.gres]]` are absent from the example file entirely (pre-existing, but cheap to fix here):

```toml
# cdi_spec_dirs = ["/opt/vendor/cdi"]   # extra dirs scanned for on-disk CDI specs
```

yansun1996 (commented):
Strong motivation, and the "match `docker run`" framing is the right call — a three-way version split between compiler, bitcode, and runtime is exactly the kind of thing that should never happen silently. A few things to work through before this lands, mostly about the flag not reaching the places it claims to.

**The flag is inert on any node that has run k0s — in both directions.** `CdiCache::DEFAULT_SPEC_DIRS` includes `/etc/cdi` (`cache.rs:11`), and `CdiCache::load` auto-detects *only* when the on-disk cache is empty, otherwise warning `"on-disk CDI specs found; auto_detect ignored"` (`cache.rs:75-81`). `write_cdi_spec` writes `/etc/cdi/amd.json` (`cluster.rs:509,516`) from `cluster.rs:751-754` whenever a non-controller node starts, and that file is removed only by `remove_cdi_spec` on `k0s down` (`cluster.rs:543`, called from `cluster.rs:816`). So:

- **Upgrade path:** a node that ran k0s under the old `spurd` has an `/etc/cdi/amd.json` that already contains the `/opt/rocm/lib{,64}` mounts (the old `build_shared_edits` pushed them unconditionally, and `discovery.rs:40-50` persists `shared_edits` into `container_edits`). New `spurd` loads that from disk, skips auto-detect, and keeps overlaying host ROCm on native container jobs with the flag off — the default flip is a silent no-op there.
- **Opt-in path:** on the same node, setting `overlay_host_rocm_libs = true` gets you nothing, because the on-disk spec shadows auto-detect.

The precedence rule itself predates this PR, but it's what makes the new flag unreliable, so it probably can't be left for later. Would regenerating `/etc/cdi/amd.json` from current config on every `spurd` start (rather than only on k0s start) work? Or stamping the spec with the policy it was written under and regenerating on mismatch? At a minimum this needs calling out in the field's doc comment and in `examples/spur.conf`, since as written both promise something the code doesn't deliver on those nodes.

**This looks like it needs `!` in the title.** The field itself is schema-compatible (`#[serde(default)]` at `config.rs:1253`) and `DevicesConfig` isn't reachable from Raft/WAL or a snapshot, so there's no replay hazard — the break is purely behavioral. But a GPU image that doesn't ship its own ROCm userspace and works today only because the host's is mounted in will fail at `dlopen` after upgrade, with nothing in the log to explain it (the new breadcrumb at `container.rs:613` only fires when mounts *are* present, so the off-path is silent). `CHANGELOG.md` is release-please-generated, so the title is the only lever — `fba350d7 fix(spurd)!: stop injecting torch rendezvous env vars` is the same class of change. Worth also emitting a one-shot `info!` when the host has `/opt/rocm/lib` but the overlay is off, naming the config key, so the failure is self-diagnosing.

**Docs.** `docs/admin-guide/configuration.rst:1102-1122` has a `[devices]` field table (`auto_detect` at 1110, `cdi_spec_dirs` at 1114, `gres` at 1119) and this PR adds no row for the new field. `AGENTS.md` asks user-facing changes to ship with docs. The blanket "**Reload: Agent restart**" note at 1098-1100 already covers the section, so a row plus a warning about the upgrade impact should be enough:

```rst
   * - ``overlay_host_rocm_libs``
     - bool
     - ``false``
     - Bind-mount the host's ``/opt/rocm/lib{,64}`` read-only over the same
       paths inside auto-detected AMD GPU containers, replacing the image's
       ROCm userspace. Off by default. Applies only to the ``auto_detect``
       path — ignored when on-disk CDI specs are found, and not applied to
       managed-k0s pods.
```

**Issue references in code comments.** `AGENTS.md` asks for these to live in git history rather than the source. `spur#779` appears in seven added comments: `config.rs:1250`, `config.rs:2847`, `discovery.rs:26`, `discovery.rs:364`, `cluster.rs:517`, `container.rs:613`, and `test_rocm_overlay.py:4`.

**Comment-block length.** Several added blocks run past the repo's 1-3 line norm and could be tightened: `config.rs:1245-1252` (8 lines, where every sibling field in `DevicesConfig` uses 1-2), `discovery.rs:364-367` (4, and largely duplicating the field doc), `discovery.rs:24-26`, `container.rs:613-615`, `examples/spur.conf:151-153` and `156-160`, and the 10-line module docstring at `test_rocm_overlay.py:4-13`.

**Nice to have:** `CdiCache::load` now takes two adjacent unnamed `bool`s (`cache.rs:63`). The `main.rs` call site names its fields so it reads fine, but the tests pass them positionally — borrowing `&DevicesConfig` or a small options struct would make a transposition a type error rather than a silent flip.

A few things I checked that are fine, for the record: there's no other ungated host-library injection path (`/opt/rocm` outside `discovery.rs` is only doc comments and test fixtures, and `oci.rs:760` unions the image's own `config.Env` rather than synthesizing a host path); no persisted-state hazard; the new `[devices]` TOML header doesn't capture any following key; and there's no config-reload staleness, since `spurd` has no SIGHUP path and the docs already say "Reload: Agent restart" for the whole section.

Rest of the notes are inline.

yansun1996 (commented):
Thanks for the turnaround. The gate test is genuinely falsifiable now, the KFD half of the precondition actually gates, the job is pinned to the node the probe lands on, and the issue references are all gone.

One thing to put back, then leftovers from last round.

**Round 2 removed more coverage than it added.** `test_shared_edits_library_overlay_is_gated` is gone, and it was the only test that touched `build_shared_edits` — it asserted `off.mounts.is_empty()` and that device-node injection is independent of the flag. What remains is `let _ = build_shared_edits(false);` with no assertion, so the call site at `discovery.rs:365` is now untested: change it to `rocm_lib_overlay_mounts(true, ROCM_LIB_DIRS)` and every unit test still passes. My note last round was aimed at the environment dependence rather than the test itself — apologies if it read as "delete this". The extracted helper is the right shape; it just needs the wrapper assertions kept alongside it.

**Docs.** `docs/admin-guide/configuration.rst:1102-1122` has the `[devices]` table with rows for `auto_detect`, `cdi_spec_dirs` and `gres`, and no `docs/` file appears in the diff. AGENTS.md asks user-facing changes to ship with docs:

```rst
   * - ``overlay_host_rocm_libs``
     - bool
     - ``false``
     - Bind-mount the host's ``/opt/rocm/lib{,64}`` over the same paths inside
       auto-detected AMD GPU containers, replacing the image's ROCm userspace.
       Off by default. Applies only to the auto-detected spec — inert on a node
       that has any on-disk CDI spec, and never applied to managed-k0s pods.
```

**Comment length.** Every block flagged last round grew rather than shrank: `config.rs` 8 -> 11 lines, the overlay doc 4 -> 6, `container.rs` 3 -> 4, the gate-test comment 2 -> 3. The repo norm is 1-3.

**The upgrade break is still unmarked.** A GPU image that ships no ROCm userspace of its own works today only because the host's is mounted in; after this it fails at `dlopen` with nothing naming the config key, since the new `info!` only fires when mounts are present and the off path logs nothing. A one-line log when the host has `/opt/rocm/lib{,64}` and the overlay is off would make it self-diagnosing, and `!` in the title would flag it for the changelog.

The k0s / on-disk-spec interaction is now described in the field's doc comment, which is a reasonable place to stop for this PR — worth carrying into the docs row above too, since that is where an admin looks rather than the source.

crates/spur-devices/src/cdi/discovery.rs
```diff
@@ -859,7 +902,7 @@ mod tests {
 
     #[test]
     fn test_build_shared_edits_kfd() {
-        let _ = build_shared_edits();
+        let _ = build_shared_edits(false);
```
yansun1996: This is the only test that reaches `build_shared_edits`, and it asserts nothing — so the line that wires the gate up (`edits.mounts = rocm_lib_overlay_mounts(overlay_host_rocm_libs, ROCM_LIB_DIRS)`) has no coverage at all. The deleted test did cover it, weakly.

Two assertions here would restore it and also cover the PR's other promise, that device nodes and GPU groups are injected either way:

```rust
let off = build_shared_edits(false);
let on = build_shared_edits(true);
assert!(off.mounts.is_empty());
assert_eq!(off.device_nodes.len(), on.device_nodes.len());
assert_eq!(off.additional_gids, on.additional_gids);
```

Both sides read the same `/etc/group`, so the `additional_gids` comparison is stable everywhere — it only fails if someone later moves that line inside the gate, which is exactly what you want it to catch.

Minor, while here: `edits.mounts = ...` is an assignment where it used to be a push, so it is also the last writer. It is correct today (nothing populates `mounts` above it), but a mount added earlier later on would be silently dropped with no compile error. `edits.mounts.extend(...)` costs nothing.

crates/spur-devices/src/cdi/discovery.rs
```diff
@@ -358,26 +361,41 @@ fn build_shared_edits() -> ContainerEdits {
...
 }
 
+/// Host ROCm library directories overlaid into a GPU container, when enabled.
+const ROCM_LIB_DIRS: &[&str] = &["/opt/rocm/lib", "/opt/rocm/lib64"];
+
+/// Read-only bind mounts overlaying the host's ROCm libraries onto a container,
+/// or empty when the overlay is off. Opt-in: the default keeps the image's own
+/// userspace (matching `docker run --device=/dev/kfd ...`); overlaying the
+/// host's runtime/math/tuning libraries is a silent version split. Only
+/// directories that exist on the host are mounted. Split out from
```
yansun1996: The last sentence — "Split out from `build_shared_edits` so the gate is testable without a real `/opt/rocm`" — explains why the refactor happened during review, which AGENTS.md asks to keep in the commit message rather than the source. Dropping it also brings the block from 6 lines to 4.

crates/spur-core/src/config.rs
```diff
@@ -1241,6 +1241,20 @@ pub struct DevicesConfig {
...
```
yansun1996: The added carve-outs are the right content — thank you. The block is now 11 lines though, where every sibling field in `DevicesConfig` uses 1-2. The `docker run` rationale is already in the PR body and `examples/spur.conf`, so it can go here:

```rust
/// Bind-mount the host's `/opt/rocm/lib{,64}` over the image's inside
/// auto-detected AMD GPU containers; off by default. Inert when on-disk CDI
/// specs exist, and never applied to managed-k0s pods. Device nodes and GPU
/// groups are injected regardless.
```

crates/spurd/src/container.rs
```diff
@@ -610,6 +610,24 @@ fn apply_container_device_plan(rootfs: &Path, plan: &spur_devices::inject::Conta
         }
     }
 
+    // Record the host-ROCm library overlay in the job's log: overlaying host
```
yansun1996: Narrowing this to the `/opt/rocm` paths addressed the first half — thanks. Two things left:

The comment says `plan.mounts` "also carries device-node ... mounts", but `ContainerInjectionPlan` keeps device nodes in a separate `device_nodes` field (`inject.rs:79-85`); `mounts` is assigned only from `unique_mounts(...)` at `inject.rs:232`. The "site-authored CDI mounts" half is right.

The `info!` still sits above the loop at `:632` that performs the binds, and that loop skips when the source is missing (`:635-636`) and only warns on failure (`:638`) — so it can still announce an overlay that was never applied. Collecting the targets that actually bound and logging after the loop fixes it. Also 4 lines, up from 3.

examples/spur.conf
yansun1996: Still "inside every GPU container" — the two carve-outs made it into the Rust doc comment but not here, which inverts it: the admin-facing file is the less accurate one. Suggest "inside auto-detected AMD GPU containers (inert where an on-disk CDI spec exists; never for k0s pods)".

Small thing on line 160: "the injected-mounts line in the job log" — the log does reach the job's own output file, but the message reads "overlaying host ROCm libraries onto the container image", so an admin grepping for "injected-mounts" finds nothing.

tests/native_host/e2e/test_rocm_overlay.py
```diff
@@ -0,0 +1,106 @@
...
```
yansun1996, line 61: Two guards worth adding around this `start`.

`cluster.root_agent_preflight()` still precedes every other `agent_as_root=True` start (`conftest.py:256`, `:495`, `:560`). Without it a host with no passwordless sudo hits the 120s `wait_ready` timeout instead of skipping cleanly.

More importantly, if the agent is not actually root the injected binds fail with only a `warn!`, and `test_overlay_off_keeps_image_libraries` then passes even with the gate reverted. `cluster.spurd_agent_user(0) == "root"` is the check the sibling tests use for this (`test_devices.py:272`), and `conftest.py:224-240` guards `cgroup_cluster` the same way for the same reason.

Minor: `cluster.devices_config(overlay_host_rocm_libs=overlay_on)` (`cluster.py:813`) builds this override and pins `auto_detect=True`, which the test depends on since the whole feature is gated on auto-detection.

PR #840, opened 2026-09-07
fix(container): reject --container-remap-root instead of silently ignoring it

### LINKED ISSUE #778: `--container-remap-root` is accepted and silently ignored (dead flag)
## Summary

**`--container-remap-root` is a dead flag.** It is parsed, propagated through `JobSpec`, and written into `ContainerConfig.remap_root`, which nothing ever reads. The submission is accepted and the requested behavior never happens.

`docker run` on the same node runs the image as the user the image declares, normally root, which is what these images are built and tested for. Spur forces the submitting UID instead and offers no supported way to opt out, while advertising a flag that claims to be exactly that.

`--container-remap-root` is accepted by `sbatch` and `srun`, propagated through `JobSpec`, and copied into `ContainerConfig`, but the native container runtime never reads it. On a normal root-run `spurd`, the container process is always dropped to the submitting host UID/GID even when remapping was requested.

## Code evidence (`b5c6ece`)

The value is parsed in `crates/spur-cli/src/sbatch.rs` and `crates/spur-cli/src/srun.rs`, and propagated in `crates/spurd/src/agent_server.rs`:

```rust
remap_root: spec.container_remap_root,
```

`ContainerConfig.remap_root` is declared in `crates/spurd/src/container.rs` and is never read anywhere else in the crate. After pivoting into the rootfs, a `spurd` running as root unconditionally executes:

```rust
if is_root {
    drop_privileges(config.uid, config.gid, &supplementary_gids)?;
}
```

There is no user-namespace UID/GID mapping step, so the flag is inert.

## Reproduction

```bash
sbatch --container-image=busybox.sqsh \
       --container-remap-root \
       --wrap 'id -u; test -r /root'
```

Expected: the submitting user is mapped to UID 0 inside the container without becoming host root.

Actual: `id -u` reports the submitting UID and access follows that UID. No warning is emitted that the flag was ignored.

## Impact on ROCm development images

Published ROCm and AI-framework images are root-layout images: they install into root-owned prefixes and write state under `/root`. Miles installs and expects assets under paths such as `/root/miles`, `/root/models`, `/root/datasets`, `/root/.cache/huggingface`, and `/root/shared_data`, and its documented ROCm runtime runs the image as root:

https://github.com/radixark/miles/blob/main/docs/getting-started/installation.md

Under Spur's non-root job UID, such an image can fail before Ray or SGLang starts — Hugging Face cache writes, venv/site-packages writes, and compile caches (Triton, hipBLASLt tuning) all target root-owned paths. Rebuilding every vendor image around a different prefix is a per-image porting effort and is not what the advertised flag promises.

## Expected behavior

Implement rootless remapping: the host submitter stays unprivileged but appears as UID/GID 0 inside the container's user namespace, consistently whether the node daemon runs as root or non-root. The job must not become host UID 0, and access to allocated GPU/RDMA supplementary groups (`video`, `render`) must be preserved.

If rootless remapping is not going to be implemented, the flag should be rejected with a clear error rather than silently accepted.

## Suggested coverage

- `id -u` is 0 inside a remapped container, while the host-side process remains the submitting UID;
- writes under `/root` succeed;
- GPU access and supplementary groups still work;
- without the flag, the process remains the submitting UID inside the container.

## Decision requested

Either implement rootless UID/GID remapping, or make `--container-remap-root` a submission-time error until it exists. A flag that parses cleanly and does nothing is worse than an unimplemented flag, because the job then fails deep inside the image with a permission error that points nowhere near the cause.

If remapping is planned but not scheduled, please say so on this issue and reject the flag in the meantime.

powderluv:
## Summary

Fixes #778. `--container-remap-root` was parsed, propagated through `JobSpec`, and copied into `ContainerConfig.remap_root`, which nothing ever read. The submission was accepted and the requested behavior never happened: on a root-run spurd the container was always dropped to the submitting host UID, so a root-layout image (ROCm/AI images that install into root-owned prefixes and write under `/root`) failed deep inside — HF cache, venv/site-packages, compile caches — with a permission error that pointed nowhere near the flag.

This PR makes the flag a clear **submission-time error** until rootless remapping exists, so a flag that parses cleanly and does nothing becomes an early, actionable failure instead of a silent one.

## Why reject now rather than implement here

Implementing rootless-root *correctly* — so the issue's "writes under `/root` succeed" actually holds — is not a small patch: the container rootfs is root-owned (squashfs overlay lower / root extraction), so a naive submitter→0 userns map reports `id -u==0` but leaves root-owned image paths unwritable, reproducing the same class of failure. Making writes work needs **idmapped mounts** (`mount_setattr(MOUNT_ATTR_IDMAP)`). That is a separate feature PR (in progress), validated on a real GPU host. Rejecting now closes the accept-and-ignore footgun immediately without blocking on it.

## Changes

- `spurctld/server.rs`: `submit_job` rejects a spec with `container_remap_root` (`invalid_argument`) — covers sbatch, srun's new-allocation path, and any direct client. + test.
- `spur-cli/srun.rs`: early rejection in `main_with_args` (mirrors the existing `--container-readonly` rejection) so srun steps into an existing allocation, which bypass the controller submit path, are also refused. + test.
- `spur-cli/{sbatch,srun}.rs`: flag help now says NOT YET IMPLEMENTED.

## Testing

spur-cli 550, spurctld 1028 unit tests pass (2 new rejection tests); clippy/fmt clean. No GPU needed — it's a rejection, no container runs.

### DIFF
--- a/crates/spur-cli/src/sbatch.rs
+++ b/crates/spur-cli/src/sbatch.rs
@@ -239,7 +239,8 @@ pub struct SbatchArgs {
     #[arg(long, overrides_with = "container_entrypoint")]
     pub container_entrypoint: Option<String>,
 
-    /// Remap user to root inside container
+    /// Remap the submitting user to root inside the container.
+    /// NOT YET IMPLEMENTED: rejected at submission rather than silently ignored.
     #[arg(long, overrides_with = "container_remap_root")]
     pub container_remap_root: bool,
 
--- a/crates/spur-cli/src/sbatch.rs
+++ b/crates/spur-cli/src/sbatch.rs
@@ -239,8 +239,7 @@ pub struct SbatchArgs {
     #[arg(long, overrides_with = "container_entrypoint")]
     pub container_entrypoint: Option<String>,
 
-    /// Remap the submitting user to root inside the container.
-    /// NOT YET IMPLEMENTED: rejected at submission rather than silently ignored.
+    /// Remap the submitting user to root inside the container (not yet implemented; rejected at submission)
     #[arg(long, overrides_with = "container_remap_root")]
     pub container_remap_root: bool,
 
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -238,10 +238,8 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
 
     if args.container_remap_root {
         anyhow::bail!(
-            "--container-remap-root is not yet implemented and is rejected rather than silently \
-             ignored. Rootless UID/GID remapping (submitter -> root inside the container, \
-             unprivileged on the host, with the rootfs owned to match) is planned but not yet \
-             available. Omit the flag."
+            "--container-remap-root is not yet implemented and has no effect on the container's \
+             user mapping, which is determined by how spurd is deployed. Omit the flag."
         );
     }
 
@@ -2981,10 +2979,8 @@ mod tests {
 
     #[tokio::test]
     async fn container_remap_root_is_rejected() {
-        // The flag was parsed, propagated, and silently ignored; it must now
-        // fail early rather than accept the flag and not honor it. The
-        // unroutable controller makes the test fail fast (not hang dialing the
-        // real controller) if this reject is ever removed.
+        // Unroutable controller: if the guard ever regresses, this errors on
+        // connect rather than submitting a live job on a dev box.
         let result = main_with_args(vec![
             "srun".into(),
             "--controller=http://127.0.0.1:1".into(),
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -179,8 +179,7 @@ pub struct SrunArgs {
     #[arg(long)]
     pub container_entrypoint: Option<String>,
 
-    /// Remap the submitting user to root inside the container.
-    /// NOT YET IMPLEMENTED: rejected at submission rather than silently ignored.
+    /// Remap the submitting user to root inside the container (not yet implemented; rejected at submission)
     #[arg(long)]
     pub container_remap_root: bool,
 
@@ -239,10 +238,10 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
 
     if args.container_remap_root {
         anyhow::bail!(
-            "--container-remap-root is not yet implemented; the container would run as the \
-             submitting user despite the flag. Rootless UID/GID remapping (submitter -> root \
-             inside the container, unprivileged on the host) is planned but not yet available. \
-             Omit the flag rather than rely on it."
+            "--container-remap-root is not yet implemented and is rejected rather than silently \
+             ignored. Rootless UID/GID remapping (submitter -> root inside the container, \
+             unprivileged on the host, with the rootfs owned to match) is planned but not yet \
+             available. Omit the flag."
         );
     }
 
@@ -2983,9 +2982,12 @@ mod tests {
     #[tokio::test]
     async fn container_remap_root_is_rejected() {
         // The flag was parsed, propagated, and silently ignored; it must now
-        // fail at submission rather than run the container as the submitting user.
+        // fail early rather than accept the flag and not honor it. The
+        // unroutable controller makes the test fail fast (not hang dialing the
+        // real controller) if this reject is ever removed.
         let result = main_with_args(vec![
             "srun".into(),
+            "--controller=http://127.0.0.1:1".into(),
             "--container-image=img.sqsh".into(),
             "--container-remap-root".into(),
             "hostname".into(),
--- a/crates/spur-cli/src/srun.rs
+++ b/crates/spur-cli/src/srun.rs
@@ -179,7 +179,8 @@ pub struct SrunArgs {
     #[arg(long)]
     pub container_entrypoint: Option<String>,
 
-    /// Remap user to root inside container
+    /// Remap the submitting user to root inside the container.
+    /// NOT YET IMPLEMENTED: rejected at submission rather than silently ignored.
     #[arg(long)]
     pub container_remap_root: bool,
 
@@ -233,6 +234,15 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
         anyhow::bail!(
             "--container-readonly is not yet implemented; the container root would be \
              writable despite the flag. Omit it rather than rely on a read-only rootfs."
+        );
+    }
+
+    if args.container_remap_root {
+        anyhow::bail!(
+            "--container-remap-root is not yet implemented; the container would run as the \
+             submitting user despite the flag. Rootless UID/GID remapping (submitter -> root \
+             inside the container, unprivileged on the host) is planned but not yet available. \
+             Omit the flag rather than rely on it."
         );
     }
 
@@ -2966,6 +2976,25 @@ mod tests {
         let msg = format!("{}", result.unwrap_err());
         assert!(
             msg.contains("--container-readonly") && msg.contains("not yet implemented"),
+            "expected a not-yet-implemented rejection, got: {msg}"
+        );
+    }
+
+    #[tokio::test]
+    async fn container_remap_root_is_rejected() {
+        // The flag was parsed, propagated, and silently ignored; it must now
+        // fail at submission rather than run the container as the submitting user.
+        let result = main_with_args(vec![
+            "srun".into(),
+            "--container-image=img.sqsh".into(),
+            "--container-remap-root".into(),
+            "hostname".into(),
+        ])
+        .await;
+        assert!(result.is_err());
+        let msg = format!("{}", result.unwrap_err());
+        assert!(
+            msg.contains("--container-remap-root") && msg.contains("not yet implemented"),
             "expected a not-yet-implemented rejection, got: {msg}"
         );
     }
--- a/crates/spurctld/src/server.rs
+++ b/crates/spurctld/src/server.rs
@@ -32,6 +32,12 @@ use crate::sched_stats::SchedStatsCollector;
 const FORWARDED_HEADER: &str = "x-spur-forwarded";
 const LEADER_HEADER: &str = "x-spur-leader";
 
+/// Shared by the controller entry points that can carry the flag, so the wording
+/// cannot drift between them.
+const REMAP_ROOT_UNIMPLEMENTED: &str =
+    "--container-remap-root is not yet implemented and has no effect on the container's user \
+     mapping, which is determined by how spurd is deployed.";
+
 /// Resolve the comm address for an agent registration.
 ///
 /// Tries the advertised address first, then the gRPC peer IP (dynamic registration
@@ -786,12 +792,9 @@ impl SlurmController for ControllerService {
         // --container-remap-root asks for rootless UID/GID remapping that is not
         // implemented; reject it rather than accept the flag and not honor it.
         if core_spec.container_remap_root {
-            return Err(Status::invalid_argument(
-                "--container-remap-root is not yet implemented; rootless UID/GID remapping \
-                 (submitter -> root inside the container, unprivileged on the host, with the \
-                 rootfs owned to match) is planned but not yet available. Resubmit without the \
-                 flag.",
-            ));
+            return Err(Status::invalid_argument(format!(
+                "{REMAP_ROOT_UNIMPLEMENTED} Resubmit without the flag."
+            )));
         }
 
         // Clamp a non-privileged caller's base priority to the configured ceiling before it reaches
@@ -2128,6 +2131,14 @@ impl SlurmController for ControllerService {
         )
         .map_err(|e| Status::permission_denied(e.to_string()))?;
 
+        // The PTY client forwards the resolved container straight to the agent, so
+        // this path needs the same guard as run_step.
+        if req.container.as_ref().is_some_and(|c| c.remap_root) {
+            return Err(Status::invalid_argument(format!(
+                "{REMAP_ROOT_UNIMPLEMENTED} Rerun the step without the flag."
+            )));
+        }
+
         if job.state != spur_core::job::JobState::Running {
             return Err(Status::failed_precondition(format!(
                 "job {} is not running (state: {:?})",
@@ -2815,16 +2826,12 @@ impl SlurmController for ControllerService {
         )
         .map_err(|e| Status::permission_denied(e.to_string()))?;
 
-        // Mirror the submit-path guard: a step carries its own ContainerSpec
-        // (srun --container-remap-root), so reject the unimplemented flag here
-        // too, or a direct gRPC client could set it and bypass the submit check.
+        // A step carries its own ContainerSpec, so a direct gRPC client would
+        // otherwise bypass the submit-path guard.
         if req.container.as_ref().is_some_and(|c| c.remap_root) {
-            return Err(Status::invalid_argument(
-                "--container-remap-root is not yet implemented; rootless UID/GID remapping \
-                 (submitter -> root inside the container, unprivileged on the host, with the \
-                 rootfs owned to match) is planned but not yet available. Rerun the step \
-                 without the flag.",
-            ));
+            return Err(Status::invalid_argument(format!(
+                "{REMAP_ROOT_UNIMPLEMENTED} Rerun the step without the flag."
+            )));
         }
 
         if job.allocated_nodes.is_empty() {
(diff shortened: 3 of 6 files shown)


crates/spurctld/src/server.rs
```diff
@@ -783,6 +783,18 @@ impl SlurmController for ControllerService {
         let mut core_spec = proto_to_job_spec(spec)?;
         Self::bind_spec_to_identity(&mut core_spec, identity.as_ref())?;
 
+        // --container-remap-root is parsed and propagated but rootless UID/GID
+        // remapping is not implemented. Reject it at submission rather than
+        // accept it and silently run the container as the submitting user.
+        if core_spec.container_remap_root {
+            return Err(Status::invalid_argument(
+                "--container-remap-root is not yet implemented; rootless UID/GID remapping \
```
yansun1996: The claim "the container runs as the submitting user" is only true when `spurd` runs as root. In the rootless path, `container_init` takes the `setup_user_namespace` branch (`spurd/src/container.rs:1083`), which writes:

```rust
std::fs::write("/proc/self/uid_map", format!("0 {} 1\n", uid))
```

so the process is already UID 0 inside the user namespace, and `drop_privileges` is never reached — it's gated behind `if is_root` at `container.rs:1110`.

That also makes the parenthetical awkward: "(submitter -> root inside the container, unprivileged on the host)" is a description of what the rootless path already does, sitting inside a message explaining that it isn't available.

Since the flag is dead in *both* paths the rejection is still right — it's just the promised alternative outcome that varies by deployment. Would something deployment-neutral work better? e.g.

> `--container-remap-root` is not yet implemented and has no effect on the container's user mapping, which is determined by how spurd is deployed. Resubmit without the flag.

The same wording change is needed in `srun.rs:242-245`.

While you're touching both: the two messages have drifted already ("Omit the flag rather than rely on it." vs "Resubmit without the flag — …"), and only the `srun` one contains the `"not yet implemented"` substring its test asserts. A shared `pub const` in `spur-core` would keep them from diverging further and let both tests assert the same string.

crates/spur-cli/src/srun.rs
```diff
@@ -236,6 +237,15 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
         );
     }
 
+    if args.container_remap_root {
+        anyhow::bail!(
+            "--container-remap-root is not yet implemented; the container would run as the \
```
yansun1996: Same wording issue as the note on `server.rs:791` — "the container would run as the submitting user" doesn't hold for a rootless `spurd`, where `setup_user_namespace` already maps the submitter to UID 0 inside the namespace.

Ordering here is correct, though: this sits before `resolve_srun_env` at `:249` and before any controller connect, so validation happens before the expensive work and the new unit test genuinely needs no live controller.

crates/spur-cli/src/sbatch.rs
```diff
@@ -239,7 +239,8 @@ pub struct SbatchArgs {
     #[arg(long, overrides_with = "container_entrypoint")]
     pub container_entrypoint: Option<String>,
 
-    /// Remap user to root inside container
+    /// Remap the submitting user to root inside the container.
+    /// NOT YET IMPLEMENTED: rejected at submission rather than silently ignored.
```
yansun1996: The sibling flag two fields up already established a one-line form for exactly this situation:

```rust
/// Read-only container rootfs (not yet implemented; rejected at submission)
```

(`sbatch.rs:226`, `srun.rs:166`). The all-caps `NOT YET IMPLEMENTED:` marker is the only one of its kind in `spur-cli`. Matching the sibling keeps `--help` consistent:

```rust
/// Remap the submitting user to root inside the container (not yet implemented; rejected at submission)
```

Same applies to `srun.rs:183`.

crates/spurctld/src/server.rs
```diff
@@ -783,6 +783,18 @@ impl SlurmController for ControllerService {
         let mut core_spec = proto_to_job_spec(spec)?;
         Self::bind_spec_to_identity(&mut core_spec, identity.as_ref())?;
 
+        // --container-remap-root is parsed and propagated but rootless UID/GID
```
yansun1996: Three consecutive lines, a touch over the repo's 1-2 line norm. The background ("parsed and propagated but never honored") is genuinely useful *why* that the code can't convey, so it's worth keeping — just tighter:

```rust
// Rootless UID/GID remapping is unimplemented and the flag is never honored,
// so reject rather than accept-and-ignore.
```

crates/spurctld/src/server.rs
```diff
@@ -5219,6 +5231,35 @@ mod tests {
...
+            container_remap_root: true,
+            ..Default::default()
+        };
+        let status = svc
+            .submit_job(Request::new(SubmitJobRequest { spec: Some(spec) }))
+            .await
+            .expect_err("container_remap_root must be rejected at submission");
+        assert_eq!(status.code(), Code::InvalidArgument);
+        assert!(
+            status.message().contains("--container-remap-root"),
```
yansun1996: Good test — real RPC handler, asserts the code as well as the message, and `expect_err` fails loudly if the guard goes away.

Tiny asymmetry with its `srun` counterpart: that one asserts both `"--container-remap-root"` and `"not yet implemented"`, this one only the flag name. Worth matching, especially if the message moves to a shared constant.

crates/spur-cli/src/srun.rs
```diff
@@ -2969,6 +2979,25 @@ mod tests {
...
+        // The flag was parsed, propagated, and silently ignored; it must now
+        // fail at submission rather than run the container as the submitting user.
+        let result = main_with_args(vec![
+            "srun".into(),
+            "--container-image=img.sqsh".into(),
+            "--container-remap-root".into(),
+            "hostname".into(),
+        ])
+        .await;
+        assert!(result.is_err());
```
yansun1996, line 2992: Drives the real parser through `main_with_args` and is properly falsifiable — the message assertion below is what makes it honest, since `is_err()` alone would still hold with the guard removed.

One thing worth guarding against: with the fix reverted, this argv has no `--jobid`, so it falls through to the submit path and would dial `SPUR_CONTROLLER_ADDR` / `localhost:6817`. On a dev box running `spurctld` that means a reverted-guard test run submits a real job while still failing the assertion. The `sbatch` sibling avoids this by testing the pure `build_sbatch_job_spec` seam (`sbatch.rs:1138`); `srun` has no equivalent seam today, so an `assert!(!msg.contains("connect"))` or an unreachable controller address in the fixture would be the cheap version.

yansun1996 (commented):
This is a good call, and the reasoning for rejecting now rather than half-implementing rootless-root is convincing — a naive submitter→0 map really would reproduce the same failure one layer down.

I traced the premise and it holds: `remap_root` is declared at `spurd/src/container.rs:147`, populated at `agent_server.rs:1632` and `:2732`, and there is no read site anywhere in the tree (only the two test fixtures at `container.rs:1622,1663`). Guard placement is right too — it sits after `proto_to_job_spec`/`bind_spec_to_identity` and before `clamp_priority` and `cluster.submit_job`, and it is *not* on the WAL apply path (`ClusterManager::apply_operation` at `cluster.rs:5308` builds `Job::new` with no validation), so an upgraded controller replays pre-existing `container_remap_root: true` entries without crashing. Worth stating that explicitly in the PR body — it's the difference between this change and an upgrade outage.

Two things to fix, then some consistency notes.

**1. The error text is wrong for rootless `spurd`** — details inline, but the short version: both messages promise "the container runs as the submitting user", and that only holds when `spurd` runs as root. In the rootless path `setup_user_namespace` (`spurd/src/container.rs:999-1007`) writes `uid_map = "0 <uid> 1"` unconditionally and `drop_privileges` is gated behind `if is_root` (`container.rs:1110-1111`), so the process already *is* UID 0 inside the namespace.

**2. Docs still document the flag as working.** `docs/user-guide/running-containers.rst:72-73` reads "Map the job user to root inside the container." — the only mention of the flag anywhere in `docs/`. The precedent is two rows up: the `--container-readonly` entry at `:61-64` already carries the right wording, and the commit that made *that* flag a rejection (`e71fc8b9`) updated this same page in the same change. Suggested patch, mirroring that row (the page uses `.. note::`/`.. warning::` only for prose asides at `:144`/`:167`, never inside this table):

```rst
   * - ``--container-remap-root``
     - Not yet implemented — rejected at submission. Rootless UID/GID
       remapping is not available yet; the flag is refused rather than
       silently ignored.
```

**The gate is client-side only for steps.** `CreateJobStepRequest` (`slurm.proto:1653-1664`) has no container field, so the bot's `create_job_step` concern doesn't quite apply — but `RunStepRequest.container` does carry `ContainerSpec.remap_root` (`slurm.proto:1014`, `:1028`), and `run_step` (`server.rs:2776`) accepts it unchecked at `server.rs:2850`, forwarding it to `spurd` via `agent_server.rs:2732`. The `srun` bail closes the stock client, not the RPC — a direct gRPC client or an older `srun` in a mixed-version cluster still gets accept-and-ignore, which is the shape this PR set out to end. No security consequence since the field is inert, but the contract is only half-delivered.

Related, and probably the cleaner fix for both: `cluster::submit_job` is the choke point every submitter reaches. REST currently calls it directly (`rest/handlers.rs:135`), bypassing the RPC guard — harmless today because `SubmitJobFields` (`rest/types.rs:148-166`) has no container fields and `handlers.rs:126` rejects `uid == 0` outright, but it shows the shape. Putting the guard there would cover REST and any future submitter for free, and pair naturally with a `run_step` check. (`spur-k8s` is fine — `job_controller.rs:214` goes through the guarded `SubmitJob` RPC.)

One thing that argues for the chokepoint over edge-guarding: `container_spec_from_job_spec` (`server.rs:3704`) still copies `spec.container_remap_root` into every step's `ContainerSpec`, so jobs submitted *before* the upgrade keep propagating `remap_root: true` into steps indefinitely. Inert, but it means edge rejection leaves a permanent second source.

**`sbatch` is inconsistent with its own sibling.** `--container-readonly` is rejected client-side in `build_sbatch_job_spec` (`sbatch.rs:834-839`) with a test at `sbatch.rs:1138`; `--container-remap-root` gets nothing, so it only fails after a controller round-trip and surfaces as `Error: job submission failed / Caused by: status: InvalidArgument, message: "..."` rather than the clean one-line bail `srun` prints. The full text does reach the user, so it's noise rather than loss — but a six-line `bail!` next to `sbatch.rs:834` plus a sibling test would match the precedent and fail faster.

**Missing e2e.** `tests/native_host/e2e/test_srun_container.py:660` has `test_container_readonly_rejected`. The remap-root analogue is a ~10-line copy and would cover the CLI→controller path end to end.

**On `!` in the title:** I'd say no. A script passing the flag does go from exit 0 to non-zero, but the flag isn't removed and its documented meaning was never honored, and the `--container-readonly` precedent (`e71fc8b9`) shipped the same accept→reject flip without `!`. Worth a line in the PR body either way. Small nit on the scope though — `cli` isn't used anywhere in this repo's history (`spur-cli` is the convention), and the change also touches `spurctld`, so `fix(spur-cli)` or `fix(container)` would fit `AGENTS.md` better.

A few smaller things I'd leave as follow-ups rather than block on: `srun.rs:690` still forwards `remap_root: args.container_remap_root` while the sibling hardcodes `readonly: false` at `:677-679` with a comment explaining the upstream rejection (dead code past the bail); `srun.rs:1592`'s `container_requested` includes `|| args.container_remap_root`, now unreachable; and both guards reject unconditionally, so `srun --container-remap-root hostname` with no `--container-image` fails even though no container is involved — though the `--container-readonly` guard directly above behaves identically, so that one is really a both-or-neither change.

crates/spurctld/src/server.rs
```diff
@@ -2804,6 +2815,17 @@ impl SlurmController for ControllerService {
         )
         .map_err(|e| Status::permission_denied(e.to_string()))?;
 
+        // A step carries its own ContainerSpec, so a direct gRPC client would
+        // otherwise bypass the submit-path guard.
+        if req.container.as_ref().is_some_and(|c| c.remap_root) {
```
yansun1996, line 2831: This checks the request's container, but the spec that actually reaches the agent is resolved further down:

```rust
let step_container = match req.container.clone().filter(|c| !c.image.is_empty()) {
    Some(mut step) => { ... }
    None => container_spec_from_job_spec(&job.spec),
};
```

and `container_spec_from_job_spec` copies `remap_root: spec.container_remap_root`. So a step with no container of its own inherits the job's flag and is never rejected. Post-upgrade only legacy jobs can carry it, and nothing reads the field, so this is inert — but it's the accept-and-ignore shape the PR is closing, and the batch launch path is the same (`scheduler_loop.rs` -> `agent_server.rs`, no guard).

Hoisting the `step_container` computation above this check and testing that would cover the inherited case, and would drop the small over-rejection here (this currently fires on an empty-image spec that the `.filter()` would have discarded anyway).

If you'd rather leave it, the comment is the thing to adjust — "a direct gRPC client would otherwise bypass the submit-path guard" is true, but reads as if the step path is now fully closed. Either way, the test only exercises the `req.container`-set path; the inherited one has no coverage.

crates/spur-cli/src/sbatch.rs
```diff
@@ -239,7 +239,7 @@ pub struct SbatchArgs {
     #[arg(long, overrides_with = "container_entrypoint")]
     pub container_entrypoint: Option<String>,
 
-    /// Remap user to root inside container
+    /// Remap the submitting user to root inside the container (not yet implemented; rejected at submission)
```
yansun1996, line 242: The help now says "rejected at submission", which is true, but `sbatch` has no client-side bail — its own sibling `--container-readonly` is rejected in `build_sbatch_job_spec` (`:833`) with a test at `:1137`, and `srun` bails locally too. So this one flag alone needs a controller round-trip and surfaces as a wrapped `InvalidArgument` rather than a clean one-line error. A matching `bail!` next to `:833` plus a mirror of that test would line all three up.

yansun1996 (commented):
Rebased onto `main` (was 11 behind) and pushed the fixes below — sorry for the branch churn, all revertible.

- **Docs/messages were wrong for rootless.** They said mapping the submitter to root "is planned", but a rootless `spurd` already does it (`setup_user_namespace` writes `uid_map "0 <uid> 1"`; `drop_privileges` is gated on `is_root`). Only the rootfs ownership is missing. The docs row, both reject messages and the two proto comments now say the flag has no effect and the mapping follows the deployment.
- **`create_job_step` was unguarded.** It gained a container field with the `--pty` work and returns the resolved spec straight to the client, which sends it to the agent. Guarded now, and steps no longer inherit the flag from the parent job.
- **E2E** for the srun and sbatch paths. The sbatch one matters most — no client-side bail there, so it's the only test driving the controller's own rejection.

I checked the legacy-job case since the guard rejects what the controller used to accept: WAL replay is unvalidated and `srun` sends `remap_root: false`, so pre-upgrade jobs keep running. **No regression.**

Only thing left is the title: `cli` isn't a scope this repo uses (`spur-cli` and `spurctld` are), and it squash-merges into the changelog. `fix(container)` would fit.

crates/spurctld/src/server.rs
yansun1996: Addressed this — pushed as `6347cb93`. Two corrections to what I wrote above.

First, my snippet was stale: the rebase pulled in the `--pty` step work from `main`, which extracted that `match` into `resolve_step_container`. Apologies for the confusion.

Second, the rebase made this bigger than I described. `CreateJobStepRequest` gained a `container` field, so `create_job_step` is now a second step-side entry point — and it was unguarded, while `CreateJobStepResponse.container` hands the *resolved* spec to the interactive-PTY client, which forwards it to the agent without coming back through the controller. So the flag had two live routes past the submit guard, not one.

What I did instead of hoisting the guard: `container_spec_from_job_spec` no longer copies `remap_root` out of the job spec, so neither step path can inherit it. Sanitising rather than rejecting keeps a pre-upgrade job's steps running — rejecting there would have turned a dead field into a broken allocation. `create_job_step` gets the same request-level guard as `run_step`, and the three messages now share one constant, which also settles the drift note in the summary.

Tests: `create_job_step_rejects_container_remap_root` and `resolve_step_container_does_not_inherit_remap_root`. I mutation-checked both — restoring the inheritance fails the second, removing the guard fails the first. Full `spurctld` suite green (1040), fmt and clippy clean.

One thing I deliberately left: `container_spec_from_job_spec` still inherits `readonly` from the job spec, and `--container-readonly` is rejected at submission too, so it has exactly the same shape. Out of scope here, but probably worth the same one-line treatment.

yansun1996 (approved):
LGTM

PR #845, opened 2026-09-07
feat(spurd): honor --container-image for srun --pty steps

sajmera-pensando:
## Summary

`srun --pty --container-image` silently ran on the **host** instead of inside the requested container. Containerized `srun` steps were fixed for the buffered path previously, but the interactive `--pty` path is a separate route (`CreateJobStep` → `InteractiveSession`/`InitSession` → agent `spawn_pty_in_job`) that carried no container config and only knew how to `nsenter` into an existing container or spawn on the host.

This makes `srun --pty --container-image` (and the inherited case — a bare `srun --pty` nested inside a containerized `sbatch --container-image` job) run inside the container, at parity with the non-pty path. Note `salloc` has no `--container-image` flag, so there is no `salloc`-inherited variant.

## Approach

- **Proto** (`proto/slurm.proto`): additive `ContainerSpec` on `CreateJobStepRequest` (tag 11), `CreateJobStepResponse` (tag 3), and `InitSession` (tag 9). No renumbering, no persisted-state change — not a breaking change.
- **Controller**: `create_job_step` resolves the effective spec (the step's own image, else inherited from a containerized allocation) via a shared `resolve_step_container` helper — also refactored out of `run_step` — and echoes it in the response so the CLI can forward it to the agent, which connects directly and cannot see the parent job spec.
- **CLI**: threads the spec through `run_interactive_pty` → `open_interactive_session` → `InitSession`; removes the now-obsolete "container options are not honored for a --pty step" warning.
- **Agent**: extracted the container fork core (namespace unshare, mounts, device injection, `pivot_root`, priv drop) out of `run_containerized_step` into shared helpers parameterized by stdio wiring, so a new `spawn_containerized_pty_step` reuses it but wires the child to a PTY slave and returns the master fd. `run_pty_bridge` now reaps via an exit-code future (a `tokio::process::Child` wait for the host/nsenter path, a blocking `waitpid` for the raw-forked container). The rootfs guard and `active_steps` registration are held for the session so a disconnect, exit, or `scancel` tears down the rootfs.

A standalone `srun --pty --container-image` builds a fresh container; a bare `--pty` step nested inside a containerized `sbatch` job enters the parent's running container via `nsenter` (the controller resolves the inherited image).

## Design choices / trade-offs

- Full parity with the non-pty path (including the inherited-image case) was chosen over an explicit-image-only fix, which is why the controller round-trip and the `CreateJobStepRequest`/`Response` fields exist — the CLI can't see the parent job's image for the inherited case. Today the only inherited path is a bare nested `srun --pty` inside a containerized `sbatch` job (resolved by the controller, entered via `nsenter`); `salloc` has no container flag, so building a *fresh* container from an inherited image isn't reachable from the CLI yet — the resolution logic keeps parity with the buffered path and is ready if `salloc` gains container support.
- Refactored the shared container-fork core rather than cloning it, so the buffered and PTY step launchers stay in sync.

## Additional fixes surfaced by the new e2e

Two pre-existing bugs the container `--pty` e2e exposed (both fixed here):

**1. Non-interactive `--pty` lost its output.** A `--pty` step run without a client TTY (a script, a pipe, redirected output, CI over SSH) was `SIGHUP`'d the instant the client's stdin hit EOF — before its output was drained (exit code still came through, so it looked half-working). The client now marks the session `non_interactive` (a new **additive** `InitSession` field — agent-internal plumbing on the `SlurmAgent` service, not the FFI/REST surface), and the agent's PTY bridge treats input-stream close as stdin-EOF for such sessions (drain the output, let the command finish) instead of a hangup. Interactive clients keep hangup-on-disconnect. This is what makes `srun --pty <cmd>` reliable in scripts/pipes.

**2. Security: spurd's environment leaked into sessions that enter a running container job.** `spur exec`, `srun --overlap`, and `srun --overlap --pty` built their child process without `env_clear()`, inheriting spurd's full environment (any daemon secrets included); for a container job, `read_proc_environ` additionally read the namespace *shepherd*, which is a spurd fork carrying spurd's env. These paths now `env_clear()` and source the session env from the **job itself** — the container's workload (PID 1) for a containerized job, else the tracked pid. Verified with a daemon-only canary that no longer appears in any of the three entry paths. Pre-existing and broader than `--pty`; bundled here since the e2e caught it. (Fresh `sbatch`/`srun --container-image` jobs were never affected — they get the submitting user's env via `--export`.)

## Test plan

- [x] `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` — clean
- [x] `cargo test --locked` — all pass, including new unit tests: effective-spec resolution, `waitpid` exit-code mapping, and the PTY-bridge drain-after-input-close guard
- [ ] **Bare-metal E2E** (requires a real cluster): `tests/native_host/e2e/test_srun_container.py::TestSrunPtyContainerStep` — container entry, tty-in-container, nested nsenter, GPU injection, host fallthrough, and a daemon-env-leak guard across `spur exec` / `srun --overlap` / `srun --overlap --pty`; plus the updated `test_srun_pty.py`

## Manual verification (single real node)

Brought up `spurctld` + `spurd` on a real node and ran an interactive `srun --pty
--container-image` against a squashfs image whose `/inside-image-marker` file does
**not** exist on the host. The shell lands inside the container (marker present),
on a real terminal (`test -t 1` → `TTY-OK`), with the job env and GPU-deny vars
injected (a zero-GPU step gets `*_VISIBLE_DEVICES=-1` / `NVIDIA_VISIBLE_DEVICES=void`):

```console
root@SCSSAJMERA01:~/src/github.com/pensando/spur# export SPUR_CONTROLLER_ADDR='http://127.0.0.1:6817'
root@SCSSAJMERA01:~/src/github.com/pensando/spur# spur srun -N1 -t 0:10 --pty --container-image=/tmp/spur-local/img/marked.sqsh bash
srun: Pending job allocation 7...
srun: job 7 running on SCSSAJMERA01
srun: opening interactive session on SCSSAJMERA01
bash-5.1# pwd
/
bash-5.1# cat /inside-image-marker
spur-container-step-marker
bash-5.1# test -t 1 && echo TTY-OK
TTY-OK
bash-5.1# env
NVIDIA_VISIBLE_DEVICES=void
PWD=/
LOGNAME=spur
SLURM_JOB_PARTITION=default
ROCR_VISIBLE_DEVICES=-1
SPUR_JOB_PARTITION=default
SLURM_JOBID=7
HOME=/home/spur
SPUR_NODELIST=SCSSAJMERA01
HIP_VISIBLE_DEVICES=-1
ZE_AFFINITY_MASK=-1
CUDA_VISIBLE_DEVICES=-1
TERM=xterm-256color
USER=spur
SLURM_NODELIST=SCSSAJMERA01
GPU_DEVICE_ORDINAL=-1
SPUR_JOBID=7
SHLVL=1
SLURM_JOB_ID=7
SPUR_JOB_ID=7
SPUR_JOB_NODELIST=SCSSAJMERA01
SPUR_JOB_GPUS=
SLURM_JOB_NODELIST=SCSSAJMERA01
_=/usr/bin/env
bash-5.1# exit
exit
```

For contrast, a `--pty` step with **no** image runs on the host (`/inside-image-marker`
absent) — confirming the image is what gates the container path, not `--pty`.

### Nested case (`sbatch --container-image` + `srun --pty`)

A batch job runs a container and writes a file into it at runtime; a nested
`srun --jobid --overlap --pty` then enters that **running** container via `nsenter`
(agent log: `step --container-image ignored: joining the parent job's running
container` → `interactive session started ... overlap=true`):

```console
root@SCSSAJMERA01:~/src/github.com/pensando/spur# spur sbatch -N1 -t 0:15 --container-image=/tmp/spur-local/img/marked.sqsh \
                        --container-env NESTED_PROOF=only-in-parent-container \
                        -o nested-sbatch-%j.out /tmp/spur-local/nested-batch.sh
Submitted batch job 9
root@SCSSAJMERA01:~/src/github.com/pensando/spur# spur srun --jobid=9 --overlap --pty bash
bash-5.1# echo NESTED_PROOF=$NESTED_PROOF; cat /container-runtime-marker; test -t 1 && echo TTY-OK; echo THIS_JOB=$SPUR_JOB_ID
NESTED_PROOF=
runtime-only-in-parent-container
TTY-OK
THIS_JOB=9
```

`/container-runtime-marker` is written at runtime by the batch script *inside* the
container, so its presence in the nested shell proves entry into the parent
container (the file exists in neither the host nor a fresh container; confirmed
absent on the host).

(`NESTED_PROOF` was empty in this first capture, taken before the env fix below:
the nsenter env came from the shepherd, not the container. The "Additional fixes"
env change corrects that — a nested session now reads the container's own PID-1
environment, so `--container-env` **does** propagate after this PR, e.g.
`srun --jobid=<id> --overlap --pty bash -c 'echo $NESTED_PROOF'` →
`only-in-parent-container`.)

### Daemon environment does not leak (security fix)

`spurd` is launched with a daemon-only var; a session that enters a running
container must never see it. All three entry paths report `CLEAN`:

```console
root@SCSSAJMERA01:~/src/github.com/pensando/spur# IMG=/tmp/spur-local/img/marked.sqsh
# start one container job and wait for it to run
root@SCSSAJMERA01:~# JID=$(spur sbatch -N1 -t 0:10 --container-image="$IMG" /tmp/spur-local/sleeper.sh | grep -o '[0-9]*$')
root@SCSSAJMERA01:~# sleep 5
# enter it and check the daemon canary is NOT there
root@SCSSAJMERA01:~# spur srun --jobid=$JID --overlap --pty env </dev/null | grep SPUR_DAEMON_ENV_CANARY && echo ">>> LEAKED" || echo "CLEAN"
CLEAN
root@SCSSAJMERA01:~# spur scancel $JID
```

The shell running `srun` never has `SPUR_DAEMON_ENV_CANARY` (only `spurd` does),
so `CLEAN` proves the daemon environment did not leak into the session.

> Note: this node runs `spurd` as root, so it exercises the **rootful** container
> path. The rootless path (unprivileged user namespace) and real GPU injection are
> covered by the bare-metal E2E rather than this smoke test.

## Notes

- The **nested** case (`sbatch --container-image` + `srun --pty` → enters the parent via nsenter) is confirmed by the new nested E2E (both the external `--jobid --overlap --pty` attach and the batch-script-internal `srun --pty`).

## Known limitation / follow-up

**Container steps nested in a `salloc` allocation run on the host under rootful spurd.** With `spurd` running as root, an allocation-only `salloc` job is treated as if it has live namespaces (`parent_namespaces=true`), so a subsequent `srun --container-image` / `srun --pty --container-image` step takes the nsenter-into-parent path instead of building a fresh container — and since the allocation has no real container, it lands on the host. This is **pre-existing and not specific to `--pty`**: it hits the buffered container-step path (from the original containerized-srun work) as well, and this PR's PTY dispatch simply mirrors it. It does **not** reproduce under rootless spurd — the existing `test_salloc_then_srun_container` (buffered `salloc` + `srun --container-image`) passes in rootless CI. Root cause to chase: why an allocation-only job reports `parent_namespaces=true` (with a live pid) in rootful mode; the fix is to route allocation-only jobs' container steps through the fresh-container path regardless of privilege. Best handled as a **follow-up PR**, or folded into a broader container-step dispatch refactor rather than expanding this one.

### DIFF
--- a/crates/spur-cli/src/interactive.rs
+++ b/crates/spur-cli/src/interactive.rs
@@ -9,6 +9,7 @@ use spur_proto::proto::{
     JobKeepaliveRequest,
 };
 use std::collections::HashMap;
+use std::io::IsTerminal;
 use tokio::io::{AsyncReadExt, AsyncWriteExt};
 use tokio::signal::unix::{signal, SignalKind};
 
@@ -195,6 +196,12 @@ pub async fn open_interactive_session(
     user: &str,
     container: Option<spur_proto::proto::ContainerSpec>,
 ) -> std::result::Result<InteractiveSessionHandle, tonic::Status> {
+    // When our stdin is not a TTY (a script, redirected input, a pipe), the
+    // client closes its input stream as soon as stdin hits EOF. Tell the agent so
+    // it treats that as stdin-EOF (drain output, let the command finish) rather
+    // than a hangup (SIGHUP). Interactive clients keep the hangup-on-disconnect
+    // behavior.
+    let non_interactive = !std::io::stdin().is_terminal();
     let init = InteractiveInput {
         msg: Some(interactive_input::Msg::Init(InitSession {
             job_id,
@@ -206,6 +213,7 @@ pub async fn open_interactive_session(
             env: HashMap::new(),
             user: user.to_string(),
             container,
+            non_interactive,
         })),
     };
 
--- a/crates/spur-cli/src/interactive.rs
+++ b/crates/spur-cli/src/interactive.rs
@@ -184,6 +184,7 @@ pub struct InteractiveSessionHandle {
 /// Open the InteractiveSession RPC, returning the raw handle.
 ///
 /// Returns `Err(tonic::Status)` on RPC failure.
+#[allow(clippy::too_many_arguments)]
 pub async fn open_interactive_session(
     agent: &mut SlurmAgentClient<crate::authclient::AuthChannel>,
     job_id: u32,
@@ -192,6 +193,7 @@ pub async fn open_interactive_session(
     winsize: spur_proto::proto::WindowSize,
     overlap: bool,
     user: &str,
+    container: Option<spur_proto::proto::ContainerSpec>,
 ) -> std::result::Result<InteractiveSessionHandle, tonic::Status> {
     let init = InteractiveInput {
         msg: Some(interactive_input::Msg::Init(InitSession {
@@ -203,6 +205,7 @@ pub async fn open_interactive_session(
             argv,
             env: HashMap::new(),
             user: user.to_string(),
+            container,
         })),
     };
 
@@ -322,9 +325,12 @@ pub async fn run_interactive_session(
     overlap: bool,
     user: &str,
 ) -> Result<i32> {
-    let handle = open_interactive_session(agent, job_id, step_id, argv, winsize, overlap, user)
-        .await
-        .map_err(|status| anyhow::anyhow!("InteractiveSession RPC failed: {}", status.message()))?;
+    let handle =
+        open_interactive_session(agent, job_id, step_id, argv, winsize, overlap, user, None)
+            .await
+            .map_err(|status| {
+                anyhow::anyhow!("InteractiveSession RPC failed: {}", status.message())
+            })?;
     drive_interactive_session(handle).await
 }
 
--- a/crates/spur-cli/src/interactive.rs
+++ b/crates/spur-cli/src/interactive.rs
@@ -196,11 +196,9 @@ pub async fn open_interactive_session(
     user: &str,
     container: Option<spur_proto::proto::ContainerSpec>,
 ) -> std::result::Result<InteractiveSessionHandle, tonic::Status> {
-    // When our stdin is not a TTY (a script, redirected input, a pipe), the
-    // client closes its input stream as soon as stdin hits EOF. Tell the agent so
-    // it treats that as stdin-EOF (drain output, let the command finish) rather
-    // than a hangup (SIGHUP). Interactive clients keep the hangup-on-disconnect
-    // behavior.
+    // Non-TTY stdin (script, pipe, redirect) closes the input stream on EOF, not
+    // on hangup. Flag it so the agent drains output instead of SIGHUP-ing the
+    // step; interactive clients keep hangup-on-disconnect.
     let non_interactive = !std::io::stdin().is_terminal();
     let init = InteractiveInput {
         msg: Some(interactive_input::Msg::Init(InitSession {
--- a/crates/spurd/src/agent_server.rs
+++ b/crates/spurd/src/agent_server.rs
@@ -4655,10 +4655,8 @@ impl AgentService {
         use tokio::io::unix::AsyncFd;
         use tokio_stream::StreamExt;
 
-        // The workload's exit is delivered by `wait_exit`: a `tokio::process::Child`
-        // wait for the host/nsenter path, or a blocking `waitpid` for a raw-forked
-        // container step. Pinned so it can be polled in the select and, if the loop
-        // exits first (EOF/disconnect), awaited once more to reap.
+        // Pinned so `wait_exit` can be polled in the select and, if the loop exits
+        // first (EOF/disconnect), awaited once more to reap.
         tokio::pin!(wait_exit);
 
         let master_raw = master.as_raw_fd();
--- a/crates/spurd/src/agent_server.rs
+++ b/crates/spurd/src/agent_server.rs
@@ -782,6 +782,206 @@ async fn run_tokio_step_to_spool(
     }
 }
 
+/// What a containerized child execs after `container_init`. Both the buffered
+/// step (a script file inside the rootfs) and the interactive PTY step (an
+/// interactive shell or inline command) route through the same fork core.
+enum StepChildCommand {
+    /// Run `[entrypoint &&] <shell> <script_path>` — the buffered step path.
+    Script {
+        script_path: String,
+        entrypoint: Option<String>,
+    },
+    /// Exec an interactive shell in place. `None` cmdline = the image's default
+    /// shell; `Some(cmdline)` runs `<shell> -c "exec <cmdline>"` so the workload
+    /// replaces the shell (keeping the PTY session leader as the real process,
+    /// which job control and Ctrl-C depend on).
+    Shell {
+        cmdline: Option<String>,
+        entrypoint: Option<String>,
+    },
+}
+
+/// Reject NUL bytes in strings that reach `execve`: a NUL would silently degrade
+/// the child's exec (e.g. to `bash -c ""`, which exits 0), reporting a step that
+/// never ran as success. Checked before the fork so the error propagates cleanly.
+fn reject_nul_bytes(cmd: &StepChildCommand) -> Result<(), Status> {
+    let strings: [Option<&str>; 2] = match cmd {
+        StepChildCommand::Script {
+            script_path,
+            entrypoint,
+        } => [Some(script_path.as_str()), entrypoint.as_deref()],
+        StepChildCommand::Shell {
+            cmdline,
+            entrypoint,
+        } => [cmdline.as_deref(), entrypoint.as_deref()],
+    };
+    for s in strings.into_iter().flatten() {
+        if s.as_bytes().contains(&0) {
+            return Err(Status::invalid_argument(
+                "container entrypoint or step command contains a NUL byte",
+            ));
+        }
+    }
+    Ok(())
+}
+
+/// Runs in a freshly forked child after its stdio has been wired (spool fds or a
+/// PTY slave): joins the job cgroup while still root, runs `container_init`
+/// (namespace unshare, mounts, device injection, pivot_root, priv drop), signals
+/// readiness on `ready_w`, then execs `cmd`. Never returns. The whole namespace
+/// setup is shared by the buffered and PTY containerized-step paths.
+fn container_child_exec(
+    container_cfg: &crate::container::ContainerConfig,
+    rootfs: &std::path::Path,
+    ready_w: std::os::fd::OwnedFd,
+    memlock: spur_core::config::MemlockLimit,
+    cgroup_join: &Option<executor::CgroupJoin>,
+    env_base: HashMap<String, String>,
+    cmd: StepChildCommand,
+) -> ! {
+    use std::os::fd::AsRawFd;
+    let ready_w_fd = ready_w.as_raw_fd();
+
+    // Join while still root, before container_init's pivot_root hides the host
+    // cgroupfs and close_inherited_fds reaps the log fd.
+    if let Some(ref join) = cgroup_join {
+        let _ = join.join();
+    }
+
+    crate::container::close_inherited_fds(ready_w_fd);
+    executor::apply_memlock(memlock);
+
+    let hook_env = match crate::container::container_init(container_cfg, rootfs) {
+        Ok(env) => env,
+        Err(e) => {
+            let msg = format!("E:{e:#}");
+            unsafe {
+                libc::write(ready_w_fd, msg.as_ptr() as *const _, msg.len());
+            }
+            std::process::exit(1);
+        }
+    };
+
+    unsafe { libc::write(ready_w_fd, b"OK".as_ptr() as *const _, 2) };
+    drop(ready_w);
+
+    let mut final_env = env_base;
+    for (k, v) in &container_cfg.container_env {
+        final_env.insert(k.clone(), v.clone());
+    }
+    for (k, v) in hook_env {
+        final_env.insert(k, v);
+    }
+
+    // Probe for a shell in the image (busybox/alpine has no /bin/bash), matching
+    // the batch path in executor.rs.
+    let shell = if std::path::Path::new("/bin/bash").exists() {
+        "/bin/bash"
+    } else {
+        "/bin/sh"
+    };
+    let c_shell = std::ffi::CString::new(shell)
+        .unwrap_or_else(|_| std::ffi::CString::new("/bin/sh").unwrap());
+    let cstr = |s: &str| std::ffi::CString::new(s).unwrap_or_default();
+
+    // NUL bytes were rejected before the fork, so CString::new cannot fail here.
+    let argv: Vec<std::ffi::CString> = match cmd {
+        StepChildCommand::Script {
+            script_path,
+            entrypoint,
+        } => match entrypoint {
+            Some(ep) => vec![
+                c_shell.clone(),
+                cstr("-c"),
+                cstr(&format!("{ep} && {shell} {script_path}")),
+            ],
+            None => vec![c_shell.clone(), cstr(&script_path)],
+        },
+        StepChildCommand::Shell {
+            cmdline,
+            entrypoint,
+        } => match (cmdline, entrypoint) {
+            (None, None) => vec![c_shell.clone()],
+            (Some(cmd), None) => vec![c_shell.clone(), cstr("-c"), cstr(&format!("exec {cmd}"))],
+            (None, Some(ep)) => {
+                vec![
+                    c_shell.clone(),
+                    cstr("-c"),
+                    cstr(&format!("{ep} && exec {shell}")),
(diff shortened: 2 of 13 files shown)


docs/user-guide/running-containers.rst
```diff
@@ -153,6 +153,25 @@ GPU allocation, bind mounts, and cancellation apply per step: a cancelled step
...
+
+.. code-block:: bash
+
+   salloc -N1 --gpus-per-node=8
+   # ...now inside the allocation shell:
+   srun --pty --container-image trainer.sqsh bash   # interactive shell in the container
+
+   # or inherit the allocation's image:
+   salloc -N1 --container-image trainer.sqsh
+   srun --pty bash                                   # same container, interactive
```
shiv-tyagi: `salloc --container-image` is not a valid flag. `SallocArgs` defines no container options, and the PR description itself notes salloc has no `--container-image` flag, so this example errors. The reachable inherited case is attaching to a running containerized `sbatch` job:

```suggestion
   # or attach an interactive shell to a running containerized job
   # (enters its container via nsenter):
   sbatch -N1 --container-image trainer.sqsh long_job.sh
   srun --jobid=<id> --overlap --pty bash            # same container, interactive
```

docs/user-guide/running-containers.rst
```diff
@@ -153,6 +153,25 @@ GPU allocation, bind mounts, and cancellation apply per step: a cancelled step
 tears down its container and cleans up its rootfs without affecting the rest of
 the allocation.
 
+Interactive steps (``srun --pty``) run inside the container too — you get a real
+terminal inside the container rather than on the host. The step uses its own
+``--container-image`` when given one, otherwise it inherits the image from a
+containerized ``salloc``/``sbatch`` allocation:
```
shiv-tyagi: A `salloc` allocation can't carry a container image (no flag exists), so it's never a source of an inherited image. Only a nested `srun --pty` inside a containerized `sbatch` job inherits.

```suggestion
``--container-image`` when given one, otherwise a nested ``srun --pty`` inherits
the image from its containerized ``sbatch`` job:
```

docs/user-guide/running-containers.rst
```diff
@@ -153,6 +153,25 @@ GPU allocation, bind mounts, and cancellation apply per step: a cancelled step
...
+
+   salloc -N1 --gpus-per-node=8
+   # ...now inside the allocation shell:
+   srun --pty --container-image trainer.sqsh bash   # interactive shell in the container
+
+   # or inherit the allocation's image:
+   salloc -N1 --container-image trainer.sqsh
+   srun --pty bash                                   # same container, interactive
+
+A ``--pty`` step nested inside a containerized ``sbatch``/``salloc`` job enters
```
shiv-tyagi: `salloc` can't be containerized, drop it here.

```suggestion
A ``--pty`` step nested inside a containerized ``sbatch`` job enters
```

crates/spurd/src/agent_server.rs
```diff
@@ -4730,6 +5191,37 @@ impl AgentService {
...
+    /// spurd's environment, secrets and all), so read the container's workload
+    /// (the shepherd's child, PID 1 inside the container) instead. For a host job
+    /// the tracked pid is the workload itself. Callers must `env_clear()` first so
+    /// spurd's own environment is not inherited on top of this.
+    fn session_environ(entry: &crate::job_entry::JobEntry) -> Vec<(String, String)> {
+        if entry.pid <= 0 {
+            return Vec::new();
+        }
+        let target = if entry.has_namespaces() {
+            Self::container_workload_pid(entry.pid as u32).unwrap_or(entry.pid as u32)
```
shiv-tyagi, line 5195: Q: this fallback fails open. If `container_workload_pid` returns `None` (empty `children`, a race before the shepherd forks init, or a kernel without `CONFIG_PROC_CHILDREN`), it reads the shepherd's `/proc/environ`, which is spurd's environment, the exact leak this PR closes. Should a container job fail closed here (return empty) rather than fall back to the shepherd's env?

tests/native_host/e2e/test_srun_container.py
```diff
@@ -685,3 +685,173 @@ def test_step_without_image_runs_on_host(self, step_container_cluster):
         ])
         assert code == 0, f"host srun step failed (exit {code}):\n{out}"
         assert "NOTFOUND" in out, f"marker found on host — assumption violated:\n{out}"
+
+
+# ---------------------------------------------------------------------------
+# Interactive PTY steps: `srun --pty --container-image` (issue #780 x #777)
```
shiv-tyagi: Nit: drop the issue refs (AGENTS.md: no issue/PR numbers in comments).

```suggestion
# Interactive PTY steps: `srun --pty --container-image`
```

shiv-tyagi (commented):
Solid, well-factored refactor. The shared container-fork core and the non-interactive drain fix both read cleanly, and the test coverage is thorough. Findings are mostly in the docs (a `salloc --container-image` flag that doesn't exist), plus one question on the daemon-env fallback.

proto/slurm.proto
```diff
@@ -1314,6 +1314,8 @@ message InitSession {
   repeated string argv = 6;  // Command to run (empty = default shell)
   map<string, string> env = 7;
   string user = 8;           // Requesting user; verified identity must own the job; the controller/admin credential overrides
+  ContainerSpec container = 9;  // Effective container the step runs in (empty image = host); resolved by the controller
+  bool non_interactive = 10;    // Client stdin is not a TTY: closing the input stream is stdin-EOF, not a hangup, so the agent drains output and lets the command finish instead of SIGHUP-ing it
```
shiv-tyagi: Nit: trim this to the essential contract.

```suggestion
  bool non_interactive = 10;    // Client stdin is not a TTY: input-stream close is stdin-EOF (drain output), not a hangup (SIGHUP)
```

crates/spur-cli/src/interactive.rs
```diff
@@ -192,7 +194,14 @@ pub async fn open_interactive_session(
     winsize: spur_proto::proto::WindowSize,
     overlap: bool,
     user: &str,
+    container: Option<spur_proto::proto::ContainerSpec>,
 ) -> std::result::Result<InteractiveSessionHandle, tonic::Status> {
+    // When our stdin is not a TTY (a script, redirected input, a pipe), the
+    // client closes its input stream as soon as stdin hits EOF. Tell the agent so
+    // it treats that as stdin-EOF (drain output, let the command finish) rather
+    // than a hangup (SIGHUP). Interactive clients keep the hangup-on-disconnect
+    // behavior.
```
shiv-tyagi: Nit: five lines for this is heavy. Tighten:

```suggestion
    // Non-TTY stdin (script, pipe, redirect) closes the input stream on EOF, not
    // on hangup. Flag it so the agent drains output instead of SIGHUP-ing the
    // step; interactive clients keep hangup-on-disconnect.
```

crates/spurd/src/agent_server.rs
```diff
@@ -4438,20 +4634,36 @@ impl AgentService {
...
     /// (stdin, resize, signal), and drains remaining output after child exit.
-    async fn run_pty_bridge<S>(
+    ///
+    /// `interactive` decides what closing the client's input stream means. An
+    /// interactive client (stdin is a TTY) only closes it when the terminal goes
+    /// away, so that is a hangup: SIGHUP the step and stop. A non-interactive
+    /// client (script/pipe/redirect) closes it as soon as stdin hits EOF while it
+    /// keeps reading output, so that is just stdin-EOF: stop forwarding input but
+    /// keep draining until the command finishes. Without this, a `--pty` step run
+    /// non-interactively is SIGHUP'd before its output is flushed.
```
shiv-tyagi: Nit: seven lines restating the same distinction. Tighten:

```suggestion
    /// `interactive` sets what closing the client's input stream means: for an
    /// interactive client (TTY) it is a hangup (SIGHUP the step and stop); for a
    /// non-interactive one (script/pipe/redirect) it is stdin-EOF, so stop
    /// forwarding input but keep draining until the command exits.
```

crates/spurd/src/agent_server.rs
```diff
@@ -4438,20 +4634,36 @@ impl AgentService {
...
     {
         use crate::pty::WindowSize as PtyWinSize;
         use std::os::fd::AsRawFd;
         use tokio::io::unix::AsyncFd;
         use tokio_stream::StreamExt;
 
+        // The workload's exit is delivered by `wait_exit`: a `tokio::process::Child`
+        // wait for the host/nsenter path, or a blocking `waitpid` for a raw-forked
+        // container step. Pinned so it can be polled in the select and, if the loop
+        // exits first (EOF/disconnect), awaited once more to reap.
```
shiv-tyagi: Nit: the host-vs-container mechanics are already spelled out where `wait_exit` is built; here only the pin rationale matters.

```suggestion
        // Pinned so `wait_exit` can be polled in the select and, if the loop exits
        // first (EOF/disconnect), awaited once more to reap.
```

crates/spurd/src/agent_server.rs
```diff
@@ -4730,6 +5191,37 @@ impl AgentService {
...
             .collect()
     }
+
+    /// The environment a session that *enters* a running job should start from —
+    /// the job's own environment, never spurd's. For a container job the tracked
+    /// pid is the namespace shepherd (a fork of spurd whose `/proc/environ` is
+    /// spurd's environment, secrets and all), so read the container's workload
+    /// (the shepherd's child, PID 1 inside the container) instead. For a host job
+    /// the tracked pid is the workload itself. Callers must `env_clear()` first so
+    /// spurd's own environment is not inherited on top of this.
```
shiv-tyagi: Nit: tighten.

```suggestion
    /// The environment a session that *enters* a running job starts from: the
    /// job's own, never spurd's. For a container job the tracked pid is the
    /// shepherd (a spurd fork carrying spurd's env), so read the container's
    /// workload (PID 1) instead. Callers must `env_clear()` first.
```

crates/spurd/src/agent_server.rs
```diff
@@ -4611,6 +4835,240 @@ impl AgentService {
...
     }
 
+    /// Launch an interactive PTY step inside a fresh container rootfs. The
+    /// effective image is resolved controller-side (the step's own
+    /// `--container-image`, else inherited from a containerized `salloc`/`sbatch`
+    /// allocation). Mirrors `run_command`'s Case 2, but wires the child's stdio to
+    /// a PTY slave (so the workload owns a real terminal) and returns the master
+    /// fd for [`Self::run_pty_bridge`] instead of collecting spooled output.
+    /// Returns `(pty_master, child_pid, rootfs_guard)`; the caller holds the guard
+    /// for the session's lifetime so the rootfs is torn down when it ends.
```
shiv-tyagi: Nit: tighten, and drop the `salloc` inheritance claim (salloc carries no image; the image is whatever the controller resolved).

```suggestion
    /// Launch an interactive PTY step in a fresh container rootfs, using the
    /// controller-resolved effective image. Like `run_command`'s Case 2, but wires
    /// the child's stdio to a PTY slave and returns the master fd for
    /// [`Self::run_pty_bridge`]. The caller holds the returned rootfs guard for the
    /// session so the rootfs is torn down when it ends.
```

crates/spurd/src/agent_server.rs
```diff
@@ -2762,6 +2853,16 @@ impl SlurmAgent for AgentService {
             // host's mount namespace, where a job's work_dir need not exist.
             cmd.current_dir(&entry.work_dir);
         }
+        // Start from an empty environment so spurd's own environment (secrets
+        // included) never leaks into the exec'd command; then apply the job's own.
```
shiv-tyagi: Nit: this same two-line rationale is repeated verbatim at L3410, L3592, and L5119. One line says it:

```suggestion
        // env_clear so spurd's own environment (secrets included) never leaks in.
```
Same trim applies to the other three copies.

shiv-tyagi (commented):
Follow-up: comment tightening. Several new comments restate the same point over 5-7 lines, or repeat the same rationale verbatim across paths. Suggestions below trim them to the 1-3 line norm; one also drops a stale `salloc` inheritance claim.

shiv-tyagi (approved):
LGTM, given this is manually tested. 

We need to rework to the job dispatch paths, step launch and containerisation later.

PR #848, opened 2026-09-08
feat(spurd): implement --container-remap-root (rootless root)

powderluv:
## Summary

Implements #778 (stacked on #840, which rejected the flag as an interim). Replaces the batch-path rejection with a real implementation: a container runs as UID/GID 0 inside a **user namespace** while the host-side identity stays the **unprivileged submitter**, and root-owned image paths (`/root`, HF cache, venv/site-packages, compile caches) are **writable** — the job never becomes host root. This is what root-layout ROCm/AI images need.

**Base this PR on `users/powderluv/fix-778-remap-root` (#840)** so the diff is just the implementation.

## Approach

The per-job rootfs runs in **extracted** mode, so it's `chown`ed to the submitter, then the container runs as UID 0 in a userns mapping `0 → submitter`; container-root then owns every path. The userns is created **last** — after all mount/pivot setup, which runs as root on the (submitter-owned) tree — and its map is written by the **root parent** via a fork handshake, since an unprivileged post-`unshare` process can only self-map its own uid. The child sends its **host-namespace PID** (read from `/proc/self/status` `NSpid` *before* the container's own `/proc` is mounted) so the parent can address `/proc/<pid>/uid_map`.

Validated the primitive in isolation on real hardware before wiring it in; the R&D also proved an idmapped-mount variant (see "Follow-ups").

## Changes

- `container.rs`: `RemapSetup`, `enter_remap_userns`, `host_pid_from_status`, `write_container_id_maps`, `chown_rootfs_to_submitter`; `container_init` takes an optional remap and, for it, drops + enters the userns after setup. `close_inherited_fds` preserves multiple fds.
- `executor.rs`: `launch_container_job` chowns the rootfs, sets up the map-handshake pipes, and writes the container id maps once the child unshares.
- `agent_server.rs`: `has_user_namespace` is true for a remapped root job (so exec/attach enter it with credential preservation); the srun-step path passes `None`.
- `spurctld/server.rs`: drop the batch-submission reject (now implemented); `sbatch` help updated. The **srun-step reject is kept** (that path is a follow-up).

## Testing (shark-a, gfx1201, root spurd)

New E2E `test_container_remap_root.py` — **2/2 pass**:
- without the flag → the job is the submitter and **cannot** write root-owned `/root`;
- with the flag → **`id -u` == 0 inside** and writes to `/root` **succeed**.

`spurd` 276 + `spurctld` 1027 unit tests pass; clippy/fmt clean.

Two issues found + fixed en route: EOVERFLOW from root writing an idmapped mount during setup (resolved by the chown ordering); and the child sending PID 1 instead of its host PID (`NSpid` reads the ns-local pid post-pivot).

## Scope / follow-ups

- **Batch containers only.** srun-step containers stay rejected at submission (#840's srun guard) — a follow-up.
- **Idmapped-mount optimization** to avoid the `chown` cost on large images (the perf-correct path) — needs overlay-mode rootfs + overlay-over-idmap; the validated primitive is recorded for that work.
- A sub-second or first-after-deploy container job races the controller's post-launch dispatch confirmation and requeues (worked around in the E2E with a warmup + trailing sleep) — likely a real product bug worth a separate look.

### DIFF
--- a/crates/spurd/src/agent_server.rs
+++ b/crates/spurd/src/agent_server.rs
@@ -458,20 +458,23 @@ async fn run_containerized_step(
                 libc::dup2(stderr_fd, libc::STDERR_FILENO);
             }
 
-            crate::container::close_inherited_fds(ready_w_fd);
+            crate::container::close_inherited_fds(&[ready_w_fd]);
 
             executor::apply_memlock(memlock);
 
-            let hook_env = match crate::container::container_init(&container_cfg, &rootfs_clone) {
-                Ok(env) => env,
-                Err(e) => {
-                    let msg = format!("E:{e:#}");
-                    unsafe {
-                        libc::write(ready_w_fd, msg.as_ptr() as *const _, msg.len());
+            // Step containers do not support --container-remap-root yet (it is
+            // rejected at submission); pass None.
+            let hook_env =
+                match crate::container::container_init(&container_cfg, &rootfs_clone, None) {
+                    Ok(env) => env,
+                    Err(e) => {
+                        let msg = format!("E:{e:#}");
+                        unsafe {
+                            libc::write(ready_w_fd, msg.as_ptr() as *const _, msg.len());
+                        }
+                        std::process::exit(1);
                     }
-                    std::process::exit(1);
-                }
-            };
+                };
 
             unsafe { libc::write(ready_w_fd, b"OK".as_ptr() as *const _, 2) };
             drop(ready_w);
--- a/crates/spurd/src/agent_server.rs
+++ b/crates/spurd/src/agent_server.rs
@@ -2333,6 +2333,16 @@ impl SlurmAgent for AgentService {
         if req.command.is_empty() {
             return Err(Status::invalid_argument("no command specified"));
         }
+        // --container-remap-root is not yet supported for job steps (only batch
+        // containers implement it). Reject rather than silently run as the
+        // submitting user: the srun CLI already rejects it, but a direct gRPC
+        // client would otherwise bypass that guard.
+        if req.container.as_ref().is_some_and(|c| c.remap_root) {
+            return Err(Status::invalid_argument(
+                "--container-remap-root is not yet supported for job steps; run it as a \
+                 batch container job (sbatch --container-image ... --container-remap-root)",
+            ));
+        }
         // Steps carry their own uid straight from the wire — gate them exactly like a batch launch.
         if let Err(msg) = crate::privdrop::check_root_execution_allowed(
             req.uid,
@@ -5214,6 +5224,38 @@ mod tests {
         assert!(
             err.message().contains("not found"),
             "expected an image-resolution failure, got: {}",
+            err.message()
+        );
+    }
+
+    #[tokio::test]
+    async fn run_command_rejects_container_remap_root_on_steps() {
+        // --container-remap-root is batch-only; a step that sets it (even a
+        // direct gRPC client bypassing the srun CLI guard) must be refused, not
+        // silently run as the submitting user.
+        let (svc, job_id) = run_command_test_setup().await;
+        let req = Request::new(RunCommandRequest {
+            command: vec!["echo".into(), "hi".into()],
+            uid: 0,
+            gid: 0,
+            work_dir: String::new(),
+            environment: HashMap::new(),
+            job_id,
+            container: Some(spur_proto::proto::ContainerSpec {
+                image: "/some/image.sqsh".into(),
+                remap_root: true,
+                ..Default::default()
+            }),
+            ..Default::default()
+        });
+        let err = svc
+            .run_command(req)
+            .await
+            .expect_err("a step with --container-remap-root must be rejected");
+        assert_eq!(err.code(), tonic::Code::InvalidArgument);
+        assert!(
+            err.message().contains("--container-remap-root"),
+            "rejection should name the flag, got: {}",
             err.message()
         );
     }
--- a/crates/spurd/src/container.rs
+++ b/crates/spurd/src/container.rs
@@ -1009,6 +1009,102 @@ fn setup_user_namespace(uid: u32, gid: u32) -> anyhow::Result<()> {
     Ok(())
 }
 
+// ---- Rootless-root (--container-remap-root) ---------------------------------
+//
+// Runs an image as UID 0 inside a user namespace while the host-side identity is
+// the unprivileged submitter, AND keeps root-owned image paths (/root, caches,
+// site-packages) writable. The extracted rootfs is a private per-job tree, so
+// the caller chowns it to the submitter; the container then runs as UID 0 in a
+// user namespace mapping 0 -> submitter, so container-root owns every path and
+// writes succeed — while the process is never host root. The user namespace is
+// created last (after all mount setup) and its map is written by the (root)
+// parent, since an unprivileged post-unshare process cannot self-map an
+// arbitrary host id (validated on hardware; see the parent/child dance in
+// `container_init` + executor's `launch_container_job`).
+//
+// (An idmapped-mount variant would avoid the recursive chown for large images,
+// but needs overlay-mode rootfs + overlay-over-idmap handling — a perf follow-up.)
+
+/// Write `0 <id> 1` uid/gid maps for `pid` (setgroups denied first). The caller
+/// must be root (it maps an arbitrary host id) AND must only call this AFTER the
+/// target has finished `unshare(CLONE_NEWUSER)` — writing earlier fails EPERM.
+pub fn write_container_id_maps(pid: i32, uid: u32, gid: u32) -> anyhow::Result<()> {
+    std::fs::write(format!("/proc/{pid}/uid_map"), format!("0 {uid} 1\n"))
+        .with_context(|| format!("write /proc/{pid}/uid_map"))?;
+    // setgroups must be denied before gid_map; ignore ENOENT on old kernels.
+    let _ = std::fs::write(format!("/proc/{pid}/setgroups"), "deny");
+    std::fs::write(format!("/proc/{pid}/gid_map"), format!("0 {gid} 1\n"))
+        .with_context(|| format!("write /proc/{pid}/gid_map"))?;
+    Ok(())
+}
+
+/// Recursively chown a private per-job extracted rootfs to the submitter, so a
+/// container-root process mapped to that uid owns every path and its writes
+/// succeed. Root-only; runs before the container forks.
+pub fn chown_rootfs_to_submitter(rootfs: &Path, uid: u32, gid: u32) -> anyhow::Result<()> {
+    // `-h` chowns symlinks themselves (never dereferences), matching how the
+    // files were owned in the image. The rootfs is a private per-job tree.
+    let status = std::process::Command::new("chown")
+        .arg("-Rh")
+        .arg(format!("{uid}:{gid}"))
+        .arg(rootfs)
+        .status()
+        .with_context(|| format!("run chown on {}", rootfs.display()))?;
+    if !status.success() {
+        bail!(
+            "chown -Rh {}:{} {} failed ({status})",
+            uid,
+            gid,
+            rootfs.display()
+        );
+    }
+    Ok(())
+}
+
+/// What `container_init` needs to run a `--container-remap-root` job. Built by
+/// the (root) parent before the fork; the fds are inherited across it.
+pub struct RemapSetup {
+    /// Child -> parent: after `unshare(CLONE_NEWUSER)`, the child writes its PID
+    /// as seen in the host pid namespace (4 bytes, native endian) so the parent
+    /// can address `/proc/<pid>/uid_map`.
+    pub map_req_fd: RawFd,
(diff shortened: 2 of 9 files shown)


crates/spurd/src/container.rs
```diff
@@ -1009,6 +1009,102 @@ fn setup_user_namespace(uid: u32, gid: u32) -> anyhow::Result<()> {
     Ok(())
 }
 
+// ---- Rootless-root (--container-remap-root) ---------------------------------
+//
```
shiv-tyagi, line 1013: Nit: this 16-line block reads like the PR description. Trim to the non-obvious constraint (userns created last, mapped by the root parent); the rest belongs in the PR body.

crates/spurd/src/container.rs
```diff
@@ -1009,6 +1009,102 @@ fn setup_user_namespace(uid: u32, gid: u32) -> anyhow::Result<()> {
...
+        .with_context(|| format!("write /proc/{pid}/gid_map"))?;
+    Ok(())
+}
+
+/// Recursively chown a private per-job extracted rootfs to the submitter, so a
+/// container-root process mapped to that uid owns every path and its writes
+/// succeed. Root-only; runs before the container forks.
+pub fn chown_rootfs_to_submitter(rootfs: &Path, uid: u32, gid: u32) -> anyhow::Result<()> {
+    // `-h` chowns symlinks themselves (never dereferences), matching how the
+    // files were owned in the image. The rootfs is a private per-job tree.
```
shiv-tyagi, line 1046: Nit: drop the second sentence, it repeats the section header and the fn doc.
```
    // `-h` chowns the symlink itself, matching image ownership.
```

crates/spurd/src/container.rs
```diff
@@ -1042,7 +1138,7 @@ fn fork_into_pid_namespace() -> anyhow::Result<()> {
 ///
 /// Prevents gRPC sockets, other jobs' output files, etc. from leaking
 /// into the container process.
-pub fn close_inherited_fds(preserve_fd: RawFd) {
+pub fn close_inherited_fds(preserve_fds: &[RawFd]) {
```
shiv-tyagi, line 1141: Nit: the doc above still says `preserve_fd (the sync pipe)`; the param is now `preserve_fds: &[RawFd]`. Update it.

crates/spurd/src/container.rs
```diff
@@ -1076,7 +1173,11 @@ pub fn container_init(
         }
     }
 
-    if is_root {
+    if is_root || remap.is_some() {
+        // Root path (including remap): create the mount + pid namespaces as root.
+        // For remap the USER namespace is created last (below), after every mount
```
shiv-tyagi, line 1178: Nit: trim to the why.
```
        // userns is created last: a userns child can't change propagation of
        // inherited mounts (EPERM).
```

crates/spurd/src/container.rs
```diff
@@ -1087,6 +1188,18 @@ pub fn container_init(
     set_mount_propagation_private()?;
     fork_into_pid_namespace()?;
 
+    // Remap: read our host-namespace PID NOW, while the host /proc is still
```
shiv-tyagi, line 1191: Nit: trim this block, and drop the one at L1200-1202 (it just restates the section header).
```
    // Read the host-ns PID before mount_filesystems remounts /proc; after that
    // /proc/self/status reports only the in-ns PID (1).
```

crates/spurd/src/executor.rs
```diff
@@ -1744,6 +1744,27 @@ async fn launch_container_job(
         -1
     };
 
+    // --container-remap-root (root spurd only): chown the private per-job rootfs
+    // to the submitter so the container-root it maps to owns every path, and set
+    // up the parent<->child map-handshake pipes BEFORE the fork so the fds are
```
shiv-tyagi, line 1749: Nit: 6 lines, and the "child can't self-map" rationale duplicates the container.rs header. Trim.
```
    // Root spurd only: chown the per-job rootfs to the submitter and set up the
    // map-handshake pipes before fork so the child inherits the fds.
```

crates/spurd/src/executor.rs
```diff
@@ -1855,6 +1885,29 @@ async fn launch_container_job(
                 debug!("pidfd_open unavailable, falling back to raw PID tracking");
             }
 
+            // --container-remap-root handshake: the container (PID 1 in its pid
```
shiv-tyagi, line 1888: Nit: trim.
```
            // Remap handshake: child sends its host PID after unshare(userns); we
            // write its 0 -> submitter id maps and release it.
```

crates/spurd/src/agent_server.rs
```diff
@@ -2330,6 +2333,16 @@ impl SlurmAgent for AgentService {
         if req.command.is_empty() {
             return Err(Status::invalid_argument("no command specified"));
         }
+        // --container-remap-root is not yet supported for job steps (only batch
```
shiv-tyagi, line 2336: Nit: trim.
```
        // Steps don't implement remap-root yet. srun rejects it client-side; a
        // direct gRPC client would bypass that, so reject here too.
```

crates/spurd/src/agent_server.rs
```diff
@@ -458,20 +458,23 @@ async fn run_containerized_step(
                 libc::dup2(stderr_fd, libc::STDERR_FILENO);
             }
 
-            crate::container::close_inherited_fds(ready_w_fd);
+            crate::container::close_inherited_fds(&[ready_w_fd]);
 
             executor::apply_memlock(memlock);
```
shiv-tyagi, line 464: Nit: one line is enough.
```
            // Steps don't support remap-root yet (rejected at submission).
```

crates/spurd/src/container.rs
```diff
@@ -1009,6 +1009,102 @@ fn setup_user_namespace(uid: u32, gid: u32) -> anyhow::Result<()> {
...
+    Ok(())
+}
+
+/// Recursively chown a private per-job extracted rootfs to the submitter, so a
+/// container-root process mapped to that uid owns every path and its writes
+/// succeed. Root-only; runs before the container forks.
+pub fn chown_rootfs_to_submitter(rootfs: &Path, uid: u32, gid: u32) -> anyhow::Result<()> {
+    // `-h` chowns symlinks themselves (never dereferences), matching how the
+    // files were owned in the image. The rootfs is a private per-job tree.
+    let status = std::process::Command::new("chown")
```
shiv-tyagi, line 1047: `chown_rootfs_to_submitter` shells out to `chown`. Two issues: it depends on `chown` being on PATH (and forks a process for what `nix::unistd::fchownat` does directly), and `chown -R` clears setuid/setgid bits on every file, so a root-layout image with a setuid binary silently loses it. Q: is stripping setuid across the image acceptable, or should the walk preserve mode bits?

crates/spurd/src/executor.rs
```diff
@@ -1744,6 +1744,27 @@ async fn launch_container_job(
...
+    // --container-remap-root (root spurd only): chown the private per-job rootfs
+    // to the submitter so the container-root it maps to owns every path, and set
+    // up the parent<->child map-handshake pipes BEFORE the fork so the fds are
+    // inherited. Rootless-root cannot be done by the child alone: an unprivileged
+    // post-unshare process may only self-map its own uid, so the (root) parent
+    // must write the container's `0 <submitter> 1` map after the child unshares.
+    let remap = config.remap_root && nix::unistd::geteuid().is_root();
+    let mut remap_pipes: Option<(OwnedFd, OwnedFd, OwnedFd, OwnedFd)> = None;
+    if remap {
+        crate::container::chown_rootfs_to_submitter(&rootfs, cfg.uid, cfg.gid)
```
shiv-tyagi, line 1756: This full-tree chown runs synchronously inside an async fn, blocking the tokio worker for the whole walk (the cost the idmapped-mount follow-up removes). Wrap it in `spawn_blocking`.

crates/spurd/src/container.rs
```diff
@@ -1107,8 +1220,19 @@ pub fn container_init(
...
 
-    if is_root {
-        drop_privileges(config.uid, config.gid, &supplementary_gids)?;
+    match remap {
+        Some(r) => {
+            // Drop to the submitter (carrying the GPU groups) FIRST, so the
+            // keep-groups user namespace preserves host device access, then
+            // enter the userns (mapped to container-root by the parent).
+            drop_privileges(config.uid, config.gid, &supplementary_gids)?;
+            enter_remap_userns(r, remap_host_pid.expect("host pid read for remap"))?;
```
shiv-tyagi, line 1229: `remap_host_pid` is read into an `Option` at L1195 then `.expect()`-ed here. The invariant (Some whenever remap is set) is carried implicitly and defended with a panic. Read the pid in this `Some(r)` arm, or store it on `RemapSetup`, so the type guarantees it and the expect goes away.

crates/spurd/src/container.rs
```diff
@@ -1009,6 +1009,102 @@ fn setup_user_namespace(uid: u32, gid: u32) -> anyhow::Result<()> {
...
+// arbitrary host id (validated on hardware; see the parent/child dance in
+// `container_init` + executor's `launch_container_job`).
+//
+// (An idmapped-mount variant would avoid the recursive chown for large images,
+// but needs overlay-mode rootfs + overlay-over-idmap handling — a perf follow-up.)
+
+/// Write `0 <id> 1` uid/gid maps for `pid` (setgroups denied first). The caller
+/// must be root (it maps an arbitrary host id) AND must only call this AFTER the
+/// target has finished `unshare(CLONE_NEWUSER)` — writing earlier fails EPERM.
+pub fn write_container_id_maps(pid: i32, uid: u32, gid: u32) -> anyhow::Result<()> {
```
shiv-tyagi, line 1031: This duplicates the `0 <id> 1` + setgroups-deny + gid_map sequence in `setup_user_namespace` (L1005-1008). Extract one helper parameterized by the proc path (`/proc/self` vs `/proc/<pid>`).

crates/spurd/src/container.rs
```diff
@@ -1009,6 +1009,102 @@ fn setup_user_namespace(uid: u32, gid: u32) -> anyhow::Result<()> {
...
+    pub map_go_fd: RawFd,
+}
+
+/// This process's PID in the OUTERMOST (host / spurd) pid namespace, which
+/// `/proc/self/status` reports first in `NSpid`. The remap child is PID 1 in its
+/// own nested pid namespace, so `getpid()` is useless to the parent.
+fn host_pid_from_status() -> anyhow::Result<i32> {
+    let status = std::fs::read_to_string("/proc/self/status").context("read /proc/self/status")?;
+    for line in status.lines() {
+        if let Some(rest) = line.strip_prefix("NSpid:") {
```
shiv-tyagi, line 1081: `NSpid`'s first field is the init-namespace pid. The parent addresses `/proc/<pid>` in spurd's own pid-namespace view, so this is correct only when spurd runs in the init pid namespace. Under a namespaced spurd (e.g. a k8s pod) it's the wrong pid. Q: is root-remap out of scope there, or should this pick the entry at spurd's namespace depth?

crates/spurd/src/executor.rs
```diff
@@ -1855,6 +1885,29 @@ async fn launch_container_job(
...
+            if let Some((req_r, req_w, go_r, go_w)) = remap_pipes {
+                // Close the child's ends so our read sees EOF if it dies early.
+                drop(req_w);
+                drop(go_r);
+                let mut pid_bytes = [0u8; 4];
+                let n =
+                    unsafe { libc::read(req_r.as_raw_fd(), pid_bytes.as_mut_ptr() as *mut _, 4) };
+                if n == 4 {
+                    let host_pid = i32::from_ne_bytes(pid_bytes);
+                    let ok = crate::container::write_container_id_maps(host_pid, cfg.uid, cfg.gid)
```
shiv-tyagi, line 1901: Nit: `map_err` for a logging side effect then `.is_ok()` is roundabout. `if let Err(e) = ... { warn!(...) }` reads clearer.

crates/spurd/src/executor.rs
```diff
@@ -1744,6 +1744,27 @@ async fn launch_container_job(
         -1
     };
 
+    // --container-remap-root (root spurd only): chown the private per-job rootfs
+    // to the submitter so the container-root it maps to owns every path, and set
+    // up the parent<->child map-handshake pipes BEFORE the fork so the fds are
+    // inherited. Rootless-root cannot be done by the child alone: an unprivileged
+    // post-unshare process may only self-map its own uid, so the (root) parent
+    // must write the container's `0 <submitter> 1` map after the child unshares.
+    let remap = config.remap_root && nix::unistd::geteuid().is_root();
```
shiv-tyagi, line 1753: On a non-root daemon this makes the flag a silent no-op: no chown, so root-owned image paths stay unwritable. But the sbatch help and docs say a non-root daemon "already runs the container as root inside a user namespace," implying parity. Q: does write-to-/root actually hold on a non-root daemon? If not, scope the docs to a root daemon (the E2E only covers root spurd).

shiv-tyagi (commented):
some comments. PTAL

PR #851, opened 2026-09-08
fix(deploy): mount spur.conf from a Secret

pre:
## What

The example Kubernetes manifests put `spur.conf` in a ConfigMap, and `spur.conf` carries the accounting database password in `accounting.database_url`. `Secret` is a more appropriate resource for credentials.

## Approach

- `examples/k8s/spurctld.yaml` mounts a Secret named `spur-config` instead of the ConfigMap.
- `examples/k8s/configmap.yaml` is gone. The configuration template is now the plain file `examples/k8s/spur.conf`, which the operator fills in and passes to `kubectl create secret generic spur-config --from-file`. A ConfigMap manifest that nothing mounts only invited a `kubectl apply` with no effect.
- The whole file goes into the Secret, not only the password. The configuration loader reads one TOML file and has no separate source for `accounting.database_url` or `auth.jwt_key`. Splitting the secrets out of the file would be a loader feature and is left for a later change.
- `docs/deployment/kubernetes.rst` gets the Secret step before the controller is applied, and the Configuration section explains why the running file is a Secret. `docs/admin-guide/configuration.rst` no longer says followers re-read a ConfigMap.

The e2e harness in `tests/k8s/e2e/` builds its own ConfigMap without accounting and is unchanged.

## Testing

A new `spur-core` test loads `examples/k8s/spur.conf` with the real configuration loader, next to the existing test for `examples/spur.conf`, so the template cannot drift into a file that does not load. Otherwise docs and manifests only.


docs/deployment/kubernetes.rst
```diff
@@ -44,6 +44,18 @@ Apply manifests in order:
    kubectl apply -f examples/k8s/configmap.yaml
```
shiv-tyagi: The apply order still runs `kubectl apply -f examples/k8s/configmap.yaml`, but after this change that manifest is `spur-config-template` and nothing mounts it. Applying it creates an orphan ConfigMap that never gets used, which contradicts the "readable template only" framing. Drop this line so readers don't apply a template that has no runtime effect.

shiv-tyagi (commented):
Some comments, please take a look.

One more, outside the diff so I couldn't anchor it inline: `docs/admin-guide/configuration.rst:111` still says followers "re-read the same ConfigMap". Update it to "Secret" to stay consistent with this change.

shiv-tyagi:
Thanks for answering my questions. @sgopinath1 @sajmera-pensando Please approve and merge if this looks good.

sgopinath1 (approved):
LGTM

PR #853, opened 2026-09-08
test(e2e): wait for registered nodes and reuse an existing CRD

pre:
Two faults in the Kubernetes e2e harness, both seen on a real three-node cluster. Neither is caused by the code under test.

## What this fixes

**The fixture returns before the controller has a node.** `ClusterFixture.deploy` waits until the controller Pods are ready and returns. The e2e manifests run no `spurd`; the operator registers the Kubernetes nodes with the controller over its own watch, which happens after the controller is ready. The first test that submits a `SpurJob` on a new cluster then hits the 60 s job timeout while the controller still has no node, and every later test passes. Measured at about three minutes on a fresh cluster. This explains `test_simple_spurjob_completes`, `test_state_survives_leader_failover` and `test_new_leader_accepts_writes` failing on a clean cluster and passing on a rerun.

**`apply_crd` fails when the CRD already exists.** The fallback from `create` to `replace` sends the manifest as is, but a replace needs the live `metadata.resourceVersion`:

```
ApiException: (422) ... metadata.resourceVersion: Invalid value: 0: must be specified for an update
```

CI never hits this path because its cluster is new. It fails on any cluster that already holds the CRD, such as a rerun after a partial teardown.

## Approach

`wait_ready` now waits, after the controller Pods, until the controller lists at least one node. The list is read with `spur nodes -N -h` inside `spurctld-0`, which a follower answers from local state, so the leader does not matter. The wait has its own timeout, `NODE_REGISTER_TIMEOUT`, instead of a larger `DEFAULT_TIMEOUT`, which guards every other wait in the suite and would hide slow tests.

`apply_crd` reads the live CRD on `AlreadyExists`, copies its `resourceVersion` into the manifest, replaces, and tries once more on a conflict, because the version can move between the read and the write. The create path for a new cluster is unchanged. When the live CRD is still terminating, because the previous session's teardown has not finished, `apply_crd` waits until it is gone and creates it anew; a replace would otherwise land on an object that vanishes during the run, after which every SpurJob call answers `404 page not found`. This was seen in every second back-to-back run on the test cluster.

The delay itself is an operator fault: its connect to the client Service hangs for the kernel's SYN retry time when the Service has no endpoint yet. That is fixed in https://github.com/ROCm/spur/pull/854. This change is still needed: the fixture must not return before the controller has a node, and with that fix the wait is seconds.

## Testing

`uvx flake8 --config tests/.flake8 tests/` is clean.

Run on a three-node k0s cluster built by `spur k8s up`, with the CRD applied before the suite so that the replace path runs: `test_spurjob.py` and `test_raft_ha.py`, twice in a row: 12 of 13 passed on the first run and 13 of 13 on the second. The one failure was `test_new_leader_accepts_writes`, which the operator caused (a stalled channel to a killed controller Pod, an operator fault fixed in https://github.com/ROCm/spur/pull/854), not the harness. Before this change the same two files failed the three tests named above on a new cluster.

CI cannot exercise the replace path, because its cluster is new.

### DIFF
--- a/tests/k8s/e2e/k8s_cluster.py
+++ b/tests/k8s/e2e/k8s_cluster.py
@@ -28,9 +28,13 @@ _DEFAULT_IMAGE = "ghcr.io/rocm/spur:ci"
 SPUR_JOB_GROUP = "spur.amd.com"
 SPUR_JOB_VERSION = "v1alpha1"
 SPUR_JOB_PLURAL = "spurjobs"
+CRD_NAME = f"{SPUR_JOB_PLURAL}.{SPUR_JOB_GROUP}"
 DEFAULT_TIMEOUT = 60
 WAIT_INTERVAL = 2
 HA_TIMEOUT = 90
+# The operator registers the nodes only after it reaches a ready controller;
+# measured at about three minutes after the controller Pods are ready.
+NODE_REGISTER_TIMEOUT = 300
 CLUSTER_SCOPED_KINDS = frozenset({"ClusterRole", "ClusterRoleBinding"})
 
 
@@ -93,6 +97,10 @@ def _is_not_found(exc: ApiException) -> bool:
 
 
 def _is_already_exists(exc: ApiException) -> bool:
+    return exc.status == 409
+
+
+def _is_conflict(exc: ApiException) -> bool:
     return exc.status == 409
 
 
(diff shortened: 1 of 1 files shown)


shiv-tyagi (approved):
LGTM

PR #854, opened 2026-09-08
fix(spur-k8s): fail fast on a controller that does not answer

pre:
## What this fixes

The operator's channel to `spurctld` had no bound of any kind, and two things stalled it for minutes on a real cluster.

**The first registration of the nodes came three minutes late.** On a new cluster the operator starts before the controller is ready, when the client Service has no endpoint. The SYN is dropped and the connect hangs for the kernel's SYN retry time, more than two minutes, before it fails with a transport error. Only then does the task's retry loop open a new connection, which succeeds at once, and the registration follows within milliseconds. Every `SpurJob` submitted in that window waited for a node that did not exist. From the operator log:

```
10:40:53 starting K8s node watcher
10:41:07 (spurctld-0 ready)
10:43:06 task failed, retrying name=node watcher error=transport error delay_secs=1
10:43:07 starting K8s node watcher
10:43:07 registering K8s node ...
```

**A killed controller Pod stalled the job controller for a minute or two.** The job controller keeps one channel open. When the Pod behind it was killed, the packets to that peer were dropped and every RPC on the channel waited for the kernel to give up on it. A `SpurJob` created meanwhile was submitted only when the connection finally failed, 70 to 110 s after the kill. This is the failure of `test_new_leader_accepts_writes` after a leader kill:

```
11:36:05 (leader Pod killed; the job controller's channel led to it)
11:37:13 heartbeat failed: Timeout expired
11:37:44 failed to report job status: Timeout expired
11:37:54 heartbeat failed: tcp connect error   <- the dead connection is finally gone
11:37:54 SpurJob submitted it-pre-failover      <- 64 s after it was created
```

## Approach

One helper, `controller::connect`, opens the channel with a 10 s connect timeout, HTTP/2 keepalive pings, and a 30 s bound on each request. No RPC on this channel streams. The node watcher, the job controller and the heartbeat sender use it; each had its own copy of the URL prefixing and the connect.

The request bound alone was not enough: it cancels the request but leaves the dead connection in place, and the next request waits on it again, 30 s at a time. The job controller's long-lived client is now a small wrapper, `ControllerClient`, that drops its channel after a transport error (`UNAVAILABLE`, `UNKNOWN`, `CANCELLED`, `DEADLINE_EXCEEDED`, `INTERNAL`) and connects again on the next call. A refusal such as `NOT_FOUND` is an answer and keeps the channel. The heartbeat sender already connects per tick, and the node watcher restarts through its retry loop.

The node watcher also ends on a registration that fails with a transport error instead of logging it. It stored the node's resource fingerprint before the call, so a node whose first registration failed was not tried again until its resources changed or the watcher restarted for another reason. Returning the error restarts the watcher through the existing retry loop, which lists every node again; a repeat registration is accepted. A refusal that is not a transport error, such as a missing admission token, no longer restarts the watcher: it is logged and the node's next event retries it, so a permanent refusal cannot stop taint and drain sync for every node. The fingerprint is stored only after the registration succeeds.

The job controller no longer connects before its first reconcile; `ControllerClient::new` starts with no channel and the first call opens it, so a controller that is not ready at start does not fail the task into its backoff. The readiness probe uses a plain `Endpoint::connect` with the same 10 s connect timeout as the channels and nothing else. kubelet's own `timeoutSeconds` (1 s by default, neither `examples/k8s/operator.yaml` nor the e2e manifest sets it) decides readiness; the connect timeout only frees the axum handler, which a dropped SYN would otherwise hold for the kernel's SYN retry window while kubelet keeps probing every 10 s.

## Testing

`cargo clippy --workspace --exclude spur-ffi --all-targets --locked` and `cargo test --locked` are clean. Unit tests cover the URL prefixing, the transport-error classification, that the wrapper drops its channel after a transport error and keeps it after a refusal, that the next call after a transport error reconnects, that a refused connect answers `UNAVAILABLE` and leaves no channel, and which registration failures restart the node watcher. One more test opens a channel to a socket that accepts the connection and never answers, which is what a killed Pod looks like until the kernel gives up, and runs the call on tokio's paused clock so the real bounds apply without waiting: the call fails with `UNAVAILABLE: http2 error`, the keepalive having closed the connection before the request bound, and the channel is dropped. The connect timeout has no unit seam, because loopback cannot drop a SYN; it is verified on a cluster.

On a three-node k0s cluster built by `spur k8s up`, with an image carrying this change:

- The operator registered the nodes 9 s after the controller became ready, with the connect failing every 10 s until then instead of once after 133 s.
- A killed leader Pod no longer stalls the job controller: the stalled report call times out, the channel is dropped, and the waiting SpurJob is submitted at once (run 7 below).

### Cluster runs with `test_raft_ha.py`

Three-node k0s cluster built by `spur k8s up`, one control plane and two workers, with #853 merged, runs back to back on the same cluster. The operator build names which part of this change the image carried; the controller image was the same throughout.

| Run | Operator build | Passed | Failed test | Cause |
|---|---|---|---|---|
| 1 | connect timeout only | 5 of 6 | `test_new_leader_accepts_writes` | job controller's channel pinned to the killed Pod; RPCs waited for the kernel |
| 2 | connect timeout only | 6 of 6 | | |
| 3 | + keepalive, request bound | 5 of 6 | `test_new_leader_accepts_writes` | channel pinned to the killed Pod; each RPC hit the 30 s bound and the next one reused the dead connection |
| 4 | + keepalive, request bound | 4 of 6 | `test_state_survives_leader_failover`, `test_new_leader_accepts_writes` | harness: the previous session's CRD was still terminating, `404 page not found` (fixed in #853) |
| 5 | + keepalive, request bound | 5 of 6 | `test_new_leader_accepts_writes` | channel pinned to the killed Pod, as in run 3 |
| 6 | + channel drop after a transport error | 4 of 6 | same two as run 4 | harness, CRD terminating |
| 7 | + channel drop after a transport error | 6 of 6 | | log shows the stalled report call time out, the channel dropped, and the waiting SpurJob submitted at once |
| 8 | + channel drop after a transport error | 4 of 6 | same two as run 4 | harness, CRD terminating |
| 9 | + channel drop after a transport error | 4 of 6 | `test_state_survives_leader_failover`, `test_new_leader_accepts_writes` | after the previous test's Pod kill a follower answered `cannot reach leader` during the election, and the job controller's error backoff had doubled to 32 s; not covered by this change, see below |
| 10 | + channel drop after a transport error | 0 of 6 | all, in setup | harness: the CRD stayed terminating for over 120 s because of leftover objects on the shared cluster |

Two focused runs of `test_state_survives_leader_failover` and `test_new_leader_accepts_writes` alone with the keepalive build: one failed on the terminating CRD, one passed 2 of 2.

The dead-channel failure, runs 1, 3 and 5, did not appear again once the channel is dropped after a transport error. The remaining failures are the harness race, fixed in #853, and the backoff after a Pod kill:

One thing this change does not cover, seen in run 9: right after a Pod kill, a follower answers `cannot reach leader` while the election runs, and the job controller's own error backoff doubles to 32 s and beyond, so a `SpurJob` created in that moment is submitted after the test's 60 s window. That is a separate matter of the backoff and is left for a follow-up.

Related:
- https://github.com/ROCm/spur/pull/853 makes the e2e fixture wait for a registered node.
- https://github.com/ROCm/spur/pull/809 re-registers a node the controller has forgotten; this change covers the first registration, which no heartbeat precedes.


crates/spur-k8s/src/controller.rs
```diff
@@ -0,0 +1,246 @@
...
+    // needless reconnect on an application Internal is harmless.
+    matches!(
+        status.code(),
+        Code::Unavailable
+            | Code::Unknown
+            | Code::Cancelled
+            | Code::DeadlineExceeded
+            | Code::Internal
+    )
+}
```
shiv-tyagi: `is_transport_error` here and `controller_rpc_retryable` in spur-proto both classify controller transport errors, and they differ only in `CANCELLED`. Two predicates that must agree but don't is what lets a timed-out `submit_job` (which surfaces as `CANCELLED`) fall through to the reject path in `job_controller.rs`. Collapse to one: have `controller_rpc_retryable` delegate to this, or drop it and use `is_transport_error` everywhere.

crates/spur-k8s/src/controller.rs
```diff
@@ -0,0 +1,246 @@
...
+
+/// The readiness probe only asks whether the controller accepts a connection,
+/// and kubelet bounds the probe itself, so it needs none of the channel bounds.
+pub async fn probe(addr: &str) -> anyhow::Result<()> {
+    Endpoint::from_shared(controller_url(addr))?
+        .connect()
+        .await?;
+    Ok(())
+}
```
shiv-tyagi, line 58: `probe` sets no connect timeout, so readiness relies entirely on kubelet's default `timeoutSeconds` of 1 s to bound a dropped SYN. That holds today since the manifests don't override it, but a probe that can otherwise hang for the kernel's SYN-retry window is a sharp edge for a one-line guard.
```suggestion
pub async fn probe(addr: &str) -> anyhow::Result<()> {
    Endpoint::from_shared(controller_url(addr))?
        .connect_timeout(CONNECT_TIMEOUT)
        .connect()
        .await?;
    Ok(())
}
```

crates/spur-k8s/src/node_watcher.rs
```diff
@@ -105,7 +108,7 @@ pub async fn run(
 
     info!(selector = %label_selector, "starting K8s node watcher");
 
-    let mut ctrl_client = connect_controller(&controller_addr).await?;
+    let mut ctrl_client = connect(&controller_addr).await?;
```
shiv-tyagi: This is a raw `SlurmControllerClient` from a single `connect`. Unlike the job controller's `ControllerClient`, it never drops a dead channel. `sync_taint_state` (L168) and the `Delete` arm's `update_node` (L186) reuse it, so after a controller failover those calls hit the same dead channel this PR fixes for the job controller, and nothing here reconnects. The only recovery is a `register_agent` transport error restarting the watcher, so a taint change on a node whose fingerprint is unchanged keeps failing until some registration event fires. Route the taint and delete calls through the reconnecting path, or wrap this client too.

crates/spur-k8s/src/node_watcher.rs
```diff
@@ -193,6 +199,14 @@ pub async fn run(
     Ok(())
 }
 
+/// A transport error means the controller may be gone; ending the watcher lets
+/// its retry loop open a fresh channel and list every node again. Any other
+/// refusal, such as a missing admission token, would repeat on every restart
+/// and stop taint and drain sync for all nodes, so it is logged instead.
+fn registration_failure_restarts_watcher(status: &Status) -> bool {
+    is_transport_error(status)
+}
```
shiv-tyagi: Nit: `registration_failure_restarts_watcher` is a pure alias for `is_transport_error`. The rationale it documents is real, but the wrapper plus its own test add indirection for a one-liner. Inline the call and keep the rationale as a comment on the match arm.

shiv-tyagi (commented):
PTAL at the comments.

crates/spur-k8s/src/controller.rs
pre: Agreed, and it was worse than a divergence: the 30 s request bound this PR adds surfaces as CANCELLED, so a timed-out `submit_job` went to the reject path and the SpurJob was Failed for good. spur-proto cannot depend on spur-k8s, so the second option: 56b43d8 adds CANCELLED to `controller_rpc_retryable`, deletes `is_transport_error`, and `ControllerClient::call` uses the shared predicate. spurd picks up the same widening for its completion and drain retries, three attempts on a fresh channel, the behaviour it already had for UNAVAILABLE.

crates/spur-k8s/src/controller.rs
pre: Done in 595b15b, with `CONNECT_TIMEOUT`; the connect stays plain. kubelet still decides readiness, the bound only frees the handler.

crates/spur-k8s/src/node_watcher.rs
pre: Done in c3a28a7: the watcher owns a `ControllerClient`, and registration, taint sync and the delete arm go through `call`, so a dead channel is dropped and the next call reconnects. The restart on a registration transport error stays, for a different reason: the fingerprint is stored only after success, and a quiet node may not produce another event for minutes, so the relist is the only prompt retry. The arm comment says so. New test: `a_taint_sync_against_an_unreachable_controller_stays_unsynced`.

crates/spur-k8s/src/node_watcher.rs
pre: Inlined in c3a28a7; the alias and its two tests are gone. The code set is tested in spur-proto.

crates/spur-k8s/src/controller.rs
shiv-tyagi: okay

shiv-tyagi:
> Tested at https://github.com/ROCm/spur/commit/77f99a3d5df3b6b8c3d8b4a8d16d4948a50a2920 on a three-node k0s cluster from spur k8s up, image built from this commit, e2e manifests in namespace spur

Looks good based on the tests performed. Thanks for addressing the comments.

@sgopinath1 @sajmera-pensando Please merge if this looks okay to you.

shiv-tyagi (dismissed):

sgopinath1 (approved):
LGTM

PR #859, opened 2026-09-09
fix(spur-cli): accept --noconvert on accounting commands

### LINKED ISSUE #695: --noconvert is rejected by sacct, sstat, sshare and sreport
## What happens

Slurm's `--noconvert` is rejected by four of the commands that accept it in Slurm. Each one exits at argument parsing, before doing any work:

```
$ sacct --noconvert
error: unexpected argument '--noconvert' found

Usage: sacct [OPTIONS]

$ sstat --noconvert
error: unexpected argument '--noconvert' found

Usage: sstat [OPTIONS] --jobs <JOB_ID>

$ sreport --noconvert
error: unexpected argument '--noconvert' found

Usage: sreport [OPTIONS] <COMMAND>
```

`sshare` does not accept it either. (Separately, `sshare` currently fails on any invocation for an unrelated argument-definition problem, so the flag cannot be exercised there yet.)

## What I expected

The commands that take `--noconvert` in Slurm accept it, so a script migrated from Slurm runs unchanged.

## The four are not equivalent

`sacct`, `sshare` and `sreport` already print raw values, so accepting the flag and ignoring it is enough, and nothing existing changes.

`sstat` is different. It prints memory with an `M` suffix today, so there the flag has something real to suppress, and honouring it changes what an existing command prints. That is a decision about the output contract rather than a parser change, and it should not be made just to silence a parse error.

## Impact

A migrated script fails at argument parsing with no output at all, which is the failure mode Slurm compatibility exists to prevent.

`squeue` and `sinfo` do accept the flag (#647), so the failure now happens further into a script rather than on the first command. That reads to a user like an intermittent problem rather than a missing flag.

## Environment

Spur 0.8.0, single-node local cluster, accounting backed by PostgreSQL.

maybeharshit:
## Summary
- `sacct`, `sstat`, `sshare`, and `sreport` rejected Slurm's `--noconvert` at clap parse time, so migrated scripts exited with no output.
- Accept the flag the same way `squeue`/`sinfo` already do, and ignore it. Those commands already print raw values; this does not strip `sstat`'s `MemAlloc` `M` suffix.
- `sreport` takes the flag as a global option so it works before or after the subcommand.

Fixes #695

## Test plan
- [ ] `cargo test -p spur-cli --locked noconvert_is_accepted`
- [ ] `sacct --noconvert` (and `sstat`/`sshare`/`sreport`) no longer print `unexpected argument '--noconvert'`
- [ ] Output without the flag is unchanged

### DIFF
--- a/crates/spur-cli/src/sacct.rs
+++ b/crates/spur-cli/src/sacct.rs
@@ -54,6 +54,10 @@ pub struct SacctArgs {
     #[arg(short = 'n', long)]
     pub noheader: bool,
 
+    /// Accepted for Slurm compatibility; has no effect
+    #[arg(long)]
+    pub noconvert: bool,
+
     /// Max records
     #[arg(long, default_value = "100")]
     pub limit: u32,
@@ -301,6 +305,13 @@ mod tests {
             derived_exit_code,
             ..Default::default()
         }
+    }
+
+    #[test]
+    fn noconvert_is_accepted() {
+        // Slurm scripts pass --noconvert to keep output machine-parseable.
+        let args = SacctArgs::try_parse_from(["sacct", "--noconvert"]).unwrap();
+        assert!(args.noconvert);
     }
 
     #[test]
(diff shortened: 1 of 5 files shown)


shiv-tyagi (approved):
LGTM

PR #860, opened 2026-09-09
feat(spur-cli): add sinfo -R with node reason provenance (user/time)

shiv-tyagi:
## What

Add `sinfo -R` / `--list-reasons` and populate the node reason provenance it depends on: who set a node's drain/down reason (`reason_uid`) and when (`reason_time`), matching Slurm's `reason_uid` / `reason_time`. This replaces the phase-1 placeholder `?` values in the USER/TIMESTAMP columns with real data.

## Compatibility

New `Node` and `WalOperation::NodeStateChange` fields use `#[serde(default)]`, so a new controller cleanly replays old WAL entries and snapshots (they load with no attribution and render as `Unknown`). Proto tags are append-only, so old clients ignore them.

## Testing

- Unit tests: serde-default snapshot replay, uid reverse-resolve fallback, attribution recorded on drain/update, system uid 0 on heartbeat-down, admin-hold precedence preserved under timeout, resume clears attribution, and `sinfo -R` grouping/splitting/long-format/empty cases.
- e2e: `sinfo -R` assertion added to `test_deregistration.py`.
- Bare-metal cluster: validated the v0.11.0 to current upgrade (old drains survive as `Unknown`), both uid paths end to end (system uid 0 via heartbeat shows `root`; authenticated non-root uid 1000 via JWT shows `vm`), and a snapshot round-trip that persists uid/time across a controller restart.

### DIFF
--- a/crates/spur-cli/src/sinfo.rs
+++ b/crates/spur-cli/src/sinfo.rs
@@ -41,6 +41,10 @@ pub struct SinfoArgs {
     #[arg(short = 'N', long)]
     pub node_oriented: bool,
 
+    /// List reasons nodes are down, drained, draining, or in error state
+    #[arg(short = 'R', long = "list-reasons")]
+    pub list_reasons: bool,
+
     /// Don't print header
     #[arg(short = 'h', long)]
     pub noheader: bool,
@@ -71,6 +75,12 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
 
     let fmt = if let Some(ref f) = args.format {
         f.clone()
+    } else if args.list_reasons {
+        if args.long {
+            format_engine::SINFO_LIST_REASONS_LONG_FORMAT.to_string()
+        } else {
+            format_engine::SINFO_LIST_REASONS_FORMAT.to_string()
+        }
     } else if args.long {
         "%#P %5a %.10l %.4D %.6t %.8c %.8m %N".to_string()
     } else if args.node_oriented {
@@ -111,7 +121,13 @@ pub async fn main_with_args(args: Vec<String>) -> Result<()> {
     if !args.noheader {
         println!("{}", format_engine::format_header(&fields));
     }
-    for line in render_sinfo_output(&fields, &partitions, &nodes, args.node_oriented) {
+
+    let lines = if args.list_reasons {
+        render_list_reasons(&fields, &nodes)
+    } else {
+        render_sinfo_output(&fields, &partitions, &nodes, args.node_oriented)
+    };
+    for line in lines {
         println!("{}", line);
     }
 
@@ -160,11 +176,14 @@ fn parse_states_arg(s: &str) -> Result<Vec<spur_proto::proto::NodeState>> {
     Ok(states)
 }
 
-fn group_nodes_by_display_state<'a>(nodes: &[&'a NodeInfo]) -> Vec<(String, Vec<&'a NodeInfo>)> {
+/// Group nodes by an arbitrary string key, returning groups sorted by key.
+fn group_nodes_by<'a>(
+    nodes: &[&'a NodeInfo],
+    key_fn: impl Fn(&NodeInfo) -> String,
+) -> Vec<(String, Vec<&'a NodeInfo>)> {
     let mut groups: BTreeMap<String, Vec<&'a NodeInfo>> = BTreeMap::new();
     for node in nodes {
-        let key = effective_state_str(node);
-        groups.entry(key).or_default().push(node);
+        groups.entry(key_fn(node)).or_default().push(node);
     }
     groups.into_iter().collect()
 }
@@ -198,7 +217,7 @@ fn render_sinfo_output(
                 .iter()
                 .filter(|n| n.partitions.contains(&part.name))
                 .collect();
-            let state_groups = group_nodes_by_display_state(&part_nodes);
+            let state_groups = group_nodes_by(&part_nodes, effective_state_str);
 
             if state_groups.is_empty() {
                 let row = format_engine::format_row(fields, &|spec| {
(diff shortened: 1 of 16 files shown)


crates/spur-cli/src/sinfo.rs
```diff
@@ -219,6 +238,31 @@ fn render_sinfo_output(
...
+                .unwrap_or(false)
+        })
+        .collect();
+
+    // Slurm splits list-reason rows when reason, setter uid, or set-time differ,
+    // so the grouping key is the composite of all three (sorted by that key).
+    group_nodes_by(&unavailable, |n| {
+        let ts = n.reason_time.as_ref().map(|t| t.seconds).unwrap_or(0);
+        format!("{}\u{1}{:?}\u{1}{ts}", n.state_reason, n.reason_uid)
+    })
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed: effective state is now part of the grouping key, so mixed-state nodes split into separate rows. Verified on the bare-metal testbed.

crates/spur-core/src/node.rs
```diff
@@ -82,6 +82,16 @@ impl NodeState {
         )
     }
 
+    /// Whether the node is unavailable for work and carries an admin reason:
+    /// the set `sinfo -R` reports. Unlike [`is_admin_hold`](Self::is_admin_hold),
+    /// this excludes `Suspended` (power management, not a fault).
```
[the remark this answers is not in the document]
shiv-tyagi: Fixed the comment wording.

crates/spurctld/src/server.rs
```diff
@@ -1800,6 +1802,7 @@ impl SlurmController for ControllerService {
                 &req.reporting_node,
                 spur_core::node::NodeState::Drain,
                 Some(req.drain_reason),
+                Some(0),
```
biluriuday: This can result in stamping root on unauthenticated wire input. req.drain_reason on req.reporting_node, both client-controlled on an RPC with no authz, recorded as Some(0). Pass None so it renders Unknown

crates/spurctld/src/server.rs
```diff
@@ -1364,12 +1364,13 @@ impl SlurmController for ControllerService {
 
         self.require_admin(&request, "update node")?;
 
+        let reason_uid = Self::verified_identity(&request).map(|id| id.uid);
```
biluriuday, line 1367: the USER column is Unknown for human drains on the default config. Permissive auth + no jwt_key + require_admin waving anonymous callers through means operator drains show Unknown while automatic ones show root. We can apply the existing authoritative_user pattern (server.rs:353)

crates/spurctld/src/server.rs
```diff
@@ -1397,6 +1398,7 @@ impl SlurmController for ControllerService {
                 }
             }
         }
+        let reason_uid = Self::verified_identity(&request).map(|id| id.uid);
```
biluriuday, line 1401: Similar one as above. the USER column is Unknown for human drains on the default config. Permissive auth + no jwt_key + require_admin waving anonymous callers through means operator drains show Unknown while automatic ones show root. We can apply the existing authoritative_user pattern (server.rs:353)

crates/spur-cli/src/sinfo.rs
```diff
@@ -302,36 +391,45 @@ fn resolve_partition_field(
...
+/// user (`%u`/`%U`), and set-time (`%H`) from the group's first node.
+fn resolve_list_reason_field(nodes: &[&spur_proto::proto::NodeInfo], spec: char) -> String {
+    if let Some(v) = resolve_node_group_field(nodes, spec) {
+        return v;
+    }
+    let first = nodes.first();
+    match spec {
+        'E' => first.map(|n| n.state_reason.clone()).unwrap_or_default(),
+        // %u: user name of who set the reason.
+        'u' => first
```
biluriuday: should we have this resolver as a separate method and re-use it? Also can unit test it that way

crates/spur-cli/src/sinfo.rs
```diff
@@ -302,36 +391,45 @@ fn resolve_partition_field(
...
+            .and_then(|n| n.reason_uid)
+            .and_then(spur_core::auth::username_for_uid)
+            .unwrap_or_else(|| "Unknown".into()),
+        // %U: user name and uid, e.g. `root(0)`; `Unknown(<uid>)` when the uid
+        // has no passwd entry, `Unknown` when no uid was recorded.
+        'U' => match first.and_then(|n| n.reason_uid) {
```
biluriuday: similar comment as above. can we move this into a separate method and have UTs?

biluriuday (commented):
Few comments in-line. PTAL

crates/spurctld/src/server.rs
shiv-tyagi: Agreed this is a real gap. On the default permissive config with no jwt_key there is no verified identity, so operator drains record no uid and show Unknown while automatic drains show root. Populating the operator would need an asserted user/uid on UpdateNodeRequest and DrainNodeRequest plus CLI wiring and server-side resolution, which grows this PR well beyond sinfo -R. It is purely additive with no regression to existing fields, so taking it as a follow-up rather than expanding this PR.

crates/spurctld/src/server.rs
shiv-tyagi: Agreed this is a real gap. On the default permissive config with no jwt_key there is no verified identity, so operator drains record no uid and show Unknown while automatic drains show root. Populating the operator would need an asserted user/uid on UpdateNodeRequest and DrainNodeRequest plus CLI wiring and server-side resolution, which grows this PR well beyond sinfo -R. It is purely additive with no regression to existing fields, so taking it as a follow-up rather than expanding this PR.

crates/spurctld/src/server.rs
shiv-tyagi: Fixed: passes None now, so it renders Unknown. Verified live via an epilog-failure drain on the bare-metal testbed.

crates/spur-cli/src/sinfo.rs
shiv-tyagi: Done: extracted into `reason_user` in a new `reason` module, reused here and in scontrol, with unit tests.

crates/spur-cli/src/sinfo.rs
shiv-tyagi: Done, same `reason_user` helper.

biluriuday (approved):
LGTM

Issue #852, opened 2026-09-08
spur k8s up on OCI: pod traffic between nodes is dropped with the default CNI


pre:
## What happened

`spur k8s up` with the default `cni = "kuberouter"` on three cloud VMs (OCI) in one subnet. Pod-to-pod traffic across nodes fails completely. CoreDNS answers only when the client Pod runs on the same node, so name resolution inside the cluster looks intermittent, and no log names the cause.

kube-router runs in its default `overlay-type=subnet` mode and routes pod packets directly between nodes of the same subnet. Those packets leave the node with a pod address as the source. An OCI VNIC drops a packet whose source is not one of its own addresses unless skip-source-dest-check is set on the VNIC.

Reproduced with `ping` from node 2 to a Pod on node 3:

```
100% packet loss
route: 10.244.0.5 via 10.0.255.204 dev enp0s9 src 10.0.255.146
```

Workaround used for the test: add `--overlay-type=full --enable-overlay=true` to the kube-router container arguments in `/var/lib/k0s/manifests/kuberouter/kube-router.yaml`. Every pod packet then travels inside the IPIP tunnel with the node address as the outer source, the same ping has 0 % loss, and every later step of the test passed. Setting skip-source-dest-check on the VNICs would work too; it was not done, because it is a network security control.

## Related observation, same code path

With `cni = "kuberouter"`, `k0s_controller_config_yaml` returns `None`, so k0s runs with its own defaults: podCIDR `10.244.0.0/16` and serviceCIDR `10.96.0.0/12`. The `[cluster]` defaults `pod_cidr = "10.42.0.0/16"` and `service_cidr = "10.43.0.0/16"` are written only for Calico. The documented CIDRs and the running CIDRs differ, and any firewall rule written from the config is wrong on a kube-router cluster.

## What was expected

A cluster built by `spur k8s up` with the default CNI passes pod traffic between nodes on OCI, or `docs/deployment/managed-kubernetes.rst` says what OCI needs.

## Environment

- k0s v1.36.2, three VMs, one subnet, control plane on node 1 and two workers
- `spur` at `2dfe3bb`, default `[cluster]` settings apart from `enabled = true`

PR #861, opened 2026-09-09
fix(spurctld): run kube-router in full overlay and pin its CIDRs

pre:
Related issue:
* #852

Related:
* https://github.com/ROCm/spur/pull/813 (merged; rebased onto it. #813 carries the CIDR pinning and `network.provider` for kube-router, this PR adds the overlay mode on top.)

## What this fixes

With the default `cni = "kuberouter"`, kube-router runs in its default `overlay-type=subnet` mode, so pod packets between nodes of the same subnet leave the node unencapsulated with a pod address as the source. A cloud NIC (OCI VNIC, AWS ENI) drops such a packet unless source/destination checking is disabled on the interface, so pod-to-pod traffic across nodes fails completely and in-cluster DNS looks intermittent. No log names the cause.

## Approach

The k0s controller config that `spurctld` generates for kube-router now sets `spec.network.kuberouter.extraArgs` with `overlay-type: full`, so every pod packet travels in the IPIP tunnel with the node address on the outside. That works on any underlay at the cost of the IPIP header between same-subnet nodes. Disabling the cloud source/destination check would also work, but it is a network security control outside Spur's reach.

Until now the only way to get full overlay was to edit the kube-router DaemonSet manifest under `/var/lib/k0s/manifests/kuberouter/` by hand after `spur k8s up`. k0s re-renders that manifest from its own config, so the edit does not survive a k0s restart and has to be repeated per cluster. Setting it in `spec.network.kuberouter.extraArgs` is the supported way and k0s renders the manifest from it.

The API stays on the underlay address for kube-router; only Calico rides the mesh. The "does this CNI route over the mesh" decision moved into one helper (`ClusterNetworking::mesh_native`) so the controller config and the worker `--node-ip` cannot drift. `nodeLocalLoadBalancing` stays Calico-only, as #813 scoped it.

## Upgrade note

A kube-router cluster built by an earlier release runs on the k0s default CIDRs in `subnet` mode. The k0s config is rendered once per control plane when it first starts, so running nodes keep working. A control plane added or re-joined after the upgrade would receive the `[cluster]` CIDRs, and k0s requires the same CIDRs on every control plane. The docs now say to reprovision (`spur k8s down --reset`, `spur k8s up`) or set the old CIDRs in `[cluster]` before adding control planes. The docs also name the firewall requirements (IPIP, IP protocol 4, and BGP on TCP 179).

## Testing

Unit tests cover the generated config for both CNIs and a meshed kube-router control plane that must not advertise the mesh IP. `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` is clean and `cargo test --locked` passes.

End-to-end on the reproduction from the issue, with the rebased branch built in release mode:

**Setup**
- Three OCI VMs (Ubuntu 24.04, 16 vCPU) in one subnet, source/destination checking left enabled on the VNICs.
- Host firewall opened for the VM subnet and the k0s CIDRs on every node.
- `spurctld` on every node (3-member Raft), `spurd` on every node, `[cluster] enabled = true`, default `cni`, default CIDRs.
- `spur k8s install-k0s` (k0s v1.36.2) on every node, then `sudo spur k8s up`. The controller picked node 1 as the control plane and nodes 2 and 3 became workers.
- One busybox pod pinned to each worker.

**Results**

| Check | Result |
|---|---|
| Generated `/etc/k0s/k0s.yaml` | `provider: kuberouter`, `podCIDR: 10.42.0.0/16`, `serviceCIDR: 10.43.0.0/16`, `kuberouter.extraArgs.overlay-type: full` |
| kube-router DaemonSet args | `--overlay-type=full --run-router=true --service-cluster-ip-range=10.43.0.0/16` |
| Pod addresses | `10.42.0.x` and `10.42.1.x`, one /24 per worker |
| Pod to pod across workers, both directions | 3 of 3 packets, 0 % loss, about 0.6 ms |
| Node 2 host to a pod on node 3 (the ping from the issue) | 3 of 3 packets, 0 % loss |
| Route to the remote pod /24 on node 2 | `10.42.1.0/24 dev tun-... proto 17` (IPIP tunnel, not the VNIC) |
| `nslookup kubernetes.default.svc.cluster.local` from a pod | resolves to `10.43.0.1` via CoreDNS at `10.43.0.10` |

Before the change, on the same kind of setup, the cross-node ping showed 100 % loss with the route going out through the VNIC (see the issue).


shiv-tyagi (approved):
> Verified on a fresh three-node OCI cluster (one controller, two workers, default cni = "kuberouter", spur k8s up with the binaries from this branch):
The rendered /etc/k0s/k0s.yaml on the controller carries provider: kuberouter, podCIDR: 10.42.0.0/16, serviceCIDR: 10.43.0.0/16 and kuberouter.extraArgs.overlay-type: full.
The kube-router DaemonSet runs with --overlay-type=full --run-router=true --service-cluster-ip-range=10.43.0.0/16. Workers got 10.42.0.0/24 and 10.42.1.0/24.
Each worker has an IPIP tunnel (tun-…, local node IP, remote peer node IP) and a route for the other worker's pod /24 through it.
Pod to pod across nodes (10.42.0.2 on worker 1 to 10.42.1.2 on worker 2): 3/3 replies in both directions, about 0.6 ms.
Worker host to a pod on the other worker: 3/3 replies.
Cluster DNS from a pod resolves kubernetes.default.svc.cluster.local to 10.43.0.1 via CoreDNS at 10.43.0.10.

Okay. Looks fine to me based on these tests.

PR #870, opened 2026-09-11
feat(spurctld): audit every mutating RPC at the middleware level

biluriuday:
Addresses SPUR-211. Stacked on #869 (now merged) and builds on #860.

## What

Spur recorded 3 of ~38 user-initiated mutating actions in the `txn` log, all of
them reservations, so "who drained this node" had no answer from Spur alone.
Auditing was opt-in per handler, so anything added next started unaudited.

## Approach

Coverage moves into the RPC pipeline. A Tower layer sits after `AuthLayer` —
each `.layer()` stacks closer to the service, so it sees the verified
`Identity` — and owns the whole record: actor, uid, verified, peer address,
timestamp, outcome, and the write. A registry classifies all 64 controller and
accounting methods as mutating, read-only, or internal, and fixes the action
and entity per method, so the layer knows what kind of action occurred without
decoding the request body. Handlers contribute only what is behind the
protobuf — the target name and parameters — via one `audit::annotate` call.

An action therefore cannot be mutating and unrecorded. The worst case is a row
with no target name, which the registry marks and the layer warns about;
compare the alternative, where a handler that forgets the call vanishes from
the log entirely. A completeness test parses the proto and fails on any
unclassified method, so adding an RPC forces a decision rather than defaulting
to unaudited.

The three reservation handlers move onto the layer, which is why `server.rs`
nets a deletion despite four newly annotated node handlers.

## Design choices

**Details are hand-written per RPC.** Deriving `Serialize` on the prost
messages and dumping the request would be less code and a security bug:
`SubmitJob` carries the job script and full environment, `CreateToken` carries
identity material.

**Rows are written only on the Raft leader.** Every mutating controller RPC
forwards to the leader when it lands on a follower, so gating on `is_leader()`
yields exactly one row, from the node that applied the change. Accounting RPCs
bypass Raft and are marked leader-exempt.

**New `peer_addr` column.** `UpdateNodeRequest` carries no user field, so under
the default `permissive` mode a node drain has no identity anywhere. The peer
address is what makes host auth logs joinable to a Spur action.
`sacctmgr show txn Peer=` matches on the `host:port` boundary, so `10.0.0.4`
does not also return `10.0.0.42`.

**Auth failures stay out of both tiers, and are logged instead.** A refused
credential previously returned 401 with no record anywhere — no `audit_rpc`
line, no `txn` row, no main-log entry. All three daemons now log it
unconditionally with path, peer, and reason. Slurm does the same: its
`AUDIT_RPCS` line prints the *authenticated* user and so runs only after
verification, while a failure is reported separately as an error. A `txn` row
would also be wrong here — there is no verified actor to attribute it to, and
it would let unauthenticated traffic grow the database.

**`verified` keys off the `Verified` marker, not the presence of an
`Identity`.** Equivalent today, but `auth.plugin = "none"` would derive an
identity from the local UNIX user with `is_admin: uid == 0`, and inferring from
presence would silently record that as a verified credential.

## Slurm parity

- `txn_table` covers only slurmdbd entities. Spur already recorded
  reservations, which Slurm's does not, but not accounts/users/QOS, which are
  exactly what Slurm's exists for. Those are now classified and recorded.
- `DebugFlags=AuditRPCs` becomes `logging.audit_rpcs` (default off, same
  reason: it is the highest-volume log), on a dedicated `audit_rpc` target so
  it can be routed to its own file. Deliberately not the existing `audit`
  target, which already carries job-submit hook decisions.
- `reason_uid`/`reason_time` on the node record arrived in #860 and stays
  complementary: the node record holds the current reason's attribution, the
  `txn` log holds the history, including resumes and refused attempts.

## Upgrade note

Reservation actions denied by `require_reservation_manager` are now audited.
The old code returned before its audit call, so those denials went unrecorded
despite the docs claiming otherwise.

## Follow-ups

Jobs, partitions, tokens, k0s, and accounting entities are classified and
recorded, but their handlers are not annotated yet, so their rows carry actor,
action, entity, peer, and outcome with no target name. This is documented, and
those are filterable by `Entity=` rather than `Name=`. Annotating them and
routing the two REST endpoints through the same path are next; REST also needs
`rest_auth` to start inserting an `Identity`, as it is gate-only today.

`spur token user` is unauditable by design — it signs a JWT locally from
`auth.jwt_key` without contacting the controller. Documented as a limitation.

## Testing

Clippy and `cargo test --locked` clean; 3751 passing, 30 of them new in the
audit module.

The load-bearing tests drive the middleware as a real `Service` rather than
unit-testing the record builder, because the design rests on the annotation
slot surviving the trip through request extensions. Outcome parsing is asserted
against tonic's real error encoding via `Status::into_http` rather than a guess
at it, and the completeness test is checked against a synthetic proto
containing an unclassified RPC so it cannot pass vacuously.

e2e adds a drain reaching the audit log with the right actor, cross-checked
against `sinfo -R`. Docs build clean; `accounting.rst` is rewritten from
reservation-specific into the general audit page.


crates/spur-k8s/src/auth_middleware.rs
```diff
@@ -16,6 +16,14 @@ use tracing::warn;
 use spur_core::auth::BearerOutcome;
 use spur_core::config::AuthMode;
 
+/// Caller address for a request, `None` when the transport did not record one.
+fn peer_addr(extensions: &http::Extensions) -> Option<String> {
+    extensions
+        .get::<tonic::transport::server::TcpConnectInfo>()
+        .and_then(|info| info.remote_addr())
+        .map(|addr| addr.to_string())
+}
```
shiv-tyagi, line 25: This is a third copy of `peer_addr` (spurctld's `rpc_middleware.rs` has one, spurd's `auth_middleware.rs` has another). spurctld's copy unwraps IPv4-mapped addresses via `canonical_peer`, so a dual-stack listener logs `10.0.0.4:51234` there. This copy doesn't, so the same listener here would log `[::ffff:10.0.0.4]:51234` instead.

Worth extracting `peer_addr`/`canonical_peer` into a shared spur-core helper so all three daemons agree on the format.

crates/spurd/src/auth_middleware.rs
```diff
@@ -34,6 +34,14 @@ use tracing::warn;
 use spur_core::auth::BearerOutcome;
 use spur_core::config::AuthMode;
 
+/// Caller address for a request, `None` when the transport did not record one.
+fn peer_addr(extensions: &http::Extensions) -> Option<String> {
+    extensions
+        .get::<tonic::transport::server::TcpConnectInfo>()
+        .and_then(|info| info.remote_addr())
+        .map(|addr| addr.to_string())
+}
```
shiv-tyagi, line 43: Same `peer_addr` function as spurctld's `rpc_middleware.rs` and spur-k8s's `auth_middleware.rs`. Only spurctld's copy normalizes IPv4-mapped addresses (`canonical_peer`), so this one and spur-k8s's will log the mapped `::ffff:` form on a dual-stack listener while spurctld logs the plain IPv4 form.

Worth extracting `peer_addr`/`canonical_peer` into a shared spur-core helper.

crates/spur-k8s/src/auth_middleware.rs
```diff
@@ -100,6 +108,14 @@ where
...
             }
             BearerOutcome::Reject(msg) => {
+                // A forged credential aimed at pod creation is worth recording on
+                // every cluster, so this is not gated behind any debug setting.
+                warn!(
+                    path = %req.uri().path(),
+                    peer = peer_addr(req.extensions()).as_deref().unwrap_or("-"),
+                    reason = %msg,
+                    "rejected an operator RPC with an invalid credential"
+                );
```
shiv-tyagi, line 118: spurctld's equivalent change (the `Reject` arm in `crates/spurctld/src/auth_middleware.rs`) got a new test asserting a rejected credential never reaches the inner service. This warn! has no equivalent test.

crates/spurd/src/auth_middleware.rs
```diff
@@ -122,6 +130,14 @@ where
...
             }
             BearerOutcome::Reject(msg) => {
+                // A forged credential aimed at a node agent is worth recording on
+                // every cluster, so this is not gated behind any debug setting.
+                warn!(
+                    path = %req.uri().path(),
+                    peer = peer_addr(req.extensions()).as_deref().unwrap_or("-"),
+                    reason = %msg,
+                    "rejected an agent RPC with an invalid credential"
+                );
```
shiv-tyagi, line 140: Same gap as spur-k8s's `auth_middleware.rs`: spurctld got a test for this rejection path (`a_rejected_credential_never_reaches_the_inner_service`), this one didn't.

tests/native_host/e2e/test_accounting.py
```diff
@@ -1128,3 +1131,55 @@ def test_denied_reservation_action_recorded(self, accounting_cluster):
...
+            drained = [r for r in rows if r[0] == "update" and r[2] == "success"]
+            assert drained, f"no successful node update recorded: {rows}"
+            action, where, _outcome, peer, info = drained[0]
+            assert where == f"node:{node}", drained
+            assert reason in info, f"the requested reason must be captured: {info}"
+            # No jwt_key here and no user field on UpdateNode, so the peer
+            # address is the only attribution left — which is why it exists.
+            assert peer, f"an unauthenticated action must still record a peer: {drained[0]}"
+
+            # #860's node-record attribution and the audit log must agree.
```
shiv-tyagi: References a PR number in a comment. Git history already tracks this; inline PR references go stale.
```suggestion
            # The node record's own attribution and the audit log must agree.
```

shiv-tyagi (commented):
Some comments. PTAL.

crates/spur-k8s/src/auth_middleware.rs
biluriuday: Good catch — fixed in 812dcb6.

Confirmed real: on a dual-stack listener the same client was logged as
`10.0.0.4:51234` by spurctld but `[::ffff:10.0.0.4]:51234` by spurd and
spur-k8s. The mapped form was observed on the bare-metal cluster, not just
inferred from the code.

`canonical_peer` now lives in `spur_core::peer` and all three daemons call it,
so the format can't diverge. I left the four-line `peer_addr` extension lookup
in each daemon rather than moving it as well: the drift you identified is
entirely in the formatting, and what remains is a `TcpConnectInfo` lookup with
no behaviour to get wrong. It also matches duplication the repo already makes
deliberately for the auth middlewares — spur-k8s's own header notes it "mirrors
spurd's `AgentAuthLayer`, duplicated since spurd is a binary crate". Moving
`peer_addr` would put a tonic server-request helper in the domain crate that
holds `job`/`node`/`partition`/`qos`.

Each copy also gained a test asserting it unwraps the mapped form, so one
drifting back is caught rather than assumed from a shared call site.

If you'd prefer zero duplication, `spur-proto` is a better home than
`spur-core` — it already owns the tonic dependency and all three daemons depend
on it. Happy to move both there instead.

crates/spurd/src/auth_middleware.rs
biluriuday: addressed as mentioned in above comment

crates/spur-k8s/src/auth_middleware.rs
biluriuday: added a_forged_credential_never_reaches_the_inner_service test for the same

crates/spurd/src/auth_middleware.rs
biluriuday: added

tests/native_host/e2e/test_accounting.py
biluriuday: done

sajmera-pensando (changes requested):
# Review: audit middleware (ROCm/spur#870)

Most of the coverage gaps below trace back to one structural choice in the current design: the generic half of each record (actor, peer, action, entity, outcome) is produced automatically by the layer, but the *per-RPC* half (which object was touched, which parameters) is contributed by a discretionary `audit::annotate` call inside each handler, assembled from the request only. Because that call is unenforced and the layer is body-blind, coverage degrades silently as RPCs are added, and a whole class of question ("which object?", "old → new value?") can't be answered at all.

So before the specific findings, I want to put an alternative structure on the table that removes those failure modes by construction. The five concrete issues follow it — some the alternate design resolves outright, others are orthogonal and need fixing regardless.

---

## Alternate design: proto-tag-driven audit in the layer (handlers untouched)

The per-RPC facts don't need to live in the handlers. They can be declared once in `slurm.proto` and read generically by the layer, so auditing becomes a property of the proto + the layer rather than of each author remembering a call. This is simpler than the current split, not more complex — it deletes the per-handler `annotate` surface entirely.

**Mechanism.**
1. Tag fields in the proto with custom options, e.g. `string name = 1 [(spur.audit_key) = true];` for the target and `string reason = 3 [(spur.audit) = true];` for a detail. Sensitive fields (`SubmitJob.script`, environment, `CreateToken` material) are left untagged — **safe by omission**, with the safe/unsafe decision sitting next to the field definition instead of in a handler's memory.
2. Emit a `FileDescriptorSet` at build time (one build-config line; e.g. prost-build's `file_descriptor_set_path`) — a single generic artifact, **not** per-message code generation.
3. The layer decodes the request generically against that descriptor (e.g. `prost-reflect`'s `DynamicMessage`), reads only the tagged fields, and assembles `entity_name` + `details` itself. One extractor, written once, works for every current and future message.

A new mutating RPC is then audited *with a named target* the moment its fields are tagged — zero handler code, nothing to forget.

**Enforcement (this is what stops a gap reaching production).** Add a build test in the spirit of the existing completeness test: parse the descriptor and fail the build if any method classified `Mutating` has no field carrying `audit_key`. A new mutating RPC cannot merge until its target field is tagged. Classification itself could move to a method option (`[(spur.audit_action) = "update"]`), collapsing the registry into the proto too.

**What this resolves vs. what it doesn't.**
- **#4 (target-less rows) — resolved by construction.** Tagging replaces the discretionary annotate call, so no handler can produce a blank `Where`, now or later.
- **#5 (before/after) — still cannot live in the layer.** The layer only ever sees the request, so the *old* value is invisible to it. The unforgettable central option for the DB-backed accounting entities is a database trigger recording old → new per changed column, with the actor threaded into the DB session so the trigger can stamp it — before/after becomes a property of the table.
- **#1, #2, #3 are orthogonal.** REST is a separate server, read authorization is a separate concern, and write durability is a separate path; the proto-tag approach doesn't touch them, so they still need the fixes below.

**Migration.** Incremental: introduce the options + generic extractor, convert the already-annotated node/reservation handlers first (removing their `annotate` calls), tag the remaining entity types, then delete the `annotate` slot once every mutating message is tagged.

---

## The five concrete issues in the current design

Items 1–4 are blocking, each a case where the log silently fails to answer the question it's built to answer. Item 5 is a capability gap to resolve or explicitly defer.

**1. REST mutations bypass the audit path entirely.**
`submit_job` and `cancel_job` under `rest/handlers.rs` call `state.cluster.*` directly and never traverse the gRPC `AuditLayer`, so an enabled REST API is an unaudited mutation surface. The PR lists this as a follow-up, but as-is the "every mutating RPC" guarantee holds only for the gRPC surface. Either route the two REST endpoints through the same handler path (preferred), or add an explicit `record_txn` at the REST handler as a stopgap. If it can't land here, the REST bypass needs to be called out in `accounting.rst`, not just the PR body — an operator reading the audit docs must not assume REST is covered.

**2. `get_transactions` has no authorization.**
Any authenticated caller — and under `auth.mode = permissive`, any caller at all — can read the entire audit log, including peer addresses and the details of every other user's actions. This is strictly more sensitive than `get_job_history`, and the code comment already concedes "confidentiality of the audit log requires auth.mode = required." Slurm restricts `sacctmgr show transaction` to `AdminLevel >= Operator`. Please gate the read behind admin/operator using the same `require_admin` already applied to the mutating accounting RPCs, rather than relying on `auth.mode` as the only control.

**3. The write is best-effort and non-transactional with the state change.**
`cluster.record_txn` -> `notifier.notify_txn` is fire-and-forget. A change committed to Raft but then dropped by the notifier (DB unavailable, controller crash between apply and write, queue overflow) is applied-but-unrecorded — which is exactly the "who drained this node" gap this PR sets out to close. The registry guarantees an action is *classified*, not that its row is *durable*. Please make the guarantee real: tie the audit write into the same transaction/commit path as the mutation, or add a durable outbox/retry so a dropped write is detectable and recovered. At minimum, failures to enqueue/persist must be surfaced (error metric + log), not silently discarded.

**4. Most entity types record a target-less row.**
Jobs, partitions, tokens, k0s, and all accounting entities (account/user/qos) are classified and recorded but produce rows with an empty `entity_name`. For the accounting entities this is the headline Slurm-parity case — the log currently cannot tell you *which* account or user was deleted, only that a deletion happened. Please annotate these handlers (the `audit::annotate` slot is already threaded; it's one call per handler) so the target name is captured — or adopt the proto-tag approach above, which removes the per-handler call altogether. If any must ship unannotated, the log should not present a blank `Where` as if it were complete.

**5. The log records the requested new value, not what actually changed (no before/after).**
`details` is built purely from the inbound request parameters, so an update records the *requested* value with no prior state. For a QOS wall-time change, the log can at best say "person x set max_wall_min=120" — it cannot say "changed from 60 to 120." For a common audit question ("what was the old value, who changed it, when") this is a real gap, and it's most acute for the accounting entities in item 4, which today record neither the target nor any before/after. Capturing before/after needs the prior state to be read where it exists (the handler or the DB), since the request-side layer cannot see it — so please decide explicitly: either capture `{field: {from, to}}` for the mutating accounting/QOS entities (a DB trigger is the unforgettable option), or document that the log is request-only and before/after is out of scope. Not silently request-only.

Happy to re-review once these are addressed.

biluriuday:
@sajmera-pensando Thank you for the thorough review — the structural framing was genuinely useful, and
it pushed me to look at parts of the write path I would otherwise have left alone.

I've fixed four of the five findings. The fifth I've addressed in part and, after
some analysis, chosen not to complete; I've explained my reasoning below and I'm
happy to be talked out of it. I've also declined the alternate design on cost
grounds, though I did adopt the enforcement goal behind it. Along the way I found
two problems the review didn't mention, which I've fixed as well, and I have two
small corrections to offer.

## What I've fixed

### 2. `get_transactions` had no authorization

*Commit: `fix(spurctld)!: require operator to read the audit log`*

You were right that this needed gating. It now uses the same `require_admin` as the
accounting mutations, which resolves to Operator-or-above and matches Slurm's
restriction on `sacctmgr show transaction`. I verified it end-to-end on a live
cluster: an admin can read the log, and an unprivileged user receives
`PermissionDenied`.

Consistent with the rest of the service, an *unauthenticated* caller is still
admitted, so `permissive` and `disabled` deployments are unaffected — those modes
trust the client by design. The gate binds identified callers, which is what
`required` mode produces. I've documented this in `accounting.rst` and in the
authoritative `privileged-operations` list in `configuration.rst`.

I marked the commit breaking, since under `mode = required` an identified caller
below Operator now receives a denial where the read previously succeeded.

### 4. Target-less rows

*Commit: `fix(spurctld): name the audited object and record the real upsert verb`*

Your numbers were exact — 7 of 32 mutating RPCs named their target. That is now
**29 of 32**. Jobs are named by id (assigned at submission), credentials by token id
(never the paired secret), and partitions, k0s node changes and all six accounting
entities by name.

The three remaining exceptions are genuinely unnamed, and I've allowlisted them
explicitly: `Reconfigure` acts on the whole configuration, and `ClusterUp` and
`ClusterDown` act on the single embedded k0s cluster.

Enforcement is a test, `every_mutating_rpc_names_its_target`, which fails the build
if any mutating RPC outside that allowlist records no target. A second test proves
the check isn't vacuous, and a third catches a stale allowlist entry.

### 1. REST mutations bypassed the audit path

*Commit: `fix(spurctld)!: audit REST mutations via the controller handlers`*

I took the option you preferred. REST submit and cancel now dispatch into the
controller handlers and inherit their authorization, validation, leader forwarding
and txn row, which also let me delete REST's duplicated ownership check and error
mappers.

One thing I'd flag, since it surprised me: routing through the same handler path
doesn't give auditing for free. `AuditLayer` is Tower middleware on the tonic stack,
so a direct handler call sits below it, with no slot in the request extensions —
`annotate` simply becomes a no-op. Delegation therefore needed a new
`audit::recorded` helper that wraps a handler call the way the layer wraps a service
call: the same registry entry, the same outcome mapping, and no row at all when the
handler forwarded, since the leader records that one itself.

Two behaviour changes are worth calling out: a denial is now 403 and an
already-terminal job 409, where both were previously 500; and a non-leader now
forwards rather than returning 503.

## What I've addressed in part

### 3. The write is best-effort and non-transactional

*Commit: `fix(spurctld): drain audit writes on shutdown and count dropped rows`*

**What I've done:**

- **Added a failure counter.** There are now two counters on `/metrics/audit`:
  `spur_audit_rows_written_total` and `spur_audit_rows_dropped_total`. A dropped row
  is alertable rather than visible only to whoever greps the controller log.
- **Added a shutdown drain.** I found that `spurctld` had no graceful shutdown at
  all, so every restart silently dropped in-flight audit writes. It now handles
  SIGTERM and Ctrl-C and waits up to five seconds for pending accounting writes,
  warning if any remain.
- **Resolved the documentation contradiction.** The page claimed both that recording
  is best-effort and that an action "cannot be mutating and unrecorded". Coverage and
  durability are now stated separately, with a new *When a row is lost* section.

**Where I stopped: the transactional write for accounting entities.**

This one is cheap in plumbing — `db::record_txn` already takes `&mut PgConnection` —
so I want to be clear that I stopped on the trade-off rather than the effort:

1. **It changes fail-open to fail-closed.** Putting the audit row inside the entity's
   transaction means that when the row can't be written, the user's `sacctmgr` command
   fails. That contradicts the documented guarantee that "a database outage never
   blocks the operation itself". It's a defensible position, but it felt like a policy
   decision rather than a bug fix, and not one I wanted to make unilaterally.
2. **It covers 6 of 32 mutating RPCs.** Only the accounting entities are `Local` scope.
   Node, job, partition, reservation, token and k0s state lives in Raft and cannot join
   a PostgreSQL transaction, so it would remain best-effort regardless.
3. **It closes the rarer failure mode.** The shutdown drain removes the routine-restart
   case entirely, and the counter makes the remaining window visible. What was left was
   a narrow crash window affecting roughly a fifth of the surface.

If this is revisited, I'd suggest one adjustment to the design you sketched. Having the
layer skip whenever a `Local` RPC succeeds fails in the dangerous direction: a handler
that forgets its transactional write would produce *zero* rows. Having the handler set
a flag meaning "already recorded" inverts that, so a mistake yields a visible duplicate
rather than a silent omission.

**Two small corrections, offered gently:**

- I don't think "silently discarded" is quite right. `notify_txn` already retries three
  times with backoff and logs `error!` with the actor, action and entity on final
  failure. What was genuinely missing was the counter, which I've now added.
- "Queue overflow" isn't a failure mode here, as far as I can tell — it's a
  `tokio::spawn` per record, with no bounded queue.

## What I've deferred

- **Durability for the Raft-backed entities** (an outbox, or carrying the record in
  `WalOperation`). The layer builds the record after the handler returns, so making it
  atomic would mean constructing the record before the propose and carrying it in the
  operation. That's a persisted-state change subject to the breaking-change rules, and
  it would move record construction into every mutating handler. I haven't attempted it.
- **5. Before/after values.** This remains open, and the log is still request-only. I owe
  you at least the documentation half of your ask, which I'll add.

  On the database-trigger suggestion, I'd offer one observation. It's strongest exactly
  where before/after matters least here, and blind where it matters most. The accounting
  service has no update RPC at all — only create and delete — so on the PostgreSQL side a
  trigger would mostly capture delete snapshots. Meanwhile every explicit `Update*` RPC
  (`UpdateNode`, `DrainNode`, `UpdateJob`, `UpdatePartition`, `UpdateReservation`,
  `Reconfigure`) is Raft-backed and structurally invisible to a trigger. Two mechanisms
  would be needed either way. Triggers also fire only on committed writes, so denied and
  failed attempts — which Spur deliberately records — would still need the layer, and the
  actor would have to be threaded into the database session by every write path, which
  reinstates the discretionary-call failure mode you're objecting to, just relocated to
  session setup.

## The alternate design: proto-tag-driven audit

I looked at this carefully, because the goal behind it is right. The mechanism would
work: I checked the two things that could have ruled it out, and no mutating RPC is
client-streaming, and `tonic_prost_build` can emit a descriptor set.

Where I'd respectfully disagree is with "simpler than the current split, not more
complex". The layer sits at the HTTP level, and the body is a pass-through stream it
never touches. Reading tagged fields would mean buffering every mutating request body,
stripping the five-byte gRPC frame, handling `grpc-encoding` compression, decoding
dynamically through a new `prost-reflect` dependency against an embedded descriptor set,
and then rebuilding the body — a second full decode of every mutating request on top of
the one tonic already performs. In exchange it deletes roughly 29 one-line calls. To me
that reads as a trade rather than a simplification, and as you note, it leaves finding 5
untouched.

That said, I did take the goal underneath it. The registry already carried a `targeted`
flag, and the layer already warned when a targeted method failed to annotate, so the only
silent-degradation path was an author choosing `untargeted`. The allowlist test I
described under finding 4 gives the same "cannot merge without a named target" guarantee
in about thirty lines, without the proto machinery.

## Two issues I found that weren't in the review

- **`sacctmgr modify` was being recorded as `create`.** Modify reaches the *create* RPCs
  as an upsert, so changing a QOS wall-time logged `action=create, entity=qos,
  name=<blank>` — a modification recorded as a creation, with no target. That struck me
  as actively misleading rather than merely incomplete, so I fixed it alongside finding 4.
  The upserts now report which half ran via `RETURNING (xmax = 0)`, which needs one
  statement with no pre-read and no race, and the handler records that verb.
- **`spurctld` had no graceful shutdown.** Covered under finding 3 above. It turned out to
  be the most frequent of the durability gaps, since it affects every routine restart.

## Verification

**Unit and static checks.** `cargo clippy --workspace --exclude spur-ffi --all-targets`
is clean, 38 test binaries pass, and the Sphinx docs build without new warnings.

**New end-to-end coverage.** I added four e2e tests in
`tests/native_host/e2e/test_accounting.py`, since findings 2 and 4 previously had only
unit coverage:

| Test | What it pins down |
|---|---|
| `TestAccountingEntityAudit::test_qos_add_modify_delete_are_named_and_distinguished` | `add` → `create`, `modify` → `update`, `delete` → `delete`, all naming `qos:<name>`, and the modify recording the value it set |
| `TestAccountingEntityAudit::test_account_modify_is_not_recorded_as_a_creation` | a second `add account` against an existing row yields exactly one `create` and one `update` |
| `TestJobAudit::test_submit_and_cancel_name_the_job_id` | both rows name `job:<id>`, the id being assigned by the very action under audit |
| `TestAuditReadAuthorization::test_identified_non_admin_cannot_read_the_log` | an admin token reads the log; a plain user's token is refused |

**Full suite run.** Against a live three-node cluster with Postgres:

sajmera-pensando (changes requested):
Thanks for the thorough turnaround — this is genuinely a substantial response, and I appreciate the depth of the reasoning as much as the fixes. #2 (operator gate) and #4 (named targets + the `xmax` verb fix) are done well, the `audit::recorded` helper correctly reconstructs the layer's record for the REST path, and the `sacctmgr modify`-as-`create` bug and missing graceful shutdown were good catches. I've re-reviewed against the code, and I agree with both your corrections: it wasn't "silently discarded" (retry + `error!` already existed), and "queue overflow" wasn't a real mode.

I'm keeping the review at changes-requested, and I want to be upfront that the three items below are blocking for my approval rather than nits — but they're all tractable, and I've tried to lay out the reasoning so we can converge on them. #2 and #4 I already consider resolved.

## 1. Durability should be fail-closed, not best-effort *(blocking)*

I'd push back — gently, but firmly — on the framing that transactional auditing is a policy call to defer. For a production system the audit log is a **requirement**, not a convenience: if a mutating action can't be recorded, it shouldn't be applied. So a mutating operation failing when its audit row can't be written is the behaviour we want, not a regression to weigh against convenience.

The "a DB outage would block operations" objection doesn't hold up on inspection:

- **For the six accounting entities, there is no new failure surface.** The accounting DB is already the system of record for those writes — if Postgres is down, `sacctmgr add/modify/delete` already fails. Writing the txn row in the *same transaction* as the entity change adds no new dependency; it just makes the row atomic with the change it describes. The plumbing already exists (`db::record_txn` takes `&mut PgConnection`), so this is wiring, not new machinery.
- **A DB hiccup that would block an admin mutation is itself an incident.** "The audit backend is down" is precisely a state where you want writes to stop, not to proceed silently unaudited. Fail-open here optimises for the wrong failure.

For the Raft-backed entities the record has to be made durable with the state change — carried in the `WalOperation` (or a durable outbox drained on commit) so replay re-emits it. I understand that's a persisted-state change under the breaking-change rules and that it moves record construction ahead of the propose; that's the expected cost of making the guarantee real, and it should be `!`-flagged like the others.

The shutdown drain and the dropped-row counter are genuinely good and worth keeping — they just *narrow* the window rather than closing it, and for an audit requirement I think the window needs to close. So I'd ask that we treat this as blocking rather than deferred.

## 2. Enforcement should catch a missing `annotate` call, in CI *(blocking)*

The new `every_mutating_rpc_names_its_target` test checks the registry's `targeted` flag — it forces a *classification decision*, which is good. But it does not verify that the handler actually calls `audit::annotate`. A handler marked `targeted = true` whose author forgets the call still ships rows with a blank `Where`; the only thing that catches it is a runtime `warn!` (layer.rs) that nobody may be watching. That's the same forgettable-call failure mode the review opened on, just moved one step over.

Please close it at build/CI time, not at runtime. Concretely, either:

- a test that drives every mutating (targeted) RPC through the handler/`recorded` path with a representative request and asserts a non-empty `entity_name` — so a forgotten `annotate` fails the build; or
- the proto-tag extraction from the original review, which removes the per-handler call entirely and makes "did you annotate" unrepresentable rather than merely tested.

Either is fine by me; a runtime warning isn't quite enough. Whatever the mechanism, I'd want it running in PR CI so a new mutating RPC can't merge with an un-annotated target.

## 3. Before/after should be captured for the PG-backed entities *(blocking)*

On the before/after deferral, I don't think the "the accounting service has no update RPC, so a trigger would mostly capture delete snapshots" argument holds for the case that matters most. There's no RPC *named* Update, but `sacctmgr modify qos … set MaxWall=…` reaches `CreateQos` as an **upsert**, which is a real SQL `UPDATE` on the row (`ON CONFLICT DO UPDATE`) — the update is already happening in the system. QOS caps, account limits and user/association limits all live in Postgres and are all modified this way, so old→new for exactly the "QOS wall time 60→120" example is fully capturable, and cheaply: the handler already performs the write and now even knows via `xmax` that it was an update, so it can read the prior row (or `RETURNING` the old values) and record `{field: {from, to}}`.

And your other points are well taken — triggers can't see denied/failed attempts (they fire only on committed writes), and the Raft-backed entities (`UpdateNode`, `UpdatePartition`, reservations, `Reconfigure`) genuinely can't be reached from Postgres — so I'd be glad to see those stay request-only for now, documented as such. It's really just the PG-backed accounting entities I think should record before/after, since the update is already occurring there.

---

To be clear on scope: #2 (gate) and #4 (targets) are resolved, and the REST and shutdown work is a real step forward. The three above are what I'd want in before approving — 1 and 3 because an audit log that can silently miss rows or can't say what changed isn't yet answering the questions it exists for, and 2 so the coverage you've built can't quietly regress. None of them look large from the code, and I'm happy to pair on the `WalOperation` durability piece specifically since that's the fiddliest. Thanks again for taking the first round so seriously.

biluriuday:
Hi @sajmera-pensando 
Thanks for pushing back on these — you were right on both counts where I'd
argued for deferring, and all three blocking items are now addressed.

**1. Fail-closed durability — done for the accounting entities.** Your point
settled it: Postgres is already the system of record for those six writes, so
putting the `txn` row in the same transaction adds no dependency that wasn't
there. The row now commits with the change or neither lands, and a mutation
whose audit row can't be written fails rather than applying unaudited.

**2. Enforcement in CI — done.** There's now a build-time check that reads the
handler sources and fails if a method marked `targeted` never records its
target. A companion test guards the check itself, so it can't pass by matching
nothing. The runtime `warn!` is no longer the only safety net.

**3. Before/after — done for the PG-backed entities.** You were right that
`sacctmgr modify` reaching `CreateQos` is a real SQL UPDATE, and that's exactly
the case now captured: modifications record `{column: {from, to}}` for the
columns they moved. One deviation from your suggestion — `RETURNING OLD.*` is
PostgreSQL 17+, and CI runs 16, so the prior row is read with `SELECT … FOR
UPDATE` inside the same transaction instead. Same guarantee, no version floor.
The Raft-backed entities stay request-only, documented as such.

**Deferred: Raft-backed durability.** I'd like to do this as a follow-up PR
rather than extend this one. It's a different shape of change: the record has
to be built before the propose and carried in the `WalOperation`, which is a
persisted-state change needing the `!` flag, and replay then has to be made
idempotent — a dedup key plus a unique index on `txn`, or a duplicate row on
every restart. That's worth its own review rather than being folded in here.

Verified with clippy, the full unit suite, and the accounting e2e suite against
a three-node cluster with Postgres (41 passed), plus a manual check that a QOS
wall-time change records `{"max_wall_min":{"from":60,"to":120}}`.

sajmera-pensando (approved):
Approving — all three blocking items are resolved, and I re-reviewed each against the code rather than the summary:

- **Fail-closed durability (accounting):** the entity write and its `txn` row now share one transaction and commit or roll back together, with the `recorded_in_band` flag (set only post-commit) correctly suppressing the layer's second write. No successful accounting mutation can commit unaudited.
- **CI enforcement:** `every_targeted_handler_annotates` scans the handler sources and fails the build if a `targeted` method never records its target, with a non-vacuity guard — so this runs in CI, not as a runtime warning.
- **Before/after (PG-backed):** `SELECT … FOR UPDATE` + JSONB diff records `{column: {from, to}}`, and the test pins the exact `max_wall_min: {from: 60, to: 120}` case. Good call using `FOR UPDATE` over `RETURNING OLD.*` for the PG16 floor.

Sanity is green across the board (clippy, fmt, test, accounting DB tests, both e2e suites).

One request before I'd call the audit story complete: please do land **Raft-backed durability as its own follow-up PR**, not leave it open-ended. As you noted, node/job/partition/reservation/config rows are still best-effort, so a crash between the Raft commit and the async write can still lose them. Carrying the record in `WalOperation` (breaking, `!`-flagged) with idempotent replay — a dedup key plus a unique index on `txn` so a restart re-emits without duplicating — is the shape, and I agree it deserves its own review. Happy to pair on it. Could you open the follow-up (or a tracking issue) so it doesn't fall off?

Non-blocking nit for that follow-up or a separate one: the accounting handlers now pass their annotation only to `write_txn_in_band`, so a *denied or failed* accounting mutation records a blank target where it previously named the account. Also depositing the annotation into the slot (letting `recorded_in_band` suppress the success duplicate) would restore naming on the failure path.

Thanks for taking both rounds so seriously — this ended up materially stronger than where it started.

PR #872, opened 2026-09-11
docs(plans): add the idle-fill scheduling design

nikhilsk:
Design document only — no code. Draft as it is under review; the intent is to implement
it once the design is agreed.

## What it proposes

**Idle-fill scheduling**: let a job that has exhausted its QOS group *node* quota run on
nodes that are otherwise idle, and take those nodes back when a job with a genuine quota
claim needs them. Behind one cluster switch, `scheduler.idle_fill_enabled`, default off.

A group node cap exists to stop one team monopolizing a contested cluster, but it is
enforced unconditionally — including when nothing is contested. On a 24-node cluster with
`grptres=node=4` per team, a team's fifth job pends with `QOSGrpNodeLimit` while twenty
nodes sit idle and nobody competes for them. The waste scales with how conservatively
quotas are set, which makes the safest quota configuration also the most wasteful one.

## What changed in this revision

The design has been through a two-step verification: every factual claim checked against
the source, then the architecture attacked adversarially. That found three blockers that
would have shipped a feature which does not work, and one prior-art question the design
had not asked. The earlier architecture is abandoned.

**The loan would have been permanently unrecallable.** A borrowed job keeps counting
toward its QOS's aggregates — `sum_running_tres` filters on nothing but "is running"
(`cluster.rs:7449`). A team at `grptres=node=4` that borrows a fifth node reads as using
5, so when a legitimate job frees a node the team's next submission is blocked at the
gate. And `retain_eligible` *removes* gate-blocked jobs from the pending list
(`cluster.rs:7099`), so that job never reaches the scheduler, never lands in
`unscheduled`, and can never trigger reclaim. The borrowed job blocks its own team
indefinitely. The `max_jobs_per_user` variant is worse: lifting the node cap does not
rescue it, so the job is neither scheduled nor collected.

**Reclaim inside `try_preempt` was a denial-of-service vector.**
`preempt_overlaps_pending_nodes` (`scheduler_loop.rs:808`) is a partition-level boolean
that never checks whether evicting a victim actually helps the reclaimer. Two existing
gates keep that inert today; the earlier draft removed both. A job requesting
`--gres=gpu:8` where the largest node has 4 is not rejected at submit, re-enters
`unscheduled` every cycle, and would requeue one borrowed job **per second, indefinitely** —
no priority edge and no allow-list entry required. Reclaim is now a self-contained routine
that evicts only a victim set which provably closes the shortfall, atomically, which fixes
the abuse and the multi-node accumulation problem together.

**Eviction frees the node before anything is killed.** `JobPreemptRequeue` deallocates
inside the Raft apply (`cluster.rs:5756`) then fires a fire-and-forget cancel, skipping the
`Completing` handshake the rest of the codebase maintains deliberately
(`scheduler_loop.rs:2115-2117`). At a 1-second cycle the reclaimer can be dispatched onto
a node whose victim process still holds GPU memory.

**Placement is now one pass, not two.** The earlier draft proposed a second `schedule()`
call over a filtered set of idle nodes. Unnecessary: `backfill.rs:718-726` already reserves
future slots for jobs it cannot place, so appending over-quota jobs to the **tail** of the
existing priority-sorted pass gets correct capacity accounting, backfill protection,
het-group atomicity, reservations and topology for free — and permits a short borrowed job
to use a reserved-but-idle node when its time limit fits before the reserved start, which
the filtered-slice approach would have stranded. `schedule()` never re-sorts and never
reads `job.priority`, so tail position is real precedence.

## Prior art this has to answer for

`docs/admin-guide/accounting.rst:942-978` already documents a **burst QOS pattern** doing
opportunistic borrowing with reclaim, using only existing features: a low-priority QOS with
`preemptmode=requeue` that normal QOSes list in `preempt`. It works today.

Idle-fill's distinct claim is therefore narrower than it first appears: it makes that
behavior **automatic and quota-derived** instead of opt-in per job and pool-derived. §2
makes the case; Q1 asks reviewers to confirm the trade rather than inherit it.

## Structure

The document carries a register of **17 numbered defects**, each with a severity, a
`path:line` citation, and a fix (§9), so it reads as a work list rather than prose. The
work breakdown (§11) is ordered against it, with one hard ordering constraint: the
aggregate fix must land before the mechanism goes live, or the loan is unrecallable.

Six open questions (§15). Q1 (build it, given the burst pattern) and Q2 (`preempt_mode =
"off"` is documented as absolute — narrow the promise, or refuse to lend without it?) are
the two that change the outcome.

## Pre-existing defects found while verifying

Not caused by this feature, affecting operators today, and worth fixing separately (§13):

- **A suspend-preempted job is never resumed.** `configuration.rst:774-776` promises it
  "continues automatically once the higher-priority job finishes". Nothing resumes it — the
  only `resume = true` call is the user-facing `resume_job` RPC (`server.rs:1138`), while
  the scheduler loop only ever suspends (`scheduler_loop.rs:767`). The job stays
  `Suspended` until a human intervenes, holding its nodes throughout, so the job that
  triggered the preemption gains nothing.
- **A typo in `preempt_mode` silently disables preemption.** The parse ends in
  `_ => PreemptMode::Off` (`config.rs:2118`) with no warning. `cancle`, or `checkpoint` out
  of Slurm habit, yields a cluster where preemption never fires and nothing says why.
- **A stale roadmap claim** about partial multi-node starts; placement is all-or-nothing.

## Testing

No code here, so nothing to run. §14 records the plan: unit coverage per layer, and a live
sequence whose first step is the regression guard — with the switch off, an over-quota job
must still pend with `QOSGrpNodeLimit` and nothing else may change. Two steps exist
specifically to catch the blockers above: a legitimate job must start while a borrowed job
of the same QOS runs, and an unplaceable job must evict **nothing** over several minutes.

### DIFF
--- a/plans/idle-fill-scheduling.md
+++ b/plans/idle-fill-scheduling.md
@@ -1,0 +1,338 @@
+# Idle-Fill Scheduling
+
+Let a job that has exhausted its QOS group *node* quota run on nodes that are
+otherwise idle, and take those nodes back by preemption when a job with a genuine
+quota claim needs them.
+
+Everything sits behind one cluster switch, `scheduler.idle_fill_enabled`, defaulting
+to off, so an existing cluster behaves exactly as it does today until an operator
+opts in.
+
+This document is the design under review. Two findings from the codebase changed its
+shape relative to the original proposal:
+
+- **The configuration surface halves, from four knobs to two.** Two of the proposed
+  knobs describe mechanisms that already exist or behavior that is already invariant
+  (§6).
+- **Reclaim is not free reuse of the existing preemption.** The proposal assumes any
+  quota-holder can call the loan in. The current `try_preempt` will not permit that
+  under default configuration, for reasons unrelated to the QOS allow list (§5.5).
+  This is the largest piece of new work and the part most worth challenging.
+
+---
+
+## 1. Problem
+
+A group node cap exists to stop one team monopolizing a contested cluster. But the
+cap is enforced unconditionally, including when nothing is contested.
+
+On a 24-node cluster where each team's QOS carries `grptres=node=4`: a team with
+four nodes busy and a fifth job queued sees that job pend with `QOSGrpNodeLimit`
+while the other twenty nodes sit idle. Nobody is competing for them. The cluster is
+enforcing a fairness rule against a party of one.
+
+The waste scales with how conservatively quotas are set, which makes the safest
+quota configuration also the most wasteful one.
+
+## 2. Goals and non-goals
+
+**Goals**
+
+- An over-quota job runs when there is genuinely spare capacity.
+- A job with a quota claim can always take that capacity back.
+- No behavior change for clusters that do not opt in.
+- No new user action: users do not opt in, mark jobs, or learn a flag.
+
+**Non-goals**
+
+- Guaranteeing *which* nodes a team gets. This is capacity-based, not
+  ownership-based; teams that own specific hardware want partitions.
+- Account and association quotas. Only the QOS group node cap is in scope for a
+  first cut (§11, Q3).
+- Partial dispatch. Running part of a multi-node job is not supported today and this
+  does not add it (§6).
+- A checkpoint-on-eviction mode. It does not exist in Spur (§6).
+
+## 3. What an operator and a user see
+
+An operator sets `idle_fill_enabled = true` under `[scheduler]`. Nothing else is
+required, though `preempt_mode` must be non-`Off` somewhere for reclaim to work,
+which is a real constraint rather than a footnote (§5.5).
+
+A user submits as normal. A job that would previously have pended with
+`QOSGrpNodeLimit` may now start, and may later be evicted. Its fate on eviction
+follows the existing `PreemptMode`, so it is the same fate any preempted job gets on
+that cluster.
+
+The one genuinely new user-visible thing is that a job can be evicted for a reason
+that is not "someone with higher priority arrived" but "you were never entitled to
+this node". Surfacing that clearly matters, and is Q5.
+
+## 4. Two tiers
+
+- **Legitimate** — within quota. Never preempted by this mechanism.
+- **Opportunistic** — running only because capacity was spare. Displaceable by any
+  legitimate job.
+
+The contract is: *you may use spare capacity, and you will lose it on demand.* Both
+halves matter. A loan that cannot be recalled is not a weaker version of this
+feature; it is a different and worse one (§10).
+
+## 5. Design
+
+### 5.1 Two scheduling passes, one cycle
+
+Both passes run inside a single scheduling cycle.
+
+**Pass 1** is today's scheduling, completely unchanged. All within-quota jobs are
+evaluated in priority order with normal limit enforcement.
+
+**Pass 2** places over-quota jobs on whatever pass 1 left empty.
+
+The implementation choice that keeps this small: **pass 2 is a second `schedule()`
+call over a filtered node slice.** `ClusterState::nodes` is a plain `&[Node]`, so
+handing the scheduler only the idle nodes reuses the entire placement engine — GPU
+matching, fabric topology, `cons_tres` packing, per-node capacity — without touching
+the backfill scheduler at all. Whole-job atomicity and priority ordering come along
+for free, because the existing engine already provides both.
+
+These are two passes within one cycle, not two phases of delivery. The document uses
+*pass* throughout to keep them distinct.
+
+### 5.2 Eligibility: the node cap must be the *sole* blocker
+
+This is the subtle part, and getting it wrong is a correctness bug rather than a
+missing feature.
+
+The obvious rule is "the job was blocked with `PendingReason::QosGrpNodeLimit`". That
+rule is **wrong**. The resource check returns only its *first* breach, and evaluates
+`grp_tres` in the order cpu, node, mem, gpu. A job over both its node cap and its
+GPU cap therefore reports the node one. Lending it idle nodes would let it escape a
+cap on a genuinely contended resource — exactly what quotas exist to prevent.
+
+Instead: re-run the whole limit check with the node dimension lifted, and treat the
+job as eligible only if it then passes. Lifting means setting that dimension to `0`,
+which the checker ignores rather than treating as a block.
+
+This is exact rather than heuristic, and it also handles the reverse case: a job over
+both its node cap and its group *CPU* cap reports CPU, and a reason-based rule would
+have wrongly excluded it from ever being considered.
+
+Eligibility is decided where the running aggregates already exist, not in the gate
+that consumes the verdict. Recomputing them in the gate would re-derive every
+running-job aggregate, undoing the single-pass aggregation the gate was built
+around. It is also gated on the config switch, so a cluster with idle-fill off pays
+nothing for the extra evaluation.
+
+### 5.3 The idle budget
+
+Idle nodes are those with no running jobs, **minus** three sets:
+
+1. Nodes pass 1 just claimed this cycle.
+2. Nodes that are idle now but hold a **future reservation for a legitimate job**.
+   The scheduler already reports these — `planned_starts()` returns exactly the nodes
+   that are empty now but reserved for a pending job's future slot. Missing this
+   would make idle-fill delay the large job that backfill was deliberately
+   protecting, reintroducing the starvation problem backfill exists to solve.
+3. Nodes under an admin reservation, or reserved for k0s.
+
+On (3): a reserved-but-empty node is excluded because an idle-fill job carries no
+promise to finish before the reservation begins. A future refinement could allow it
+when the job's time limit provably fits the gap; the machinery for that reasoning
+exists, but it is not needed for a first cut.
+
+Node state alone is not a sufficient test — a node can be idle *and* reserved.
+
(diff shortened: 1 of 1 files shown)


sajmera-pensando (changes requested):
## Requested changes — idle-fill design decisions

Reviewed with the design-summary canvas alongside. Resolving the open questions, flagging one correctness prerequisite, and adding one capability. Please fold the following into the doc before this merges.

### 1. Q1 (build it) — resolved
Decided: build it. The burst pattern in §2 is not a reason to defer — idle-fill's distinct value is making opportunistic borrowing automatic and quota-derived instead of opt-in per job, which is exactly the adoption gap §2 identifies. Close Q1 in §15 as resolved.

### 2. Q2 (`preempt_mode = "off"`) — reclaim regardless
Decided: Option 1. Reclaim of a borrowed job is intrinsic to idle-fill — a job is "borrowed" by definition of being reclaimable. The only guard is `idle_fill_exempt_secs`; `preempt_mode` does not shield a borrowed job. Reject Option 2 (refuse-to-lend): it reintroduces the adoption cliff §2 exists to avoid and makes idle-fill inert on a default cluster.

Consequences to bake in:
- Breaking change in the AGENTS.md sense → the implementation PR title takes the `!` marker.
- The documented `off` guarantee (`configuration.rst:779-785`) narrows from "running jobs in this partition are never kicked out" to "jobs with a quota claim are never kicked out." Ship that docs edit with the change.

### 3. Q5 (burst tier) — keep `idle_fill_preemptable`
Reversing the inclination to drop it. A burst QOS has quota = whole cluster, so it is never over-quota, never fails the §6.2 sole-blocker test, and is never stamped `idle_fill` — meaning the §8.2 victim set can never include a burst job. Without this flag, idle-fill can never reclaim burst.

Keep the flag *because* burst is being phased out, not despite it: during the burst→idle-fill migration, live burst jobs hold cluster-wide-quota nodes that a legitimate reclaim must be able to call back. Expectation is that burst usage decays as users stop submitting to burst explicitly.

Design consequence: §4 and §8.2 need a second source of "borrowed." Victim set = `idle_fill`-stamped jobs ∪ jobs in a QOS marked `idle_fill_preemptable`, evicted under one tail-priority ordering. Update the §4 tier definition and §8.2 victim selection accordingly.

### 4. Document the full flag set
Add a single consolidated config table:
- `idle_fill_enabled` — cluster only
- `idle_fill_preemptable` — per-QOS (per [3])
- `idle_fill_atomic` — cluster → partition → QOS
- `idle_fill_exempt_secs` — cluster (per-QOS planned)

Currently `idle_fill_preemptable` only lives inside Q5 and there is no consolidated table.

### 5. D1 (unrecallable loan) — exclusion is the hard prerequisite, and it is all aggregates
Affirming D1 as ordering-critical. Make the fix complete:
- Exclude `idle_fill` jobs from every quota aggregate — `sum_running_tres`, `occupied_nodes`, and the running/submitted counts behind `max_jobs_per_user` (`qos.rs:217-221`) and `array_max_concurrent` (`cluster.rs:3666-3691`) — not only node TRES.
- The `max_jobs_per_user` variant is why partial exclusion is insufficient: it is checked before the TRES breach, so lifting the node dimension does not rescue the job — it fails the sole-blocker test too and pends forever. Keep this as the one hard ordering constraint in §11.

### 6. Q3 (borrow cap) — add a `borrowed_tres` counter + two cap policies
Decided: add a borrow cap.
- Introduce a per-QOS `borrowed_tres` counter, distinct from the quota aggregate D1 excludes borrowed jobs from.
- Hard constraint: this counter gates borrow-admission only. It must never feed the legitimate-job gate or `retain_eligible`, or D1 returns through the back door — a borrow-counter value blocking a legitimate job would again strip it before scheduling and kill reclaim. Strictly one-directional: it can only deny a new borrow.
- Two configurable cap dimensions:
  - multiple-of-quota — per-team blast radius (e.g. borrow ≤ N × grp_nodes). Avoid fraction-of-own-quota, which rounds small teams to zero.
  - fraction-of-cluster — cross-team fairness ceiling (e.g. no team borrows > X% of total nodes).
- Sequence after the D1 exclusion fix — additive, behind its own config, does not block the core feature.
- Note that un-capped borrowing is self-limiting (borrowed capacity is reclaimable by definition), so the cap is a churn/predictability knob, not a correctness fix.

shiv-tyagi:
Converting to draft since it a doc review only PR. We do not need to run e2es on this PR on every push. 

We can continue to review even if it is marked draft.

nikhilsk:
## Draft 4 pushed — all six points folded in, three consequences to confirm

`f263916` and `968b0cc`. New sections: §3.1 the consolidated config table, §4.1 the two
sources of "borrowed", §7.1 the borrow cap. §7 now enumerates every aggregate the exclusion
covers. §8.6 is a decision rather than a question. §15 records Q1, Q2, Q3 and Q5 as
resolved.

### One correction to the Q5 reasoning

The premise was that a burst QOS "has quota = whole cluster". Checking
`accounting.rst:952-964`, the documented burst QOS is created with only a priority and a
preempt mode — `sacctmgr add qos name=burst priority=-5000 preemptmode=requeue` — with no
group TRES cap at all. So the reason a burst job is never stamped is that it has no group
node cap to exceed, not that its cap is large.

The conclusion is unaffected and the flag is still needed for exactly the reason given. But
the distinction matters for the implementation: a burst-style QOS that *does* carry a group
node cap can be stamped and flagged at once, so the two sources are a union rather than a
partition. §4.1 now says such a job is borrowed once and takes the stricter treatment where
the rules differ.

### Three consequences of the Q5 decision, which are calls rather than transcription

Keeping `idle_fill_preemptable` means the victim set contains jobs that were never stamped,
and three mechanisms have to treat the two sources differently. I took the conservative
side in each case, on the principle that idle-fill should not change how the burst pattern
behaves for operators who have not adopted idle-fill. Flagging them because they are
decisions, and better confirmed than discovered in the diff.

1. **The quota exclusion keys on the stamp only.** A burst job is not over any quota and
   keeps counting toward the ones it has. Excluding it would misreport the burst QOS's usage
   and silently change the burst pattern's own accounting.
2. **The stale-stamp re-check does not apply to the QOS source.** The flag is read live, so
   clearing `idle_fill_preemptable` stops that QOS's jobs being victims immediately, with no
   per-job re-evaluation. Only the stamp needs re-deriving at reclaim time.
3. **The fairshare exemption covers stamped jobs only.** Burst runs ended by preemption are
   charged to usage today; exempting them would alter existing behavior for operators who
   never turned idle-fill on.

Separately, and worth stating plainly because it reaches beyond idle-fill's own borrowers:
since reclaim consults no preemption setting, enabling idle-fill makes burst jobs
reclaimable on a cluster that never set `preempt_type` or listed `preempt=burst`.
`idle_fill_exempt_secs` still applies to them and the master switch is off by default, but
it is a genuine widening of who can be evicted. §4.1 says so and the new docs page will
too.

### Q4 — worth closing before implementation starts

Which cap do the teams asking for this actually hit: the QOS group node cap, or the
association/account cap? This is the one open question that could invalidate work rather
than adjust it. The account gate runs first and returns early, so an
account-blocked job never reaches the eligibility test at all — extending to account quotas
needs a second predicate and a second collection point, not a parameter change
(`cluster.rs:3716` is the account gate, ahead of the QOS gate in the same closure).

The plan builds for the QOS cap and lists account quotas as a non-goal. If the requesting
teams are actually hitting association limits, that non-goal is the feature.

### Q6 — the Q3 decision has largely answered this

The tension was: exempt reclaimed runs from fairshare and borrowed compute becomes free,
which invites deliberate over-quota submission; charge them and an evicted borrower loses
priority and becomes the preferred next victim, which is self-reinforcing.

Adding the borrow cap changes the balance, because it bounds the abuse case directly and by
construction, where fairshare only discourages it after the fact.

Checking how the charge actually lands strengthens the case further. Usage is recorded at
job end as elapsed time times tasks times CPUs per task, with no reference to *how* the job
ended (`db.rs:440-458`), and the resulting factor is `target_share / actual_share`
(`fairshare.rs:11`), so more usage means a lower factor and a lower priority. The factor is
keyed on **(user, account)**, not on QOS. So charging a reclaimed run does not merely make
that borrowed job a likelier next victim — it lowers the priority of every job that user
runs under that account, including their in-quota work, as a consequence of capacity being
taken back from them. That is hard to defend under the loan model.

Proposal: exempt reclaimed runs of stamped jobs from usage. Burst victims keep today's
treatment, per [3] above.

### Q7 — new: `idle_fill_atomic` has no behavior to configure

It is in the flag table as requested, but flagged as reserved. Placement is already
whole-job-or-nothing (`backfill.rs:577-580`), so `true` describes what the scheduler does
today and `false` would mean building partial dispatch — a much larger change and an
explicit non-goal.

Leaving the field out is the worse option, counter-intuitively: nothing uses
`deny_unknown_fields`, so an operator who writes `idle_fill_atomic = false` gets a config
that parses cleanly and silently ignores them. The proposal is therefore to accept the
field and fail startup with a clear message if it is set to `false`, so the limitation is
stated at the point of use rather than only in documentation. Happy to just drop it if you
would rather not carry a field that only ever rejects one value.

### Unchanged

The hard ordering constraint is affirmed and unchanged: the aggregate exclusion lands
before the mechanism goes live, or the loan is unrecallable. The defect register is up to
17 entries with the severities and citations intact, and the work breakdown is now 13 steps
with the borrow cap last.

PR #881, opened 2026-09-15
feat(spurctld)!: add idle-fill scheduling

nikhilsk:
Implements idle-fill scheduling: a job that has exceeded its QOS group node quota may
run anyway, on capacity no job with a quota claim wants, and that capacity is recovered
when a job which does hold a claim needs it. Design spec in #872.

A quota is a ceiling on what a team may *claim*, not an instruction to leave nodes
empty when nobody wants them. Today an over-quota job waits while nodes sit idle, and
those node-hours are simply lost — the job is not waiting for a resource, it is waiting
for a number.

Off by default (`idle_fill_enabled`), so no existing cluster changes behaviour until an
operator enables it.

## Breaking change

Reclaim ignores `preempt_mode`. On a partition with `preempt_mode = "off"` a borrowed
job can still be evicted, which contradicts a guarantee the shipped docs state
absolutely. The documented promise narrows from "running jobs in this partition are
never kicked out" to **"jobs with a quota claim are never kicked out"**, and that edit
ships in this PR rather than after it.

This was the deliberate choice of two. A borrowed job is *defined* by being
reclaimable, so a setting that shielded it would not produce a safer job — it would
produce capacity that was lent out and can never be recovered, which is worse than
never lending it. The alternative, refusing to lend wherever reclaim is not permitted,
would make the feature inert on a default-configured cluster.

Two things bound the change: the master switch is off by default, and for every job
running inside its quota `preempt_mode = "off"` still means exactly what it says.

## Approach

**Eligibility is exact, not heuristic.** A job qualifies only when its group node cap
is the *sole* limit it breaches, decided by re-running the full QOS check with that cap
lifted. The tempting rule — "blocked with `QOSGrpNodeLimit`" — is wrong in both
directions: the limit check reports only its first breach in cpu/node/mem/gpu order, so
a job over its node cap *and* its GPU cap reports the node one and would be lent nodes,
letting it escape a cap on a genuinely contended resource; while a job over its node
cap *and* its group CPU cap reports CPU and would be refused forever.

**Placement reuses the existing pass rather than adding one.** Candidates are appended
to the tail of the priority-sorted pending list, below every in-quota job. The pass
already walks in caller order, either placing a job and reserving its nodes on the
timeline or reserving the future slot it needs, so a tail job is offered exactly what is
left after every claim-holder has taken and reserved — the definition of spare capacity,
computed by the code that already owns the question. A second filtered pass, which the
first draft proposed, would have been blind to the first pass's assignments and
double-booked partially allocated nodes.

The append point is load-bearing: it must follow the node-set construction, because that
set is derived *from* the pending list, and a candidate's `--nodelist` would otherwise
unpin a node held back by dispatch cooldown for an in-quota job.

**Borrowed runs sit outside every quota aggregate, at both admission gates.** This is
what makes the loan recoverable at all. A borrowed job that counted toward its own
team's quota would block that team's next legitimate job at a gate — and a job blocked
there is stripped from the pending list before the scheduler sees it, so it could never
trigger reclaim. Covering only the QOS gate is not enough: the account gate runs first
and returns early, so the same defect is reachable through it.

**Reclaim is its own routine, not a relaxation of `try_preempt`.** Victim selection there
never checks that an eviction helps — it treats any node the victim occupies in any of
the reclaimer's partitions as an overlap, with no GPU-type, memory, feature or topology
check, and nothing afterwards verifies the shortfall closed. That is harmless only
because the outer loop short-circuits on `preempt_mode = Off` and a hardcoded priority
gap. Reclaim honours neither, so reusing that selection would let a job that can never be
placed anywhere — `--gres=gpu:8` where the largest node has 4 — requeue one borrowed job
every cycle indefinitely, at no cost to whoever submitted it.

Instead the set to evict must provably close the shortfall or nothing is evicted. A node
counts as recovered only when every job on it is reclaimable, making the outcome atomic
and leaving an unplaceable job with no such set at all, permanently.

## Design choices worth reviewing

- **Requeue, not cancel or suspend.** Requeue frees the nodes, closes the accounting
  run, and returns the job with its spec intact. Suspend cannot work: it signals SIGSTOP
  but leaves the allocation held, so capacity is never released and nothing resumes the
  job. Cancel is gratuitous — the job never had a claim, so taking the run back is the
  price, destroying the work is not.
- **The stamp is not trusted at reclaim time.** Raising a cap or a sibling finishing can
  leave a stamped run comfortably inside quota, so the question is re-asked against live
  state; evicting it then would break the very promise above.
- **A separate exempt window.** `preempt_exempt_time` is unbounded and a user can raise
  their own by submitting to several partitions, since the most protective one wins. A
  cluster-wide hour would make borrowed capacity unrecoverable for an hour.
  `idle_fill_exempt_secs` defaults to 60 and doubles per eviction, capped, because the
  existing anti-thrash hold is five seconds at defaults — exactly the agent's
  SIGTERM-to-SIGKILL grace — so a job could otherwise be lent and evicted forever while
  completing nothing.
- **Eviction awaits the cancel, then holds the nodes.** The requeue deallocates inside
  the Raft apply and only then fires a fire-and-forget cancel, so at a one-second cycle
  the reclaimer would land on a node whose previous occupant still holds GPU memory and
  its cgroup — failing in precisely the case the feature exists for.
- **Reclaimed borrowed runs are not charged to fairshare.** Charging them creates a
  self-reinforcing loop: the borrower is billed for work it was not allowed to finish,
  its priority drops, and that makes it the preferred next victim.
- **Four kinds of job are never lent to**: no effective time limit (mandatory — an
  unbounded borrowed job makes its node look busy indefinitely), het components, burst
  buffer, and licenses. The last three because the gate concerned is either unsound or
  not a pure predicate to re-run speculatively.

## Consequences to be aware of

Once borrowed nodes leave the aggregates, neither the QOS nor the association node cap
bounds how much a scope may borrow — only idle capacity does. Bounding it is the borrow
cap's job, which is a follow-up on its own counter. This does not let an account-capped
job borrow: a job must pass the account gate to be considered at all.

A group node cap counts distinct *occupied nodes*, not jobs, so non-exclusive jobs that
pack onto one node may never reach the cap. Expect idle-fill to engage mainly for
exclusive or whole-node work.

## Follow-ups, deliberately not here

- `idle_fill_preemptable` has no accounting schema column, so an operator cannot set it
  yet. The reclaim path honours it; it is documented as not-yet-settable rather than
  presented as a working knob.
- Metrics (nodes-on-loan gauge, reclaim counter) and the borrow cap, which is additive,
  sits behind its own configuration, and controls churn rather than correctness.

## Testing

`cargo clippy --workspace --exclude spur-ffi --all-targets --locked` — warning-free.
`cargo test --locked` — all suites pass, 23 of them idle-fill specific.

The unit coverage that matters most:

- An unplaceable job finds no set to evict, with a control proving a job that *does* fit
  finds one — so the refusal is caused by unplaceability, not an unsatisfiable fixture.
- A set that cannot close the shortfall evicts nothing, and a node shared with an
  in-quota job is never counted as recovered.
- Every excluded aggregate, at *both* gates, each with an unstamped sibling that does
  block — a node-only exclusion would otherwise pass a node-only test.
- Three cross-user cases. The earlier tests all used one mock user, which left the
  per-user aggregates only trivially exercised: with one identity, an exclusion scoped to
  the wrong user would have passed every one of them.
- The stale-stamp rule, extracted into a pure function so it is testable without a
  cluster, including boundary and saturation cases.

### Live cluster

Built from this branch and deployed to a 4-node lab — controller, CLI and all four
agents, every binary reporting the branch commit, so nothing in the path is stale:

```
$ spurctld --version                                    # and spurd on all 4 nodes
spur 0.12.0 (4fd19a8e)
```

**Borrowing, across two real users.** `ifalice` and `ifbob` share a `team` account under
a QOS capped at one node. Alice holds the shared quota; bob, with no jobs at all, borrows
the idle node:

```
$ squeue -o "%i %j %u %t %N %W"
JOBID NAME         USER     ST NODELIST BORROWED
18    alice-legit  ifalice  R  spur-w3  no
19    bob-borrow   ifbob    R  spur-w4  yes
```

**A job with no time limit is refused.** The first attempt did *not* borrow, because the
lab configures no default time limit — the mandatory exclusion, observed rather than
assumed. Adding `-t 10` made it borrow immediately.

**Reclaim, for a third user's claim.** A job holding a genuine quota claim arrived, and
the borrowed node was recovered for it:

```
INFO spurctld::scheduler_loop: reclaimed a borrowed job for a job with a quota claim
     reclaimed_job=19 for_job=20 nodes=["spur-w4"]

$ scontrol show job 19
UserId=ifbob Account=team
JobState=PENDING Reason=QOSGrpNodeLimit
Requeue=0 Restarts=1 BatchFlag=1 Exclusive=1
PreemptedBy=20 PreemptMode=Requeue PreemptQOS=N/A
```

Requeued with its spec intact, and **alice's legitimate job was never touched** — the
narrowed guarantee held exactly as documented.

**Fairshare exemption, with a control.** Bob's reclaimed borrowed run added `0`
cpu-seconds to the `team` account. Alice's ordinary run in the *same* account then
charged `199`. Same table, same code path, so the zero is the exemption rather than a
dead write path.

**Persistence.** An earlier deployment on this branch replayed 68 pre-existing Raft log
entries written by a binary that never heard of `idle_fill`, becoming leader with no
deserialize error — the `#[serde(default)]` risk, which would crash a controller on
upgrade rather than fail a test.

### Docs

Built with the same flags CI uses (`sphinx-build -W --keep-going`) — succeeds. This
caught a genuine failure: navigation here is `docs/sphinx/_toc.yml.in`, not a `toctree`,
and a `:doc:` link alone does not register a page, so the new page would have failed the
build. `pyspelling -c .spellcheck.yaml` reports nothing for any file this PR touches;
the `runtime`/`datasets` hits it does report also occur on a clean checkout of `main`,
verified in a separate worktree, and are a local aspell dictionary difference.


sajmera-pensando (changes requested):
# Review: idle-fill scheduling — blockers + design conformance

Reviewed on a branch rebased onto `rocm:spur/main` (`cargo build -p spurctld` passes). This focuses on the two blocking items and on how the implementation maps to the idle-fill design. Overall the correctness core is faithfully built and the feature is opt-in (`idle_fill_enabled` defaults `false`); the notes below are what to resolve before merge and where the implementation stops short of the full design.

## Blocking

### 1. MAJOR — reclaim's stale-stamp check evaluates each borrowed run in isolation, so a QOS with ≥2 borrowed runs is never fully reclaimable

`crates/spurctld/src/cluster.rs:1692`

`borrowed_run_still_over_quota` re-derives a stamped run's eligibility at reclaim time (so a run that *became* legitimate — cap raised, or a sibling finished — isn't evicted). It uses `legitimate + own > cap`, where `legitimate` counts only non-borrowed nodes in the QOS and `own` is just *this* run's nodes. There is no total-borrowed term — every *other* borrowed run in the QOS is invisible, so each borrower measures itself against the same slice of headroom.

That formula is correct for a single borrower (then `legitimate + own` is the team's whole footprint). With two or more it undercounts: each reads as "became legitimate" when only one actually did.

**Failure:** QOS cap `node=1`, 0 legitimate jobs, two borrowed single-node runs B1/B2 (team physically holds 2 nodes on a cap of 1). A legitimate claim arrives needing one node, cluster full. Reclaim's pool-building step skips any run where the predicate is false: B1 → `0 + 1 = 1`, not `> 1` → skipped; B2 identical → skipped. The reclaimable pool (`scheduler_loop.rs:1062`) is empty, reclaim frees nothing, and the legitimate claim is blocked indefinitely by its own team's borrowed runs. Correct outcome: exactly one borrower spared, one reclaimed. Same shape at `cap=4` with 2 legitimate + 3 borrowed.

This is the expected regime, not an edge case — the design's own safety argument is that "reclaim makes borrowing recoverable, and tail ordering shares it out among borrowers, so uncapped borrowing is self-limiting and not a correctness problem." If reclaim can't recover a multi-borrow QOS, that self-limiting property fails and the promise that a legitimate job can always reclaim its quota is violated. The re-evaluation requirement doesn't prescribe this per-run formula, and the "borrowed must not count" rule is scoped to the admission-gate aggregates, not this reclaim test — so counting the team's total borrowed footprint here is consistent with the design.

**Suggested direction:** re-derive "has this become legitimate?" against the team's *total* borrowed footprint — spare borrowers only down to the cap, treat the surplus above it as reclaimable — or, closer to the routine's design, let all stamped borrowers be eligible and let the minimal satisfiable-victim-set size the eviction, applying the became-legitimate protection at the team level rather than per run.

### 2. MAJOR — docs claim `sacct` shows borrowed status, but no such `sacct` field exists

`docs/admin-guide/idle-fill-scheduling.rst:201` and `docs/user-guide/monitoring-jobs.rst:552-553`

Both pages state `sacct` records whether a completed run was borrowed. The flag is persisted (`jobs.idle_fill`) and exposed via gRPC/REST and `squeue %W`, but `crates/spur-cli/src/sacct.rs` has no format field for it — its field spec covers only JobID…QOS/PreemptedBy/PreemptMode/PreemptQOS. `monitoring-jobs.rst` is also internally inconsistent: its `sacct` field table (~lines 357-360) lists no borrowed field while the prose claims one. A user following the docs cannot display borrowed status via `sacct`. Please confirm whether this needs a missing CLI field (the squeue/sacct commit title implies one was intended) or the docs should drop the `sacct` claim.

## Design conformance

The correctness core of the design is implemented and matches:

- Single priority-ordered pass with over-quota jobs appended to the tail (`scheduler_loop.rs:167-187`), inheriting capacity accounting, backfill, reservations, and topology.
- Atomic placement (whole job or nothing) — inherited from the scheduler's all-or-nothing backfill. Because atomicity is intrinsic, the design's separate `idle_fill_atomic` knob is unnecessary: `false` (partial dispatch) isn't a behavior this scheduler can produce, so a knob for it would be dead config. Recommend dropping it from the design rather than adding it here.
- Legitimate tier is never preempted by idle-fill; a legitimate job reclaims from the opportunistic tier via a self-contained reclaim routine (`scheduler_loop.rs:1020`), orthogonal to the QOS preempt allow-list.
- Borrowed nodes excluded from every quota aggregate (`sum_running_tres`, `occupied_nodes`, the `max_jobs_per_user`/array counts) — the hard prerequisite.
- Reclaim runs regardless of `preempt_mode`, guarded only by the exempt window — the narrowed `preempt_mode = off` guarantee (the `!` breaking change).
- `idle_fill_enabled` (cluster) and `idle_fill_exempt_secs` (default 60) present; automatic, normal path unchanged.

**Gaps to close in this PR.** These are part of the design's core model, not separable follow-ups. Some are framed as follow-ups in the PR description; this review's position is that they belong in this change, with the reasoning given per item.

1. **Borrow bound (`borrowed_tres` counter + the two caps) is missing.** There is no dedicated per-QOS borrow counter and no cap — neither multiple-of-quota (per-team blast radius) nor fraction-of-cluster (cross-team ceiling). The design tracks borrowing in a `borrowed_tres` counter distinct from the quota aggregate, precisely so borrowing can be bounded without re-entering the quota gate. As implemented, borrowing is bounded only by idle capacity, so one QOS can borrow every idle node in the cluster. That is not merely a churn knob here: combined with blocker #1 (a multi-borrow QOS cannot be fully reclaimed), an unbounded borrow becomes an unrecoverable one. The counter and at least the per-QOS cap should land with the mechanism, not after it.
2. **Eviction fate is hardcoded to `PreemptMode::Requeue`** (`scheduler_loop.rs:1088`). The design says a reclaimed job's fate is governed by the existing `PreemptMode` (REQUEUE/CANCEL/CHECKPOINT), "no new configuration needed." The implementation ignores that setting. Either honor `PreemptMode` on reclaim, or, if requeue-always is the deliberate choice (suspend can't release the allocation; cancel destroys work the job had no claim to lose), correct the design/docs so the two agree — but that reconciliation should happen in this PR, not be left open.
3. **`idle_fill_preemptable` is not settable.** The QOS field exists (`accounting.rs:318`) and reclaim honors it (`scheduler_loop.rs:1044`), but it has no accounting schema column and is never read from or written to the DB (nothing in `accounting/db.rs` touches it), so it is hardcoded `false` in practice — an operator cannot enable the burst tier at all. The design makes this flag the *only* way a whole-cluster-quota QOS becomes borrowable, so without settability the burst source of the reclaim victim set (`idle_fill`-stamped ∪ `idle_fill_preemptable` QOS) is unreachable. Add the schema column and the read/write path in this PR rather than shipping an inert knob.

## Test coverage — required in this PR

Every change here must ship with proper end-to-end tests **in this same PR**, not as a follow-up. The PR currently adds solid Rust unit coverage but **no automated e2e** — nothing under `tests/native_host/e2e/` or `tests/k8s/e2e/` is touched, and the live-cluster validation in the description is manual only. That is not sufficient for a scheduler change of this scope.

Specifically, add automated e2e that covers:

- **Regression guard (switch off):** with `idle_fill_enabled = false`, an over-quota job still pends with `QOSGrpNodeLimit` and nothing else changes.
- **Borrow:** an over-quota job runs on idle capacity when the switch is on, and is flagged borrowed.
- **Reclaim for a legitimate claim:** a job with a real quota claim recovers a borrowed node; the borrowed job is requeued with its spec intact and the legitimate claim is never touched.
- **Multi-borrow reclaim** (directly exercises blocker #1): a single QOS holding **two or more** borrowed runs, where a legitimate claim must still be able to reclaim. Both unit and e2e coverage today use effectively single-borrow scenarios, which is why the ≥2-borrower defect went unnoticed — this case must be a permanent test.
- **Unplaceable job evicts nothing:** a job that can never be placed (e.g. `--gres` larger than any node) must not requeue any borrowed job, verified over several cycles.
- **Exclusions:** a job with no effective time limit is refused idle-fill.
- **Burst tier:** once `idle_fill_preemptable` is settable (gap #3), a legitimate job reclaims a burst job whose QOS opted in.

The unit tests are welcome and should stay, but the design (§14) called for this live sequence explicitly, and the multi-borrow blocker is concrete evidence that unit coverage alone was insufficient. Please add the e2es before merge.

nikhilsk:
Thanks for the review. Both blockers are fixed, all three gaps are closed, and the e2e suite found something. Details and evidence below.

### [1] Multi-borrow reclaim (blocker)

Your trace was exactly right. Replaced the per-run test with a team-level one, `team_over_quota(cap, legitimate, borrowed)`, evaluated once per QOS and cached across that QOS's runs. That is the second of your two suggested directions: the became-legitimate protection now applies at team level, and the minimal satisfiable victim set still sizes the eviction.

I deleted the unit test that had pinned the per-run behaviour as correct, since it was the reason this survived review-by-tests. In its place is a matrix over 0-3 borrowers against caps 1-4, plus a regression test for your exact `cap=1` two-borrower case, asserted through the real accessor rather than the pure helper.

Verified live on a 4-node cluster, same cluster and same pending job, with only the controller binary swapped. Before, on the pre-fix build, the claimant sat `PD / Resources` for 90 seconds and no reclaim decision was logged at all. After:

```
reclaimed a borrowed job for a job with a quota claim reclaimed_job=54 for_job=57 nodes=["spur-w2"]
```

One borrower was evicted, not both. The other kept its node and neither in-quota job was touched.

### [2] sacct borrowed field (blocker)

Added the field rather than dropping the claim, since the commit title did intend one. `Borrowed` is letter `W`, matching `squeue %W`. I added a consistency test that walks every named `sacct` field and asserts each has both a header and a resolver; it immediately caught a pre-existing bug where `ReqMem` had a header but no resolver and rendered `?`, fixed in the same commit. The `monitoring-jobs.rst` field table now lists both, so prose and table agree.

Live:

```
$ sacct -j 54,55 --format=JobID,User,State,Borrowed
54 ifcarol PREEMPTED yes
55 ifdave  RUNNING   no
```

### [3] Borrow bound

Added the per-QOS borrowed-node counter, kept separate from the quota aggregates a borrowed run sits outside of, seeded from running jobs and then charged as the pass admits candidates so a single cycle cannot admit a batch that collectively breaks the ceiling. Both dimensions from the design are there: `idle_fill_max_borrow_factor` as a multiple of the QOS's own group node cap, and `idle_fill_max_cluster_fraction` as a share of the registered nodes. The tighter wins, and fractions floor so a ceiling never overshoots. It is consulted only where a new loan is admitted, which keeps it one-directional: lowering it denies the next borrow and never evicts a run that is already borrowing.

One judgement call to flag. Both default to `0.0`, meaning unbounded. With the multi-borrow gate fixed, reclaim can recover a multi-borrow QOS again, so the design's own position holds and the cap is a churn and predictability control rather than a correctness fix. Any non-zero default would be guessing at an operator's policy. Happy to ship one if you would rather.

Live, against a cap of `node=1`: at factor `1.0` the second loan was refused with `QOSGrpNodeLimit` while a node sat idle; at `2.0` the same job borrowed that node.

### [4] PreemptMode on reclaim

Took the reconcile-the-docs option. The other two modes cannot serve reclaim: suspend never releases the allocation, which is the one thing reclaim needs, and cancel destroys work the run had no claim to lose when requeue preserves it at no cost to the reclaimer. The page stated the requeue rule without ever naming the setting it overrides, so an operator who had set `preemptmode=cancel` had no way to know it did not apply. It now names `cancel` explicitly, and the reasoning is recorded at the call site.

Verified live and also pinned by an e2e test: a QOS with `preemptmode=cancel` had its borrowed run requeued with its stamp intact, and the claimant took the freed node 10 seconds later.

### [5] idle_fill_preemptable settability

You were right that it was inert. `limits_cache.rs` hardcoded it to `false`, so no operator action could have enabled the tier. It is now carried end to end: an additive column defaulting to `FALSE`, proto tag 19 on `QosInfo` and `CreateQosRequest`, the accounting read and update paths, the limits cache, and `sacctmgr` as `idlefillpreemptable=yes|no` with an `IdleFillPreemptable` display field. `parse_bool` rejects unrecognised values rather than reading them as off, so a typo cannot silently disable the tier.

Live: `no` on all eight pre-existing QOS rows, sets and clears correctly, refuses `maybe` without changing the stored value, and survives a controller restart.

### [6] idle_fill_atomic

Agreed, and nothing to change here. `rg idle_fill_atomic` finds no hit in code, config or docs on this branch, so the knob was never introduced. Removing it from the design document is yours to make.

### Test coverage

Six of your seven cases are now automated in `tests/native_host/e2e/test_idle_fill.py` and pass on real hardware: the switch-off regression guard, borrowing, reclaim for a legitimate claim, multi-borrow reclaim in two shapes, unplaceable-evicts-nothing, and the no-time-limit exclusion. The second multi-borrow shape is the one a per-run test provably mishandles, because it removes the team's legitimate run so that each borrower fits the cap alone.

Every test fills the cluster before asking for the behaviour under test. Without that step a waiting job is satisfied by the spare node, no reclaim ever fires, and the test passes while asserting nothing.

### The burst-tier case is the seventh, and it is not here because it found a bug

Reclaim deallocates the node inside Raft before the evicted run's completion report arrives. The report is then refused with `node <n> is not allocated to job <j>`, the agent treats that as non-retryable and gives up, and the controller never records the run as finished. The job is requeued and relaunched repeatedly (I observed `Restarts=5`), and the reclaimer is planned for a future slot instead of the node just freed for it, while `sinfo` reports that node idle throughout.

```
05:48:56 spurd:    reported completion to controller job_id=1 exit_code=0
05:50:17 spurctld: reclaimed a borrowed job ... reclaimed_job=1 for_job=4 nodes=["spur-w4"]
05:50:17 spurctld: node_complete failed job_id=1 node=spur-w4
                   error=node spur-w4 is not allocated to job 1
spurd:             ReportJobStatus failed with non-retryable error job_id=1
```

`scontrol show job 4` then reports `SchedNodeList=spur-w4` with a `StartTime` set to when the unrelated filler jobs end, rather than now.

That undercuts the promise that a job with a quota claim takes the capacity back promptly, so it wants fixing rather than papering over. The test is correct and stays red until the ordering is fixed. CI runs the whole e2e directory, so it lands together with that fix rather than red, and I would rather not soften it with an xfail. It is my immediate follow-up.

One more thing from the same debugging, unrelated to this change: when a rootless agent cannot use `/var/spool/spur` it falls back to `/tmp/spur`, but the launch error reports the permission failure against `/var/spool/spur` and never names the path that actually blocked it. A stray file at `/tmp/spur` cost an evening of chasing the wrong path.

sajmera-pensando (changes requested):
## Opportunistic-tier preemption does not follow the design's priority model

I checked this against the design summary canvas (*Idle-Fill Scheduling: Design
Summary & Comparison*), which is the source of truth. The canvas is explicit that the
opportunistic tier is **priority-ordered**, and the PR does not implement that. Two
gaps, plus missing tests. The exempt window, by contrast, is implemented correctly.

### What the design mandates

From the canvas, *Preemption Changes*:

> **Within the opportunistic tier, priority determines order: a higher-priority
> idle-fill job outranks a lower-priority burst job. Burst cannot displace idle-fill
> unless burst has equal or higher effective priority.**
>
> When a **legitimate** (within-quota) job arrives and needs nodes: […] **Eviction
> order: lowest-priority opportunistic jobs first (typically burst), then
> higher-priority idle-fill if still insufficient.**
>
> When an **opportunistic** job arrives and needs nodes: It can only preempt other
> opportunistic jobs of strictly lower priority. […] **A burst job (low priority,
> `idle_fill_preemptable = true`) cannot displace a higher-priority idle-fill job.**

So the design has two distinct priority rules inside the opportunistic tier:

- **(A) Legitimate reclaim is priority-ordered among victims** — evict the
  lowest-priority opportunistic job first (typically burst), and only climb to a
  higher-priority idle-fill job if that is still not enough.
- **(B) Opportunistic-vs-opportunistic preemption** — a higher-priority idle-fill job
  can displace a strictly-lower-priority burst job; the reverse is forbidden.

This is not incidental. It is the resolution to the burst edge case raised in design
review: a burst QOS has cluster-wide quota and very low priority, so without a
priority rule it would be a permanent free victim in one direction and never yield in
the other. The whole point of considering priority here is that burst is being phased
out and idle-fill takes over, so idle-fill must outrank burst, not the other way
round.

### The design doc (#872) does not capture this use case

Before the PR gaps, the written design in #872 is itself the root of the problem: it
does not carry the opportunistic-tier priority model that the canvas resolved. The
canvas states both rules (A) and (B) plainly; #872 silently narrows them:

- **Rule (B) is dropped entirely.** #872 §5.1/§8.2 model reclaim as *in-quota jobs
  only* evicting borrowed jobs. There is no notion of a higher-priority idle-fill job
  displacing a lower-priority burst job — the burst edge case that motivated the whole
  priority discussion is simply absent from the doc.
- **Rule (A) is stated too weakly to implement.** #872 §4.1 disposes of victim
  ordering in one line — "a burst QOS sits at a large negative priority by
  construction, so it already sorts to the tail alongside the stamped candidates, and
  §8.2 evicts in one ordering across both sources." That assumes a priority-sorted
  victim list without ever requiring one, and §8.2's actual steps never sort by
  priority. An implementer following §8.2 literally gets node-order eviction — which is
  exactly what shipped.

So the burst-vs-idle-fill priority guarantee lives only in the canvas and never made it
into the design doc the implementation was built from. The doc should be corrected to
carry rules (A) and (B) explicitly — with the burst phase-out rationale — so the
canvas and #872 agree and the implementation has an unambiguous spec to target. Right
now the doc, if read on its own, describes a feature that does not honor priority in
the opportunistic tier at all.

### Gap 1 — an idle-fill job can never preempt a burst job (rule B is absent)

The reclaimer set explicitly excludes idle-fill candidates
([scheduler_loop.rs:258](crates/spurctld/src/scheduler_loop.rs#L258)):

```rust
.filter(|p| !borrow_ids.contains(&p.job_id))
```

with the comment "An unplaced borrow candidate is deliberately absent from this list.
It has no claim on the capacity it wanted, so it must not evict anyone." As a result
an unplaced idle-fill job reclaims nothing — it simply pends. A high-priority idle-fill
job that needs nodes currently held by a low-priority burst job will **not** displace
it, which is the exact scenario the canvas says must work ("a higher-priority idle-fill
job outranks a lower-priority burst job"). Rule (B) is not implemented at all.

### Gap 2 — legitimate reclaim does not evict lowest-priority-first (rule A is unordered)

`satisfiable_victim_set` picks victims by **node iteration order**, not priority
([scheduler_loop.rs:1177-1197](crates/spurctld/src/scheduler_loop.rs#L1177)): it walks
`cluster_state.nodes`, collects evacuable nodes, and takes them until it has enough.
Both borrowed sources are eligible — stamped idle-fill and `idle_fill_preemptable`
burst ([scheduler_loop.rs:1047-1056](crates/spurctld/src/scheduler_loop.rs#L1047)) —
but nothing orders them by priority.

Failure case: a legitimate claim needs one node; a low-priority **burst** job sits on
node A and a high-priority **idle-fill** job on node B. The design requires the burst
job to be evicted (lowest-priority first). The PR evicts whichever node sorts first in
the list — so it can requeue the higher-priority idle-fill job and spare the
lower-priority burst job, the precise inversion the canvas rules out ("Burst cannot
displace idle-fill").

Note this is also a mismatch against the narrower GitHub design doc (#872 §4.1), which
still requires victims to be evicted "in one ordering" with burst sorting to the tail
by its large negative priority. Node-order selection satisfies neither document.

### Gap 3 — no e2e (or unit) coverage for any of this

No test sets `idlefillpreemptable=yes`, so the burst tier is entirely unexercised end
to end. Specifically missing:

- **Legitimate reclaim prefers burst over idle-fill** — a legitimate claim with one
  burst and one higher-priority idle-fill victim available must evict the burst one.
  (Directly covers Gap 2.)
- **High-priority idle-fill displaces low-priority burst** — covers Gap 1 / rule (B).
- **Low-priority burst cannot displace higher-priority idle-fill** — the negative
  control.

**Policy: every new flag ships with an e2e.** More broadly, each new knob this PR
introduces should land with an end-to-end test that exercises it, not just unit
coverage. Right now that bar is not met:

- `idle_fill_preemptable` (per-QOS) — no e2e sets it at all.
- `idle_fill_max_borrow_factor` — no e2e; needs one asserting a borrow is refused at
  the cap and admitted below it.
- `idle_fill_max_cluster_fraction` — no e2e; needs one, including the tighter-of-the-two
  interaction with the factor.
- `idle_fill_exempt_secs` — only used as a delay; no e2e asserts protection *before* the
  window or the doubling backoff after a requeue.

A flag with no e2e is a flag whose real-cluster behavior is unverified — the burst tier
here is the concrete example, where the missing test is exactly why the priority gaps
above went unnoticed.

### Implemented correctly: the exempt window

`idle_fill_exempt_secs` (the "idle-fill preemption exempt time" in the design) is
present and is the sole guard on reclaim: config default 60
([config.rs](crates/spur-core/src/config.rs)), enforced in `idle_fill_exempt_window`
with per-eviction doubling backoff
([scheduler_loop.rs:998](crates/spurctld/src/scheduler_loop.rs#L998)), checked before a
victim is eligible ([scheduler_loop.rs:1065](crates/spurctld/src/scheduler_loop.rs#L1065)).
The per-QOS override the canvas lists as "planned" is not implemented, which matches
the design's own framing. The doubling backoff has unit coverage but no e2e; worth one.

### Ask

Two things, and they are linked:

1. **Fix the design doc (#872) first.** Bring rules (A) and (B) over from the canvas —
   priority-ordered victim eviction (burst first) and higher-priority idle-fill
   displacing lower-priority burst — with the burst phase-out rationale. Without this
   the doc and the canvas disagree and the implementation has no unambiguous spec.

2. **Then make the PR match.** Either implement the opportunistic-tier priority model
   (rules A and B) with e2e coverage, or — if intra-tier priority is being deliberately
   deferred — say so in the doc and the PR, and still fix Gap 2, because evicting a
   higher-priority borrower while a lower-priority burst job keeps running is a behavior
   an operator would reasonably treat as a bug. Gap 2 is the minimum bar for this PR.

nikhilsk:
Thanks for the second pass — the opportunistic-tier gaps were real and both rules are now implemented and tested.

**Eviction order.** A legitimate claim now spends the cheapest opportunistic run first, and takes only as many nodes as it needs rather than everything reclaimable.

**Opportunistic against opportunistic.** An opportunistic reclaimer may now only displace work strictly cheaper than itself, so a burst job can no longer evict a dearer borrowed run. The reverse, a higher-priority idle-fill job displacing lower-priority burst, works.

Worth flagging how the first attempt failed, because anyone implementing from the design would hit the same thing. I first compared `Job::priority`, which is `base * fair_share * age_factor * partition_tier` and therefore grows while a job waits. Instrumentation showed a long-queued burst job at 10010 against a borrowed run at 6000, so the guard concluded the burst job was the expensive one and protected it. Standing within a tier is a property of the QOS, not of queue position, so both rules now compare `Qos::priority`. That is a stable `i32`, and a burst QOS sits at a large negative delta by construction, so it sorts below stamped idle-fill without special-casing.

That work also exposed a third defect worth its own mention: reclaim over-evicted whenever a freed node was still inside its kill-grace cooldown. The node was invisible to the next scheduler pass, so the pass reclaimed a second victim one second later. That silently defeated the ordering rule even once the comparison was right, so there is now a short in-flight hold per reclaiming job.

**E2E.** Four of the new knobs had none, which was a fair call. Twelve tests now cover the switch being off, borrowing, single and multi-node reclaim, the unplaceable job evicting nothing, the no-time-limit refusal, eviction order, the borrow factor ceiling with a control below it, the cluster fraction as the tighter of the two, and the exempt window protecting a fresh borrow and then releasing it. All twelve pass on a three-node cluster.

sajmera-pensando (dismissed):
## Approving — priority model and knob coverage are in

The two blockers from my earlier review are resolved, and cleanly:

- **Rule (A) — eviction order.** `satisfiable_victim_set` now spends the cheapest
  opportunistic run first, ties broken on node name, and cost is **QOS priority**
  rather than `Job::priority` — the right call, since `Job::priority` climbs with age
  and a long-queued burst job would otherwise have outranked a fresh borrowed run. The
  `reclaim_in_flight` guard also stops the second-victim over-eviction that would have
  defeated the ordering.
- **Rule (B) — opportunistic-vs-opportunistic.** The reclaimer set now includes borrow
  candidates, gated by a ceiling at the reclaimer's own QOS priority, so a
  higher-priority idle-fill job can displace strictly-lower-priority `idle_fill_preemptable`
  work and a low-priority burst job cannot displace a dearer borrowed run. The victim
  side correctly requires the `idle_fill` stamp **or** an `idle_fill_preemptable` QOS,
  so an unflagged burst QOS is never touched.
- **Flag coverage.** New e2es for the eviction order, both borrow caps (with a
  below-ceiling control), and the exempt window (both halves — protected inside,
  reclaimed after). The master switch is off by default and the feature is opt-in.

Good to merge. The items below are **not blockers** — please pick them up in a
follow-up PR.

### Follow-up: tests

- **An e2e for rule (B) proper.** The current priority e2e uses a *legitimate* claim as
  the reclaimer (rule A). The case where a **higher-priority idle-fill job itself**
  reclaims from a lower-priority `idle_fill_preemptable` burst job — and the negative
  control, that a low-priority burst job cannot displace a higher-priority idle-fill
  job — is only unit-tested. Since rule (B) is the specific behavior this review turned
  on, it deserves end-to-end coverage.

### Follow-up: docs

- **The priority model is undocumented.** `idle-fill-scheduling.rst` describes reclaim
  as "a job with a claim evicts borrowed jobs" but never states rules (A) or (B): that
  the lowest-priority opportunistic job (typically burst) is evicted first, or that a
  higher-priority idle-fill job outranks a lower-priority burst job while the reverse is
  forbidden. This is the heart of the burst-vs-idle-fill behavior and should be on the
  page.
- **One line needs qualifying.** `idle-fill-scheduling.rst:154` says "Reclaim consults
  none of ... the priority gap." That is true for a *legitimate-claim* reclaim but not
  for the opportunistic-vs-opportunistic path, which now consults QOS priority as a
  ceiling. Please distinguish the two so the page matches the shipped code.
- **Borrow-cap knobs on the feature page.** `idle_fill_max_borrow_factor` and
  `idle_fill_max_cluster_fraction` are in `configuration.rst` but absent from the
  dedicated page, so a reader learning the feature there won't see that borrowing can be
  bounded.
- **Reconcile the design doc (#872).** It still does not carry rules (A)/(B); the
  guarantee lives only in the design-summary canvas. Bringing the doc in line keeps the
  written spec and the implementation from drifting.

sajmera-pensando (approved):
Please raise a JIRA for tracking the follow up items. @nikhilsk

PR #883, opened 2026-09-16
fix(spurctld): stop charging QOS/account grp-node quota for unplaceable jobs

yansun1996:
## Problem

A job's QOS/account `grp_tres` node cap can stay permanently exhausted even
when zero jobs are actually `Running`. The scheduler's admission gate
(`classify_pending_jobs`) reserves a job's node charge against the cap the
moment it passes the QOS/account check, before real node placement is
attempted. If every node matching that job is administratively down, the job
can never place (it eventually surfaces as `ReqNodeNotAvail`/`NodeDown` via a
later, separate check), but nothing ever releases the charge it already
reserved. The same stuck job re-claims the cap on every scheduling pass,
starving lower-priority jobs in the same QOS with `QOSGrpNodeLimit` even
though nothing is running.

## Fix

Add a structural pre-check (`structural_unplaceable_reason`) that runs before
the QOS/account charge: if every node matching a job's placement is down, or
a required listed node in an additive nodelist (a short `-w` list padded by
`--nodes`) is down, it's blocked with the real node-state reason and never
reaches the charge.

The additive-nodelist rule was previously duplicated across `find_suitable_nodes`
(real placement) and `update_pending_reasons` (post-hoc reason tagging); it's
now factored into `NodePlacement::additive_listed_node_unavailable` in
`spur-sched`, shared by all three call sites so they can't drift apart.

Deliberately scoped to the down-node case only. A broader version that also
treated "too few eligible nodes registered" as structural regressed several
existing tests that submit jobs against a cluster with no nodes registered yet
(where a job's eligibility legitimately depends on nodes that may still join)
— so that case is left charging normally, unchanged from today's behavior.

## Known limitations (follow-up, not fixed here)

- A job pinned via a non-additive `--nodelist` naming a mix of a down node and
  a node that was never registered at all falls into the "too few eligible
  nodes" bucket and is not caught by this fix — same bug class, narrower
  trigger. This exact reason-classification imprecision already exists today
  via `update_pending_reasons`'s identical logic (verified), so this fix
  doesn't introduce or worsen it.
- `PartitionNodeLimit` and `BadConstraints` (a job whose partition/constraint
  can never be satisfied by current topology) are equally permanent but also
  fall outside this fix's scope, for the same regression reason above.
- The main "all eligible down" check verifies node state but not resource
  capacity — a job whose eligible nodes are all up but structurally too small
  to ever satisfy the request isn't caught either. This matches existing
  behavior: today's codebase already classifies that case as `Resources`
  (contention-shaped), not a structural reason, so this fix doesn't change it,
  just doesn't extend to it.

All three are real but rarer paths than the reported down-node case; recommend
a follow-up once "cluster still filling in" can be distinguished from
"topology that structurally can never satisfy this" with more confidence.

## Testing

- Four new unit tests, each confirmed red without the fix and green with it:
  a job that starts out Pending pinned to an already-down node; a job that
  was genuinely `Running`, fails, and requeues onto a node that has since
  gone down (driven through the real `start_job`/`requeue_job` production
  path, not a direct state injection); a job pinned via an additive nodelist
  to one specific down node padded out with otherwise-idle capacity; and a
  guard asserting the real cap still blocks a job that can genuinely place
  (both against the end-to-end outcome and directly against the new
  structural-check function).
- Fixed one existing test that relied on a single node which the new gate now
  correctly keeps blocked after a simulated failure drains it — added a
  second healthy node so the test's actual intent (verifying a release clears
  a hold) still holds.
- The refactored additive-nodelist predicate inherits `find_suitable_nodes`'s
  own 8 existing tests (multi-node lists, exclude/constraint, reservations,
  hostlist expansion) in addition to the new dedicated test, since it's now
  the same shared code.
- `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` clean,
  `cargo fmt --all` clean, full `spurctld` (1108) and `spur-sched` (140) suites
  passing.
- Validated end-to-end on an isolated deployment, two scenarios:
  1. A job pinned to a drained node correctly shows `ReqNodeNotAvail` instead
     of consuming a 1-node QOS cap, while a second job under the same QOS
     reaches `RUNNING` on the other node rather than being blocked by
     `QOSGrpNodeLimit`.
  2. A job genuinely running under the same 1-node-cap QOS, whose node is
     then killed for real (not just admin-drained), correctly requeues to
     `ReqNodeNotAvail` — and a second job submitted under that QOS while the
     first is failing over still reaches `RUNNING` on the surviving node
     rather than getting stuck behind the requeued job's charge.


yansun1996:
Added coverage for the requeue path specifically: a job that was genuinely
\`Running\` (consuming real QOS quota), fails, and requeues onto a node that
has since gone down. New unit test (\`qos_grp_node_does_not_charge_after_a_running_job_is_requeued_onto_a_down_node\`)
confirms it — red without the fix, green with it.

Also validated end-to-end on the isolated deployment through the actual
requeue mechanism (not just initial submission): a job running on a node,
that node failing for real, the job requeuing to \`ReqNodeNotAvail\`, and a
second job under the same 1-node-cap QOS reaching \`RUNNING\` on the other
node rather than being blocked.

yansun1996:
Checked both suppressed Copilot findings against the actual code (not just the review text):

**Additive-nodelist finding (\`cluster.rs:7207\`, the \`nodelist=n1\` + \`--nodes=2\` example)** — confirmed, real gap. Traced \`find_suitable_nodes\` (\`spur-sched/src/backfill.rs:262-271\`): it rejects the whole job when a required *listed* node is down, regardless of other idle capacity. The structural gate didn't mirror that — fixed by adding the same additive-nodelist check, with a new red/green-proven test.

**Non-additive nodelist finding (\`cluster.rs:7207\`, the \`nodelist=n1,n2\` unregistered-node example)** — confirmed as described, but pre-existing: \`update_pending_reasons\` has the identical \`eligible.len() < needed\` logic already, so this exact misclassification happens today too, just later in the pipeline (after the charge, not before). This fix doesn't introduce or worsen it. Documented as a known limitation rather than expanding scope further.

crates/spurctld/src/cluster.rs
```diff
@@ -7181,6 +7187,47 @@ fn license_block(job: &Job, pool: &HashMap<String, u64>) -> Option<spur_core::jo
...
+    if placement.additive_listed_node_unavailable(
+        eligible.iter().copied(),
+        reservations,
+        now,
+        &required,
+    ) {
+        return Some(PendingReason::ReqNodeNotAvail);
+    }
+
+    if eligible.len() < needed || eligible.iter().any(|n| n.state.is_up()) {
```
biluriuday: is_up() ignores k0s reservation, but the additive branch right above it (and real placement, NodePlacement::matches) exclude k0s-reserved nodes. So a job whose only up eligible node is k0s-reserved passes this escape → charges → then fails placement (matches() drops the k0s node) → the exact leak this PR fixes, via a different node class. Align the escape with real placeability (and with the additive branch, which already uses this):

if eligible.len() < needed
    || eligible.iter().any(|n| placement.matches_for_reservation(n, reservations, now))
{
    return None;
}
matches_for_reservation is !is_k0s_reserved() && eligible && is_up(). I verified this doesn't regress the empty-/filling-cluster case (the len() < needed arm still short-circuits first). Reachability is narrow (needs a k0s worker sharing a batch job's partition/features and being the sole up-eligible node), but given Spur's k8s/GPU focus it's plausible.

biluriuday (commented):
Correct root-cause fix, well-scoped, with real red/green tests. The refactor is provably behavior-preserving. One issue highlighted in-line. PTAL

crates/spurctld/src/cluster.rs
yansun1996: Good catch — fixed, using `matches_for_reservation` like the additive branch above it. Added a regression test (`qos_grp_node_does_not_charge_when_the_only_up_node_is_k0s_reserved`), confirmed red without the fix. Pushed.

biluriuday (dismissed):
LGTM. Please cross-check the e2e failure

PR #884, opened 2026-09-16
feat(spurd): add --container-remap-root for Enroot-style rootless root

shiv-tyagi:
## What this adds

`--container-remap-root` lets a container workload run as uid 0 *inside* the container while the host identity stays the unprivileged submitter. This is the Enroot model: the process gets root-in-container (can write root-owned image paths, install into the rootfs, etc.) without ever becoming real host root. Works for both `sbatch` and `srun`.

Stacked on [#889](https://github.com/ROCm/spur/pull/889) (userns-first namespace ordering + workload-pid tracking). Review that first; the remap commit sits on top of it.

## Approach

Namespaces follow the runc/Podman order from #889: the user namespace is created and mapped first, then the mount and PID namespaces are created inside it, so they are owned by the user namespace.

`user_namespace_mode(is_root, remap_root)` selects one of three modes:

- **`Host`** (root daemon, no remap): no user namespace. The workload drops from real root to the submitter via `drop_privileges`, which `setgroups` the submitter's full supplementary group list (`video`/`render` included). Unchanged existing behavior.
- **`CurrentUserAsRoot`** (rootless daemon): the daemon's own uid self-maps to namespace root (`0 <uid> 1`, `setgroups=deny`). `--container-remap-root` is a no-op here since the submitter is already root-in-container.
- **`SubmitterAsRoot`** (root daemon + `--container-remap-root`): the new path. The shepherd `unshare`s the user namespace and blocks on a socketpair while the root parent writes its `0 <submitter> 1` uid/gid maps over `/proc/<pid>/`, then acks. Only the root parent still holds `CAP_SETUID` in the init userns, so it (not the child) must write the maps.

Because the shepherd is real host root, it is unmapped in the new namespace (only the submitter is, at id 0) and shows as `nobody`. `assume_container_root()` does `setgid(0)`/`setuid(0)` inside to become container root (the submitter on the host). It runs last, after the privileged mount/pivot setup, since the uid change clears the effective capability set.

**Submitter-owned rootfs.** Under remap the per-job rootfs is extracted submitter-owned, forced into extracted mode (no overlay fast path), and `--container-name` reuse is rejected, so container-root can write every path. The recursive chown uses `lchown` (never follows symlinks) to prevent escape via crafted images.

**Nested / exec entry.** A step entering a running remap container over `nsenter` lets nsenter's default reset uid/gid to 0 (the submitter), using the caps it holds in the namespace. No in-container binary is required. Rootless keeps `--preserve-credentials` because its unprivileged userns has `setgroups=deny` and the daemon uid already maps to 0.

## Compatibility

No proto renumbering, no config field changes, no persisted-state (Raft/WAL/snapshot) changes. `remap_root` is an in-memory launch-config field. Existing non-remap paths (rootless, and root-without-remap `Host` mode) are unchanged, gated entirely behind the flag.

## Limitations

- Single-uid/gid mapping: files owned by any id other than the mapped one appear as `nobody` inside.
- Group-gated device access (GPU `video`/`render`) does not carry through under remap: `setgroups=deny` plus the single-gid map collapses supplementary groups. Full-fidelity groups are available on the non-remap `Host` path. (Follow-up below.)

## Testing

**Unit:** mode selection, id-map handshake, `assume_container_root`, and `build_nsenter_argv` (all three modes) covered. `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` clean, `cargo fmt` clean, `cargo test -p spurd` green.

**E2E** (`tests/native_host/e2e/test_container.py`): `srun` and `sbatch` remap tests for uid 0 in-container, `/root` writable with remap, `/root` denied and non-zero uid without the flag, and a bind-mount proof that container-root-written files are submitter-owned on the host.

**Manual matrix, 4-node bare-metal testbed, root daemons:**

| Case | Scenario | Result |
|---|---|---|
| P1 | `srun --container-remap-root` | uid 0, `/root` writable |
| N1 | `srun` no flag (root daemon) | uid 1000, `/root` denied |
| P2 | remap + bind mount | root inside, submitter-owned on host |
| P3 | `sbatch` remap (parent/child map handshake) | uid 0, `/root` writable, host submitter-owned |
| P4 | step `nsenter`s into running remap job | uid 0, sees batch job's `/root` marker, writable |
| N2 | remap + `--container-name` | rejected, workload never runs |

Core security property confirmed: container root is the unprivileged submitter on the host, never real root.

Two bugs were found on the testbed and fixed in this PR:
1. Remap workload came up as `nobody` (uid 65534) instead of 0. A root daemon's shepherd is unmapped in the namespace, so it must explicitly assume container root. Added `assume_container_root()`.
2. `nsenter` into a running remap job failed with `failed to execute setpriv`. The old approach ran `setpriv` after `nsenter`, but nsenter resolves the command in the container's mount namespace, so it required `setpriv` inside every image. Replaced with nsenter's built-in credential reset (no in-container binary).

Rootless collapse was not re-run on the testbed this round (the flag is a no-op there); mode selection is covered by unit tests.

- The `--container-remap-root` + `--container-name` rejection currently fails the job at container setup with a bare exit code 1; the reason does not reach the client. Validate at submit time for a clear error.
- Remap + GPU group-gated devices: a root daemon could install a multi-range gid map and allow `setgroups` for the (root-created) namespace to carry `render`/`video` through. Not attempted here.


biluriuday (commented):
One query on the architecture decision:
Generic model (Enroot, podman/crun rootless): create the user namespace first, on the process that becomes the container leader, then mount inside it. Two consequences fall out for free: (1) /proc/<leader>/ns/user is authoritative, so nsenter-in and re-exec are naturally correct; (2) the mount namespace is owned by the new userns, so container-root holds CAP_SYS_ADMIN over its own mounts.

This PR (root daemon): inverts the order — real-root mount/pivot → drop to submitter → unshare(CLONE_NEWUSER) last, in the forked workload. The motivation is legitimate: keep unrestricted real-root mount capability during setup and sidestep userns mount limits. But it produces two deviations:

- Container-root's mount ns is owned by the host userns, so container-root can't mount/remount/unmount — a "neutered" root. 
- The userns is created on a non-tracked, post-fork process. The generic model doesn't have this failure mode because the leader is in the userns. The code even re-derives this exact hazard for the pid namespace ("that process stays in its original pid namespace … enter through pid_for_children") and handles it — but the identical reasoning for the user namespace (the unsharer stays behind; only the child enters) was not applied. There is no ns/user_for_children, so the only correct target is a process actually in the remapped userns — the workload.

So: the deviation from the generic pattern is defensible for setup, but it was carried into the entry/credential path in the wrong way. Aligning with the generic model on that one axis — the process you nsenter into must be inside the remapped userns

shiv-tyagi:
## Testing

Validated on the 4-node bare-metal testbed in both daemon modes (root and rootless), plus local clippy/unit/e2e-compile. Container image was a root-layout squashfs (`/root` owned root:root). Leak checks submit a sleeper, cancel, and verify no leftover processes, mounts, or per-job rootfs dirs.

### Smoke matrix

| Area | Case | Checks | Root daemon | Rootless daemon |
|---|---|---|---|---|
| Non-container | A1 srun stream | uid=submitter, rc=0 | PASS | PASS |
| Non-container | A2 srun `--pty` | uid=submitter, tty=yes | PASS | PASS |
| Non-container | A3 sbatch to file | uid=submitter | PASS | PASS |
| Container (non-remap) | B1 srun stream | uid, rc=0 | PASS | PASS |
| Container (non-remap) | B2 srun `--pty` | uid, tty=yes | PASS | PASS |
| Container (non-remap) | B3 sbatch to file | uid | PASS | PASS |
| Container (remap) | C1 srun stream | uid=0, /root writable | PASS | PASS |
| Container (remap) | C2 srun `--pty` | uid=0, tty=yes, /root writable | PASS | PASS |
| Container (remap) | C3 sbatch to file | uid=0, /root writable | PASS | PASS |
| Container (remap) | C4 bind-mount ownership | host file owned by submitter, not root | PASS | PASS |
| Container (remap) | C5 host-side groups | supplementary groups == submitter's, no gid 0 | PASS | N/A (remap no-op) |
| Negative | D1 remap + `--container-name` | rejected under root; no-op under rootless | PASS | PASS |
| Negative | D2 non-container exit code | exit 7 propagated | PASS | PASS |
| Negative | D3 container exit code | exit 9 propagated | PASS | PASS |
| Multi-node | E1 `srun -N4` | ran on all 4 distinct nodes | PASS | PASS |
| Breakout | G isolation battery | see table below | PASS | N/A (remap no-op) |
| Leak | L1 non-container cancel | no leftover procs/dirs | PASS | PASS |
| Leak | L2 container cancel | no leftover procs/dirs | PASS | PASS |
| Leak | L3 remap cancel | no leftover procs/dirs | PASS | PASS |
| Leak | L4 rapid submit+cancel x6 | no accumulation | PASS | PASS |
| **Total** | | | **21/21** | **18/18** |

### Container breakout / isolation (remap, root daemon)

A workload running as container-root (uid 0, mapped to the unprivileged submitter on the host) attempts each escape. All must be blocked.

| Vector | Result |
|---|---|
| `mknod` block device | blocked |
| raw disk read (`/dev/sda`) | blocked |
| write host `kernel.core_pattern` | blocked |
| write `/proc/sysrq-trigger` | blocked |
| mount cgroupfs (CVE-2022-0492 release_agent) | blocked |
| `insmod` kernel module | blocked |
| read root:root 0640 file via bind (group-root leak) | blocked |
| append root:root 0660 file via bind | blocked, host file intact |
| overwrite/chown/chmod host root-owned file via rw bind | blocked |
| host PIDs via fresh `proc` mount | isolated (only container PIDs) |
| `/proc/1/root` | container rootfs, not host |
| file created on shared mount | owned by submitter (uid 1000) on host |

Host-side identity of a running remap workload: `Uid: 1000`, `Groups: 4 24 27 30 46 101 109 1000` (the submitter's real groups, no gid 0). The group-root read/write cases above were a real leak before this change and are now blocked.

### Local validation

| Check | Result |
|---|---|
| `cargo clippy -p spurd --all-targets --locked` | clean (0 warnings) |
| `cargo test -p spurd --lib` (container) | 74/74 pass |
| e2e `test_container.py` compile | ok |

PR #885, opened 2026-09-16
feat(spurd): name step cgroups like Slurm and contain task hooks

biluriuday:
Build on the already-nested job_<id>_<attempt>/step_<id> topology run by spurstepd to reach Slurm parity for step naming and task hooks.

- Naming: render reserved steps as batch/extern/interactive via spur_core::step::step_display_name (scontrol) and step_dir_name (cgroup leaves: step_batch/step_extern/step_interactive/step_<n>).
- Task hooks: run task_prolog/task_epilog as the job user inside the step cgroup -- in spurstepd for batch/supervised srun, in-agent for legacy nsenter steps -- with the Slurm TaskProlog export/unset/print protocol.
- Hardening: run_hook (and run_submit_hook) validate an absolute path plus owner/mode and exec the fd via /proc/self/fd (shared secure_hook_command).

Deferred: interactive-primary -> STEP_INTERACTIVE scoping (flips supervisor_owns_teardown for pty jobs; needs cluster validation).

Tested: workspace fmt/clippy/test; bare-metal native-host E2E -- test_cgroup_limits.py 21/21 and new test_step_cgroups.py 4/4 (step_batch naming, numbered srun leaf, container step_batch, and task hooks run as the job user in the step cgroup with the export/print protocol).


crates/spurd/src/agent_server.rs
```diff
@@ -4575,6 +4576,7 @@ impl SlurmAgent for AgentService {
             gid: spec.gid,
             container: container_launch,
             prolog_script: None,
+            task_prolog_script: self.hooks.task_prolog.clone(),
```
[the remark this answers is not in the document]
biluriuday: deferring it to follow-up

crates/spurd/src/agent_server.rs
```diff
@@ -6174,9 +6216,17 @@ impl SlurmAgent for AgentService {
...
             }
         };
+        let mut stdout = read_back(stdout_path).await;
+        if !task_prolog_printed.is_empty() {
+            stdout = format!(
+                "{}{}",
+                String::from_utf8_lossy(&task_prolog_printed),
+                stdout
+            );
+        }
```
shiv-tyagi: Legacy `RunCommand` splices TaskProlog `print` into the gRPC `stdout` after spool read-back. Host `srun` is supervised and writes the prefix onto the fd, so it is fine. Standalone `srun --container-image` stays on this path. `srun` live-tails the spool; when that tail succeeds it skips `emit_step_output`, so `print` never reaches the user.

Write the prefix onto the spool stdout fd after TaskProlog and before dispatch (same as `JobIo::write_prefix`). Then live tails and read-back both see it, and this splice can go.

crates/spurd/src/task_hook.rs
```diff
@@ -0,0 +1,387 @@
...
+
+/// Run a task hook as the job user, joined to the step cgroup leaf. Captures and
+/// caps stdout/stderr while draining both, logs stderr, and fails on a non-zero
+/// exit or a capped overflow.
+async fn run_task_hook(
+    script_path: &str,
+    context: &HookContext,
+    task_environment: &HashMap<String, String>,
+    cgroup_path: Option<&Path>,
+) -> anyhow::Result<TaskHookOutput> {
```
shiv-tyagi, line 120: Nit: `run_hook` logs `info!(..., "running hook")` before exec. Task hooks have no start log, only stderr `warn`s and failure errors. A hung TaskProlog has nothing in the agent or stepd log that the hook started.

crates/spurd/src/executor.rs
```diff
@@ -1336,6 +1456,17 @@ impl CgroupJoin {
...
 
+    /// Like [`join`](Self::join), but a failed join is an error rather than a
+    /// silent degrade — for a short-lived child (a task hook) that may exit
+    /// before parent-side membership verification is meaningful. Async-signal-safe.
+    pub(crate) fn join_required(&self) -> std::io::Result<()> {
+        if self.join() {
+            Ok(())
+        } else {
+            Err(std::io::Error::from_raw_os_error(libc::EPERM))
+        }
```
shiv-tyagi: Nit: `join_required` maps every failed join to `EPERM`. `join()` already collapses open/write failures to `false`, so a missing `cgroup.procs` or `EBUSY` surfaces as `Permission denied` on TaskProlog. Preserve the real errno from the join (or at least distinguish ENOENT/EBUSY from EPERM).

tests/native_host/e2e/test_step_cgroups.py
```diff
@@ -0,0 +1,209 @@
...
+        )
+        out_path = f"{cluster.remote_dir}/th.out"
+        sb = cluster.sbatch(
+            ["-J", "th", "-N", "1", "-w", cluster.node_names[0], "-t", "2",
+             "--cpus-per-task=1", "--mem=256", "-o", out_path, payload]
+        )
+        job_id = parse_job_id(sb)
+        assert job_id is not None, f"sbatch failed: {sb}"
+        wait_job(cluster, job_id, timeout=120)
+        out = cluster.wait_output(out_path, "PAYLOAD", timeout=120)
```
shiv-tyagi, line 175: Task-hook E2E covers batch only (`export` / `print` / uid / `step_batch`). Numbered `srun` is asserted for leaf naming, not for hooks. A single-node, unredirected container `srun` hook test asserting `print` output would catch the legacy live-tail path, where `print` is spliced into the RPC body after spool read-back.

crates/spurd/src/stepd.rs
```diff
@@ -1998,6 +2015,22 @@ pub async fn run_process(args: &[String]) -> anyhow::Result<i32> {
...
+                    &task_epilog_context,
+                    &task_environment,
+                    cgroup.as_deref(),
+                )
+                .await
+                {
+                    tracing::warn!(job_id, %error, "TaskEpilog failed");
+                }
+            }
+        }
```
shiv-tyagi, line 2034: Q: TaskProlog runs inside `launch_job`. If it succeeds and the payload spawn then fails, `launch_job` returns `Err` and this teardown never runs, so TaskEpilog is skipped. Is TaskEpilog only meant to run after a started task, or should a hook that allocated something in TaskProlog still get paired cleanup on that failure path?

shiv-tyagi (commented):
Some comments. PTAL.

crates/spurd/src/agent_server.rs
biluriuday: Fixed. Write task_prolog_printed to the spool stdout file right after TaskProlog and before dispatch (step_files.stdout is an unbuffered std::fs::File, and all three legacy arms wire the workload's stdout from that fd) — so the live tail and the read-back both include it. Removed the response splice (read-back now already contains it, so keeping the splice would double it). It's a no-op for supervised steps: their stepd writes the prefix via JobIo::write_prefix, and task_prolog_printed is empty here.

crates/spurd/src/task_hook.rs
biluriuday: run_task_hook now logs a start line before spawning, mirroring run_hook:
Added info to the tracing import. Since both TaskProlog and TaskEpilog go through run_task_hook, each now emits running task hook with hook=prolog_task/epilog_task before exec — so a hung hook is visible in the stepd/agent log

crates/spurd/src/executor.rs
biluriuday: join_cgroup_self now returns std::io::Result<()> carrying the real errno instead of collapsing to bool:

open failure → last_os_error() (e.g. ENOENT for a missing cgroup.procs).
write failure → last_os_error() (e.g. EBUSY); a short count (no errno) → WriteZero.
errno captured before close() so it isn't clobbered. Remains async-signal-safe (last_os_error() just reads thread-local errno + wraps it, no alloc).
join() keeps its bool contract (.is_ok(), for the best-effort callers), and join_required() now simply returns the Result — so a TaskProlog join failure surfaces its true errno rather than a blanket EPERM. The two best-effort `let _ = join_cgroup_self(...)` raw-fork call sites are unaffected.

tests/native_host/e2e/test_step_cgroups.py
biluriuday: added a test

crates/spurd/src/stepd.rs
biluriuday: Yes — TaskEpilog should (and now does) pair with a successful TaskProlog even when the launch then fails. I wrapped launch_job's post-prolog body in an async block and, on any failure after TaskProlog ran, run the paired TaskEpilog before returning the error
Design points:

- The epilog context + post-prolog environment + step-cgroup leaf are captured before the launch consumes them, so a failed launch can still reach the leaf. It uses the same post-prolog env and cgroup the success-path epilog does.
- It fires for every post-prolog failure (container spawn, namespace-wrapper write, command spawn, the cgroup-required check) — not just the payload spawn — since the wrap catches them uniformly.
- It runs inside launch_job, before the stepd's cleanup_failed_launch reaps the cgroup, so the epilog still runs in the live leaf. No double-run: success → stepd teardown; failure-after-prolog → here; paired_epilog is None unless both prolog and epilog are configured (and reaching this point means TaskProlog already succeeded).
- Added task_epilog_script to JobLaunchConfig, routed from hooks.task_epilog in into_launch_config and set at every other construction site.

shiv-tyagi (approved):
LGTM

PR #886, opened 2026-09-17
feat(spur-cli): implement sbatch --wait

### LINKED ISSUE #882: sbatch --wait not supported (Slurm parity gap for blocking job submission)
## Summary

Spur's `sbatch` does not accept `--wait`, which Slurm has supported for years as the standard way to block until a batch job reaches a terminal state.

```bash
$ spur sbatch --wait job.sbatch
Error: unexpected argument '--wait'
```

On Slurm clusters we rely on `sbatch --wait --parsable` so the submit call itself does not return until the job finishes (success or failure), and we get a bare job id for logging.

## Why this matters for us

We run the **coco** RCCL test harness against Spur clusters (Crusoe `crs-spur`, Spur 0.12.0). coco's Slurm backend submits build and test jobs with `--wait --parsable` when talking to real Slurm. For Spur we had to set `submit_blocks=False` and poll `squeue` instead — and that polling is unreliable.

### Use case 1: RCCL build jobs marked done while still running

On Crusoe nightly perf (`crusoe-rccl-perf-nightly`), coco reported an RCCL build as **Built** while the sbatch job had not even started executing:

| Event | Time (UTC) |
|-------|------------|
| coco marks build complete | 00:19:27 |
| sacct Start | 00:19:28 |
| sacct End (actual compile) | 00:20:31 |

Job **141520**: coco moved on after `squeue` went empty; the build log had 10k+ compile lines but no `::coco-build-end::` marker and no `all_reduce_perf` binary. Downstream: **216/216 tests blocked**, 0 executed.

Root cause: `squeue` can be empty while the job is still pending or between state transitions. coco's current rule ("once seen in queue, empty = done") fires too early.

### Use case 2: Non-blocking submit + poll is fragile

Without `--wait`, callers must implement their own completion detection. Options we evaluated:

| Approach | Result |
|----------|--------|
| Poll `squeue` (current coco workaround) | False completion (use case 1) |
| Poll `sacct`, treat empty output as done | **Also fails** — sacct returns nothing while job is PENDING |
| Poll `sacct`, wait for terminal state only | Works in manual test (job 141761, 40s sleep → COMPLETED) |

So sacct polling *can* work but requires careful semantics; `sbatch --wait` would eliminate the race entirely and match Slurm behaviour our tooling already expects.

### Use case 3: Exit code propagation

With non-blocking submit, `sbatch` returns 0 on successful **submission**, not on script success. We only learn the build failed later (missing binaries, incomplete logs). `--wait` would propagate the batch script exit code at submit time, as Slurm does.

## Request

Add Slurm-compatible `sbatch --wait` (and ideally `--wait` together with `--parsable`) so blocking submission works the same as on Slurm:

```bash
spur sbatch --wait --parsable job.sbatch   # blocks until terminal; prints job id
```

Related: `--parsable` already exists and works; only `--wait` is missing.

## Environment

- Spur CLI: 0.12.0 (`/usr/local/bin/spur`)
- Cluster: Crusoe `crs-spur.crusoe.amd.com`, partition `amd-spur`
- Harness: [tRCCL/coco](https://github.com/ROCm/tRCCL) SpurBackend (`submit_blocks=False`)

## Workaround (partial)

coco could switch from `squeue` to `sacct` polling with "empty sacct = still in flight, terminal states only = done". That fixes the false-early-completion case but still doesn't give blocking submit or script exit codes at the CLI boundary.

bhatturu:
# PR: feat(spur-cli): implement sbatch --wait

## Problem

Spur's `sbatch` rejects `--wait` with "unexpected argument". Slurm has supported `sbatch --wait` for years as the standard way to block a submission until the job reaches a terminal state and to propagate the job's exit code back to the caller. Test harnesses like tRCCL/coco use `sbatch --wait --parsable` against real Slurm; against Spur they had to fall back to polling `squeue`, which is racy — `squeue` can go empty during state transitions, causing false-positive completion signals. On Crusoe `crs-spur` 0.12.0, this caused a build job (141520) to be marked `Built` at 00:19:27 UTC while `sacct` shows the compile didn't finish until 00:20:31, blocking 216/216 downstream RCCL tests with missing binaries.

`sacct` polling isn't a reliable fallback either: accounting records don't appear until the job enters `RUNNING` (empty during `PENDING`), and `sacct -X` fails entirely on current Spur. `--wait` removes the need for any caller-side polling.

## Fix

Add a `-W`/`--wait` boolean flag to `SbatchArgs` in `sbatch.rs`. When set, after `SubmitJob` returns and the job ID is printed (so `--parsable` output is emitted at submission time, matching Slurm), the CLI polls `GetJob` using `tokio::time::interval` (immediate first tick, then 1-second cadence) until the job reaches a terminal state, then exits with the job's exit code. Stdout is flushed before entering the poll loop so piped consumers receive the job ID at submission time.

Terminal state detection uses `JobState::try_from(job.state)` with `spur_core::job::JobState::from_proto(state).is_terminal()`, the same pattern as `srun.rs:670`. This structurally excludes `Preempted`, `Requeued`, and `Suspended` — no raw integer matching.

Exit code mapping (`terminal_exit_code`):
- `COMPLETED` — propagate the job's own `exit_code`.
- `FAILED` — `exit_code.max(1)`, clamping the controller's `-1` sentinel to 1 (matches `srun.rs:1171`).
- All other terminal states (`CANCELLED`, `TIMEOUT`, `NODE_FAIL`, `DEADLINE`, `OUT_OF_MEMORY`) — exit 1, aligned with `srun`.

Requeue race handling: requeue is not atomic (`cluster.rs:2057-2086` proposes `Failed` then re-pends). When a requeueable job lands on a non-`Completed` terminal state, the poll loop re-confirms once before returning, so a 1 Hz poller can't exit during the transient `Failed` window.

RPC error handling uses a transient allowlist: `Unavailable`, `DeadlineExceeded`, `Aborted`, `Internal`, and `Unknown` are retried with warn-once. All other codes (`NotFound`, `PermissionDenied`, `Unauthenticated`, etc.) are fatal — exit 1 with error message.

Ctrl-C during the wait calls `CancelJob` via the shared `interactive::install_ctrl_c_cancel` helper, which was promoted from `srun.rs` (where it was private) into `interactive.rs` as part of this PR. All three call sites (`sbatch`, `srun`, `salloc`) now use the single shared implementation.

The `submit_user` for cancel RPCs is cloned from `job_spec.user` before `SubmitJob`, matching `salloc.rs:118`, so a lookup failure can't orphan a submitted job.

No proto changes, no server changes, no new RPCs.

## Known limitations

- **Array jobs**: Spur's controller reports the array parent job as `COMPLETED` once all tasks finish and `FAILED` if any task fails. `--wait` polls the parent job ID and propagates that composite exit code. This matches Slurm's `--wait` behavior for arrays (blocks until all tasks reach a terminal state). Not tested with `%N` throttling.
- **Poll interval**: fixed at 1 second, same as `salloc`. For very long-running jobs this is fine; for a future high-frequency submission pipeline it could be made configurable or use exponential backoff. Slurm's `--wait` also polls at a fixed interval internally so this is parity.
- **No `WatchJob` RPC**: Spur has no streaming/subscribe mechanism for job state changes, so polling `GetJob` is the only option. If `WatchJob` is added later, `--wait` could switch to it for lower latency.
- **`SBATCH_WAIT` env var**: Slurm honours this; Spur does not wire it into the env layer. No other sbatch bool flag has an env var either, so this is follow-up material rather than a gap.

## Testing

**Unit tests** — 5 flag/directive parsing tests (`--wait`, `-W`, default false, `--parsable` combo, directive merge), 4 `terminal_exit_code` tests (completed propagation, failed clamping including `-1` sentinel, other-states-exit-one, non-terminal exclusion of `Suspended`/`Preempted`/`Requeued`).

**Mock controller integration tests** — 7 tests driving `wait_for_job` through the in-process mock with scripted `JobInfo` sequences:
- `wait_polls_through_pending_to_completed` — Pending → Running → Completed(0), verifies ≥3 polls and exit 0
- `wait_propagates_nonzero_exit_code` — Pending → Failed(42), returns 42
- `wait_clamps_negative_exit_code` — Failed(-1) sentinel, returns 1 (not 255)
- `wait_skips_suspended_and_preempted` — Running → Suspended → Preempted → Pending → Running → Completed(7), verifies ≥6 polls
- `wait_exits_on_permanent_rpc_error` — NotFound, immediate exit 1, single poll
- `wait_cancelled_exits_one` — Pending → Cancelled, returns 1
- `wait_timeout_exits_one` — Running → Timeout, returns 1

`cargo test --package spur-cli`: 531 passed, 0 failed. `cargo clippy --workspace --all-targets --locked` and `cargo fmt --all --check` clean.

**Live E2E on banff-cyxtera-cx58-11 (8× MI300X)** — 20 scenarios, all pass:

| # | Test | Exit | Expected |
|---|------|------|----------|
| 1 | success (`echo hello`) | 0 | 0 |
| 2 | exit 42 | 42 | 42 |
| 3 | exit 1 (`false`) | 1 | 1 |
| 4 | exit 255 (8-bit boundary) | 255 | 255 |
| 5 | exit 127 (command not found) | 127 | 127 |
| 6 | `-W` short flag | 0 | 0 |
| 7 | `--wait` without `--parsable` | 0 | 0 |
| 8 | `--wait --parsable` (coco pattern) | 0 | 0 |
| 9 | `scancel` during `--wait` | 1 | 1 |
| 10 | timeout (`--time=0:00:10`) | 1 | 1 |
| 11 | without `--wait` baseline | 0, 94ms | 0, <5s |
| 12 | script file with `--wait` | 7 | 7 |
| 13 | `#SBATCH --wait` directive | 3 | 3 |
| 14 | 3 concurrent `--wait` (0,11,22) | 0,11,22 | 0,11,22 |
| 15 | array job `--array=0-3` | 0 | 0 |
| 16 | array job, task 2 fails | 1 | nonzero |
| 17 | `--dependency=afterok:N` | 0 | 0 |
| 18 | `--hold` + `scontrol release` | 0 | 0 |
| 19 | output file still written | 0, content ok | 0, has content |
| 20 | `kill -9 $$` sentinel clamp | 1 | 1 (not 255) |

Closes #882.


crates/spur-cli/src/sbatch.rs
```diff
@@ -1044,9 +1050,85 @@ pub async fn main_with_args(cli_args: Vec<String>) -> Result<()> {
...
 }
 
+/// Map a terminal job state to a process exit code.
+///
+/// COMPLETED uses the job's own exit_code. FAILED does too (the batch script's
+/// non-zero code). CANCELLED mirrors the standard SIGINT convention (128 + 2).
+/// All other terminal states exit 1.
+fn terminal_exit_code(state: i32, exit_code: i32) -> i32 {
+    match state {
+        3 => exit_code,                   // JOB_COMPLETED
```
yansun1996: These raw proto tags are the root of a couple of the other findings. Would it be worth matching on the typed enum instead — either `spur_proto::proto::JobState::try_from(state)` with named variants (the `srun.rs:670` idiom) or `spur_core::job::JobState::from_proto_i32(state)` (`scancel.rs:161`, `squeue.rs:351`)?

That would give you three things at once: the `JOB_PREEMPTED` inclusion goes away because `is_terminal()` already excludes it, the terminal set stops being duplicated between here and the range guard at line 1119, and the `// JOB_*` trailing comments become unnecessary.

Worth flagging that the two enums are numbered differently — `spur_core`'s has `Cancelled = 5, Failed = 6` while the proto has `JOB_FAILED = 4, JOB_CANCELLED = 5` — so it's easy for hand-written integers to silently mean the wrong state if this code ever moves between the two.

crates/spur-cli/src/sbatch.rs
```diff
@@ -1044,9 +1050,85 @@ pub async fn main_with_args(cli_args: Vec<String>) -> Result<()> {
...
+        4 if exit_code != 0 => exit_code, // JOB_FAILED (script's exit code)
```
yansun1996: `exit_code: -1` is the controller's sentinel for machine-finalized jobs (`cluster.rs:2059`, `3747`, `5158`), and it reaches the client unclamped — `server.rs:3957` does `exit_code: job.exit_code.unwrap_or(0)` with no range check. So this arm passes `-1` through to `std::process::exit`, which the shell reports as 255 — indistinguishable from a script that genuinely exited 255.

`srun` already guards this with `std::process::exit(exit_code.max(1))` (`srun.rs:1171`). Could this do the same, or clamp to `1..=255`?

crates/spur-cli/src/sbatch.rs
```diff
@@ -1044,9 +1050,85 @@ pub async fn main_with_args(cli_args: Vec<String>) -> Result<()> {
...
 
+    if wait {
+        let submit_user = crate::interactive::current_user()?;
+        std::process::exit(wait_for_job(&mut client, job_id, &submit_user).await);
+    }
+
     Ok(())
 }
 
+/// Map a terminal job state to a process exit code.
```
yansun1996: Two small things on this block.

It's five lines and mostly restates the match arms just below it, which reads as a *what* comment — the repo convention is to reserve comments for the non-obvious *why*. Switching to named enum variants (see the note below) would make most of it redundant anyway.

The SIGINT part isn't quite accurate either: `CANCELLED` → 130 is right for Ctrl-C, but a job cancelled via `scancel`, or cancelled by the controller itself (e.g. an unsatisfiable dependency, `cluster.rs:3745`), also lands here and gets 130 without any SIGINT involved. For comparison `srun` exits 1 on `JobCancelled` (`srun.rs:1173`). Worth aligning the mapping with `srun`, or keeping 130 and rewording the comment to say why.

crates/spur-cli/src/sbatch.rs
```diff
@@ -1044,9 +1050,85 @@ pub async fn main_with_args(cli_args: Vec<String>) -> Result<()> {
...
+
+/// Poll `GetJob` until the job reaches a terminal state, cancelling on Ctrl-C.
+async fn wait_for_job(
+    client: &mut spur_proto::proto::slurm_controller_client::SlurmControllerClient<
+        crate::authclient::AuthChannel,
+    >,
+    job_id: u32,
+    submit_user: &str,
+) -> i32 {
+    let cancel_client = client.clone();
```
yansun1996: This block is essentially identical to `install_ctrl_c_cancel` in `srun.rs:626-653` — same clone-then-`let mut`, same `resolve_job_owner_for_cancel`, same `signal: 2`, same `exit(130)`. `salloc.rs:144-166` is a third inline copy of the same thing.

Since the helper already exists, would you be up for promoting it (it's currently private to `srun.rs`) into `interactive.rs` and calling it from all three? That gives the cancel path a single home, and the pre-existing rough edges I mentioned in the summary — the discarded `cancel_job` result and the two racing `process::exit` calls — become fixable in one place instead of four.

crates/spur-cli/src/sbatch.rs
```diff
@@ -1044,9 +1050,85 @@ pub async fn main_with_args(cli_args: Vec<String>) -> Result<()> {
...
+                    signal: 2, // SIGINT
+                    user: cancel_user,
+                })
+                .await;
+            std::process::exit(130);
+        }
+    });
+
+    loop {
+        tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
```
yansun1996: Minor: the sleep is at the top of the loop, so even a job that's already finished by the time we get here costs a full second before the first `GetJob`. `srun.rs:658` uses `tokio::time::interval(..)` with `.tick()` at the top of the loop instead, whose first tick completes immediately — same cadence afterwards, but no dead first second.

crates/spur-cli/src/sbatch.rs
```diff
@@ -1044,9 +1050,85 @@ pub async fn main_with_args(cli_args: Vec<String>) -> Result<()> {
...
+            Ok(resp) => {
+                let job = resp.into_inner();
+                match job.state {
+                    3..=8 | 10 | 11 => {
+                        return terminal_exit_code(job.state, job.exit_code);
+                    }
+                    _ => {}
+                }
+            }
+            Err(e) => {
```
yansun1996, line 1138: Every `GetJob` error is retried here indefinitely — no bound, no backoff, and one stderr line per second. If the controller is down, mid-election, or returns something non-transient, `sbatch --wait` becomes an infinite spinner that a caller can't distinguish from a long-running job.

Would it make sense to treat `Code::NotFound` and `Code::PermissionDenied` as fatal (exit non-zero with a clear message — note `server.rs:919` also surfaces the hidden/not-authorized case as `not_found`), and for genuinely transient codes add backoff plus a warn-once latch? `srun.rs:691` already uses a `warned_unknown_state` flag for the warn-once part.

crates/spur-cli/src/sbatch.rs
```diff
@@ -1044,9 +1050,85 @@ pub async fn main_with_args(cli_args: Vec<String>) -> Result<()> {
         println!("Submitted batch job {}", job_id);
     }
 
+    if wait {
+        let submit_user = crate::interactive::current_user()?;
```
yansun1996: Small ordering nit: `current_user()?` runs after the job has already been submitted, so if it fails here the `?` propagates and `sbatch` exits with an error while the job stays queued and unwaited — the caller gets no ID handling and no wait.

`build_sbatch_job_spec` already resolved the same value at line 905, so this is also a second lookup. `salloc.rs:118` handles it by reusing `job_spec.user.clone()` right after building the spec — could this do the same?

crates/spur-cli/src/sbatch.rs
```diff
@@ -1809,4 +1891,90 @@ echo "hello world"
...
+    }
+
+    #[test]
+    fn test_wait_directive_fallback() {
+        let args = parse_merged(&["--wait"], &["sbatch"]);
+        assert!(args.wait);
+    }
+
+    #[test]
+    fn test_wait_cli_overrides_directive() {
```
yansun1996: This one passes an empty directive list (`parse_merged` takes directives first, CLI argv second), so it only asserts that a CLI `--wait` sets `wait` — which `test_wait_flag_parsed_long` above already covers. It would still pass with `fallback!(wait, "wait")` deleted, so it isn't testing the merge path its name describes.

Since a set-only bool can't be overridden back to false from the CLI, there may not be a meaningful override case to test — the directive-only direction is already covered by `test_wait_directive_fallback`. Probably simplest to drop this one, or rename it to match what it actually asserts.

yansun1996 (commented):
Nice, focused change — `--wait` is a real gap for harnesses that currently poll `squeue`, and the exit-code mapping plus Ctrl-C handling are the right shape. A few things worth another look before merge.

The three existing bot comments all look correct to me: `JOB_PREEMPTED` doesn't belong in the terminal set, the new async path has no test coverage, and the flag needs a docs entry. I've left some additional notes inline.

The biggest one is that the terminal set is written as raw proto tags (`3..=8 | 10 | 11`) and duplicated between the range guard and `terminal_exit_code`. The repo has two established idioms for this — `spur_core::job::JobState::from_proto_i32(..).is_terminal()` (`scancel.rs:161`, `squeue.rs:351`) and `spur_proto::proto::JobState::try_from(..)` with named variants (`srun.rs:670`). Worth noting the two enums are numbered *differently* (core has `Failed = 6, Cancelled = 5`; proto has `JOB_FAILED = 4, JOB_CANCELLED = 5`), so hand-written integers here are exactly the hazard those helpers exist to avoid. Switching to named variants would fix the PREEMPTED case structurally and collapse the duplication at the same time.

One design question: requeue isn't atomic with finalization. `cluster.rs:2057-2086` proposes `JobComplete { state: Failed, exit_code: -1 }` and only then proposes the requeue back to `Pending`; `cluster.rs:1954-1959` does the same for `Timeout`/`NodeFail` via `maybe_requeue`. A 1 Hz poller can land in that window and return exit 1 for a job that is about to rerun. Would it be worth re-polling once after observing a non-`Completed` terminal state (or checking the restart count) before returning?

On tests — the fifteen added tests all exercise either clap parsing or the pure `terminal_exit_code` mapper. `wait_for_job` itself has no coverage: not the `3..=8 | 10 | 11` dispatch, not the poll loop, not the error branch. States 9 (`SUSPENDED`) and 12 (`REQUEUED`), whose *exclusion* is the whole correctness argument, aren't asserted anywhere — deleting the range guard would leave the suite green. `mock_controller.rs` is close to being able to host this: it already has a `get_job` with a settable response and a call counter, but it only returns `user`, so the response would need to carry `state`/`exit_code` (ideally a small queue so a test can script `Pending → Running → Completed(42)` and assert the return after exactly three calls). `tokio::time::pause()` would keep it fast without adding a timeout. A pytest e2e would fit well too — `tests/native_host/e2e/cluster.py:435` already has `srun_with_exit()` to model an `sbatch_with_exit()` on.

One heads-up: the PR description mentions `mock_controller.rs` changes (`submit_job`, `cancel_job`, `get_job_sequence`) that aren't in the current diff — `submit_job` and `cancel_job` are still in the `unimplemented { }` block and there's no `get_job_sequence`. Looks like they were dropped somewhere along the way, so the description may need updating.

Finally, a few things here are copied from `srun`/`salloc` rather than new problems — the discarded `cancel_job` result, both tasks racing to `process::exit`, SIGINT-only handling, and `signal: 2` (the server hardcodes `0` at `server.rs:952` and never reads the field). Not asking you to fix those in this PR, but they would all get a single home with the shared-helper suggestion below.

Two minor ones not worth their own threads: the `--wait` help text at `sbatch.rs:259` ("exit with the job's exit code") is only accurate for COMPLETED/FAILED, since CANCELLED gives 130 and the rest give 1; and `SBATCH_WAIT`, which Slurm honours, isn't wired into the env layer — though no other sbatch bool is either, so that's follow-up material rather than a gap here.

crates/spur-cli/src/sbatch.rs
bhatturu: Fixed. `terminal_exit_code` now takes `JobState` enum. Poll loop uses
`JobState::try_from(job.state)` + `spur_core::job::JobState::from_proto(state).is_terminal()`,
same pattern as `srun.rs:670`. Eliminates raw integers and the duplicated

crates/spur-cli/src/sbatch.rs
bhatturu: Fixed. `FAILED` arm uses `exit_code.max(1)`, matching `srun.rs:1171`.

crates/spur-cli/src/sbatch.rs
bhatturu: Fixed. Switched from `sleep()` at loop top to `tokio::time::interval()` with
immediate first tick, matching `srun.rs:658`.

crates/spur-cli/src/sbatch.rs
bhatturu: Fixed. `NotFound` and `PermissionDenied` are now fatal (exit 1 with error
message). Other errors use warn-once pattern from `srun.rs:691`
(`warned_unknown_state`), avoiding infinite spam.

crates/spur-cli/src/sbatch.rs
bhatturu: Fixed. `submit_user` is now cloned from `job_spec.user` before `SubmitJob`,
matching `salloc.rs:118`. No failure window between submit and wait.

crates/spur-cli/src/sbatch.rs
bhatturu: Replaced with `non_terminal_states_excluded_from_is_terminal` — asserts that
`Pending`, `Running`, `Completing`, `Suspended`, `Preempted`, and `Requeued` are
all non-terminal. This is the correctness property the reviewer identified as
unasserted.

crates/spur-cli/src/sbatch.rs
bhatturu: Aligned with srun. `CANCELLED` now exits 1 (was 130). The Ctrl-C handler still
exits 130 directly (same as srun's `install_ctrl_c_cancel`), but a `scancel`
from another shell produces exit 1. Verified on live cluster — T06 confirms.

crates/spur-cli/src/sbatch.rs
bhatturu: Fixed. Promoted `install_ctrl_c_cancel` into `interactive.rs` as a shared
`pub fn`. srun, salloc, and sbatch all call through it now. Net -25 lines
across 4 files. The pre-existing rough edges (discarded `cancel_job` result,
racing `process::exit` calls) now have a single home to fix.

crates/spur-cli/src/sbatch.rs
```diff
@@ -256,6 +256,10 @@ pub struct SbatchArgs {
     #[arg(long)]
     pub parsable: bool,
 
+    /// Block until the job terminates and exit with the job's exit code
+    #[arg(short = 'W', long)]
+    pub wait: bool,
```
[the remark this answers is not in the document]
bhatturu: Will add a docs entry in `docs/user-guide/submitting-jobs.rst`.

crates/spur-cli/src/sbatch.rs
```diff
@@ -1044,9 +1050,84 @@ pub async fn main_with_args(cli_args: Vec<String>) -> Result<()> {
...
+    });
+
+    loop {
+        tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+
+        match client.get_job(GetJobRequest { job_id }).await {
+            Ok(resp) => {
+                let job = resp.into_inner();
+                match job.state {
+                    3..=8 | 10 | 11 => {
```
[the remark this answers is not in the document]
bhatturu: Fixed. The raw `3..=8 | 10 | 11` range is gone entirely. The poll loop now uses
`JobState::try_from(job.state)` with `spur_core::job::JobState::from_proto(state).is_terminal()`,
which structurally excludes `Preempted` (and `Requeued` and `Suspended`).

crates/spur-cli/src/sbatch.rs
```diff
@@ -1044,9 +1050,84 @@ pub async fn main_with_args(cli_args: Vec<String>) -> Result<()> {
...
+        _ => 1,
+    }
+}
+
+/// Poll `GetJob` until the job reaches a terminal state, cancelling on Ctrl-C.
+async fn wait_for_job(
+    client: &mut spur_proto::proto::slurm_controller_client::SlurmControllerClient<crate::authclient::AuthChannel>,
+    job_id: u32,
+    submit_user: &str,
+) -> i32 {
```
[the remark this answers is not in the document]
bhatturu: The mock_controller extensions were removed from this PR (they introduced
dead-code CI failures). The async `wait_for_job` path requires a live gRPC
connection — the pure logic (terminal state classification and exit code
mapping) is covered by unit tests, and the integration path is validated by the
20-test E2E suite on live MI300X hardware. The
`non_terminal_states_excluded_from_is_terminal` test asserts the key correctness
property: that `Suspended`, `Preempted`, and `Requeued` are all excluded from
the terminal set.

crates/spur-cli/src/sbatch.rs
```diff
@@ -1044,9 +1051,91 @@ pub async fn main_with_args(cli_args: Vec<String>) -> Result<()> {
...
+                }
+            }
+            Err(e)
+                if e.code() == tonic::Code::NotFound
+                    || e.code() == tonic::Code::PermissionDenied =>
+            {
+                eprintln!("sbatch: error: {}", e.message());
+                return 1;
+            }
+            Err(e) if !warned_unknown_state => {
```
yansun1996: **Must-fix:** This still retries permanent RPC failures forever. In particular, user JWTs default to a 24-hour lifetime; after one expires, `GetJob` returns `Unauthenticated`, and a long-running `sbatch --wait` will warn once and then silently loop forever even after the job finishes. Could this retry only explicitly transient codes such as `Unavailable`/`DeadlineExceeded`, and return an error for permanent statuses such as `Unauthenticated`, `InvalidArgument`, and `Unimplemented`?

crates/spur-cli/src/sbatch.rs
```diff
@@ -1044,9 +1051,91 @@ pub async fn main_with_args(cli_args: Vec<String>) -> Result<()> {
...
+fn terminal_exit_code(state: JobState, exit_code: i32) -> i32 {
+    match state {
+        JobState::JobCompleted => exit_code,
+        JobState::JobFailed => exit_code.max(1),
+        _ => 1,
+    }
+}
+
+/// Poll `GetJob` until the job reaches a terminal state, cancelling on Ctrl-C.
+async fn wait_for_job(
```
yansun1996, line 1072: **Should-fix:** The added tests still never execute `wait_for_job`, so regressions in polling, terminal dispatch, transient recovery, fatal errors, or requeue confirmation leave the suite green. The existing in-process `mock_controller` already serves `GetJob`; extending it with a queued `JobInfo`/error response and using it from tests would exercise the real client loop without an external dependency or dead code.

yansun1996 (commented):
Thanks for addressing the earlier feedback. The typed state handling, exit-code mapping, shared Ctrl-C helper, immediate first poll, submit-user ordering, output flush, requeue confirmation, and docs all look much better. I found two remaining issues inline. Also, please update the PR description: it still says CANCELLED returns 130, lists PREEMPTED as terminal, claims mock_controller.rs extensions that were removed, and reports stale test counts.

crates/spur-cli/src/sbatch.rs
bhatturu: Fixed. Flipped from blocklist to transient allowlist. Only `Unavailable`, `DeadlineExceeded`, `Aborted`, `Internal`, `Unknown` are retried (warn-once). All other codes (`Unauthenticated`, `NotFound`, `PermissionDenied`, `InvalidArgument`, etc.) are fatal — exit 1 with error message.

crates/spur-cli/src/sbatch.rs
bhatturu: Fixed. Extended `mock_controller.rs` with `get_job_sequence` (queued `JobInfo`
responses with `state`/`exit_code`), `submit_job`, and `cancel_job`
implementations. Added 7 integration tests that drive `wait_for_job` through the
mock:

- `wait_polls_through_pending_to_completed` — Pending → Running → Completed(0)
- `wait_propagates_nonzero_exit_code` — Pending → Failed(42) → returns 42
- `wait_clamps_negative_exit_code` — Failed(-1) → returns 1 (not 255)
- `wait_skips_suspended_and_preempted` — Running → Suspended → Preempted → Pending → Running → Completed(7)
- `wait_exits_on_permanent_rpc_error` — NotFound → immediate exit 1, single call
- `wait_cancelled_exits_one` — Pending → Cancelled → returns 1
- `wait_timeout_exits_one` — Running → Timeout → returns 1

531 total tests pass (530 spur-cli + 1 broken_pipe).

yansun1996 (approved):
lgtm

PR #893, opened 2026-09-18
feat(auth): add native spur credentials, mint, and RBAC

sgopinath1:
## Summary

Adds a Spur-owned native auth plugin (`plugin = "spur"`) so callers are identified from local Unix identity instead of a username on the wire, and enforces User / Operator / Administrator on gRPC, REST, and agent job surfaces.

JWT remains the default plugin. Node JWT is still valid for node identity; user login is native (SO_PEERCRED + per-RPC HMAC). Job/step launch uses signed Ed25519 credentials.

## Approach

- `spurauthd` mints UserRpc credentials over a Unix socket using `SO_PEERCRED` and NSS (at mint only).
- Controller and agent verify HMAC envelopes with nonce replay protection; audience is `spur/{cluster}/controller|agent/{hostname}`.
- Signing JWKS lives on the controller; agents get verify-only material. Agent→controller RPCs (register, heartbeat, drain, completion) mint UserRpc after Ping when native is required.
- RBAC: `resolve_role` + `operates_jobs` / `check_job_caller` for cancel, hold, drain, accounting, and PTY attach (`CreateJobStep`, agent attach/stream/exec). Job list pins non-operators to self; `job_info_visibility` redacts or hides others’ jobs.
- CLI: native Ping then mint when `SPUR_AUTH_PLUGIN=spur`. REST Bearer is Identity, not a free-form JWT user.

Default `plugin` is still `jwt`. Native `mode=required` rejects missing bearer except Ping.

## Docs

- `docs/admin-guide/rbac.rst`
- `docs/developer/native-credential-mint.rst`
- configuration / interactive / toc updates

- Coordinator role / grant catalog
- Live JWKS rotation editor
- Dual-path CLI `srun` identity
- Raft HA under native required
- K8s operator native path beyond the wiring already here

## Test plan

- [x] `cargo test --locked`
- [x] Crusoe bare-metal: `plugin=spur` `mode=required`, heartbeats on reachable agents, RBAC matrix (62 pass; 1.3 CLI Ping-audience with config loaded is harness/env, not a product bug)


crates/spur-cli/src/auth_keys.rs
```diff
@@ -0,0 +1,94 @@
...
+    }
+}
+
+fn write_mode_0600(path: &Path, bytes: &[u8]) -> std::io::Result<()> {
+    let mut f = OpenOptions::new()
+        .write(true)
+        .create(true)
+        .truncate(true)
+        .mode(0o600)
+        .open(path)?;
```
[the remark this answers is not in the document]
sgopinath1: Agreed. OpenOptions.mode only applies on create, so we chmod the opened file to 0600 before writing. Overwriting a world-readable JWKS no longer leaves signing material at 0644.

crates/spur-core/src/native_peer.rs
```diff
@@ -0,0 +1,245 @@
...
+            nonce,
+            key_id: self.keys.default_kid().to_string(),
+        };
+        let payload = env.to_signing_bytes()?;
+        let (kid, signature) = self
+            .keys
+            .sign(&payload, now)
+            .map_err(|_| CredentialError::BadSignature)?;
+        env.key_id = kid.clone();
+        let payload = env.to_signing_bytes()?;
```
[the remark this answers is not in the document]
sgopinath1: Agreed. We set the final key_id first, serialize, sign that payload, and reject if the signer returns a different kid. The token payload and signature now cover the same bytes.

crates/spurd/src/main.rs
```diff
@@ -338,6 +338,15 @@ async fn main() -> anyhow::Result<()> {
...
         }
     };
+    match config.as_ref() {
+        Some(c) => controller_auth::install(&c.auth.plugin, &c.cluster_name),
+        None => {
+            let plugin = std::env::var("SPUR_AUTH_PLUGIN").unwrap_or_default();
+            let cluster = std::env::var("SPUR_CLUSTER_NAME").unwrap_or_default();
+            controller_auth::install(&plugin, &cluster);
+        }
+    }
```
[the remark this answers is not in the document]
sgopinath1: We are not embedding the mint. spurauthd is required on every host that mints (login CLI and each compute node). Docs now say that; native plugin start warns if the socket is missing, and RPCs still fail closed.

crates/spur-core/src/native_service.rs
```diff
@@ -0,0 +1,244 @@
...
+    keys.verify(&signed.key_id, &signed.payload, &signed.signature, now)
+        .map_err(|_| CredentialError::BadSignature)?;
+    let cred = ControllerRpcCredential::from_signing_bytes(&signed.payload)?;
+    if cred.key_id != signed.key_id {
+        return Err(CredentialError::Malformed("kid"));
+    }
+    if cred.cluster_id != cluster_id {
+        return Err(CredentialError::ClusterMismatch);
+    }
+    if cred.audience != CONTROLLER_TO_AGENT_AUDIENCE {
```
biluriuday: When the controller talks to an agent, it presents a signed token proving "I am the controller." The problem: that token is identical for every agent — it carries a fixed label "controller-to-agent" instead of naming which agent it's for, and the anti‑reuse counter is 0
Each agent only remembers tokens it has seen (its own replay cache), so a token already used on agent A is still "fresh" to agent B. On top of that, this token is sent in plaintext with no TLS (confirmed: no rustls/ClientTlsConfig in agent_client.rs or spur-client; it's just an authorization header at agent_client.rs:86,137). And an agent trusts any valid controller token for its most powerful RPCs (it only checks is_controller()).

Result: anyone who sees one of these tokens — a single malicious/compromised node (which is handed the token legitimately), or anyone sniffing the wire — can copy it and present it to every other agent for the ~5 minutes it stays valid, and those agents will obey it as if it were the controller (cancel/kill jobs, run controller‑only commands). No signing key is stolen; the token is simply reused. That directly breaks the design's promise (§13) that a compromised node can't act as the controller.

Fix: make each token usable on only one agent — stamp it with that agent's name and boot epoch (spur/{cluster}/agent/{host}), which the controller already learns from Ping, exactly like the CLI does — or put mTLS on the controller↔agent channel. As long as the token is generic and each agent tracks reuse on its own, copying it to another agent will keep working.

crates/spurctld/src/main.rs
```diff
@@ -285,33 +292,40 @@ async fn main() -> anyhow::Result<()> {
...
+    let bearer = spur_core::auth::BearerAuth::from_config(
+        &config,
+        jwt_key.as_bytes(),
+        spur_core::auth::VerifierKind::Controller,
+    )
+    .map_err(|e| anyhow::anyhow!("native auth key set: {e}"))?;
+
     if config.rest_api.enabled {
         let rest_addr: std::net::SocketAddr = config.controller.rest_addr.parse()?;
         if !rest_addr.ip().is_loopback() {
```
biluriuday, line 310: if REST is enabled on anything other than localhost, we currently just warn and start anyways. Should we refuse to start unless it's behind a trusted gateway?

crates/spurctld/src/cluster.rs
```diff
@@ -1322,21 +1322,32 @@ impl ClusterManager {
...
     ///
     /// These are control-plane paths where `user` has already been bound to the verified identity
-    /// (or is a daemon-internal call that leaves it empty). An empty `user` is the daemon caller and
-    /// a literal `"root"` is the admin override, so both are treated as internal here; the raw
-    /// [`spur_core::auth::check_job_owner`] no longer infers that itself.
-    fn check_job_owner(
+    /// (or is a daemon-internal call that leaves it empty). An empty `user` is the daemon caller, a
+    /// literal `"root"` is the admin override, and `operates_jobs` is a verified Operator or
+    /// Administrator. The raw [`spur_core::auth::check_job_owner`] no longer infers that itself.
+    fn check_job_owner_for(
```
biluriuday, line 1326: There's a shared helper, check_job_owner, that was deliberately hardened to stop treating a blank username or the literal name "root" as a trusted internal caller (it even has tests asserting both must be rejected). But the wrapper the real cancel/suspend/resume paths actually call quietly puts those two shortcuts back:
Here, user.is_empty() or user == "root" makes the caller count as "internal," which skips the ownership check entirely.
This is safe when auth is set to required, because there the username is always replaced with the verified identity before this runs, so it's never blank. But in permissive or disabled mode, an unauthenticated caller has no identity, so user is just the empty string — which flips is_internal to true and lets them cancel/suspend/resume any job. A plain DELETE /job/{id} over REST with no credentials would go through. (The REST submit path, by contrast, correctly refuses anonymous callers — so the two paths disagree.)

Result: the hardening on check_job_owner is effectively undone on every path that actually stops or cancels work, and the safety now depends entirely on each caller remembering to overwrite user first.

Fix: decide "is this an internal/privileged caller" from a verified signal — an identity that is actually present and is the controller, an admin, or an operator — never from the raw user string. And make the REST cancel/suspend/resume endpoints reject anonymous callers, the same way submit already does.

crates/spur-core/src/rbac.rs
```diff
@@ -0,0 +1,233 @@
...
+    if identity.is_admin {
+        return true;
+    }
+    if name_listed(&auth.cluster_admins, &identity.user) {
+        return true;
+    }
+    if group_listed(&auth.admin_groups, nss_groups) {
+        return true;
+    }
+    if auth.allow_uid_zero_administrator && identity.uid == 0 && identity.trusted_unix {
```
biluriuday, line 102: Spur has a setting, allow_uid_zero_administrator, that lets an admin say "root on a host is not automatically a cluster admin." The role logic honors it carefully — it only treats uid 0 as Administrator when that flag is on and the identity was verified by the kernel
But the job‑ownership shortcut from I1 (cluster.rs:1335) has a second, cruder check: if the caller's name is literally "root", it's treated as internal and allowed to act on anyone's job — with no regard for allow_uid_zero_administrator. So an admin who deliberately set that flag to false (expecting root to be an ordinary user) still finds a verified root able to cancel and manage everyone's jobs through this path.

Result: two different pieces of code answer "is root special?" and they disagree. It's low‑risk (you need to actually be root, which is already privileged), but it's a confusing policy split where one config knob is silently ignored.

Fix: route the root / uid‑0 decision through resolve_role (which respects the flag) instead of matching the string "root", so a single setting governs it everywhere.

biluriuday (commented):
This is strong, coherent work. The cryptographic core is done right: deterministic length‑prefixed signing bytes with an authenticated kind+context, peek_kind read from the signed payload, constant‑time HMAC and vetted Ed25519 (aws-lc-rs, no == on any MAC/sig), strict atomic JWKS loading with fail‑closed startup and key‑domain separation, and a correct user‑RPC audience + random boot‑epoch + atomic replay path. The controller RBAC is centralized through one resolve_role, role is never taken from the wire, authoritative_user/authoritative_unix rebind identity from the verified credential, job‑list scoping pins non‑operators to self, and the k0s root‑string spoof is closed once a caller is verified.

crates/spur-core/src/native_service.rs
sgopinath1: Agreed. Tokens are now minted per agent: Ping, then spur/<cluster>/agent/<hostname> plus that agent’s boot epoch. A token captured on one node is not valid on another.

crates/spurctld/src/cluster.rs
sgopinath1: Native plugin=spur + mode=required now refuses to start unless REST is loopback or rest_api.allow_non_loopback = true (gateway only). Other plugins still warn.

crates/spurctld/src/main.rs
sgopinath1: Those string shortcuts are gone. Privilege is only operates_jobs (verified Operator/Administrator). REST cancel/suspend/resume reject anonymous callers the same way submit does.

crates/spur-core/src/rbac.rs
sgopinath1: Same path: job control no longer matches the name "root". UID 0 is Administrator only when resolve_role says so, which honors the flag.

sgopinath1:
> This is strong, coherent work. The cryptographic core is done right: deterministic length‑prefixed signing bytes with an authenticated kind+context, peek_kind read from the signed payload, constant‑time HMAC and vetted Ed25519 (aws-lc-rs, no == on any MAC/sig), strict atomic JWKS loading with fail‑closed startup and key‑domain separation, and a correct user‑RPC audience + random boot‑epoch + atomic replay path. The controller RBAC is centralized through one resolve_role, role is never taken from the wire, authoritative_user/authoritative_unix rebind identity from the verified credential, job‑list scoping pins non‑operators to self, and the k0s root‑string spoof is closed once a caller is verified.

Thanks for reviewing @biluriuday. I have addressed your comments Please take a look.

biluriuday (approved):
LGTM

PR #900, opened 2026-09-20
fix(spurd): reclaim a supervised job's ledger on cancel when its stepd is confirmed dead

yansun1996:
## Summary

- When a cancel for a supervised job arrives after its stepd is already
  confirmed dead, actively reap the job's ledger (running-set entry and
  GPU/CPU allocation) instead of waiting on a teardown that will never come.
- Reuses the existing crash-watchdog liveness check (pid + start-ticks) to
  confirm the stepd is actually gone, not just unreachable.
- Reuses the existing release path the watchdog already uses, so a stale
  cancel naming a superseded attempt can never evict a newer, legitimate run.
- Applies to both the graceful-cancel and explicit-signal RPC paths.

## Root cause

A cancel for a supervised job only signals its stepd and waits for the
stepd's own teardown to release the local GPU/CPU allocation ledger. If the
stepd is already dead — crashed, killed, or its completion report was lost
— that teardown never happens, so the ledger entry is stuck until a manual
node drain. This holds even though the controller has already given up on
the job and keeps re-sending the same cancel on its own reclaim heartbeat:
every one of those retries is a no-op, because nothing on the agent side
ever checks whether the stepd is still actually there.

## Validation

- `cargo fmt --all --check`
- `cargo clippy -p spurd --all-targets --locked`: clean
- `cargo test -p spurd --lib`: 570 passed, 0 failed, 2 ignored
- 6 new regression tests, each proven red on the code before its fix and
  green after:
  - graceful cancel reclaims a confirmed-dead supervised job's ledger
  - graceful cancel does *not* reclaim a live supervised job's ledger (no
    regression on the normal teardown-owns-release path)
  - graceful cancel naming a superseded attempt spares the newer run's
    ledger
  - `send_explicit_signal` reclaims a confirmed-dead supervised job's
    ledger (same gap, second entry point)
  - a job with one live sibling stepd is not reaped just because another
    sibling died
  - an attempt-less cancel resolved only against a dead stepd must not
    release a reused job_id's newer reservation (added during review; see
    below)

## New e2e test coverage

Two new pytest e2e tests in
`tests/native_host/e2e/test_stepd_restart_survival.py`
(`TestSupervisedGpuReclaim`), covering both ends of the reclaim path on
real GPU hardware:

- `test_cancelling_a_job_whose_supervisor_died_frees_its_gpu_promptly` —
  the agent side: a job's supervisor is killed, the job is cancelled, and
  the GPU must free promptly instead of waiting on the crash watchdog.
- `test_reconnected_agent_reporting_a_cancelled_job_gets_its_gpu_reclaimed`
  — the controller side, previously untested: the node's agent goes dark,
  the controller cancels the job anyway (its own bookkeeping doesn't need
  the node reachable to do that), and the agent comes back up still
  holding the job's supervisor and GPUs, unaware it was ever cancelled.
  Verifies the controller's own reclaim-heartbeat notices the mismatch
  unprompted and frees both the process *and* the GPU ledger — checked by
  dispatching a same-sized job to the same node afterward, not just by
  checking the process is gone.

Both were hand-verified end to end on 2-node GPU hardware before being
committed: killed the agent, cancelled from the controller, restarted the
agent, and confirmed the stale supervisor died and a full-size job
immediately redispatched to the same node — the controller reclaimed the
mismatch on the agent's very next reconnect, without waiting on any
periodic tick.

### Live end-to-end validation (2-node GPU lab)

Reproduced the field scenario this fix targets: an agent loses its
connection to the controller, its supervised stepd is killed while
partitioned (simulating a crash the agent can't observe immediately), the
controller cancels the job, and — once connectivity returns — the
controller's own reconnect-driven reclaim heartbeat re-delivers that same
cancel to the agent, exactly as it does in production.

To isolate this fix's effect from the independent ~15s crash watchdog
(which would eventually release the same allocation on its own timer
regardless of this fix), the watchdog interval was temporarily widened and
the heartbeat interval shortened for the duration of the test only, so the
reconnect-driven cancel is guaranteed to be the first thing that can act.
Neither change was committed; both were reverted immediately after.

**Before this fix (RED):**
- Stepd killed, network partitioned, cancel issued, connectivity restored.
- The reconnect-driven `CancelJob` arrived every ~3s (the heartbeat
  cadence) and was a no-op every time (`Connection refused` — stepd
  already gone, nothing released).
- A second job resubmitted to the same node was rejected and exhausted
  retries into `JobHoldMaxRequeue` — the exact stuck state this fix
  targets.

**After this fix (GREEN):**
- Identical sequence, same timing.
- The reconnect-driven `CancelJob` arrived once (~9s after the kill, ~2s
  after connectivity was restored) and released the ledger immediately —
  no retries needed.
- A job resubmitted to the same node dispatched and was running within
  ~1s.

| Stage | Before (RED) | After (GREEN) |
| --- | --- | --- |
| Reconnect-driven cancel | loops every ~3s, always a no-op | resolves on first delivery |
| Resubmitted job on the same node | rejected, exhausts retries → `JobHoldMaxRequeue` | running in ~1s |

## Known limitations (not addressed here)

- The crash watchdog stays the fallback for a stepd that dies with no
  cancel ever issued against it — that's expected, it's the general safety
  net for exactly that case. This fix is scoped to the cancel path: once a
  cancel does arrive, it no longer has to wait on the watchdog's own ~15s
  tick to notice and clean up.


crates/spurd/src/agent_server.rs
```diff
@@ -7698,6 +7711,25 @@ impl AgentService {
...
     }
 
+    /// A cancel normally just signals a supervised job's stepd and waits for
+    /// its teardown. If every stepd for this attempt is already confirmed
+    /// dead, no teardown is coming — release the ledger ourselves.
+    async fn reap_dead_supervised_stepds(&self, runtimes: &[crate::stepd::StepdDescriptor]) {
+        if runtimes.is_empty() || !runtimes.iter().all(stepd_confirmed_dead) {
+            return;
+        }
+        for descriptor in runtimes {
```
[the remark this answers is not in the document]
yansun1996: Real race, but one level up — `reap_dead_supervised_stepds` is already generation-safe; the gap is `cancel_job`'s trailing cleanup blindly releasing on `run_attempt == 0`. Fixed by snapshotting the doomed attempt before the signal call and using `release_job_if` instead. Added a regression test.

crates/spurd/src/agent_server.rs
```diff
@@ -5072,6 +5084,31 @@ impl SlurmAgent for AgentService {
...
+                None => match stepds_for_job(&*self.stepds.lock().await, job_id)
+                    .first()
+                    .map(|descriptor| descriptor.run_attempt)
+                {
+                    Some(attempt) => Some(attempt),
+                    None => self.allocation.lock().await.owner_attempt(job_id),
+                },
+            },
+            named => Some(named),
+        };
```
shiv-tyagi: `cancel_job` now binds `doomed_attempt` twice. This one snapshots the attempt before the signal/cancel dispatch (used for the `release_job_if` at L5131). A second `doomed_attempt` at L5147 is recomputed after the cancel from `tracked_attempt`/`supervised_attempt` and drives the cgroup cleanup at L5152. Same name, different inputs, different timing, same function scope.

Rename this one to make the distinction explicit, e.g. `pre_cancel_attempt`.

shiv-tyagi (commented):
One comment. PTAL.

crates/spurd/src/agent_server.rs
yansun1996: Confirmed, renamed the pre-cancel snapshot to `pre_cancel_attempt` (5aefcc31).

sajmera-pensando:
Merging the PR as all the comments have been addressed.

PR #902, opened 2026-09-21
docs(config): rewrite [auth] by plugin instead of a union table

sgopinath1:
## Summary

The `[auth]` reference mixed every plugin into one field table, so required vs optional was unreadable. This rewrite documents `none`, `jwt`, and `spur` as independent subsections (what you get, what you do not, fields, examples, setup), then roles/bindings and privileged operations in the same section.

`accounting.rst` now matches `resolve_role`: `adminlevel=Operator` is Operator; `Admin` is Administrator. `rbac.rst` points at the `[auth]` roles section as the long form.


biluriuday (dismissed):
LGTM

biluriuday (approved):

PR #909, opened 2026-09-22
fix(spur-core)!: use Slurm's exact reserved step-id sentinels

biluriuday:
Align STEP_BATCH/EXTERN/INTERACTIVE with slurm.h (SLURM_BATCH_SCRIPT 0xfffffffb, SLURM_EXTERN_CONT 0xfffffffc, SLURM_INTERACTIVE_STEP 0xfffffffa). These ids are persisted (Raft snapshot, stepd runtime state, PMIx namespaces) and cross the gRPC wire, so upgrading requires a drained cluster; see docs/deployment/upgrading.rst.


shiv-tyagi (approved):
LGTM

PR #910, opened 2026-09-22
test(e2e): add MPI-launched RCCL collective tests

nikhilsk:
## What this adds

Five e2e tests for **MPI-launched RCCL collectives**, which nothing currently covers.

Spur already has the two halves separately. `test_mpi.py` proves `srun --mpi=pmix` hands out a correct rank map, and `fixtures/distributed_test.py` drives RCCL through PyTorch on a single node — and states outright that its multi-node mode does no cross-node RCCL. Neither exercises an MPI-bootstrapped RCCL communicator, which is what every MPI-based collective workload actually runs.

| Test | Shape | What only this test can catch |
|---|---|---|
| `two_ranks_one_node` | 2 ranks, 1 host | local rank map and communicator setup |
| `each_rank_gets_a_distinct_device` | 2 ranks, 1 host | device assignment regressing to always device 0 |
| `two_nodes_one_rank_each` | 1 rank per host | RCCL traffic on the wire; transport or interface problems |
| `two_nodes_two_ranks_each` | 2 ranks per host | intra-node and inter-node paths in one communicator |
| `survives_back_to_back_launches` | 2 runs in a row | a leaked communicator or PMIx namespace |

The payload is a C fixture that bootstraps ranks through MPI, builds an RCCL communicator and all-reduces. It asserts the **arithmetic**, not just the status: each rank contributes `rank + 1`, so a correct sum is `n(n+1)/2`. That distinguishes a collective that moved the wrong data from one that hung. Device distinctness is a separate assertion, because two ranks sharing one GPU would still produce the right sum.

## Node requirements

Every test skips with a message naming exactly what is missing, so the suite is inert on a CPU-only runner rather than failing it. To make them actually execute, each GPU e2e node needs:

| Requirement | Probed by | Typical package |
|---|---|---|
| A GPU (`/dev/kfd`) | `gpu_preflight` | `amdgpu-dkms` / working driver |
| `hipcc` | `rccl_preflight` | `rocm-hip-sdk` (or `hip-dev`) |
| `librccl` on the loader path | `rccl_preflight` | `rccl` |
| `/opt/rocm/include/rccl/rccl.h` | `rccl_preflight` | `rccl-dev` |
| `mpicc` | `mpi_preflight` | `libopenmpi-dev` |
| `libpmix` | `mpi_preflight` | `libpmix-dev` |
| `spur_mpi_pmix.so` | `mpi_plugin_dir` | built and shipped by #908 |

Note `mpicc` is needed on the controller node too, since the fixture is compiled there as well as on each compute node — the same rule the MPI docs give for MPI binaries.

If the MPI include path is not discoverable via `mpicc -showme:incdirs` (that flag is Open MPI specific and absent on MPICH), set `SPUR_TEST_MPI_INCLUDE` to override it.

## Also in here

Registers the `mpi` pytest marker. `test_mpi.py` has used it since it was written but never declared it, so it only worked because `--strict-markers` is off. Declaring it means `-m mpi` and `-m "not mpi"` select reliably, which matters for routing these to GPU runners.

## How this was tested

Honestly, and with a gap worth stating plainly:

- **Verified:** the suite collects (5 tests), `flake8 --config tests/.flake8 tests/` is clean, and all five skip cleanly against a live 3-node Spur cluster built from this branch, each naming the missing prerequisite:

  ```
  SKIPPED e2e/test_rccl.py:79:  MPI preflight failed: libpmix on spur-w2; mpicc on controller node spur-w2
  SKIPPED e2e/test_rccl.py:127: MPI preflight failed: libpmix on spur-w2; libpmix on spur-w3; mpicc on controller node spur-w2
  ```

- **Not verified:** the RCCL path itself. Our lab nodes have no GPU, no ROCm and no MPI, so the fixture has never been compiled or run, and the collective assertions have never executed. They need a provisioned GPU e2e node. Treat the fixture as unproven until it runs there — if `hipcc` rejects it or the include flags need adjusting for the local MPI, that will only surface on real hardware.

Raised as a draft for that reason: ready for the provisioned runners to be pointed at it, not ready to be called green.


nikhilsk:
## First run on real GPU hardware

Ran this suite on a 4-node AMD MI355X bed (ROCm 7.0.1, Open MPI 4.1.6, OpenPMIx 5.0.1). Headline: **the RCCL collective works**, and the fixture and its assertions are correct. Verified directly on a node before trusting the harness:

```
rank=0 size=2 host=crsuse2-m2m-141 device=0
rank=1 size=2 host=crsuse2-m2m-141 device=1
allreduce rank=0 expected=3.0 actual=3.0 status=OK
allreduce rank=1 expected=3.0 actual=3.0 status=OK
comm_size=2
```

Two ranks on distinct GPUs, and the arithmetic is right rather than merely returning success. The existing `test_mpi.py` single-node suite also passes on the same bed, so the PMIx layer underneath is sound.

## What the bed exposed, and what changed here

Every item below was a hard failure on first contact with real hardware, not a tidy-up. Six of them, in the order they surfaced:

**1. CLI calls inherited the host's `/etc/spur/spur.conf`.** The daemons get the throwaway cluster's config via `-f`, but CLI calls got nothing, so they fell back to the system file. On a node previously configured as a real Spur cluster that file carried `[auth] plugin = "spur"`, `mode = "required"`, and every `sinfo`/`sbatch` refused to talk to the test controller: `controller did not advertise a native auth audience`. This is the one I would flag beyond this PR — it breaks the whole suite on any node that has ever been a real cluster, and the error gives no hint that a stale config is involved. Fixed by passing `SPUR_CONF` at the single point every CLI call goes through.

**2. `hipcc` inherits none of `mpicc`'s paths**, and compiles a `.c` payload as plain C, so no HIP platform macro and no HIP includes. The build now asks `mpicc` for its include *and* lib dirs, probes for the HIP include root, and states `-D__HIP_PLATFORM_AMD__`.

**3. The RCCL header spelling is version-dependent.** ROCm moved it into an `rccl/` subdirectory, so the fixture now includes `<rccl.h>` against a `-I` at whichever directory actually holds it. Hardcoding either spelling breaks on the other layout.

**4. Tests must request GPUs.** Without `--gres` the ranks died with `HIP error no ROCm-capable device is detected` on a node with eight of them. Now every `srun` requests them, per node.

**5. PMIx needs two things the harness never set.** Hash GDS on the agent, since Ubuntu's `libpmix2t64` fails against Spur's PMIx server with the default shmem GDS; and `SPUR_TEST_PMIX_LIB_DIR` for a site with two PMIx installs, where the server picks the distro copy while the ranks use a vendor MPI's bundled client and `MPI_Init` dies in `pmix3x_client.c`.

**6. Locally built `spur_mpi_pmix.so` can be a stub.** Building it on a machine without libpmix development packages yields a plugin that fails at runtime with `spur_mpi_pmix.so was built without libpmix`. Not a code change — worth knowing, since the failure only appears on the node. Use a nightly tarball's plugin, as the deployment guide already advises.

## Still open: `srun` hangs on a step that requests GPUs

The suite is **not** green end to end. With `--gres`, the ranks run and exit but the job never reaches a terminal state, so `srun` never returns.

What is ruled out: device isolation is not involved. On a rootless agent the log says `cgroup unavailable; job runs without isolation`, all eight GPUs stay visible, and the allocation itself is correct — `registered srun allocation job_id=1 cpus=2 memory_mb=0 gpus=[0, 1]`. The same payload completes normally under `mpirun` outside Spur, and `test_mpi.py` (which requests no GPUs) passes on the same cluster.

So the difference is a GPU-requesting step specifically. Tracking separately; I have not changed any test to accommodate it, and deliberately have not marked anything xfail.

## Notes for whoever runs this next

On a bed with two PMIx installs, Spur requires PMIx >= 4.1.0 (`[mpi].pmix_min_version`), so a vendor MPI bundling PMIx 3.2.5 cannot serve as the server side. The working shape was: system PMIx 5.0.1 for the server, hash GDS set on both the agent and the ranks, and `$HOME/spur/mpi/env.sh` present on every node with `OPAL_PREFIX` pointed at the MPI the binary is built against and Open MPI's TCP pinned to the fabric NIC. Overlay interfaces from Flannel and CNI are present on these hosts and will be picked otherwise, which the deployment guide warns hangs `MPI_Init`.

Environment used:

```
SPUR_TEST_MPICC=/opt/openmpi/bin/mpicc
SPUR_TEST_BINARIES_DIR=<unpacked nightly>/bin      # plugin found at ../lib/spur
SPUR_TEST_REMOTE_BIN_DIR=/tmp/spur-e2e-nightly-bin # cache uploads across runs
```

tests/native_host/e2e/cluster.py
```diff
@@ -104,6 +104,21 @@ def __init__(self, host: str, user: str, password: str | None = None, key_path:
...
+        socks_proxy = os.environ.get("SPUR_TEST_SSH_SOCKS", "").strip()
+        if socks_proxy:
+            import socks  # PySocks; only needed on this path
+
+            proxy_host, _, proxy_port = socks_proxy.rpartition(":")
+            sock = socks.socksocket()
+            sock.set_proxy(socks.SOCKS5, proxy_host or "127.0.0.1", int(proxy_port))
+            sock.settimeout(60)
+            sock.connect((host, 22))
+            connect_kwargs["sock"] = sock
```
shiv-tyagi: Why do we need this?

shiv-tyagi:
> The suite is not green end to end

We need to make it green. 

Also our CI testbed has only VMs without any AINICs with only one GPU on each machine. Please check which tests can run on the CI testbed. 

cc @sgopinath1 @sajmera-pensando This is a testbed limitation we can't overcome.

nikhilsk:
## Inter-node RCCL resolved

The transport guess in my last comment was half right. Pinning `NCCL_SOCKET_IFNAME` alone changes nothing — the run still dies at communicator init. The actual blocker is that RCCL loads AMD's `RCCL-ANP` plugin and takes all eight `ionic` AINICs over RoCE with GPU Direct RDMA, and that path fails on this bed:

```
NET/IB : Using [0]ionic_0:1/RoCE ... [7]ionic_7:1/RoCE [RO]; OOB ens3:10.245.148.188<0>
Using network RCCL-ANP
NET/RCCL-ANP : GPU Direct RDMA Enabled for HCA 0 'ionic_0'
...
RCCL error unhandled system error at rccl_all_reduce.c:92
```

`NCCL_IB_DISABLE=1` makes it fall back to sockets. *Then* the interface pinning becomes necessary, because the socket fallback otherwise picks the Flannel or `cni0` overlay and hangs. Each alone is insufficient: `NCCL_SOCKET_IFNAME` by itself fails at init, and `NCCL_IB_DISABLE` by itself works for one rank per host and hangs at two.

Both are documented as site configuration in this PR rather than set by the tests, so a bed with working RDMA still exercises the fast path. `NCCL_IB_DISABLE=1` is written up as a last resort with its cost stated.

## Where the suite stands on real hardware

Four of the five shapes now pass, driven entirely by `$HOME/spur/mpi/env.sh` with nothing in the submitter's environment:

```
[1] rank=1 size=2 host=crsuse2-m2m-143 device=0 bus=0002:00:02.0
[1] allreduce rank=1 expected=3.0 actual=3.0 status=OK
[0] rank=0 size=2 host=crsuse2-m2m-141 device=0 bus=0002:00:02.0
[0] allreduce rank=0 expected=3.0 actual=3.0 status=OK
[0] comm_size=2
```

and four ranks over two hosts likewise, each rank on its own GPU. I have not re-run all five as a suite off a single build, so I am not claiming the suite is green yet.

Two product fixes are prerequisites, both raised with live before/after evidence:

- #915 — a multi-task GPU step failed with `no ROCm-capable device is detected` because `ROCR_VISIBLE_DEVICES` and `CUDA_VISIBLE_DEVICES` were both set to absolute ids while ROCr renumbers what it leaves visible. This blocks both intra-node cases.
- #917 — `srun --mpi=pmix -N 2 -n 2` never sourced `env.sh`, because the agent only built a launch wrapper when a node had more than one task and that wrapper is the only place the file is read. This blocks the inter-node cases on any bed where MPI is not on the default loader path.

The `device=0` in the output above is also why the distinctness check in this PR now compares PCI bus ids: Spur gives each task its own narrowed and renumbered device list, so every rank correctly reports ordinal 0 while sitting on different hardware. Ordinals could never have told them apart.

tests/pytest.ini
```diff
@@ -9,3 +9,4 @@ addopts = -p no:xdist
 markers =
     rootful: run the cluster fixture with spurd as root (via sudo)
     gpu: requires GPU hardware on the nodes (auto-applied to tests using GPU fixtures)
+    mpi: requires an MPI toolchain on the nodes (mpicc plus a PMIx-capable launcher)
```
shiv-tyagi: Let us remove this marker. I will enable mpi toolchain by default on the GPU hosts.

tests/native_host/e2e/test_rccl.py
```diff
@@ -0,0 +1,197 @@
...
+            f"rank {rank} expected {expected}, arithmetic says {want}:\n{out}"
+        )
+        assert float(actual) == want, f"rank {rank} got {actual}, want {want}:\n{out}"
+
+    assert f"comm_size={expected_size}" in out, (
+        f"rank 0 did not report comm_size={expected_size}:\n{out}"
+    )
+
+
+@pytest.mark.mpi
```
shiv-tyagi: Please remove this as well.

tests/native_host/e2e/fixtures/rccl_all_reduce.c
```diff
@@ -0,0 +1,148 @@
+// Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// MPI + RCCL all-reduce. Exercises the one combination Spur's e2e never covered:
+// ranks bootstrapped by Spur's PMIx plugin, then a GPU collective over RCCL.
+//
+// The existing distributed_test.py drives RCCL through PyTorch on a single node,
+// and hello_mpi.c drives PMIx with no GPU. Neither proves that a rank map handed
+// out by `srun --mpi=pmix` can bring up an RCCL communicator, which is what every
+// real MPI-launched RCCL workload depends on.
```
shiv-tyagi: This comment is unnecessary

tests/native_host/e2e/conftest.py
```diff
@@ -136,11 +136,31 @@ def _ensure_bins(ssh_nodes, remote_bin_dir):
...
+# the distro one and the ranks fail at MPI_Init with
+# "OPAL ERROR: Error in file pmix3x_client.c". Point this at the vendor lib dir.
+def _pmix_server_env() -> dict[str, str]:
+    env = {"PMIX_MCA_gds": "hash"}
+    lib_dir = os.environ.get("SPUR_TEST_PMIX_LIB_DIR", "").strip()
+    if lib_dir:
+        env["LD_LIBRARY_PATH"] = f"{lib_dir}:$LD_LIBRARY_PATH"
+    return env
+
+
```
shiv-tyagi: This is not needed for CI runs and unconditionally applying it to existing MPI fixtures can cause issues.

tests/native_host/e2e/conftest.py
```diff
@@ -136,11 +136,31 @@ def _ensure_bins(ssh_nodes, remote_bin_dir):
...
 def _deploy_cluster(ssh_nodes, remote_bin_dir, *, agent_as_root: bool = False,
                     config_overrides: dict | None = None,
-                    agent_labels: dict[int, dict[str, str]] | None = None):
+                    agent_labels: dict[int, dict[str, str]] | None = None,
+                    agent_env: dict[str, str] | None = None):
```
shiv-tyagi: From what I can see, `agent_env` is only being used to inject pmix server env, which I am not sure if we really need it. Please see and remove it and similar elsewhere.

docs/deployment/native-host.rst
```diff
@@ -735,6 +735,12 @@ agents (``ip -br addr``). Unpinned ``btl=tcp`` may pick docker, overlay, or
 IPv6-only NICs; **several tasks per node across two nodes** then hangs in
 ``MPI_Init`` even though a single-node or 1-task-per-node run may succeed.
 
+GPU collectives need the same treatment one layer up. RCCL chooses its transport
+independently of Open MPI, so ``OMPI_MCA_btl_tcp_if_include`` does not constrain
+it, and ``NCCL_SOCKET_IFNAME`` must name the same NIC. Left unpinned, the symptom
+matches the Open MPI one: one rank per host works and several ranks per host
+across two hosts hangs.
+
```
shiv-tyagi: Is this needed? Does it pollute the docs?

docs/deployment/native-host.rst
```diff
@@ -744,8 +750,15 @@ IPv6-only NICs; **several tasks per node across two nodes** then hangs in
...
    export OMPI_MCA_oob_tcp_if_include=eth0
+   export NCCL_SOCKET_IFNAME=eth0                    # RCCL sockets; the same NIC
    EOF
 
+Where the host has RDMA devices that are not usable end to end, RCCL still
+prefers them and ``ncclCommInitRank`` fails with ``unhandled system error``
+rather than falling back on its own. ``NCCL_IB_DISABLE=1`` forces sockets. Set it
+only after confirming RDMA is genuinely unavailable, since it gives up the fast
+path for every collective on the node.
+
```
shiv-tyagi: This is only specific to your testbed I guess? should we commit it to general docs? Can you cause hallucinations in other agents?

tests/native_host/e2e/cluster.py
```diff
@@ -422,6 +437,12 @@ def _cli_env_assignments(self, controller_addr: str | None = None) -> list[str]:
         parts = [
             f"SPUR_CONTROLLER_ADDR={shlex.quote(controller_addr or self.controller_addr)}",
             f"PATH={shlex.quote(self.bin_dir)}:$PATH",
+            # Point the CLI at this cluster's own config. The daemons already get it
+            # via -f, but the CLI would otherwise fall back to /etc/spur/spur.conf and
+            # inherit whatever the host is configured for -- on a node set up for
+            # native auth it refuses to talk to a throwaway controller at all
+            # ("controller did not advertise a native auth audience").
+            f"SPUR_CONF={shlex.quote(self.etc_dir)}/spur.conf",
```
shiv-tyagi: Why is this needed in these tests? In CI env we expect not to create throwaway controller or anything. This looks junk.

tests/native_host/e2e/cluster.py
```diff
@@ -422,6 +437,12 @@ def _cli_env_assignments(self, controller_addr: str | None = None) -> list[str]:
         parts = [
             f"SPUR_CONTROLLER_ADDR={shlex.quote(controller_addr or self.controller_addr)}",
             f"PATH={shlex.quote(self.bin_dir)}:$PATH",
+            # Point the CLI at this cluster's own config. The daemons already get it
+            # via -f, but the CLI would otherwise fall back to /etc/spur/spur.conf and
+            # inherit whatever the host is configured for -- on a node set up for
+            # native auth it refuses to talk to a throwaway controller at all
+            # ("controller did not advertise a native auth audience").
```
shiv-tyagi: remove this comment please.

shiv-tyagi (changes requested):
Please keep testbed-specific plumbing out of the shared e2e harness. If a test needs extra setup on a particular bed, that belongs in the host configuration (PMIx MCA params, ld.so.conf.d, $HOME/spur/mpi/env.sh), not in fixtures or cluster.py. The tests should pass on a stock node that meets the documented requirements.

On the CI bed (1 GPU per VM), only two_nodes_one_rank_each and survives_back_to_back_launches can run. I'll add RCCL to the GPU host image so those run.

nikhilsk:
All addressed, in two passes: `eb383e8` removes the plumbing you named, `84b9641` is the wider "remove what isn't needed" sweep. The PR is 339 added lines, down from 533.

Gone: the PMIx server env the MPI fixtures injected unconditionally, the `agent_env` parameter that only carried it, the SOCKS5 dialling, the `mpi` marker, and the comments.

The sweep went further than your list, by diffing against what already passes rather than guessing. `OMPI_CC` points the MPI wrapper at hipcc so it supplies its own include, link and rpath flags — that deleted both `mpicc -showme` probes. Including `<rccl/rccl.h>` against one ROCm include root deleted two more, since that path also resolves `hip/hip_runtime.h`. `librccl` was already in the loader cache. Six `SPUR_TEST_*` overrides went with them: a node either has the toolchain or the preflight names what is missing.

Your instinct on host configuration checked out rather than being taken on faith. Hash GDS now comes only from `/etc/pmix/pmix-mca-params.conf` on the nodes, and both suites still pass on a two-node MI355X bed: `test_rccl.py` 5 passed, `test_mpi.py` 20 passed. Worth recording that the MCA params file is the right home because it reaches `spurd`, which hosts the PMIx server — `$HOME/spur/mpi/env.sh` cannot, since the rank wrapper sources it after the server is already up.

Two judgement calls. On the marker I removed only what this PR added; `main` already applies `@pytest.mark.mpi` in three places in `test_mpi.py` without declaring it, so dropping the declaration restores `main`'s status quo rather than creating a new inconsistency. Happy to clean up all four separately. CI shards select on `gpu` only either way.

On the docs I kept the `NCCL_SOCKET_IFNAME` line and dropped the `NCCL_IB_DISABLE` paragraph. The first sits under the existing Open MPI multi-NIC warning and is the same advice with the same symptom, since RCCL picks its own sockets and `OMPI_MCA_btl_tcp_if_include` does not constrain it. The second described one bed's broken RDMA, which does not generalise.

For the GPU host image work: the suite needs #915 and #917 to pass on real hardware, both open with before/after evidence. Without them the intra-node and inter-node cases respectively cannot pass anywhere.

PR #911, opened 2026-09-22
fix(spur-cli): make --help print to stdout with exit 0

sajmera-pensando:
## What this fixes

`--help` was broken across the entire CLI. Every subcommand printed its help text to **stderr** prefixed with `Error:` and exited **1**, and the top-level `spur --help` printed usage to stderr too. Anything relying on standard help behavior — `spur squeue --help | less`, `--help` in exit-0 scripts, tooling that scrapes usage — got nothing on stdout and a non-zero exit.

Two independent causes:

1. **Subcommands** parsed args with `try_parse_from(..)?`. Clap reports `--help` as an `Err(clap::Error)`; the `?` propagated it into `anyhow`, so `main` printed it as `Error: <help>` on stderr with exit 1 instead of clap's normal stdout + exit 0. Genuine parse errors were likewise routed through anyhow rather than clap's exit-2 path.
2. **Top-level** `spur --help` / `help` / `-h` printed usage via `eprintln!` (stderr), even when help was explicitly requested.

## Approach

- Add a small `clap_exit` helper (`parse_or_exit`, `matches_or_exit`, `from_matches_or_exit`) that finishes parsing via `clap::Error::exit`, and route all 24 subcommands through it — plus `salloc`/`srun` (matches-based) and `sbatch` (directive-aware resolver).
- Route explicit top-level help to stdout with exit 0; keep usage on stderr for the error paths (no command given, unknown command). Extract the text into a testable `write_usage<W: Write>`.

After this: help → stdout, exit 0, empty stderr; parse errors → stderr, exit 2 (subcommands) / exit 1 (unknown top-level command).

## Not included

`spur self-update` / `update` still ignore `--help` (they run the updater directly, bypassing clap). Left as a follow-up since it lives in the hand-rolled dispatch, not the clap path.

Subcommand `--version` is intentionally out of scope: the subcommand parsers do not register a version flag, so `spur squeue --version` still errors (unchanged from `main`); only the top-level `spur --version` path handles it. This PR is limited to `--help` routing.

## Testing

- New unit test `usage_lists_every_top_level_command` (buffers `write_usage`, asserts every command is listed).
- New integration test `tests/help_output.rs`: all 24 subcommands `--help` → stdout/exit-0/empty-stderr, plus a parse-error case.
- `tests/broken_pipe.rs` updated to break the stdout pipe (top-level help now writes there).
- Verified live across 24 canonical commands, 33 friendly aliases, and nested subcommands.
- `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` clean; `cargo test -p spur-cli --locked` = 603 unit + 3 integration tests pass.


crates/spur-cli/src/clap_exit.rs
```diff
@@ -0,0 +1,33 @@
...
+//!
+//! Clap reports `--help` and `--version` as an `Err(clap::Error)`. Propagating
+//! that error with `?` turns it into an `anyhow::Error`, which `main` prints as
+//! `Error: <help text>` on stderr and exits with code 1. Routing it through
+//! `clap::Error::exit` instead prints help/version to stdout with code 0, and
+//! genuine parse errors to stderr with code 2 — matching every standard CLI.
+
+use clap::{ArgMatches, Command, FromArgMatches, Parser};
+
+/// Parse a clap `Parser`, exiting cleanly on help/version or parse errors.
```
yansun1996: This helper/documentation says it routes both --help and --version through clap’s normal exit path, but the affected Parser definitions do not register a version flag. As a result, native forms such as spur squeue --version and spur srun --version still exit 2 with unexpected argument --version; only spur --version and actual symlink invocations use the early version path in main.rs.\n\nPlease either add #[command(version)] to the subcommand parsers and test the native/symlink forms, or scope this change and its documentation to --help.

sajmera-pensando: Good catch — you're right, `--version` isn't registered on the subcommand parsers, so `spur squeue --version` still exits 2. I've scoped this PR to `--help` per your suggestion:

- Reworded the `clap_exit` module/function docs to reference `--help` only, with an explicit note that subcommand `--version` is not registered here and is handled solely by the top-level `spur --version` path in `main`.
- Updated the PR description to drop the `--version` claim and call out that subcommand `--version` is intentionally out of scope (behavior unchanged from `main`).

I can file a follow-up for adding `#[command(version)]` across the subcommand parsers (with the version string wired to `spur_core::version::version_string()` so it matches the top-level format) if we want that Slurm-compat improvement separately.

yansun1996 (approved):
LGTM

PR #912, opened 2026-09-23
feat(spur-cli): add squeue %Q priority specifier

shiv-tyagi:
## What

Adds the `%Q` numeric-priority format specifier to `squeue`, closing part of SPUR-249.

Slurm's `squeue` accepts `%Q` for the integer priority. Spur only recognized `%p`, so `-o "%Q"` rendered `?` with no header, and `-S Q` was rejected as an invalid sort field.

## Approach

Three match arms keyed on `'Q'`:

- Header map (`squeue_header`): `'Q' => "PRIORITY"`.
- Value resolver (`resolve_job_field`): `'Q' => job.priority.to_string()`.
- Sort comparator (`compare_field`): folded into the existing `'p'` arm.

Adding the header arm makes `parse_sort_arg` accept `-S Q` automatically, since it rejects only specs whose header is `?`.

## Note on `%p` vs `%Q`

Slurm splits these: `%Q` is the integer priority, `%p` is a normalized float in 0.0-1.0. Spur represents priority only as an integer (`JobInfo.priority`, and even the scheduler's effective priority is a `u32`), so it has no normalized float to expose. `%p` and `%Q` therefore render the same integer today. The docs call this out. If Spur later adds a normalized priority, `%p` can diverge to match Slurm; that is a follow-up.

## Scope

`%Q` only. The other `squeue` parity gaps from SPUR-249 (`--me`, `-O/--Format`, `-q`, `-R`, `-s`, `--json/--yaml`, `-M`, `-i`, `-r`, `-a`) are tracked separately. `--me` is deliberately split into its own PR to keep review focused.

## Tests

- Unit: `q_specifier_renders_integer_priority` and `sort_by_q_orders_by_priority`. The existing `every_header_letter_resolves` test now also covers `%Q`.
- `cargo test --locked -p spur-cli`: 580 passed. `cargo clippy --workspace --exclude spur-ffi --all-targets --locked`: clean.
- Bare-metal cluster (4 nodes): `%Q` renders the integer (not `?`), and `-S Q` / `-S -Q` order jobs by priority in both directions (verified with jobs set to priorities 500/1500/3000).

## Docs

`docs/user-guide/monitoring-jobs.rst`: added the `%Q` row and a note on the `%p`/`%Q` relationship.


yansun1996 (approved):
LGTM

biluriuday (approved):

PR #918, opened 2026-09-23
fix(spurd): drain node on stepd launch enospc across dispatch paths

yansun1996:
## Problem

A node whose local spool root (`/var/spool/spur`) is full stays reported as `IDLE` and keeps getting jobs dispatched to it. Every dispatch fails the same way (agent-side ENOSPC preparing the job's stepd/session state), and the affected job eventually gets held/cancelled after enough retries instead of the node getting pulled out of rotation.

## Root cause

An existing classifier (`classify_spool_error` / `LaunchError::NodeFault`) already drains a node on a real OS-level ENOSPC (or similar) under its own spool root — but three write paths in `launch_stepd` (shared by both the `LaunchJob` and allocation-only `RegisterAllocation` RPCs) bypassed it and fell through to a generic, non-draining error instead:

1. Preparing the job's session directory.
2. Writing the runtime launch specification into that directory.
3. (Same underlying function, so both the batch-launch and allocation-only paths were affected identically.)

## Fix

Route all of `launch_stepd`'s node-owned-spool writes through the existing classifier instead of adding new drain logic. A real errno under the owned spool root now condemns the node exactly as designed; anything under the job's own `work_dir` (user-controlled, often a shared mount) is untouched, so a user's own disk usage can never trigger a drain.

## How tested

- Unit tests exercising the classification and self-drain behavior directly, each verified to fail against the unfixed code and pass against the fix.
- Live end-to-end validation on an isolated single-node deployment: forced a real ENOSPC on the node's spool root, confirmed both the batch (`sbatch`) and allocation-only (`salloc`/`srun`) dispatch paths now drain the node with an accurate reason, and confirmed a healthy node running a job that fills its own work directory does **not** get a false-positive drain.

`cargo test`, `cargo clippy --workspace --exclude spur-ffi --all-targets --locked`, and `cargo fmt --all --check` all clean.


crates/spurd/src/agent_server.rs
```diff
@@ -174,6 +174,19 @@ fn execv_reported_failure(pipe_r: &std::os::fd::OwnedFd) -> bool {
...
+/// rest of a job's spool files, so a failure to create it condemns the node
+/// exactly like the unsupervised launch path already does.
+fn session_dir_prepare_error(
+    dir: &std::path::Path,
+    error: std::io::Error,
+) -> executor::LaunchError {
+    executor::classify_spool_error(
+        dir,
+        anyhow::Error::from(error).context("prepare stepd directory"),
+    )
```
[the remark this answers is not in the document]
yansun1996: Confirmed and fixed in ffba3391 — both call sites now route through `drain_on_node_fault` before converting to a `Status`.

crates/spurd/src/executor.rs
```diff
@@ -98,7 +98,7 @@ fn is_node_owned_spool(dir: &Path) -> bool {
 /// Only spool writes may reach this. Writes to the job's `work_dir` must not use
 /// it: that path is user-controlled and frequently a shared mount, where one user
 /// filling their quota would otherwise drain every node in turn.
-fn classify_spool_error(dir: &Path, err: anyhow::Error) -> LaunchError {
+pub(crate) fn classify_spool_error(dir: &Path, err: anyhow::Error) -> LaunchError {
```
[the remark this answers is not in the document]
yansun1996: Same fix, ffba3391.

crates/spurd/src/agent_server.rs
```diff
@@ -16562,24 +16589,162 @@ mod tests {
...
+        timeout_ms: u64,
+    ) -> Vec<spur_proto::proto::DrainNodeRequest> {
+        let deadline = tokio::time::Instant::now() + tokio::time::Duration::from_millis(timeout_ms);
+        loop {
+            let captured = drains.lock().expect("drain requests").clone();
+            if !captured.is_empty() || tokio::time::Instant::now() >= deadline {
+                return captured;
+            }
+            tokio::time::sleep(tokio::time::Duration::from_millis(50)).await;
+        }
```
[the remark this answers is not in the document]
yansun1996: Replaced the sleep-poll helper with a Notify-backed wait for the positive case, and made the two no-drain assertions synchronous (no spawn happens for a non-NodeFault error, so there's nothing to wait for) in ffba3391.

crates/spurd/src/agent_server.rs
```diff
@@ -174,6 +174,19 @@ fn execv_reported_failure(pipe_r: &std::os::fd::OwnedFd) -> bool {
...
 }
 
+/// The session directory lives under the same node-owned spool root as the
+/// rest of a job's spool files, so a failure to create it condemns the node
+/// exactly like the unsupervised launch path already does.
+fn session_dir_prepare_error(
+    dir: &std::path::Path,
+    error: std::io::Error,
+) -> executor::LaunchError {
+    executor::classify_spool_error(
```
spraveenio: **Correctness: the drain fix silently no-ops when the spool is relocated.**

`session_dir_prepare_error` (and the launch-spec-write path at line 248) route the failing path through `executor::classify_spool_error`, which only returns `NodeFault` (→ drain) when the path passes `is_node_owned_spool` (`executor.rs:90`). That helper compares against a **hardcoded** `const SPOOL_ROOT = "/var/spool/spur"` (`executor.rs:117`).

But the path passed here is `intended_session_dir`, rooted under the stepd **state dir**, which is independently configurable (`--state-dir` / `SPUR_STEPD_STATE_DIR` / config) and documented as relocatable (`docs/admin-guide/configuration.rst:192`).

Failure scenario: an admin sets `SPUR_STEPD_STATE_DIR=/data/spur` (supported). Disk fills. `prepare_session_dir` / `launch.json` write fails with ENOSPC at `/data/spur/runtime/...`. `is_node_owned_spool("/data/spur/...")` is `false` → `LaunchError::Other` → `drain_on_node_fault` is a no-op. The broken node stays in rotation and every subsequent launch to it fails identically — the walk-the-cluster failure this PR targets.

It's also split-brained: `RunCommand`'s *job* spool still hardcodes `/var/spool/spur` and *would* drain, so the two spool paths on the same node disagree on whether to drain. Ironically, the admin who moves the spool onto a dedicated disk *because* of disk-full concerns is exactly who loses the fix.

Suggest classifying against the resolved spool/state root rather than the hardcoded constant (e.g. thread the configured root into `is_node_owned_spool`, or compare against `store`'s state dir).

crates/spurd/src/agent_server.rs
```diff
@@ -14947,6 +14970,47 @@ mod tests {
...
+        let status = svc
+            .run_command(Request::new(RunCommandRequest {
+                command: vec!["echo".into(), "unreachable".into()],
+                uid: 0,
+                gid: 0,
+                job_id,
+                ..Default::default()
+            }))
+            .await
+            .expect_err("spool dir blocked on both candidates; launch must fail");
```
spraveenio: **Test depends on the runner being non-root.**

Under root (common in container CI), `create_job_spool_dir`'s first candidate `/var/spool/spur/job9001` is created successfully instead of failing with `EACCES`, so `open_step_output_files` returns `Ok`, `run_command` proceeds past the spool step, and this `.expect_err(...)` panics → spurious failure.

The test's own comment ("this test runs as a non-root user, so `/var/spool/spur` always fails first anyway") admits the dependency. This violates the repo rule in AGENTS.md: *"Do not write tests that depend on the test runner's environment. Use explicit fixtures so tests pass identically everywhere."*

Use an explicit unwritable fixture path (e.g. a `tempdir` entry made read-only, or a path under a read-only dir you create in the test) so the assertion holds regardless of uid. It also writes to a shared `std::env::temp_dir()/spur/job9001` rather than an isolated tempdir and only cleans up on the success path — an isolated `TempDir` would fix both.

spraveenio (commented):
Two findings from a high-effort review of this branch (both verified against the checked-out code, not just the diff). Detail is in the inline comments; summary:

1. **Correctness** — the new stepd session-dir / launch-spec node-fault draining is gated on the hardcoded `/var/spool/spur`, but the stepd state dir is a documented, relocatable config. When relocated, real ENOSPC faults classify as `Other` and the node never drains — the exact failure this PR set out to stop.
2. **Test env-dependency** — `run_command_drains_the_node_when_its_spool_root_is_unwritable` depends on the runner being non-root and fails under root (container CI).

crates/spurd/src/agent_server.rs
yansun1996: Confirmed — `is_node_owned_spool` needs the resolved state-dir root, not the hardcoded constant. Fixed by threading `state_dir` through at both call sites; internal executor.rs callers keep using the job-spool constant unchanged.

crates/spurd/src/agent_server.rs
yansun1996: Confirmed, this test's failure mode depended on runner uid. Replaced with an injectable job-spool-root override that blocks two isolated tempdirs, so it's deterministic regardless of who runs it.

yansun1996:
Re-validated after the review fixes on an isolated single-node deployment:

- Relocated `--state-dir` (off `/var/spool/spur`) + forced ENOSPC on the session-dir/launch-spec write → node now drains with an accurate reason (previously a no-op, per [1] above).
- Job-specific failure (e.g. missing binary) on the same node → does not drain, as expected.
- Healthy job on the relocated-root node → runs to completion, no false-positive drain.

`cargo test`/`clippy`/`fmt` clean; CI green.

spraveenio (approved):
## Approve

Re-reviewed at `4428e77` against the checked-out branch. Both earlier findings are resolved with real fixes, not band-aids.

**Finding 1 (correctness) — fixed.** `is_node_owned_spool` / `classify_spool_error` now take an explicit `owned_root`; `launch_stepd`'s session-dir prepare and launch-spec write pass the agent's configured `state_dir`, while job-scratch-spool callers keep `SPOOL_ROOT` (correct — those paths genuinely live there). A relocated `state_dir` (`--state-dir` / `SPUR_STEPD_STATE_DIR`) hitting ENOSPC now classifies as `NodeFault` and drains.

**Finding 2 (test env-dependency) — fixed.** The test injects isolated tempdir candidates via a `#[cfg(test)] with_job_spool_roots` seam and blocks them with regular files, so `create_dir_all` fails with ENOTDIR regardless of uid. Verified passing under a mapped-root (uid=0) user namespace — the exact condition that previously broke it.

**No regression.** `cargo clippy --workspace --exclude spur-ffi --all-targets --locked` is clean; no stale callers of the renamed `open_step_output_files`, and the new `job_spool_roots` field doesn't break any `AgentService` construction. Production `RunCommand` keeps the same `[SPOOL_ROOT, temp/spur]` candidate order via `default_job_spool_candidates()` — behavior identical except the intended classification change.

**Tests + evidence.** New regression test `stepd_session_dir_enospc_under_a_relocated_state_dir_is_a_node_fault` asserts NodeFault + drain on a relocated `/data/spur`. All targeted classifier/drain tests pass.

No CLI/config/proto/REST surface change, so no doc update required.

spraveenio (commented):
### Non-blocking follow-up (approval stands)

A deeper pass surfaced one latent inconsistency plus two low-severity nits. None is an active bug or a regression — the diff is correct as merged — but the first is worth folding in since it's the same failure class this PR eliminates, just dormant.

**1. Latent: `step_output_open_error` classifies against a hardcoded root while its caller is parameterized — `executor.rs:1814`.**
`step_output_open_error` passes `Path::new(SPOOL_ROOT)` as `owned_root`, but its enclosing `open_step_output_files_under` is parameterized on `candidates`. Today these agree only because the production default `job_spool_roots[0]` *is* `/var/spool/spur`. The moment `job_spool_roots` is wired to a relocated root (the parameterization is clearly staged for that; currently only the `#[cfg(test)] with_job_spool_roots` override uses it), `create_job_spool_dir_under` would succeed under the relocated root and return `X/jobN`, but a real ENOSPC/EIO opening `X/jobN/stepK.out` would classify against `SPOOL_ROOT` → `X/jobN` fails `starts_with("/var/spool/spur")` → `LaunchError::Other` → node not drained. That's the dir-creation-drains-but-file-open-doesn't split this PR set out to remove, reintroduced one layer down. Suggest threading `candidates.first()` into `step_output_open_error` instead of the constant.

**2. Low: test doesn't actually cover the new drain call site — `agent_server.rs:16788`.**
`register_job_allocation_reaches_the_drain_check_on_a_stepd_launch_failure` forces a missing-`spurstepd` failure (`LaunchError::Other` → no drain) and asserts the node is *not* drained. That assertion holds whether or not the new `drain_on_node_fault` wiring at line 6302 exists, so reverting line 6302 leaves this test green — the NodeFault→drain path from `register_job_allocation` is never exercised end-to-end. Consider a case that drives a `NodeFault` through that call site so it genuinely guards line 6302.

**3. Nit: stale rustdoc after rename — `executor.rs:1782`.**
Doc comments still reference `open_step_output_files` (renamed to `open_step_output_files_under` in this diff), producing a broken intra-doc link and a stale name reference.

All three are fine as fast-follow if you'd rather not reopen this PR.

